PR #66, opened 2017-07-11
Enable Travis CI

texasaggie97:
* Enable Travis CI integration on GitHub
* Add support for using tox for testing
* We do not use tox-travis integration
  * Travis "builds" with one version - 3.4
  * The build then invokes tox which iterates over the listed versions
* Adds 'make test' target
  * Copies tox.ini to bin/<driver> and then invokes tox at that location
* Make all still invokes pytest using the same python that is used to build

* Automation is good

* [Travis CI](https://travis-ci.org/ni/nimi-python)


build/Makefile
```diff
@@ -22,15 +22,17 @@ endef
...
       $(eval $(call per_driver_per_target,$(d),$(t)))))
 
 define per_target
 $1:
 	$(foreach d,$(DRIVERS),make --no-print-directory -f src/$(d)/$(d).mak DRIVER=$(d) $1 &&)  echo
 endef
 
-$(foreach t,$(POSSIBLE_TARGETS),$(eval $(call per_target,$(t))))
+# We add test when we create the $(POSSIBLE_TARGETS). This is because it shouldn't be
+# run when all us used, but we want it to work
```
marcoskirsch: us -> is

build/templates/attributes.py.mako
```diff
@@ -1,17 +1,23 @@
 <%
```
marcoskirsch: We should remove this file, it's not being used. It should be its own PR.

marcoskirsch:
There appears to be changes in the diff that are not mentioned in the PR description. Are we missing stuff?

build/Makefile
texasaggie97: Fixed

build/templates/attributes.py.mako
texasaggie97: Done in #68

PR #75, opened 2017-07-13
Store generated files - Fixes #63

### LINKED ISSUE #63: Generated code is not available in repo for browsing
A client that wants to look at the generated code, but doesn't want to install nor clone and build, has no way to do it today.

@epage suggests putting the generated code in the repo for this purpose.

texasaggie97:
* Fixes #63 
* Puts generated files into generated folder
  * The commit is not done as part of the build
  * Any changes will show up in git and can be committed as part of a regular commit
* Move command_history.txt into generated file and no longer ignore it from git

* Makes it easy to see what the generated code will look like without having to build

* Travis


marcoskirsch:
For command_history, we should consider turning it into .sh or .bat or perhaps both, and also put a comment on top about what it is, something like

> This is the list of commands that make invoked in order to build nimi-python. If you want to reproduce the build but don't have GNU Make setup in your system, you can invoke this script.

(can be different PR)

What do you think?

texasaggie97:
I was going to leave it to whomever might need it to make it into either a .sh or a .bat since I didn't know which would be needed.  However,  can change it to create both though.

PR #79, opened 2017-07-14
Create a context manager object for acquisition and/or generation

texasaggie97:
* Create context manager object that can be used for controlling the acquisition and/or generation
  * Configurable per driver
  * Allows both if needed
  * Use: 
```
with nidmm.Session('dev1') as session:
    with session.acquisition:
        reading = session.fetch(1000)
    with session.acquisition:
        reading = session.fetch(1000)
```
* Update system test with additional test for acquisition context manager
* Add acquisition context manager test

* Next step for API

* Travis & system test


marcoskirsch:
```
with nidmm.Session('dev1') as session:
    with session.acquisition:
        reading = session.fetch(1000)
    with session.acquisition:
        reading = session.fetch(1000)
```

I was thinking something like:

```
    with session.initiate() as acquisition:
        reading = acquisition.fetch(1000)
```

So initiate would be "special", return an Acquisition object, and this Acquisition object would be code-generated to have all the flavors of fetch().

If we want fetch() to be called on session - which is simpler on us - how about:

```
with nidmm.Session('dev1') as session:
    with session.initiate():
        reading = session.fetch(1000)
    with session.initiate:
        reading = session.fetch(1000)
```

build/templates/session.py.mako
```diff
@@ -100,6 +104,24 @@ class AttributeViSession(object):
 
     def __set__(self, obj, value):
         raise TypeError('Attributes of type ViSession are unsupported in Python')
+% for c in config['context_manager']:
+<%
+context_name = 'acquisition' if c['direction'] == 'input' else 'generation'
```
marcoskirsch: should be uppercase Acquisition.

build/templates/session.py.mako
```diff
@@ -100,6 +104,24 @@ class AttributeViSession(object):
...
+exit_function = next(f for f in functions if f['name'].upper() == c['exit'].upper())
+%>\
+
+
+class ${context_name.title()}(object):
+    def __init__(self, session):
+        self.session = session
+
+    def __enter__(self):
+        self.session.${enter_function['python_name']}()
```
marcoskirsch: aren't the enter/exit function names all the same?

generated/nidmm/session.py
```diff
@@ -186,6 +196,7 @@ def __init__(self, resource_name, id_query=0, reset_device=False, options_string
         self.library = library.get_library()
         self.vi = 0  # This must be set before calling _init_with_options.
         self.vi = self._init_with_options(resource_name, id_query, reset_device, options_string)
+        self.acquisition = Acquisition(self)
```
marcoskirsch: intersting. I don't think it's appropriate for acquisition to be a property, it should be a method since it is semantically an action.

build/templates/session.py.mako
texasaggie97: Case doesn't matter here. I use title() case when I am using the name as a class name and lower() when I am using it as a member name.

build/templates/session.py.mako
texasaggie97: Are they? If they are I can hard code them, but I like it being part of the code generation.

generated/nidmm/session.py
texasaggie97: Methods aren't context managers. As far as I have been able to determine, it has to be a class with __enter__ and __exit__ methods or a function (not a method) with a \@contextmanager decoration.

Inner classes do not have access to instance variables like in other languages. This is why I create an object and give it a session object to hold on to. Otherwise it was looking like `session.Acquisition(session)` which was redundant. As I was looking around, either not using inner classes at [all](https://stackoverflow.com/questions/2024566/access-outer-class-from-inner-class-in-python) or using a [property](https://stackoverflow.com/questions/36601006/python-accessing-outer-private-functions-from-inner-class) was the answer:

texasaggie97:
Methods cannot be a context manager. It needs to be either a class or a function.

Using an inner class has no advantage since there is no special access. See reply to comment for more information. Consequently, I came up with two solutions, both using a class to act as the context manager.
1. Class only. invocation looks like `with session.Acquisition(session):`
1. Class with session property. invocation looks like `with session.acquisition:`

I thought about returning an object with operations happening on the object like you suggested, but we would need to know for each driver which operations can happen during an acquisition (or generation):
1. All permutations of Fetch
1. On the fly attributes
1. Other functions that can work while an acquisition is going on such as points available, etc.

The issue I see with that is getting that information and then keeping it up to date.

build/templates/session.py.mako
marcoskirsch: Yes, I think it's always initiate and abort.
I don't want to make it configurable (add complexity) until we know for sure there's a case where it breaks down.

generated/nidmm/session.py
marcoskirsch: Sure, but can you have a method that instantiates the class and returns it?

marcoskirsch:
I thought about what you said about having methods like fetch and read on a new acquisition class. Makes sense to leave those on the Session.


with nidmm.Session('dev1') as session:
    with session.initiate() as acquisition: // initiate() returns a an object called Acquisition.
        reading = session.fetch(1000)
        reading = session.fetch(1000)

generated/nidmm/session.py
texasaggie97: Yes. That is what I had originally but it looks odd because of the redundancy:
```
with nidmm.Session('dev1') as session:
    with session.Acquisition(session):
        print(session.fetch(1000))
```
Note how session is in that line twice. If we are okay with that, then I can go back to that.

And do we need to return the Acquisition object? If so, why? What would we do with it?

marcoskirsch:
> Methods cannot be a context manager. It needs to be either a class or a function.

This illustrates what I meant better, you will have to remove .txt from the file in order to run it.
[cm_test.py.txt](https://github.com/ni/nimi-python/files/1178066/cm_test.py.txt)

texasaggie97:
Okay, yes something like that works.

I now have this:

Context Manager object definition:
```
class Acquisition(object):
    def __init__(self, session):
        self.session = session

    def __enter__(self):
        self.session._initiate()
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        self.session._abort()
```

How Session uses it:
```
    def initiate(self):
        return Acquisition(self)
```

Called like this:
```
        with nidmm.Session('dev1') as session:
            with session.initiate():
```
I have some questions:
1. I am hard coding `_initiate` and `_abort`. Is that reasonable?
1. I am hard coding the name of the function that returns the CM to `initiate`. Is that okay?
1. Is using the name initiate going to be confusing to someone who tries to use the API? They call `session.initiate()` and nothing happens. There's nothing to show that it returns a context manager, other than documentation, and initiate is the function you would normally call.

PR #86, opened 2017-07-20
Add enum documentation plus update session documentation

texasaggie97:
* Add generated documentation for enums
  * add docstring to enums.py for use in editors
  * Create explicit documentation for enums using Sphinx Python domain commands
  * Some enums' names don't match the name from the documentation so add a lookup key that can be used for the ones that do differ
  * Don't use common enum for attributes:
    * EnabledSetting became AutoZero, ADCCalibration, OffsetCompensatedOhms
    * Slope became TriggerSlope, SampleTrigSlope, MeasurementDestinationSlope
  * Added specific enums for terminals
    * TriggerSource
    * MeasurementCompleteDest
    * SampleTrigger
* Update generated session documentation to also use Sphinx Python domain commands
  * Simple list of attributes with link to more complete help
  * If the attribute type is an enum, add link to enum help

* Documentation improvement

* Travis


src/nidmm/metadata/enums.py
```diff
@@ -1,148 +1,254 @@
...
-    ],
-
-    'AcquisitionStatus': [
-        {'name': 'RUNNING', 'value': 0},
-        {'name': 'FINISHED_WITH_BACKLOG', 'value': 1},
-        {'name': 'FINISHED_WITH_NO_BACKLOG', 'value': 2},
-        {'name': 'PAUSED', 'value': 3},
-        {'name': 'NO_ACQUISITION_IN_PROGRESS', 'value': 4},
-    ],
+    'ApertureTimeUnits': {
```
marcoskirsch: These were arrays so that we can control the order of the enum values, which we may want to display in an order different than their ascending numeric values.

What was behind this change?

src/nimodinst/metadata/attribute_docs.py
```diff
@@ -2,216 +2,216 @@
...
+The following table lists the characteristics of this property.
+
++------------------+-------------+
+| Characteristic   | Value       |
++------------------+-------------+
+| Datatype         | string      |
++------------------+-------------+
+| Permissions      | Read Only   |
++------------------+-------------+
+| High Level VI    | N/A         |
```
marcoskirsch: We should remove this

PR #87, opened 2017-07-26
Added more system tests

injaleea:
Adding more system tests for nidmm for checking function calling with different function types

[ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/<reponame>/blob/master/CONTRIBUTING.md).

TODO: Check the above box with an 'x' indicating you've read and followed [CONTRIBUTING.md](https://github.com/ni/<reponame>/blob/master/CONTRIBUTING.md).

- Added System tests for nidmm
    - test_parameter_count0
    - test_parameter_count1
    - test_parameter_count1_error
    - test_parameter_count2

  More System test coverage

Ran added system tests before submission


src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -116,6 +116,39 @@ def test_ViSession_attribute(device_info):
...
+
+def test_parameter_count0(device_info):
+    with nidmm.Session(device_info['name']) as session:
+        assert session.get_aperture_time_info()[1] == 0 # Assuming default aperture time unit will be seconds
+
+                
+def test_parameter_count1(device_info):
+    with nidmm.Session(device_info['name']) as session:
+        session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 1, 5.5)
+        session._initiate()
```
texasaggie97: This can be rewritten in terms on the initiate context. 

with session.initiate():

texasaggie97:
I'm not clear on what these tests are trying to verify. We probably talked about this but I don't remember.

src/nidmm/system_tests/test_system_nidmm.py
marcoskirsch: I don't think there's any good reason to use any private methods ever in system tests.

marcoskirsch:
> I'm not clear on what these tests are trying to verify.

I agree. Ideally the test name makes that perfectly clear. If this cannot be accomplished then add a docstring to the test.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -66,6 +66,7 @@ def test_warning(device_info):
 
 def test_ViBoolean_attribute(device_info):
     with nidmm.Session(device_info['name']) as session:
+
```
marcoskirsch: remove empty line to match other tests

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -94,43 +95,35 @@ def test_Enum_attribute(device_info):
...
     with nidmm.Session(device_info['name']) as session:
         assert session.get_aperture_time_info()[1] == 0 # Assuming default aperture time unit will be seconds
-
-                
-def test_parameter_count1(device_info):
+   
+        
+def test_function_call_with_one_parameter(device_info):
     with nidmm.Session(device_info['name']) as session:
         session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 1, 5.5)
```
marcoskirsch: The test is test_function_call_with_one_parameter but in the body a lot of stuff is happening.
Can you pare it down? For example, a lot happens when we call initiate().

There must be some 1 param method you can use.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -94,43 +95,35 @@ def test_Enum_attribute(device_info):
...
-        session._initiate()
-        assert session.fetch(1000) != 0 # Assumes DMM doesn't Fetch is not exactly zero to support non-connected modules and simulated modules.
-        session._abort()
+        with session.initiate():
+            assert session.fetch(1000) != 0 # Assumes DMM doesn't Fetch is not exactly zero to support non-connected modules and simulated modules.
+
         
-    
-def test_parameter_count1_error(device_info):
+def test_function_call_with_one_parameter(device_info):
```
marcoskirsch: again?

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -94,28 +94,50 @@ def test_Enum_attribute(device_info):
...
         try:
-            session.io_session = 5
-            assert false
-        except TypeError as e:
-            print(e)
-            pass
-        try:
-            value = session.io_session
-            assert false
+            session.fetch()
```
marcoskirsch: This would raise an exception even if fetch() didn't take a parameter because the session is not in the running state. Use a different function please.
Also, fetch() or read() are bad examples because we are planning on having a default value for the timeout parameter.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -94,28 +94,50 @@ def test_Enum_attribute(device_info):
...
+            print(session.fetch(1000))
+
+			
+def test_method_call_with_zero_parameter(device_info):
+    with nidmm.Session(device_info['name']) as session:
+        assert session.get_aperture_time_info()[1] == 0 # Assuming default aperture time unit will be seconds
+   
+        
+def test_method_call_with_one_parameter(device_info):
+    session.configure_power_line_frequency(60)
```
texasaggie97: You are missing the with statement in this test

PR #101, opened 2017-07-28
Create and acquire lock before referencing or updating cfunc information

### LINKED ISSUE #92: (Suspected) Spurious exceptions with multi-threaded calls
DAQmx at least ran into this where two threads tried to assign `argtypes` at the same time and got strange exceptions.

See https://github.com/ni/nidaqmx-python/commit/82f7ee56580abc5adf44fc8069eac5a9eeff7e4d

DAQmx previously re-assign `argtypes` on every call.  nimi-python does this once.  This leaves the race to be between the check and the assignment.

```python3
    def niDMM_close(self, vi):  # noqa: N802
        if self.niDMM_close_cfunc is None:
            self.niDMM_close_cfunc = self._library.niDMM_close
```

1. Thread A performs if-check
2. Thread B performs if-check
3. Thread A assigns
4. Thread B assigns

https://github.com/ni/nimi-python/blob/master/generated/nidmm/ctypes_library.py#L101

texasaggie97:
* Fixes #92 
* Create and acquire lock before referencing or updating cfunc information
* List functions sorted by name
  * This will prevent spurious changes because Python iterated over objects in a different order

* Fix issue

* Travis & system test

### DIFF
--- a/build/templates/ctypes_library.py.mako
+++ b/build/templates/ctypes_library.py.mako
@@ -27,7 +27,7 @@ import ${module_name}.python_types
 class ${module_name.title()}CtypesLibrary(object):
     def __init__(self, library_name, library_type):
         # We cache the cfunc object from the ctypes.CDLL object
-% for f in functions:
+% for f in helper.sorted_functions(functions):
         self.${c_function_prefix}${f['name']}_cfunc = None
 % endfor
 
@@ -36,7 +36,7 @@ class ${module_name.title()}CtypesLibrary(object):
         else:
             assert library_type == 'cdll'
             self._library = ctypes.CDLL(library_name)
-% for f in functions:
+% for f in helper.sorted_functions(functions):
 <%
     func_name = c_function_prefix + f['name']
     params = f['parameters']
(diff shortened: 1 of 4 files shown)


build/templates/ctypes_library.py.mako
```diff
@@ -19,15 +19,17 @@ functions = helper.add_all_metadata(functions)
...
 from ${module_name}.ctypes_types import *  # noqa: F403,H303
 import ${module_name}.python_types
 
 
 class ${module_name.title()}CtypesLibrary(object):
     def __init__(self, library_name, library_type):
+        self._func_lock = threading.Lock()
         # We cache the cfunc object from the ctypes.CDLL object
-% for f in functions:
+% for f in helper.sorted_functions(functions):
```
marcoskirsch: What's with the sorting, is that needed by this fix?

texasaggie97: No, it's not required.

It is because iterating over dictionaries is [non deterministic](https://stackoverflow.com/questions/14956313/why-is-dictionary-ordering-non-deterministic)

By making this change, there won't be arbitrary changes to generated files because Python chose to iterate in a different order.

PR #104, opened 2017-07-28
Fix system test

### LINKED ISSUE #95: test_system_nidmm.py test_ViSession_attribute fails
Test expects session.io_session = 5 to raise TypeError.
This test used to work, although someone could argue that we _should_ allow setting a numeric value on a ViSession.

```
__________________________ test_ViSession_attribute ___________________________

device_info = {'name': '4070', 'sn': '00D82BA5'}

    def test_ViSession_attribute(device_info):
            try:
                session.io_session = 5
>               assert False
E               assert False

src\nidmm\system_tests\test_system_nidmm.py:108: AssertionError
```

texasaggie97:
* Fixes #95 
* Remove support for ViSession attributes
  * Remove ViSession attribute test
* Use new, correct enum values

* Fixes issues

* Travis & system test

### DIFF
--- a/src/nidmm/system_tests/test_system_nidmm.py
+++ b/src/nidmm/system_tests/test_system_nidmm.py
@@ -94,23 +94,6 @@ def test_Enum_attribute(device_info):
         assert session.function == nidmm.Function.AC_CURRENT
         assert type(session.function) is nidmm.Function
         try:
-            session.function = nidmm.LCCalculationModel.CALC_MODEL_SERIES
-            assert false
-        except TypeError as e:
-            print(e)
-            pass
-
-
-def test_ViSession_attribute(device_info):
-    with nidmm.Session(device_info['name']) as session:
-        try:
-            session.io_session = 5
-            assert false
-        except TypeError as e:
-            print(e)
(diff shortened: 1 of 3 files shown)


src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -52,7 +52,7 @@ def test_wrong_parameter_type(device_info):
 
 def test_warning(device_info):
     with nidmm.Session(device_info['name']) as session:
-        session.configure_measurement_digits(nidmm.Function.RES_2_WIRE, 1e6, 3.5)
+        session.configure_measurement_digits(nidmm.Function._2_WIRE_RESISTANCE, 1e6, 3.5)
```
marcoskirsch: Weird to use _ because this enum value happens to start with number.
DAQmx spells it out when needed so would be: TWO_WIRE_RESISTANCE

I don't think it's for this PR but I will open an issue about it.

marcoskirsch: https://github.com/ni/nimi-python/issues/106

PR #105, opened 2017-07-28
Bug/bug96

### LINKED ISSUE #96: Disallow inserting attributes into an instance of Session
Say you have this code:

```
def test_take_simple_measurement_works(device_info):
        session.rannge = 50 # Coerces up!
        value = session.read(1000) != 0 # Assumes DMM reading is not exactly zero to support non-
```

Interestingly, this appears to work, but notice the typo when setting the range property! The range property was never set on the session and no error occurred.

Given that:

* Our Python bindings are abstractions on top of pre-established APIs
* We'll release as 0.1 and there's a chance that attribute names will get tweaked (as happened in https://github.com/ni/nimi-python/issues/95)

I propose that we disallow inserting any attributes into an instance of Session. How to do that is TBD.

texasaggie97:
* Fixes #96 
* "Freeze" the class so no more attributes can be set after __init__ runs
* Add unit test for freeze
* Add ModInst unit test for iteration

* Fixes issue

* Travis & system test

### DIFF
--- a/build/templates/session.py.mako
+++ b/build/templates/session.py.mako
@@ -126,6 +126,8 @@ class ${context_name.title()}(object):
 class Session(object):
     '''${config['session_description']}'''
 
+    __isfrozen = False
+
 % for attribute in helper.sorted_attrs(attributes):
     %if attributes[attribute]['enum']:
     ${attributes[attribute]['name'].lower()} = AttributeEnum(${attribute}, enums.${attributes[attribute]['enum']})
@@ -143,6 +145,13 @@ class Session(object):
         self.library = library.get_library()
         self.vi = 0  # This must be set before calling _init_with_options.
         self.vi = self._init_with_options(resource_name, id_query, reset_device, options_string)
+
+        self.__isfrozen = True
+
+    def __setattr__(self, key, value):
+        if self.__isfrozen and key not in dir(self):
+            raise TypeError("%r is a frozen class" % self)
+        object.__setattr__(self, key, value)
 % for c in config['context_manager']:
 <%
 context_name = 'acquisition' if c['direction'] == 'input' else 'generation'
--- a/build/templates/session.py.mako
+++ b/build/templates/session.py.mako
@@ -126,7 +126,8 @@ class ${context_name.title()}(object):
 class Session(object):
     '''${config['session_description']}'''
 
-    __isfrozen = False
+    # This is needed during __init__. Without it, __setattr__ raises an exception
+    _is_frozen = False
 
 % for attribute in helper.sorted_attrs(attributes):
     %if attributes[attribute]['enum']:
@@ -146,10 +147,10 @@ class Session(object):
         self.vi = 0  # This must be set before calling _init_with_options.
         self.vi = self._init_with_options(resource_name, id_query, reset_device, options_string)
 
-        self.__isfrozen = True
+        self._is_frozen = True
 
     def __setattr__(self, key, value):
-        if self.__isfrozen and key not in dir(self):
+        if self._is_frozen and key not in dir(self):
             raise TypeError("%r is a frozen class" % self)
         object.__setattr__(self, key, value)
 % for c in config['context_manager']:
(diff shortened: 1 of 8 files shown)


build/templates/session.py.mako
```diff
@@ -126,6 +126,8 @@ class ${context_name.title()}(object):
 class Session(object):
     '''${config['session_description']}'''
 
+    __isfrozen = False
```
marcoskirsch: I'm guessing this idea came from:
https://stackoverflow.com/questions/43460776/python-prevent-dynamic-addition-of-variables-in-static-classes

So far we use single _ for private. I'd suggest following that convention and calling this _is_frozen. If the convention needs to change we can do that for the whole thing in a single PR.

build/templates/session.py.mako
```diff
@@ -143,6 +145,13 @@ class Session(object):
         self.library = library.get_library()
         self.vi = 0  # This must be set before calling _init_with_options.
         self.vi = self._init_with_options(resource_name, id_query, reset_device, options_string)
+
+        self.__isfrozen = True
```
marcoskirsch: __isfrozen on line 129 is a class variable, but this is an instance variable.
Perhaps the class variable is redundant.

build/templates/session.py.mako
```diff
@@ -143,6 +145,13 @@ class Session(object):
...
+
+    def __setattr__(self, key, value):
+        if self.__isfrozen and key not in dir(self):
+            raise TypeError("%r is a frozen class" % self)
```
marcoskirsch: Should be KeyError or AttributeError (not sure which makes more sense semantically, would need to check docs)

generated/nimodinst/tests/test_modinst.py
```diff
@@ -50,3 +50,25 @@ def test_context_manager(self):
             self.patched_ctypes_library.niModInst_OpenInstalledDevicesSession.assert_called_once_with(b'', ANY, ANY)
             self.patched_errors._handle_error.assert_called_once_with(session, self.patched_ctypes_library.niModInst_OpenInstalledDevicesSession.return_value)
         self.patched_ctypes_library.niModInst_CloseInstalledDevicesSession.assert_called_once_with(SESSION_NUM_FOR_TEST)
+
+    def test_freeze(self):
```
marcoskirsch: test_cannot_add_properties_to_session

Better be more explicit in our test names.

generated/nimodinst/tests/test_modinst.py
```diff
@@ -50,3 +50,25 @@ def test_context_manager(self):
...
+        with nimodinst.Session('') as session:
+            try:
+                session.io_session = 5
```
marcoskirsch: consider using something other than io_session as this can confuse someone familiar with NI-DMM in general.

build/templates/session.py.mako
texasaggie97: Changed to _is_frozen

build/templates/session.py.mako
texasaggie97: They are both needed. Comment added.

Without the class one, then __setattr__ raises an exception during __init__

build/templates/session.py.mako
texasaggie97: Is it? I thought TypeError was reasonable since you can't modify this type.

marcoskirsch: https://docs.python.org/2/library/exceptions.html#exceptions.AttributeError
exception AttributeError
Raised when an attribute reference (see Attribute references) or assignment fails. (When an object does not support attribute references or attribute assignments at all, TypeError is raised.)

https://docs.python.org/2/library/exceptions.html#exceptions.TypeError
exception TypeError
Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.

https://docs.python.org/2/library/exceptions.html#exceptions.KeyError
exception KeyError
Raised when a mapping (dictionary) key is not found in the set of existing keys.

I suppose TypeError is the right one after all! My bad.

PR #109, opened 2017-07-28
Documentation changes

texasaggie97:
* Changes how documentation (rst files) is generated
  * Don't store complete long_description
  * Generate characteristics table during generation
  * Show attributes in session differently
* Changes can be previewed on [read the docs](http://nimi-python.readthedocs.io/en/docs-build_docs/index.html)
 
* Clean up documentation

* Travis


build/helper.py
```diff
@@ -240,3 +241,121 @@ def get_rst_header_snippet(t, header_level='='):
...
+        return 'floating point'
+    elif visa_type == 'ViChar':
+        return 'string'
+    elif visa_type == 'ViString':
+        return 'string'
+    elif visa_type == 'ViConstString':
+        return 'const string'
+    elif visa_type == 'ViBoolean':
+        return 'bool'
+    elif visa_type == '':
```
marcoskirsch: I don't understand these last two.

build/helper.py
```diff
@@ -240,3 +241,121 @@ def get_rst_header_snippet(t, header_level='='):
...
+
+    for d in data[1:-1]:
+        table.append(template.format(*d))
+        if full:
+            table.append(separator)
+    table.append(template.format(*data[-1]))
+    table.append(separator)
+    return '\n'.join(table)
+
+def get_python_type_from_visa_type(visa_type):
```
marcoskirsch: We have classes for each one of these clases. Why don't we store the "python" type as a class member of those? We should info about types centralized in one place.

texasaggie97: I thought about that and immediately saw two problems. 
1. This is a different context than the classes we have. Those are for during runtime of the driver, this is for build time when we are generating the driver. While I think it would work, it seems odd to include something from the templates folder during build. 
2. I am not instantiating any of these to then query the base type. I would need something like an orb (object request broker) to instantiate an object from just a name.

build/helper.py
texasaggie97: Oops. To many copies. I will remove them.

build/helper.py
marcoskirsch: > 1. This is a different context than the classes we have. Those are for during runtime of the driver, this is for build time when we are generating the driver. While I think it would work, it seems odd to include something from the templates folder during build.

I agree it's weird. But I think given the tradeoff, it's better than having a big if/else/if/else structure with the information spread out. If anything we should look into consolidating ctypes_types.py and python_types.py (I'll open an issue).

> 2. I am not instantiating any of these to then query the base type. I would need something like an orb (object request broker) to instantiate an object from just a name.

ORB? This is Python!

```
class Foo(object):
    def __init__(self):
        print("Hello World!")

# Create a Foo by name
foo = globals()["Foo"]()
```

build/helper.py
```diff
@@ -240,3 +241,119 @@ def get_rst_header_snippet(t, header_level='='):
...
+    elif visa_type == 'ViChar':
+        return 'string'
+    elif visa_type == 'ViString':
+        return 'string'
+    elif visa_type == 'ViConstString':
+        return 'const string'
+    elif visa_type == 'ViBoolean':
+        return 'bool'
+    else:
+        return 'Unknown'
```
thejcannon: I bet you want to raise an Exception here rather than returning 'Unknown'

PR #118, opened 2017-08-08
Add support for "IVI dance" buffer size buffered reads

texasaggie97:
* Add support for "IVI dance" buffer size buffered reads
  * To use the "ivi-dance" style of getting a buffer, the size key for the parameter that will use it should be set to `'ivi-dance,<variable that is used for size>'`
  * Example:
    * `niDMM_GetAttributeViString(vi, channelName, attributeID, bufSize, value)`
    * The "size" key should look like `'size': 'ivi-dance,bufSize',`
* Removed hardcoded get_atttribute_vi_string functions since generated function now works

* Needed for API

* Travis and system tests


build/helper.py
```diff
@@ -179,28 +198,23 @@ def _get_output_param_return_snippet(output_parameter):
...
+def get_method_return_snippet(parameters):
     '''Returns a string suitable to use as the return argument of a Session method, i.e. "return reading_ctype.value"'''
-    if len(output_parameters) == 0:
-        return 'return'
     snippets = []
-    for x in output_parameters:
-        snippets.append(_get_output_param_return_snippet(x))
-    return 'return ' + ', '.join(snippets)
+    for x in parameters:
+        if x['direction'] == 'out' or x['ivi-dance']:
```
marcoskirsch: isn't it the case that ivi-dance parameters are always direction 'out'?

build/helper.py
```diff
@@ -38,13 +38,24 @@ def extract_input_parameters(parameters, sessionName = 'vi'):
...
 
 def extract_output_parameters(parameters):
-    '''Returns list of parameters only with output parameters'''
-    return [x for x in parameters if x['direction'] == 'out']
+    '''Returns list of parameters only with output parameters, not including the ivi-dance parameter if it exists'''
+    return [x for x in parameters if x['direction'] == 'out' and not x['ivi-dance']]
 
 def extract_enum_parameters(parameters):
-    '''Returns a dictionary with information about the output parameters of a session method'''
+    '''Returns a list with information about the output parameters of a session method'''
```
texasaggie97: It was wrong before, but this is still wrong. In a different branch, docs/function_help, I have changed it to:

> Returns a list of parameter dicts that have an associated enum

build/helper.py
```diff
@@ -38,13 +38,24 @@ def extract_input_parameters(parameters, sessionName = 'vi'):
     return [x for x in parameters if x['direction'] == 'in' and x['name'] != sessionName]
 
 def extract_output_parameters(parameters):
```
texasaggie97: How does it no longer match? Because it is output-params minus ivi-dance-params?

build/helper.py
```diff
@@ -90,13 +101,19 @@ def _add_python_return_type(f):
...
-def _add_is_buffer(parameter):
-    '''Adds 'is_buffer' key/value pair to the parameter metadata iff not already populated'''
+def _add_buffer_info(parameter):
+    '''Adds buffer information to the parameter metadata iff 'size' is defined else assume not a buffer'''
     try:
-        parameter['is_buffer']
+        parameter['size']
+        parameter['is_buffer'] = True
+        parameter['ivi-dance'] = False
+        if type(parameter['size']) is str and parameter['size'].startswith('ivi-dance,'):
```
texasaggie97: Metadata is under control of the developer. I am okay with being strict about the format.

I was avoiding RegExs.

Though I am open on how to improve this.

build/helper.py
```diff
@@ -90,13 +101,19 @@ def _add_python_return_type(f):
...
+            parameter['ivi-dance'] = True
+            parameter['size'] = camelcase_to_snakecase(parameter['size'].split(',')[1])
     except KeyError:
         # Not populated, assume False
         parameter['is_buffer'] = False
+        parameter['ivi-dance'] = False
```
texasaggie97: No, 'size' only matters when is_buffer == True.

Though we could just for completeness.

build/helper.py
```diff
@@ -153,7 +170,9 @@ def get_library_call_parameter_snippet(parameters_list, sessionName = 'vi'):
                     snippet += '.encode(\'ascii\')'
         else:
             assert x['direction'] == 'out', pp.pformat(x)
-            if x['is_buffer']:
+            if x['ivi-dance']:
+                snippet = x['ctypes_variable_name']
```
texasaggie97: Even though _add_buffer_info() prevents that from ever happening?

build/helper.py
texasaggie97: Theoretically yes. I can remove the or.

This brings up a question though. Should we run some sort of validation on the metadata? To ensure there aren't places that don't meet these type of assumptions?

PR #120, opened 2017-08-10
Added test for writeonly attribute

bhaswath:

Adds a test case for the test test_readonly_attribute, which tests that appropriate exception is thrown when trying to write into a read-only attribute.

To increase coverage of tests

Tested locally on a machine with DMM and test passes

### DIFF
--- a/src/nidmm/system_tests/test_system_nidmm.py
+++ b/src/nidmm/system_tests/test_system_nidmm.py
@@ -139,5 +139,5 @@ def test_writeonly_attribute(device_info):
         try:
             session.channel_count = 5
         except nidmm.Error as e:
-            assert e.code == -1074135027
+            assert e.code == -1074135027 #Error : Attribute is read-only.
 
--- a/src/nidmm/system_tests/test_system_nidmm.py
+++ b/src/nidmm/system_tests/test_system_nidmm.py
@@ -132,3 +132,10 @@ def test_get_dev_temp(device_info):
         temperature = session.get_dev_temp('')
         print(temperature)
         assert 20 <= temperature <= 50
+
+def test_writeonly_attribute(device_info):
+    with nidmm.Session(device_info['name']) as session:
+        try:
+            session.channel_count = 5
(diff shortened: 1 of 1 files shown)


src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -132,3 +132,10 @@ def test_get_dev_temp(device_info):
         temperature = session.get_dev_temp('')
         print(temperature)
         assert 20 <= temperature <= 50
+
```
injaleea: Add one more empty space to keep the consistency with other tests

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -132,3 +132,10 @@ def test_get_dev_temp(device_info):
...
+def test_writeonly_attribute(device_info):
+    with nidmm.Session(device_info['name']) as session:
+        try:
+            session.channel_count = 5
+        except nidmm.Error as e:
+            assert e.code == -1074135027
```
injaleea: Add a newline at the end

texasaggie97:
When you fill in the PR description, you should check the box next to "This contribution adheres to CONTRIBUTING.md." and remove the TODOs as you take care of them.

src/nidmm/system_tests/test_system_nidmm.py
texasaggie97: I would also like to see a comment about what the error is. I think the description should work.

This way if the error code ever changes, there will at least be some information to help hunt for the new appropriate error code.

texasaggie97: According to pep8 (coding convention) there should not be an blank line at the end of the file.

However, having said that, I have disabled that particular check from the flake8 (pep8 enforcement) checks.
1. There were some generated files I couldn't figure out how to keep mako from adding it
1. I don't happen to agree with it

Number 1. was the real reason but 2. was just a happy coincidence.

PR #125, opened 2017-08-13
New/ivi dance cleanup

marcoskirsch:

Cleanup of code generator:

* Improved metadata format for specifying size of buffers.
    * Size now point to a dictionary with 'mechanism' and 'value' keys
    * No more need to parse the size string.
    * No more need to make assumptions about strings.
* Moved code from templates into helper functions.
* Removed hidden contract about how python_name is determined used in several places with actual lookups in the metadata. This is cleaner and will help us should we ever change rules or override how python names are determined.
* Other minor cleanup, renames, annotations.

Overall improvements to code, takes care of several TODOs.

Unit tests only, as well as diffing generated code.

### DIFF
--- a/build/helper.py
+++ b/build/helper.py
@@ -118,7 +118,7 @@ def _add_python_return_type(f):
     f['returns_python'] = f['returns']
     return f
 
-def _add_is_buffer(parameter):
+def _add_buffer_info(parameter):
     '''Adds buffer information to the parameter metadata iff 'size' is defined else assume not a buffer'''
     try:
         parameter['size']
@@ -140,7 +140,7 @@ def add_all_metadata(functions):
             _add_python_type(p)
             _add_ctypes_variable_name(p)
             _add_ctypes_type(p)
-            _add_is_buffer(p)
+            _add_buffer_info(p)
     return functions
 
 # Python 2/3 compatibility
--- a/build/helper.py
+++ b/build/helper.py
@@ -157,7 +157,7 @@ def normalize_string_type(d):
 
 # Functions that return snippets that can be placed directly in the templates.
 
-def get_method_parameters_snippet(parameters, skip_session_handle, skip_ivi_dance_size_parameter, skip_output_parameters):
+def get_method_parameters_snippet(parameters, skip_session_handle, skip_output_parameters, skip_ivi_dance_size_parameter, session_name = 'vi'):
     '''
     Returns a string suitable for the parameter list of a method given a list of parameter objects.
     You can optionally skip session handle parameter, the parameter used for an output
@@ -171,7 +171,7 @@ def get_method_parameters_snippet(parameters, skip_session_handle, skip_ivi_danc
             skip = True
         if x == ivi_dance_size_parameter and skip_ivi_dance_size_parameter:
             skip = True
-        if x['name'] == 'vi' and skip_session_handle:
+        if x['name'] == session_name and skip_session_handle:
             skip = True
         if not skip:
             snippets.append(x['python_name'])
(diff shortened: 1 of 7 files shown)


build/helper.py
```diff
@@ -136,23 +143,38 @@ def add_all_metadata(functions):
...
         elif type(d) is str:
             d = d.decode('utf-8')
-
     return d
+
 # Functions that return snippets that can be placed directly in the templates.
 
-def get_method_parameters_snippet(parameters):
-    '''Returns a string suitable for the parameter list of a method given a list of parameter objects'''
+def get_method_parameters_snippet(parameters, skip_session_handle, skip_ivi_dance_size_parameter, skip_output_parameters):
```
marcoskirsch: This is one of the more interesting changes I think. Instead of caller having to filter out unwanted parameters, now caller can pass all function parameters and specify what to skip. What do you think?

build/helper.py
```diff
@@ -136,23 +143,38 @@ def add_all_metadata(functions):
...
+    '''
+    Returns a string suitable for the parameter list of a method given a list of parameter objects.
+    You can optionally skip session handle parameter, the parameter used for an output
+    buffer size (i.e. you don't want it in a Session method), and output parameters.
+    '''
     snippets = ['self']
+    ivi_dance_size_parameter = find_size_parameter(extract_ivi_dance_parameter(parameters), parameters)
```
texasaggie97: find_size_parameter() uses ['size']['value'] as the name to search for in find_parameter() which will assert if nothing is found. 

You call it here unconditionally, so the default value of 1 can be used, right? Or am I missing something?

build/helper.py
```diff
@@ -200,15 +222,15 @@ def get_library_call_parameter_types_snippet(parameters_list):
...
+def _get_output_param_return_snippet(output_parameter, parameters):
     '''Returns the snippet for returning a single output parameter from a Session method, i.e. "reading_ctype.value"'''
     assert output_parameter['direction'] == 'out', pp.pformat(output_parameter)
     if output_parameter['is_buffer']:
         if output_parameter['type'] == 'ViChar' or output_parameter['type'] == 'ViString':
             snippet = output_parameter['ctypes_variable_name'] + '.value.decode("ascii")'
         else:
-            # TODO(marcoskirsch): I don't like calling camelcase_to_snakecase here, it relies on contract that parameter name where the size is stored was created with that function.
-            snippet = '[' + output_parameter['ctypes_variable_name'] + '[i].value for i in range(' + camelcase_to_snakecase(output_parameter['size']) + ')]'
+            size_parameter = find_size_parameter(output_parameter, parameters)
```
texasaggie97: This seems like it would have the same issue as above regarding not finding a parameter when ['size']['value'] is a size, and not another parameter name

marcoskirsch: So far we don't have any use cases in the metadata in which we return a buffer of fixed size that is other than a string. So we never hit that case, so I can't test it correctly.

If you are ok with it, I'm inclined to leave it alone (maybe TODO it?) until we can actually exercise the codepath and need to fix it. This is one example in which having a fake driver with one of each "cases" covered in metadata would be *very* helpful.

build/helper.py
```diff
@@ -136,23 +143,38 @@ def add_all_metadata(functions):
...
     snippets = ['self']
+    ivi_dance_size_parameter = find_size_parameter(extract_ivi_dance_parameter(parameters), parameters)
     for x in parameters:
-        snippets.append(x['python_name'])
+        skip = False
+        if x['direction'] == 'out' and skip_output_parameters:
+            skip = True
+        if x == ivi_dance_size_parameter and skip_ivi_dance_size_parameter:
+            skip = True
+        if x['name'] == 'vi' and skip_session_handle:
```
marcoskirsch: done

PR #126, opened 2017-08-15
Add function help to Read The Docs as well as adding docstrings

texasaggie97:
* Add documentation to metadata for functions and function parameters
* Add docstrings to generated functions in session.py
* Add function reference on [readthedocs](http://nimi-python.readthedocs.io/en/docs-function_help/)

* Updates documentation

* Travis & System tests
* Visual inspection of documentation


docs/nidmm/functions.rst
```diff
@@ -63,15 +63,15 @@ NI-DMM Functions
         :py:data:`nidmm.function` attribute to AC measurements. The driver sets the
         :py:data:`nidmm.AC\_MAX\_FREQ` attribute to this value. The valid range is 1
         Hz–300 kHz for the NI 4080/4081/4082 and the NI 4070/4071/4072, 10
-        Hz–100 Hz for the NI 4065, and 20 Hz–25 kHz for the NI 4050 and NI 4060.
+        Hz–100 Hz for the NI 4065, and 20 Hz–25 kHz for the NI 4050 and NI 4060.
```
marcoskirsch: any idea why GitHub thinks this line is different?

build/helper.py
```diff
@@ -177,10 +177,16 @@ def get_method_parameters_snippet(parameters, skip_session_handle, skip_output_p
             snippets.append(x['python_name'])
     return ', '.join(snippets)
 
-def get_function_parameters_snippet(parameters):
-    '''Returns a string suitable for the parameter list of a method given a list of parameter objects'''
+def get_function_parameters_snippet(parameters, sessionName=None):
```
marcoskirsch: should be session_name
Weird that default is different than in other functions. We should make it match or stop using default parameter value everywhere.

marcoskirsch: In my last PR, I added skip_foo bool parameters. I think we should follow a consistent pattern.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -146,7 +146,8 @@ def test_multi_point_acquisition(device_info):
         session.configure_multi_point(4, 2, nidmm.SampleTrigger.IMMEDIATE, 0)
         session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 1, 5.5)
         measurements, numberOfMeasurements = session.read_multi_point(-1, 8)
-        print(measurements)
+        for measurement in measurements:
```
marcoskirsch: Less important, but we should keep the PRs focused. This doesn't seem to be relevant to this PR.

docs/nidmm/functions.rst
texasaggie97: To make it work with Python2, I change non-breaking spaces '\xa0' to normal spaces.

build/helper.py
texasaggie97: So to skip the session name you would need to parameters? skip_session_name=True and session_name="vi"?

Right now it is setup that only one parameter is needed: session_name.

src/nidmm/system_tests/test_system_nidmm.py
texasaggie97: I agree this isn't related to this PR. 

How should we deal with these type of small improvements though? It seems like a lot of overhead to try create a new branch, make a change like this, create a PR and then wait for it to be approved.

PR #133, opened 2017-08-16
Run the unit tests every time

### LINKED ISSUE #124: Unit tests aren't re-run if they fail in next make
Say you run 

    make

...and the unit tests fail.

Then next time you run

    make

the unit tests won't be re-run and the build will complete successfully.

This is probably because the make rule uses file test_results.log alone as a target in order to determine if it needs to execute.

texasaggie97:
* This is not invoking tox, just run a quick pytest on the generated files
* Fixes #124 

* Fixes issue

* Travis
* Building multiple time under different circumstances


texasaggie97:
I prefer to merge this and start another issue about building too much. Wheels also build every time.

PR #135, opened 2017-08-16
Fix return value types

### LINKED ISSUE #127: Functions returning wrong type
get_self_cal_supported function is supposed to return value of type Boolean instead it returns Integer.

```
with nidmm.Session(args.name) as session:
        print( session.get_self_cal_supported() )

>> 1
```

### LINKED ISSUE #128: Methods suppose to return enum is returning Int
read_status method suppose to return acquisition status as enum type . Instead it is returning Int.
 
```
with nidmm.Session(args.name) as session:
     print(session.read_status())
     print(type(session.read_status()[0]))
     print(type(session.read_status()[1]))
 
>>> 
(0, 4)
<class 'int'>
<class 'int'>
```

texasaggie97:
* Fixes #127 
* Fixes #128 
* Cast return values to the underlying intrinsic type (bool, float, int, str)
  * This is done using Visa types
  * Visa types now return instance of appropriate intrinsic type on new
* Force ViBoolean to return a bool type
  * Then need special case when converting from Visa type to Python type
  * ViBoolean has to be of type int since bool can't be subclassed
* All documentation should now reference Python intrinsic types or enums
  * No more Visa types in API

* Fixes issues

* Travis
* System tests
* Added unit tests


build/templates/python_types.py
```diff
@@ -64,10 +64,16 @@ def python_type(self):
...
     def from_param(cls, param):
-        return 1 if bool(param) else 0
+        return True if bool(param) else False
 
     def python_type(self):
-        return 'integer'
+        return 'bool'
+
+    def __new__(cls, val=0):
+        if val:
```
marcoskirsch: You should rewrite this in one line like in 67, it's odd to use two styles within a few lines of code.

build/helper.py
```diff
@@ -711,7 +717,11 @@ def get_python_type_from_visa_type(visa_type):
     else:
         p_types = importlib.import_module('build.templates.python_types')
     v_type = getattr(p_types, visa_type)
-    p_type = v_type().python_type()
+    # We have a special case for bool
```
marcoskirsch: This is unfortunate...
We have python_types.py so that we can centralize type information... then there is this one special case.
I'm not sure what the best way to solve it is though.

generated/nidmm/session.py
```diff
@@ -2327,7 +2328,7 @@ def get_measurement_period(self):
         period_ctype = ctypes_types.ViReal64_ctype(0)
         error_code = self.library.niDMM_GetMeasurementPeriod(self.vi, ctypes.pointer(period_ctype))
         errors._handle_error(self, error_code)
-        return period_ctype.value
+        return python_types.ViReal64(period_ctype.value)
```
marcoskirsch: Were these changes intentional? I don't fully understand the implications.
We are now returning an object of a different type than we used to, and in the case of booleans we aren't really returning True/False.

To clients, is this?
* Transparent
* Helpful
* Detrimental
And how?
(mostly I want to understand it)

generated/nidmm/session.py
```diff
@@ -2852,7 +2853,7 @@ def read_multi_point(self, maximum_time, array_size):
         actual_number_of_points_ctype = ctypes_types.ViInt32_ctype(0)
         error_code = self.library.niDMM_ReadMultiPoint(self.vi, maximum_time, array_size, ctypes.cast(reading_array_ctype, ctypes.POINTER(ctypes_types.ViReal64_ctype)), ctypes.pointer(actual_number_of_points_ctype))
         errors._handle_error(self, error_code)
-        return [reading_array_ctype[i].value for i in range(array_size)], actual_number_of_points_ctype.value
+        return [python_types.ViReal64(reading_array_ctype[i].value) for i in range(array_size)], python_types.ViInt32(actual_number_of_points_ctype.value)
```
marcoskirsch: If we determine at some point that we need to address performance concerns with creating this list; and want to return the ctypes object directly because it has the right interface, or want to return a numtypes object because it's more performant... We most likely won't be able to return ViReal64s anymore. Is that ok?

texasaggie97:
I rethought how to solve this.
Now the functions return the intrinsic type instead of the Visa type. I still instantiate the Visa type (ViReal64) but they now return an instance of the underlying intrinsic type on new()

generated/nidmm/session.py
```diff
@@ -986,7 +986,7 @@ def configure_frequency_voltage_range(self, voltage_range):
         measurements.
 
         Args:
-            voltage_range (ViReal64):
+            voltage_range (float):
```
marcoskirsch: I like this better. The Python developer shouldn't need to know or care about ViReal64.

generated/nidmm/session.py
```diff
@@ -1481,7 +1481,7 @@ def configure_thermocouple(self, thermocouple_type, reference_junction_type):
                 +---------------------+---------------------+
                 | NIDMM_VAL_TEMP_TC_T | Thermocouple type T |
                 +---------------------+---------------------+
-            reference_junction_type (ViInt32):
+            reference_junction_type (int):
```
marcoskirsch: So this should be the enum, right?
It was wrong before too.

build/helper.py
```diff
@@ -98,6 +98,14 @@ def _add_python_type(parameter):
         parameter['python_type'] = 'enums.' + parameter['enum']
     return parameter
 
+def _add_intrinsic_type(parameter):
+    '''Adds a intrinsic key/value pair to the parameter metadata'''
```
marcoskirsch: can you expand on what intrinsic type is?

build/helper.py
```diff
@@ -711,7 +726,19 @@ def get_python_type_from_visa_type(visa_type):
     else:
```
marcoskirsch: Please add docstring to this function.

build/helper.py
```diff
@@ -711,7 +726,19 @@ def get_python_type_from_visa_type(visa_type):
...
         p_types = importlib.import_module('build.templates.python_types')
     v_type = getattr(p_types, visa_type)
-    p_type = v_type().python_type()
+    # If we just take type() and make it a string, we get <class 'float'> which is not what we want
```
marcoskirsch: Use type(x).__name__

I don't think the if/elses pass the "pythonic" smell test

build/helper.py
texasaggie97: Added

build/helper.py
texasaggie97: Added

build/helper.py
texasaggie97: Changed. Actually __name__

Also renamed function to be explicit about returning the underlying intrinsic (basic python) type

generated/nidmm/session.py
texasaggie97: Added enum information to functions_addon.py

build/helper.py
```diff
@@ -231,14 +240,20 @@ def get_library_call_parameter_types_snippet(parameters_list):
 def _get_output_param_return_snippet(output_parameter, parameters):
     '''Returns the snippet for returning a single output parameter from a Session method, i.e. "reading_ctype.value"'''
     assert output_parameter['direction'] == 'out', pp.pformat(output_parameter)
+    return_type_snippet = ''
+    if output_parameter['enum'] is not None:
+        return_type_snippet = 'enums.' + output_parameter['enum'] + '('
```
marcoskirsch: this is fine... but if we find ourselves appending enums. and python_types. all over the place, then perhaps we should store the appended string in the metadata.

PR #139, opened 2017-08-18
Add system tests

bhaswath:
Adding two system tests

Adds a test for checking correct error is thrown when invalid value is passed to a property
Adds a test for verifying correct operation of a function that returns a Int32 datatype

Increases coverage of the systemtests

Has been run on a local machine and is passing


marcoskirsch:
If you look at the diff, it appears as if the whole file changed.
My guess is that that's not the case and something like line endings was changed. Can you revert that the way it was?

PR #149, opened 2017-08-22
New/ni fake

marcoskirsch:

Adds a new fake MI driver "NI-FAKE" to the set of Python bindings generated by nimi-python. The goal is to use this fake driver for testing the code-generator and unit-testing the generated code. As such, NI-FAKE should contain one but no more than one example for every type of attribute/function and unit-tests that verify that the Python code is calling into the mock library correctly.

Neither the metadata nor the unit tests have been fully fleshed out.

This is a refinement of our testing strategy. There's no point writing unit tests for the real bindings (such as NI-DMM, NI-SCOPE, etc.) since that wouldn't validate the metadata anyway (we're mocking the calls to the library). Validating the metadata requires system tests.
Also, by attempting to write unit tests for the individual bindings we will either risk lots of repetition which is bad or missing coverage which is worse.

By having a concise "for unit testing" driver that has one example for every thing the code-generator needs to handle, we can ensure that we have good unit-test coverage. By not having redundancy (i.e. multiple attributes of the same type) the use of code coverage tools becomes viable.

Once this is merged and things have been fleshed out, we can remove the unit-tests for the individual MI drivers.

Everything builds correctly.
Upon inspection, things look correct.
Some unit tests were added (derived from NI-DMM unit tests actually) and run correctly.


src/nifake/metadata/attributes.py
```diff
@@ -0,0 +1,59 @@
...
+
+attributes = {
+    1000000: {
+        'access': 'read-write',
+        'channel_based': 'False',
+        'enum': None,
+        'lv_property': 'Fake attributes:Read Write Bool',
+        'name': 'READ_WRITE_BOOL',
+        'resettable': 'No',
+        'short_description': '''An attribute of type bool with read/write access.''',
```
texasaggie97: This is not how documentation is formatted now. It should be something like:
```
    'documentation': {
        'description': 'An attribute of type bool with read/write access.',
    },
```

build/defines.mak
```diff
@@ -21,6 +21,7 @@ WHEEL := $(OUTPUT_DIR)/dist/$(DRIVER)-$(VERSION)-py2.py3-none-any.whl
 SDIST := $(OUTPUT_DIR)/dist/$(DRIVER)-$(VERSION).tar.gz
 
 MKDIRECTORIES += \
+                 $(DRIVER_DOCS_DIR) \
```
marcoskirsch: It exercises the documentation code generation and templates. I lean towards leaving it in.

src/nifake/metadata/attributes.py
marcoskirsch: Will fix. Does this apply only to attributes.py?

texasaggie97: It applies to enums as well.

marcoskirsch: Done.

Issue #154, opened 2017-08-24
We should use niXXX_error_message, not niXXX_GetErrorMessage


marcoskirsch:
When retrieving errors from the driver in session.py, [we call niXXX_GetErrorMessage](https://github.com/ni/nimi-python/blob/master/build/templates/session.py.mako#L208) (where niXXX is replaced at code-generation with the correct driver prefix).

This works for NI-DMM, but turns out that not all MI drivers support this function - it is an NI-specific extension and not part of the IVI specification. For example, [NI-SWITCH](http://zone.ni.com/reference/en-XX/help/375472A-01/TOC87.htm) doesn't have it.

What all drivers do have is niXXX_error_message, which is equivalent but limits the buffer size for the string to 256 characters (no IVI-dance!), which can lead to truncation if the error message is longer. Using this function for the Python bindings is probably acceptable.

The alternatives are:

- Code generate this part of the bindings differently for each driver (added complexity)
- File NI-internal Corrective Action Requests to drivers that don't support niXXX_GetErrorMessage, wait for them to implement it and ship, then bindings won't work on older drivers. This is likely unacceptable.

PR #160, opened 2017-08-24
Fix error handling

texasaggie97:
* Fix how error handling is done
  * Use generated functions
  * Only mock error handling functions when required
  * Return mock return value immediately when set

* Allows writing unit tests properly

* Travis & system tests


marcoskirsch:
Is there an issue for this?

build/templates/session.py.mako
```diff
@@ -180,38 +180,12 @@ context_name = 'acquisition' if c['direction'] == 'input' else 'generation'
...
-            '''
-            error_code = buffer_size
-            buffer_size = self.library.${c_function_prefix}GetErrorMessage(self.vi, error_code, 0, None)
-            error_message = (ctypes_types.ViChar_ctype * buffer_size)()
-            self.library.${c_function_prefix}GetErrorMessage(self.vi, error_code, buffer_size, ctypes.cast(error_message, ctypes.POINTER(ctypes_types.ViChar_ctype)))
-
-        # TODO(marcoskirsch): By hardcoding encoding "ascii", internationalized strings will throw.
-        #       Which encoding should we be using? https://docs.python.org/3/library/codecs.html#standard-encodings
-        return new_error_code.value, error_message.value.decode("ascii")
+        try:
```
marcoskirsch: Fixes https://github.com/ni/nimi-python/issues/163

build/templates/session.py.mako
```diff
@@ -180,38 +180,12 @@ context_name = 'acquisition' if c['direction'] == 'input' else 'generation'
...
+            (new_error_code, new_error_string) = self._get_error()
+            return new_error_code, new_error_string
+        except errors.Error:
+            (new_error_code, new_error_string) = self._get_error_message(error_code)
```
marcoskirsch: Fixes https://github.com/ni/nimi-python/issues/154

build/templates/session.py.mako
```diff
@@ -180,38 +180,12 @@ context_name = 'acquisition' if c['direction'] == 'input' else 'generation'
...
-            error_message = (ctypes_types.ViChar_ctype * buffer_size)()
-            self.library.${c_function_prefix}GetErrorMessage(self.vi, error_code, buffer_size, ctypes.cast(error_message, ctypes.POINTER(ctypes_types.ViChar_ctype)))
-
-        # TODO(marcoskirsch): By hardcoding encoding "ascii", internationalized strings will throw.
-        #       Which encoding should we be using? https://docs.python.org/3/library/codecs.html#standard-encodings
-        return new_error_code.value, error_message.value.decode("ascii")
+        try:
+            (new_error_code, new_error_string) = self._get_error()
+            return new_error_code, new_error_string
+        except errors.Error:
```
marcoskirsch: I think this deserves some of the explanation, maybe a reduced version of what we had before.

build/templates/session.py.mako
```diff
@@ -245,7 +219,7 @@ context_name = 'acquisition' if c['direction'] == 'input' else 'generation'
         error_code = self.library.${c_function_prefix}${func_name}(${helper.get_library_call_parameter_snippet(f['parameters'])})
         # Don't use _handle_error, because positive value in error_code means size, not warning.
         if (errors._is_error(error_code)):
-            raise errors.Error(self.library, self.vi, error_code)
+            raise errors.Error(self, error_code)
```
marcoskirsch: Wasn't there an issue for this, or didn't @Fladolcetta fix this separately?

generated/nidmm/tests/test_session.py
```diff
@@ -75,18 +88,29 @@ def test_multiple_return_params(self):
...
+    def test_get_string_attribute_with_error(self):
+        self.patched_ctypes_library.niDMM_GetAttributeViString.side_effect = self.side_effects_helper.niDMM_GetAttributeViString
+        self.side_effects_helper['GetAttributeViString']['attributeValue'] = 'Testing is fun?'
+        self.side_effects_helper['GetAttributeViString']['return'] = -1234
+        self.patched_ctypes_library.niDMM_GetError.side_effect = self.side_effects_helper.niDMM_GetError
+        self.side_effects_helper['GetError']['errorCode'] = -1234
+        self.side_effects_helper['GetError']['description'] = "ascending order"
+        with nidmm.Session('dev1') as session:
+            try:
+                session._get_attribute_vi_string("", 5)
```
marcoskirsch: I don't think we need to have unit tests for private methods (although I'm not sure about that).
We should test this using the public API.

marcoskirsch:
It seems like this PR can be broken up in several smaller PRs. It fixes a bunch of issues (not noted). I think it'd be best to break it up.

Also: I see no need to keep improving NI-DMM unit tests. Work should go into NI-FAKE at this point with the goal of deleting NI-DMM unit tests in the future.

build/templates/session.py.mako
texasaggie97: Added

build/templates/session.py.mako
texasaggie97: He had a fix, but since I needed to fix it as part of fixing error handling in general, I said I would put the fix in here.

I don't see an open issue for it though

texasaggie97:
All these fixes are needed to be able to unit test error handling.

New tests moved to niFake

_get_error_message fall back does not work. Will be addressed in a separate PR since it is more complicated and not required for the usual situation.

marcoskirsch:
> _get_error_message fall back does not work. Will be addressed in a separate PR since it is more complicated and not required for the usual situation.

Should open en issue for this.

src/nifake/metadata/functions.py
```diff
@@ -1041,6 +1041,32 @@
             'note': 'Splinter is not supported.',
         },
     },
+    'reset': {
```
marcoskirsch: I added simpleFunction function with this exact signature. We should use that for the tests - no need to have a separate reset() function.

texasaggie97: Using simple_function()

Fladolcetta:
Fixes #163

PR #190, opened 2017-08-28
Add cmd line options to system tests

### LINKED ISSUE #99: Update systemTest to accept device name at runtime
As of now SystemTest will be running on first device found by niModInst. There should be an easier way to specify the device name while running system tests.

Fladolcetta:
[ ] This contribution adheres to [CONTRIBUTING.md]

Added command line option for device name in NI-DMM 
Added command line option for ModInst to select device family

fixes #99 

Ran system tests


marcoskirsch:
Don't merge this until we make a decision on https://github.com/ni/nimi-python/issues/192 which if approved would mean we discard this PR.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -7,16 +7,20 @@
...
-    device_name = "unknown"
+    device_name = name
     device_sn = "unknown"
     try:
         with nimodinst.Session('nidmm') as session:
             if len(session) > 0:
-                device_name = session.device_name[0]
-                device_sn = session.serial_number[0]
-
+                for x in range(0, len(session)):
```
texasaggie97: This is where having some static documentation would help. Right now the only way to know this is by looking at the example.

You can write this loop as `for device in session:`

Then you just use `device_name = device.device_name`

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -7,16 +7,20 @@
...
+                    if (session.device_name[x] == device_name):
+                        device_sn = session.serial_number[x]
+                        break
+                if (device_sn == "unknown"):
```
texasaggie97: Instead of checking after the fact to see whether to use information from modinst or not instead do something like this:
```
with nimodinst.Session('nidmm') as session:
    device_name = None
    if len(session) > 0:
        device_name = session.device_name[0]
    for device in session:
        if device.device_name == name
            device_name = name
    assert device_name is not None, "No NI-DMM HW installed"
```

src/nimodinst/system_tests/test_system_nimodinst.py
```diff
@@ -4,28 +4,29 @@
...
 
-def test_bad_device_family():
+def test_bad_device_family(family):
     with nimodinst.Session('FAKE') as session:
         assert len(session) == 0
 
 
-def test_no_device_family():
-    with nimodinst.Session('') as session:
+def test_no_device_family(family):
```
texasaggie97: This doesn't seem to do what the name says. It says no device family but then opens the session using the family

src/nimodinst/system_tests/test_system_nimodinst.py
```diff
@@ -4,28 +4,29 @@
...
+    with nimodinst.Session(family) as session:
         assert len(session) > 0, 'Must have hardware for ModInst tests to be valid.'
 
 
-def test_device_family_string_with_dashes():
-    with nimodinst.Session('NI-DMM') as session:
-        assert len(session) > 0, 'Must have NI-DMM hardware for this tests to be valid.'
+def test_device_family_string_with_dashes(family):
```
texasaggie97: Since you can take the family from the command line, you should verify there isn't already a dash.

src/nimodinst/system_tests/test_system_nimodinst.py
```diff
@@ -4,28 +4,29 @@
...
+    modified_family = family.upper()[:2] + '-' + family.upper()[2:]  # Changing nidmm to NI-DMM
+    with nimodinst.Session(modified_family) as session:
+        assert len(session) > 0, 'Must have hardware for this tests to be valid.'
 
 
-def test_device_family_string_without_dashes():
-    with nimodinst.Session('nidmm') as session:
-        assert len(session) > 0, 'Must have NI-DMM hardware for ModInst tests to be valid.'
+def test_device_family_string_without_dashes(family):
+    with nimodinst.Session(family) as session:
```
texasaggie97: Similar here, since you can take the family on the command line, make sure there is no dash.

Fladolcetta:
Deleting as this is no longer needed with the option to use simulated devices

PR #193, opened 2017-08-29
Change how tox is used

texasaggie97:
* Uses tox for pytest, flake8 & sphinx
  * Now these dependencies are no longer required to be installed on the build machine
  * All three are ran as part of the test target
  * Runs after all drivers are complete
* Only unit test nifake
* Run flake8 on all drivers at the same time
* Fix rerunning tests when not required
  * Use tracking file similar to other things
  * Item builds that can change testing results trigger the tests to be ran by deleting the tracking file if it exists
* Moves common functions into tools.mak
  * Allows use in multiple places without multiple defines
  * There was already a discrepancy between two versions of one of them

* Improve overall build times

* Travis

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -13,11 +13,6 @@ export COMMAND_LOG_BATCH
 COMMAND_LOG_SH := $(GENERATED_DIR)/command_history.sh
 export COMMAND_LOG_SH
 
-define log_command
-	@echo '$1' >> $(COMMAND_LOG_BATCH)
-	@echo '$1' >> $(COMMAND_LOG_SH)
-	$1
-endef
 
 ifeq (,$(PRINT))
 _hide_cmds := @
--- a/build/Makefile
+++ b/build/Makefile
@@ -3,9 +3,6 @@ include $(BUILD_HELPER_DIR)/tools.mak
 DEFAULT_TARGETS := module unit_tests update_generated_files update_system_tests update_examples wheel sdist
 export DEFAULT_TARGETS
 
-ADDITIONAL_TARGETS := 
-export ADDITIONAL_TARGETS
-
 GENERATED_DIR := $(ROOT_DIR)/generated
 export GENERATED_DIR
 
@@ -38,7 +35,7 @@ endef
 
 $(foreach d,$(ALL_DRIVERS),$(eval $(call per_driver_all,$(d))))
 $(foreach d,$(ALL_DRIVERS),\
-   $(foreach t,$(DEFAULT_TARGETS) $(ADDITIONAL_TARGETS),\
+   $(foreach t,$(DEFAULT_TARGETS),\
       $(eval $(call per_driver_per_target,$(d),$(t)))))
 
 define per_target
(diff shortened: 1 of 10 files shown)


build/Makefile
```diff
@@ -1,8 +1,9 @@
+include $(BUILD_HELPER_DIR)/tools.mak
 
-DEFAULT_TARGETS := module unit_tests update_generated_files update_system_tests update_examples run_unit_tests wheel sdist
+DEFAULT_TARGETS := module unit_tests update_generated_files update_system_tests update_examples wheel sdist
 export DEFAULT_TARGETS
 
-ADDITIONAL_TARGETS := test flake8
+ADDITIONAL_TARGETS :=
```
marcoskirsch: Can we get rid of this? It doesn't appear to be in use.

build/Makefile
```diff
@@ -102,20 +97,19 @@ printvar:
...
-SOURCEDIR     = $(ROOT_DIR)/docs
-BUILDDIR      = $(BIN_DIR)/docs
+.PHONY: help Makefile
 
-html:
-	$(_hide_cmds)$(call log_command,$(SPHINXBUILD) -M html "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS))
 
-.PHONY: help Makefile
+# Any step that any driver build does that would invalidate unit testing, flake8 or generated html
+# needs to delete this file. This will trigger a tox run.
```
marcoskirsch: Isn't the normal way to do this with GNU Make is to add anything that invalidates these things be a dependency of $(TOX_RUN_DONE) ?

build/rules.mak
```diff
@@ -46,84 +22,77 @@ $(foreach d,$(MKDIRECTORIES),$(eval $(call mkdir_rule,$(d))))
...
 	$(call trace_to_console, "Updating",$(GENERATED_DIR)/$(DRIVER)/)
-	$(_hide_cmds)$(call log_command,touch $@)
-	$(_hide_cmds)$(call log_command,rm $@)
-	$(_hide_cmds)$(call log_command,rm -Rf $(GENERATED_DIR)/$(DRIVER))
-	$(_hide_cmds)$(call log_command,mkdir -p $(GENERATED_DIR)/$(DRIVER))
-	$(_hide_cmds)$(call log_command,cp -Rf $(MODULE_DIR)/* $(GENERATED_DIR)/$(DRIVER))
-	$(_hide_cmds)$(call log_command,cp -Rf $(OUTPUT_DIR)/setup.py $(GENERATED_DIR)/$(DRIVER))
-	$(_hide_cmds)$(call log_command,touch $@)
+	$(_hide_cmds)$(call make_with_tracking_file, $@, \
+      rm -Rf $(GENERATED_DIR)/$(DRIVER) && \
```
marcoskirsch: this is pre-existing... but

if you rm dir/* instead, then the folder is emptied but you don't need to mkdir it right after...

build/tools.mak
```diff
@@ -0,0 +1,34 @@
...
+# $1 is tracking file path
+# $2 is command to run
+define make_with_tracking_file =
+	$(_hide_cmds)$(call log_command,touch $1)
+	$(_hide_cmds)$(call log_command,rm $1)
+	$(_hide_cmds)$(call log_command,$2)
+	$(_hide_cmds)$(call log_command,touch $1)
+endef
+
+# Delete the file that will trigger tox to run
```
marcoskirsch: Delete the file whose absence will trigger tax to run

build/Makefile
texasaggie97: Removed

build/rules.mak
texasaggie97: Changed. Also had to add to MKDIRS to ensure it existed

build/tools.mak
texasaggie97: Changed

build/Makefile
texasaggie97: Yes, that would be the normal way. In order to do that, we would need to somehow dynamically generate that from the driver builds so that the list would be available here.

PR #209, opened 2017-08-31
Added system test for initializing with all non default values

bhaswath:
[ X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/<reponame>/blob/master/CONTRIBUTING.md).

Adds test that initializes a session with all non-default parameters

To increase coverage of system tests

Ran on local machine, all tests passing


src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -178,3 +178,8 @@ def test_vi_int32_output_function(session):
     assert isinstance(backlog, int)
     assert backlog == 0
 
+
+def test_initialize_nondefault():
+    with nidmm.Session('FakeDevice', False, False, 'Simulate = 1,RangeCheck=0,QueryInstrStatus=1,Cache=0,DriverSetup=Model:4080; BoardType:PXIe') as session:
+        assert session.simulate is True
+
```
Fladolcetta: This seems like it could be a unit test right? The defaults are set by __init__ in generated code. Additionally, It seems that False, False is the default for the python code. (excluding resource name and option string)

    def __init__(self, resource_name, id_query=0, reset_device=False, options_string=""):
        self.library = library_singleton.LibrarySingleton.get()

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -178,3 +178,8 @@ def test_vi_int32_output_function(session):
     assert isinstance(backlog, int)
     assert backlog == 0
 
+
+def test_initialize_nondefault():
+    with nidmm.Session('FakeDevice', False, False, 'Simulate = 1,RangeCheck=0,QueryInstrStatus=1,Cache=0,DriverSetup=Model:4080; BoardType:PXIe') as session:
+        assert session.simulate is True
```
Fladolcetta: Shouldnt we also assert on the other items (false, false, device name, etc)

bhaswath: I don't think we can assert whether Reset_device is true or false
And for Simulated devices, ID Query will not be returned, an exception will be thrown instead (can add a system test for this if needed?), so we cant assert that either.
Another system test already checks device name assertion so it can be omitted.

src/nidmm/system_tests/test_system_nidmm.py
bhaswath: I think it is possible.
Unit tests would mean that this would go into the niFake folder right?
I noticed that session.simulate etc were not available under niFake

texasaggie97:
Can one of the admins verify this patch?

bhaswath:
Closing to make this into a unit test for NI Fake.

PR #213, opened 2017-08-31
Move rst files to generated folder hierarchy

texasaggie97:
* Moves generated rst files into generated/docs
  * All files generated as part of the build now in generated
  * References in toctree cannot reference out of own hierarchy, so reference *._link.rst that use rst `.. include::` directive to pull in the generated files

* Keeps generated files in single location

* Travis & [ReadTheDocs](http://nimi-python.readthedocs.io/en/build-move_rst_to_generated/)


marcoskirsch:
After discussing with @texasaggie97 , we aren't convinced that overall, this is better than just living with generated files in docs.

We're going to close, may revisit at some point in the future.

PR #220, opened 2017-09-01
Bug/bug217

### LINKED ISSUE #217: Remove NI-DMM Unit Tests and add them to NI-FAKE
We currently have unit tests for NI-DMM. Given that we are using NI-FAKE as the primary unit test "driver" we should move these tests to NI-FAKE and remove duplicates.

Fladolcetta:
[x ] This contribution adheres to [CONTRIBUTING.md]

Fixes #217 
Removes NI-DMM unit tests

It reduces redundency in unit testing and has a single central place for unit testing to be performed. 

Ran Unit tests
Ran System Tests for DMM and ModInst


src/nimodinst/tests/test_modinst.py
```diff
@@ -1,74 +0,0 @@
-
```
texasaggie97: Don't delete the ModInst unit tests. I filed issue #221 because we need to re-enable running these tests.

PR #231, opened 2017-09-02
Don't raise exception when driver returns warning

### LINKED ISSUE #198: Warnings returned by the driver should not raise
If the underlying driver runtime returns a warning, as it stands today, the Python binding raises.
One example of this is when a DMM reads over-range. It is undesirable to change flow control of the program in these scenarios.

[nidaqmx](https://github.com/ni/nidaqmx-python) handles warnings by logging them to file. We should look into their approach and deal with warnings in the same manner.

We should also log when driver's close function errors.

texasaggie97:
* Fixes #198 
* Renames <driver>.Warning exception class to <Driver>Warning.
  * This is to disambiguate from built in Warning exception class
* <Driver>Warning now inherits from (warnings.Warning) instead of _ErrorBase
* When warning occurs, log to warning mechanism
* Add unit test for warnings
* Still need to update CHANGELOG.md once merged into master

* Fixes issue
* More correct behavior

* Travis

### DIFF
--- a/build/templates/errors.py.mako
+++ b/build/templates/errors.py.mako
@@ -15,41 +15,41 @@ import platform
 import warnings
 
 
-def _is_success(error_code):
-    return (error_code == 0)
+def _is_success(code):
+    return (code == 0)
 
 
-def _is_error(error_code):
-    return (error_code < 0)
+def _is_error(code):
+    return (code < 0)
 
 
-def _is_warning(error_code):
-    return (error_code > 0)
+def _is_warning(code):
+    return (code > 0)
 
 
 class _ErrorBase(Exception):
 
-    def __init__(self, error_code, error_description):
+    def __init__(self, code, description):
 
-        self.code = error_code
-        self.description = error_description
+        self.code = code
+        self.description = description
         super(_ErrorBase, self).__init__(str(self.code) + ": " + self.description)
 
 
 class Error(_ErrorBase):
     '''An error originating from the ${driver_name} driver'''
 
-    def __init__(self, error_code, error_description):
-        assert (_is_error(error_code)), "Should not raise Error if error_code is not fatal."
-        super(Error, self).__init__(error_code, error_description)
+    def __init__(self, code, description):
+        assert (_is_error(code)), "Should not raise Error if code is not fatal."
+        super(Error, self).__init__(code, description)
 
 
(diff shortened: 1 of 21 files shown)


build/templates/errors.py.mako
```diff
@@ -42,12 +44,12 @@ class Error(_ErrorBase):
         super(Error, self).__init__(error_code, error_description)
 
 
-class Warning(_ErrorBase):
+class ${module_name_class}Warning(Warning):
```
marcoskirsch: Shame to rename class, since it's already scoped inside the nidmm (or whatever) module. It's inconsistent with Error and redundant. I suppose there was no way around it that you could find?

build/templates/errors.py.mako
```diff
@@ -42,12 +44,12 @@ class Error(_ErrorBase):
...
     '''A warning originating from the ${driver_name} driver'''
 
-    def __init__(self, error_code, error_description):
-        assert (_is_warning(error_code)), "Should not raise Warning if error_code is not positive."
-        super(Warning, self).__init__(error_code, error_description)
+    def __init__(self, warning_code, warning_description):
```
marcoskirsch: Alternatively, we can remove warning_ and error_ prefixes from all. More consistent, and we are within the context of Warning or Error class anyway so it's not ambiguous.

I think it's cleaner.

build/templates/errors.py.mako
```diff
@@ -42,12 +44,12 @@ class Error(_ErrorBase):
...
+        assert (_is_warning(warning_code)), "Should not raise Warning if error_code is not positive."
```
marcoskirsch: Reword, since we aren't going to raise these I though.

build/templates/errors.py.mako
texasaggie97: Not that I could find.

build/templates/errors.py.mako
texasaggie97: Remove all 'error_' and 'warning_' prefixes

build/templates/errors.py.mako
texasaggie97: Reworded

marcoskirsch:
I've opened https://github.com/ni/nimi-python/issues/253 as a followup to this PR

PR #264, opened 2017-09-07
Added Unit test for initializing with all non default parameters

bhaswath:

Adds a unit test that checks Initializing with non default parameters

It increases areas being tested

Ran on a local machine and it passes

### DIFF
--- a/src/nifake/tests/test_session.py
+++ b/src/nifake/tests/test_session.py
@@ -333,3 +333,29 @@ class TestSession(object):
         session = nifake.Session('FakeDevice', 1, True, 'Some string')
         assert(session.vi == SESSION_NUM_FOR_TEST)
         self.patched_library.niFake_InitWithOptions.assert_called_once_with(b'FakeDevice', 1, True, ANY, ANY)
+
+    def test_read(self):
+        test_maximum_time = 1000
+        test_reading = 5
+        self.patched_library.niFake_Read.side_effect = self.side_effects_helper.niFake_Read
+        self.side_effects_helper['Read']['reading'] = test_reading
+        with nifake.Session('dev1') as session:
+            session.read(test_maximum_time)
+            from mock import call
+            calls = [call(SESSION_NUM_FOR_TEST, test_maximum_time, ANY)]
+            self.patched_library.niFake_Read.assert_has_calls(calls)
+            assert self.patched_library.niFake_Read.call_count == 1
+
+    def test_read_multipoint(self):
+        test_maximum_time = 1000
+        test_reading_array = 1.2
+        test_actual_number_of_points = test_array_size = 2
+        self.patched_library.niFake_ReadMultiPoint.side_effect = self.side_effects_helper.niFake_ReadMultiPoint
+        self.side_effects_helper['ReadMultiPoint']['readingArray'] = test_reading_array
+        self.side_effects_helper['ReadMultiPoint']['actualNumberOfPoints'] = test_actual_number_of_points
+        with nifake.Session('dev1') as session:
+            session.read_multi_point(test_maximum_time, test_array_size)
+            from mock import call
+            calls = [call(SESSION_NUM_FOR_TEST, test_maximum_time, test_array_size, ANY, ANY)]
+            self.patched_library.niFake_ReadMultiPoint.assert_has_calls(calls)
+            assert self.patched_library.niFake_ReadMultiPoint.call_count == 1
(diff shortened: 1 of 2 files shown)


texasaggie97:
Can one of the admins verify this patch?

texasaggie97:
add to whitelist

texasaggie97:
test this please

src/nifake/tests/test_session.py
```diff
@@ -328,3 +328,34 @@ def test_multiple_return_params(self):
             assert self.patched_errors.handle_error.call_count == 2
             self.patched_errors.handle_error.assert_called_with(session, self.patched_library.niFake_GetCalDateAndTime.return_value)
     '''
+
+    def test_init_with_options_nondefault(self):
+        session = nifake.Session('FakeDevice', 1, True, 'Some string')
+        assert(session.vi == SESSION_NUM_FOR_TEST)
+        self.patched_library.niFake_InitWithOptions.assert_called_once_with(b'FakeDevice', 1, True, ANY, ANY)
```
marcoskirsch: Don't use ANY, use the parameters values you expect. Otherwise we aren't ensuring the DLL gets called correctly.

src/nifake/tests/test_session.py
```diff
@@ -328,3 +328,34 @@ def test_multiple_return_params(self):
...
+
+    def test_read(self):
+        test_maximum_time = 1000
+        test_reading = 5
+        self.patched_library.niFake_Read.side_effect = self.side_effects_helper.niFake_Read
+        self.side_effects_helper['Read']['reading'] = test_reading
+        with nifake.Session('dev1') as session:
+            session.read(test_maximum_time)
```
marcoskirsch: Check the returned value from the read, that it matches what you wanted.

src/nifake/tests/test_session.py
```diff
@@ -328,3 +328,34 @@ def test_multiple_return_params(self):
...
+
+    def test_read_multipoint(self):
+        test_maximum_time = 1000
+        test_reading_array = 1.2
+        test_actual_number_of_points = test_array_size = 2
+        self.patched_library.niFake_ReadMultiPoint.side_effect = self.side_effects_helper.niFake_ReadMultiPoint
+        self.side_effects_helper['ReadMultiPoint']['readingArray'] = test_reading_array
+        self.side_effects_helper['ReadMultiPoint']['actualNumberOfPoints'] = test_actual_number_of_points
+        with nifake.Session('dev1') as session:
+            session.read_multi_point(test_maximum_time, test_array_size)
```
marcoskirsch: assert returned data is correct

src/nifake/tests/test_session.py
```diff
@@ -328,3 +328,34 @@ def test_multiple_return_params(self):
...
+        with nifake.Session('dev1') as session:
+            session.read(test_maximum_time)
+            from mock import call
+            calls = [call(SESSION_NUM_FOR_TEST, test_maximum_time, ANY)]
+            self.patched_library.niFake_Read.assert_has_calls(calls)
+            assert self.patched_library.niFake_Read.call_count == 1
+
+    def test_read_multipoint(self):
+        test_maximum_time = 1000
+        test_reading_array = 1.2
```
marcoskirsch: shouldn't this be a list?

src/nifake/tests/test_session.py
```diff
@@ -328,3 +328,34 @@ def test_multiple_return_params(self):
...
+        test_actual_number_of_points = test_array_size = 2
```
marcoskirsch: get rid of this, use len(test_reading_array) elsewhere

src/nifake/tests/test_session.py
bhaswath: I initially tried it as a list, but I get a typeError saying it should be a float.
In session.py, it is mentioned that this is a float.

Should it actually be a list instead?

src/nifake/tests/test_session.py
```diff
@@ -328,3 +328,34 @@ def test_multiple_return_params(self):
             assert self.patched_errors.handle_error.call_count == 2
             self.patched_errors.handle_error.assert_called_with(session, self.patched_library.niFake_GetCalDateAndTime.return_value)
     '''
+
```
texasaggie97: When there are changes to a test file, either system tests or unit tests, there should be a corresponding change in the generated files.

When you don't have the changed generated file as part of your change, it ends up being an unrelated change in a different pull request.

PR #269, opened 2017-09-07
Use shared implementation for merge_dicts

### LINKED ISSUE #255: Move metadata processing to build module
Right now, each Mako template is responsible for taking the raw metadata as found in source and processing it (merging with add-on metadata, adding missing defaults by calling helper.py functions, etc).

We should improve this. Have all Mako templates receive the metadata fully processed and ready to be consumed. This means some templates will receive more metadata than they need, but the performance/memory hit is not important because it happens during code generation.

Move all the metadata processing functions out of helper.py into its own file (metadata.py? metadata_processing.py?). 

Add unit tests for these functions.

texasaggie97:
* Fixes #255 
* Create shared implementation of merge_dicts used in all metadata modules
* Use new shared implementation

* Reduces redundancy

* Travis & system tests

### DIFF
--- a/build/metadata_utilities.py
+++ b/build/metadata_utilities.py
@@ -1,0 +1,28 @@
+# Useful functions for use in the metadata modules
+
+import re
+def merge_dicts(into, outof):
+    for item in sorted(outof):
+        if type(outof[item]) is dict:
+            if item in into and type(into[item] is list):
+                merge_dicts(into[item], outof[item])
+            elif item in into:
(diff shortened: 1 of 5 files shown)


build/metadata_utilities.py
```diff
@@ -0,0 +1,28 @@
+# Useful functions for use in the metadata modules
+
+import re
+def merge_dicts(into, outof):
```
marcoskirsch: add docstring

Issue #205, opened 2017-08-30
Getting a error when calling GetError causes a recursion failure


Fladolcetta:
If you are calling GetError and this returns an error. This throws a recusion failure.

marcoskirsch:
If I fix https://github.com/ni/nimi-python/issues/166 I would be introducing this problem to NI-ModInst.
Should I tackle it or should I hold off? @texasaggie97

texasaggie97:
I think it is fine to go ahead and fix #166. The reason I am okay with that is the only time we have hit this recursion failure so far is when we try to write a unit test to test that condition. I.e. it isn't very common.

PR #272, opened 2017-09-07
Simpler metadata merge

### LINKED ISSUE #271: Allow adding to, not just replacing into metadata using function_addons.py
As implemented, metadata_utilities.merge_dicts() ignores (read: does not merge) keys in the dictionary that don't already exist in the dictionary being merged into.

In practical terms, this means that our code-generated function metadata must have all keys that we care about.

For example: functions.py marks *every* function as 'codegen_method': 'public'.
Then we go and replace some of those with 'private' or 'no'.

This is silly, the generator should decide those defaults, rather than doing it by hardcoding it in the functions.py metadata.

This is a bigger problem as we add more metadata. For example, as part of the solution to https://github.com/ni/nimi-python/issues/205 I want function metadata to specify for each function if it is used in the context of error reporting. But as it is, if metadata contains info only for the small subset where the value is True, it doesn't get merged because the key didn't exist in the first place in the source metadata.

marcoskirsch:

* Simpler code
* Fixes #271 
* Adds tests for merge_dict function
* Adds make dependency so when metadata_utilities.py changes, files are rebuilt.
* Prerequisite for intended solution to #205 

It makes code better

Building project yields no changes.
Unit tests pass


build/test_metadata_utilities.py
```diff
@@ -0,0 +1,54 @@
...
+    expected = {'a':1, 'b':{'b1':5, 'b2':6, 'b3':7}, 'c':4}
+    do_the_test(a,b, expected)
+
+def test_merge_6():
+    a        = {'a':1, 'b':{'b1':5, 'b2':6}, 'c':['x', 'y', 'z']}
+    b        = {'b':{'b3':7, 'b1':8}, 'c':{0:'r', 1:'s', 2:'t'}}
+    expected = {'a':1, 'b':{'b1':8, 'b2':6, 'b3':7}, 'c':['r', 's', 't']}
+    do_the_test(a,b, expected)
+
+def test_merge_6():
```
texasaggie97: You have two test_merge_6's

build/test_metadata_utilities.py
```diff
@@ -0,0 +1,54 @@
+import metadata_utilities
```
texasaggie97: We should call this as part of the build

marcoskirsch: Done, too includes it now.

build/test_metadata_utilities.py
marcoskirsch: Fixed.

PR #277, opened 2017-09-08
Move generation of html for coverage tool into bin/ folder.

marcoskirsch:

Moves the HTML that the coverage tool generates into bin/htmlcov, rather than the default htmlcov/

* make clean now get rids of this potentially stale directory
* removes need for developer to remember that htmlcov/ is special (we already learned bin/ is special)
* Simplifies .gitignore

Locally ran tox and HTML was created in desired location.
make clean still works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,5 @@ bin/
 .tox/
 .coverage
 commands.log
-htmlcov/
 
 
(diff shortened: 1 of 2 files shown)


marcoskirsch:
@Fladolcetta this will be a change for you - since you've been looking at these files a lot.
I think this is more correct that what we had. Although one difference is that before this, bin/ was only written to by Make. Now the coverage tool also writes to it.

PR #296, opened 2017-09-11
Context manager name in metadata

### LINKED ISSUE #267: Context manager is "Acquisition" or "Generation" for all APIs
In each driver's metadata/config.py we specify "direction" (input or output).
The code generator keys off of that in order to decide what to call the Context Manager that we return on initiate().

```
def test_acquisition(session):
    session.configure_measurement_digits(nidmm.Function.DC_CURRENT, 1, 5.5)
    with session.initiate():
        session.fetch(1000)
```

It's not a big deal because the user rarely sees the class name. But for bidirectional APIs like NI-HSDIO (which won't be added anytime soon) and for NI-SWITCH, these names are meaningless. I propose the name of the context manager itself is specified in metadata rather than we try to deduce it based on direction.

Also to consider: If the classes are to be considered private, then perhaps they should be prefixed with _.

texasaggie97:
* Fixes #267 
* Driver metadata no longer has list of directions.
  * Only one context manager per driver
  * Name explicitly in metadata

* Fixes issue
* Simplifies code

* Travis
* System tests

### DIFF
--- a/generated/nifake/tests/test_session.py
+++ b/generated/nifake/tests/test_session.py
@@ -284,6 +284,47 @@ class TestSession(object):
             except AttributeError as e:
                 pass
 
+    def test_get_vi_real64_attribute(self):
+        self.patched_library.niFake_GetAttributeViReal64.side_effect = self.side_effects_helper.niFake_GetAttributeViReal64
+        test_number = 1.5
+        self.side_effects_helper['GetAttributeViReal64']['attributeValue'] = test_number
+        with nifake.Session('dev1') as session:
+            attr_double = session.read_write_double
+            assert(attr_double == test_number)
+            from mock import call
+            calls = [call(SESSION_NUM_FOR_TEST, b"", 1000001, ANY)]
+            self.patched_library.niFake_GetAttributeViReal64.assert_has_calls(calls)
+            assert self.patched_library.niFake_GetAttributeViReal64.call_count == 1
+
+    def test_get_vi_bool_attribute(self):
+        self.patched_library.niFake_GetAttributeViBoolean.side_effect = self.side_effects_helper.niFake_GetAttributeViBoolean
+        test_boolean = True
+        self.side_effects_helper['GetAttributeViBoolean']['attributeValue'] = test_boolean
+        with nifake.Session('dev1') as session:
+            attr_bool = session.read_write_bool
+            assert(attr_bool == test_boolean)
+            from mock import call
+            calls = [call(SESSION_NUM_FOR_TEST, b"", 1000000, ANY)]
+            self.patched_library.niFake_GetAttributeViBoolean.assert_has_calls(calls)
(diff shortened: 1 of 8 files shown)


src/niswitch/metadata/config.py
```diff
@@ -15,10 +15,6 @@
             '64bit': {'name': 'libniswitch.so', 'type': 'cdll'},
         },
     },
-    'context_manager': [
-        {
-            'direction': 'input'
-        },
-    ]
+    'context_manager_name': '_Acquisition',
```
marcoskirsch: This should be _Scan

src/nidmm/metadata/config.py
```diff
@@ -15,10 +15,6 @@
             '64bit': {'name': 'libnidmm.so', 'type': 'cdll'},
         },
     },
-    'context_manager': [
-        {
-            'direction': 'input'
-        },
-    ]
+    'context_manager_name': '_Acquisition',
```
marcoskirsch: This name for the variable is fine for now. I think it will be a little confusing once we add a second context manager to Session for channels.

src/niswitch/metadata/config.py
texasaggie97: Changed

src/nidmm/metadata/config.py
texasaggie97: Changed 'context_manager_name' to be a dictionary. This way we can list different context manager names.

I am calling this one 'session'

generated/nifake/tests/test_session.py
```diff
@@ -284,6 +284,47 @@ def test_cannot_add_properties_to_session(self):
             except AttributeError as e:
                 pass
 
+    def test_get_vi_real64_attribute(self):
```
marcoskirsch: This is independent of this PR right?

src/nidmm/metadata/config.py
```diff
@@ -15,10 +15,8 @@
             '64bit': {'name': 'libnidmm.so', 'type': 'cdll'},
         },
     },
-    'context_manager': [
-        {
-            'direction': 'input'
-        },
-    ]
+    'context_manager_name': {
+        'session': 'acquisition',
```
marcoskirsch: I think 'session' is the wrong key. We have a Session class, and _it is_ a context manager.
This is something else. Best I can come up with is "task".

generated/nifake/tests/test_session.py
texasaggie97: It is. Which is why I requested changes on other PRs so things like this aren't required any more

src/nidmm/metadata/config.py
texasaggie97: This is the name of the object returned by the session context manager. Shouldn't that then be session?

texasaggie97: No, you're not the one confused. I am.

Changed to task.

PR #303, opened 2017-09-12
Consolidate different functions into single get_params()

texasaggie97:
* Take the different flavors of get parameter list snippet and consolidate into single function
* Pick which flavor using enum
  ``` python
  class ParamListType(Enum):
      '''Type of parameter list to return'''
      API_METHOD = 1
      '''Used for methods param list for the public API'''
      IMPL_METHOD = 2
      '''Used for methods param list for implementation'''
      DISPLAY_METHOD = 3
      '''Used for methods param list for display (rst)'''
      LIBRARY_METHOD = 4
      '''Used for methods param list when calling library'''
      LIBRARY_CALL = 5
      '''Used for methods param list when calling into the DLL'''
      LIBRARY_CALL_TYPES = 6
      '''Used for methods param list types when calling into the DLL'''
  ```
* Add extra types to metadata
* Add session name to config metadata
* Each flavor has defaults for options:
  * skip_self (bool)
  * skip_session_handle (bool)
  * skip_output_parameters (bool)
  * skip_ivi_dance_size_parameter (bool)
  * session_name (str)
* Default options can be overridden by passing an additional dictionary
  ``` python
  get_params_snippet(f, helper.ParamListType.LIBRARY_CALL_TYPES, {'session_name': config['session_name']})
  ```
* This will simplify parameter list generation when adding default parameters

* Travis
* System tests

### DIFF
--- a/build/helper.py
+++ b/build/helper.py
@@ -103,10 +103,10 @@ def normalize_string_type(d):
 # Functions that return snippets that can be placed directly in the templates.
 class ParamListType(Enum):
     '''Type of parameter list to return'''
-    METHOD = 1
-    '''Used for methods'''
-    FUNCTION = 2
-    '''Used for functions'''
+    API_METHOD = 1
+    '''Used for methods param list for the public API'''
+    IMPL_METHOD = 2
+    '''Used for methods param list for implementation'''
 
 
 def get_params_snippet(function, param_type, options={}):
--- a/build/helper.py
+++ b/build/helper.py
@@ -107,17 +107,59 @@ def normalize_string_type(d):
 class ParamListType(Enum):
     '''Type of parameter list to return'''
     API_METHOD = 1
-    '''Used for methods param list for the public API'''
+    '''Used for methods param list for the public API
+    
+    'skip_self': False,
+    'skip_session_handle': True,
+    'skip_output_parameters': True,
+    'skip_ivi_dance_size_parameter': True,
+    'session_name': 'vi',
+    '''
     IMPL_METHOD = 2
-    '''Used for methods param list for implementation'''
+    '''Used for methods param list for implementation
+    
+    'skip_self': False,
+    'skip_session_handle': False,
+    'skip_output_parameters': False,
+    'skip_ivi_dance_size_parameter': False,
+    'session_name': 'vi',
+    '''
     DISPLAY_METHOD = 3
-    '''Used for methods param list for display (rst)'''
+    '''Used for methods param list for display (rst)
+    
+    'skip_self': True,
+    'skip_session_handle': True,
+    'skip_output_parameters': True,
+    'skip_ivi_dance_size_parameter': True,
+    'session_name': 'vi',
+    '''
     LIBRARY_METHOD = 4
-    '''Used for methods param list when calling library'''
+    '''Used for methods param list when calling library
+    
+    'skip_self': True,
+    'skip_session_handle': False,
+    'skip_output_parameters': False,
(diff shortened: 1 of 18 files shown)


build/helper.py
```diff
@@ -98,80 +102,170 @@ def normalize_string_type(d):
...
-            snippets.append(x['python_name'])
-    return ', '.join(snippets)
-
 
-def get_function_parameters_snippet(parameters, session_name='vi'):
-    '''Returns a string suitable for the parameter list of a method given a list of parameter objects
 
-    If session_name set, skip that parameter
+ParamListTypeDefaults = {}
+ParamListTypeDefaults[ParamListType.API_METHOD] = {
```
marcoskirsch: Some of the things in this dictionary are "filtering/sorting" options, and some are "other" options. I think we should split it up.

And in general, I think we can break up this in two logical steps:

1. filter / reorder parameters
2. render the text snippet 

What do you think?

build/metadata_utilities.py
```diff
@@ -248,7 +279,8 @@ def test_add_all_metadata_simple():
                         'mechanism': 'fixed',
                         'value': 1
                     },
-                    'type': 'ViSession'
+                    'type': 'ViSession',
+                    'library_call_name': 'self.vi',
```
marcoskirsch: What is this supposed to be? The name is not descriptive.

docs/nidmm/functions.rst
```diff
@@ -1019,7 +1019,7 @@ nidmm.Session methods
     
 
 
-.. function:: fetch(maximum_time, reading)
+.. function:: fetch(maximum_time)
```
marcoskirsch: nice catch!

src/nifake/metadata/config.py
```diff
@@ -5,6 +5,7 @@
     'c_function_prefix': 'niFake_',
     'driver_name': 'NI-FAKE',
     'session_description': 'An NI-FAKE session to a fake MI driver whose sole purpose is to test nimi-python code generation',
+    'session_name': 'vi',
```
marcoskirsch: Is this the name the generator will use for the member that stores the session handle?
Anything else?

src/nidmm/metadata/config.py
```diff
@@ -5,6 +5,7 @@
     'c_function_prefix': 'niDMM_',
     'driver_name': 'NI-DMM',
     'session_description': 'An NI-DMM session to a National Instruments Digital Multimeter',
+    'session_name': 'vi',
```
marcoskirsch: Suggested renames for clarity:

* session_description  -> session_class_description
* session_name -> session_handle_parameter_name

src/nifake/metadata/config.py
marcoskirsch: It's not, it's the parameter name used by the C API as the session handle.

marcoskirsch:
Right now in session.py we hard-coded the name of the Session member that holds the session handle "vi". This needs to match the new config key. It's a hidden contract. We should instead use the same config key so that this always matches.

build/helper.py
texasaggie97: Broken into two loops

build/metadata_utilities.py
texasaggie97: This is the name used for the call into the library (DLL)

src/nidmm/metadata/config.py
texasaggie97: Changed

PR #306, opened 2017-09-13
Refactor Attribute classes to support channel-based set/get

marcoskirsch:

* Created base class Attribute
* Attribute is initialized with a default channel which is used on __set__ and __get__
* Derived attributes have a standard interface get/set methods with channel input.
* More code reuse

Preparation for support of channel-based attributes.

Ran unit tests and system tests.


build/templates/session.py.mako
```diff
@@ -32,77 +32,70 @@ from ${module_name} import library_singleton
...
+    def set(self, session, channel, value):
+        session._set_attribute_vi_boolean(channel, self._attribute_id, value)
 
-class AttributeEnum(object):
 
-    def __init__(self, attribute_id, enum_meta_class):
-        self.attribute_id = attribute_id
-        self.attribute_type = enum_meta_class
-        self.channel = ''
+class AttributeEnum(AttributeViInt32):
```
texasaggie97: I don't think you want to inherit from ViInt32. I believe Scope will have some float enums. Specifically for input impedance.

Or Maybe name it AttributeEnumViInt32?

build/templates/session.py.mako
```diff
@@ -32,77 +32,70 @@ from ${module_name} import library_singleton
 from ${module_name} import python_types
 
 
-class AttributeViInt32(object):
```
texasaggie97: Should these move to a separate file instead of being in session.py?

build/templates/session.py.mako
marcoskirsch: Let's fix it when we need to so we can test exactly the right scenario. Is that ok?

build/templates/session.py.mako
marcoskirsch: I agree. I'll move it on its own PR so as to not make the diff too difficult.

PR #309, opened 2017-09-13
Don't copy system tests or examples to bin/ folder

### LINKED ISSUE #286: Stop copying system tests to bin/
We copy system tests at build time, from:

    src/niswitch/system_tests/test_system_niswitch.py

to:

    bin/niswitch/system_tests/test_system_niswitch.py

If memory serves me right, we do this for historical reasons: we used to play import path tricks rather than install the modules so we needed the tests to be in a specific location relative to the built modules. This is no longer true - now we install the modules in the virtual environment. So we should be able to stop copying the files and run them from their location in src/.

texasaggie97:
* Fixes #286 
* Don't copy system tests or examples to bin/
* Run flake8 tests on them in src/\<driver\>/\[system_tests|examples\]

* Fixes issue

* Travis & system tests


tox.ini
```diff
@@ -42,16 +42,22 @@ commands =
     test: coverage html --directory=bin/htmlcov
     clean: make clean {posargs}
     codegen: make {posargs}
-    flake8: flake8 {posargs}
+    flake8: flake8 bin/ 
+    flake8: flake8 src/nidmm/system_tests/
```
marcoskirsch: what about the unit tests, are those being flake8ed?

texasaggie97: Yes, they are still copied into bin for the 'flake8 bin/' command will test them.

marcoskirsch: Do they have to be copied, or can they be run from src/?

texasaggie97: They need to be copied anyway for pytest because we are not installing anything. We rely on the relative positions in bin for the driver module to load correctly. 

For flake8 they can be tested in src.

marcoskirsch: We could install, however, since we run inside a virtual environment, right?
I ask so I can understand, not because we necessarily need to change things.

PR #322, opened 2017-09-15
Adding unit tests for Boolean and string attribute set methods

injaleea:

Adding more unittest for nifake to cover boolean and String set methods.

- test_set_bool_attribute
- test_set_vi_string_attribute

More unit test coverage

Passed on dev machine

### DIFF
--- a/generated/nifake/tests/test_session.py
+++ b/generated/nifake/tests/test_session.py
@@ -395,4 +395,4 @@ class TestSession(object):
         attrib_string = 'This is test string'
         with nifake.Session('dev1') as session:
             session.read_write_string = attrib_string
-            self.patched_library.niFake_SetAttributeViString.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, bytes(attrib_string())
+            self.patched_library.niFake_SetAttributeViString.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, bytes(attrib_string)
--- a/generated/nifake/tests/test_session.py
+++ b/generated/nifake/tests/test_session.py
@@ -392,7 +392,7 @@ class TestSession(object):
     def test_set_vi_string_attribute(self):
         self.patched_library.niFake_SetAttributeViString.side_effect = self.side_effects_helper.niFake_SetAttributeViString
         attribute_id = 1000002
-        attrib_string = 'This is test string'
+        attrib_string = b'This is test string'
         with nifake.Session('dev1') as session:
             session.read_write_string = attrib_string
             self.patched_library.niFake_SetAttributeViString.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, bytes(attrib_string))
--- a/generated/nifake/tests/test_session.py
+++ b/generated/nifake/tests/test_session.py
@@ -395,4 +395,4 @@ class TestSession(object):
         attrib_string = 'This is test string'
         with nifake.Session('dev1') as session:
             session.read_write_string = attrib_string
-            self.patched_library.niFake_SetAttributeViString.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, bytes(attrib_string)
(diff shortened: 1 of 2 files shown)


generated/nifake/tests/test_session.py
```diff
@@ -380,3 +380,19 @@ def test_init_with_options_nondefault(self):
...
+            session.read_write_bool = attrib_bool
+            self.patched_library.niFake_SetAttributeViBoolean.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, attrib_bool)
+
+    def test_set_vi_string_attribute(self):
+        self.patched_library.niFake_SetAttributeViString.side_effect = self.side_effects_helper.niFake_SetAttributeViString
+        attribute_id = 1000002
+        attrib_string = 'This is test string'
+        with nifake.Session('dev1') as session:
+            session.read_write_string = attrib_string
+            self.patched_library.niFake_SetAttributeViString.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, attrib_string)
```
injaleea: `assertion error value = AssertionError("Expected call: niFake_SetAttributeViString(42, b'', 1000002, 'This is test string')\nActual call: niFake_SetAttributeViString(42, b'', 1000002, b'This is test string')",) `

Should I use bytes(variable) in expected call ?

texasaggie97: Using bytes() will make the test pass.

Now whether that is correct or not, I am not sure. 

If the underlying driver doesn't care if it is a string or bytes in Python, then I think adding bytes() is an acceptable solution to this test.

If it does care, then we need to change things so that we pass in a string and not bytes, in which case you wouldn't need to change this.

marcoskirsch: I think it's correct that the expectations for the ctypes mock is that it receives a buffer of bytes.

src/nifake/tests/test_session.py
```diff
@@ -380,3 +380,19 @@ def test_init_with_options_nondefault(self):
...
         assert(session.vi == SESSION_NUM_FOR_TEST)
         self.patched_library.niFake_InitWithOptions.assert_called_once_with(b'FakeDevice', True, True, b'Some string', ANY)
+
+    def test_set_bool_attribute(self):
+        self.patched_library.niFake_SetAttributeViBoolean.side_effect = self.side_effects_helper.niFake_SetAttributeViBoolean
+        attribute_id = 1000000
+        attrib_bool = True
+        with nifake.Session('dev1') as session:
+            session.read_write_bool = attrib_bool
+            self.patched_library.niFake_SetAttributeViBoolean.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, attrib_bool)
```
marcoskirsch: the expectation should be that it gets called with 1 not with True. This is what we are sending the C DLL (ViBool is 16 bit signed int).

marcoskirsch:
need to resolve conflict

injaleea:
Submitted test works fine for dev machine and passing while tox . When building on github failes. Any thoughts ?

Fladolcetta:
This fails on my build machine with which uses Python3.6. I am assuming you are using Python 2.7?

A stack overflow comment here https://stackoverflow.com/questions/16750561/python-3-subprocess-error-in-bytes seems to indicate:

> In Python3 bytes can be decoded to str, and str can be encoded to bytes, but bytes can never be encoded, and str can never be decoded. That is why Python3 is complaining.

This appears to fail because you are sending a bytes object instead of just a string:

`attrib_string` = b'This is test string'

I removed "b" and ensured that we return the correct string and it passed.

    def test_set_vi_string_attribute(self):
        self.patched_library.niFake_SetAttributeViString.side_effect = self.side_effects_helper.niFake_SetAttributeViString
        attribute_id = 1000002
        attrib_string = 'This is test string'
            session.read_write_string = attrib_string
            self.patched_library.niFake_SetAttributeViString.assert_called_once_with(SESSION_NUM_FOR_TEST, b'', attribute_id, b'This is test string')

marcoskirsch:
This PR has been opened for a while and apparently doesn't work. I'm going to close it so it's off our list. Please reopen when ready.

PR #326, opened 2017-09-15
Add support for default values in the code generator

### LINKED ISSUE #88: Support default parameter values in generated APIs
We want some of the generated methods to have default values.
We need to add support for this to our code generator through functions_addons.py.

Examples:

`ViStatus _VI_FUNC niDMM_InitWithOptions(ViRsrc resourceName, ViBoolean IDQuery, ViBoolean   resetDevice, ViString optionsString, ViSession *newVi);
`
IDQuery, resetDevice could have defaults. Maybe a bad example because we are going to handcode constructors anyway.

```
ViStatus _VI_FUNC niDMM_GetDevTemp(
   ViSession   vi,
   ViString    reserved,
   ViReal64    *temperature);
```
Reserved should be omitted or at least defaulted to something for the Python user.

There may be other examples.

texasaggie97:
* Fixes #88 
* Allows setting default values to parameters
* Add default values for InitWithOptions
* Includes changes from #316 

* Prerequisite for different functions for __init__

* Travis & System tests


build/helper/codegen_helper.py
```diff
@@ -123,9 +145,19 @@ class ParamListType(Enum):
...
+    'skip_session_handle': False,
+    'skip_output_parameters': False,
+    'skip_ivi_dance_size_parameter': False,
+    'reordered_for_default_values': False,
+    'session_name': 'vi',
+    'name_to_use': 'python_name',
+}
 
 
 def get_params_snippet(function, param_type, options={}):
```
marcoskirsch: This function is 50 lines long... we should consider breaking up at some point (not needed yet, but an observation)

build/helper/metadata_add_all.py
```diff
@@ -121,6 +121,19 @@ def _add_library_call_name(parameter, session_name):
     parameter['library_call_name'] = library_call_name
 
 
+def _add_default_value_name(parameter):
```
marcoskirsch: IMO: Concatenating two strings in metadata probably doesn't warrant inserting into metadata.

marcoskirsch: If you disagree with this, you can merge. PR accepted.

texasaggie97: The advantage of adding this to metadata is that the render portion of get_params_snippet() has no name specific code. Right now it is three lines:
``` python
for x in params_to_use:
    snippets.append(x[name_to_use])
return ', '.join(snippets)
```

If this isn't added to metadata, then there would need to be something like this:
``` python
for x in params_to_use:
    if name_to_use == 'python_name_with_default':
        <do the concatenation here>
    else:
        snippets.append(x[name_to_use])
return ', '.join(snippets)
```
This would make that function even longer, which I was trying to avoid.

marcoskirsch: Well, that function needs to be broken up / refactored.

PR #327, opened 2017-09-15
Support for driver specific init call

### LINKED ISSUE #150: Need to a way to customize the session constructor for a specific driver
The default initialize methods for NI-SWITCH is Initialize with topology, and for FGEN and DCPower is Initialize with Channels. We need a way to make these the default session constructor. The current template forces the user to open a session with Initialize with Options.

texasaggie97:
* Fixes #150 
* Includes changes from #326 
* Add support for having __init__ call different InitWith* functions depending on what is appropriate for that driver
  * __init__ parameter list will be based on parameter list for configured function
  * Uses default values
* Switch uses InitWithTopology
  * Set default values

* Fixes issue
* Needed for Iteration

* System test
* Travis


src/niswitch/metadata/functions_addon.py
```diff
@@ -86,8 +86,12 @@
 
 # Default values for method parameters
 function_default_value = {
-    'InitWithOptions':  { 'parameters': { 1: { 'default_value': False, },
-                                          2: { 'default_value': False, },
-                                          3: { 'default_value': '', }, }, },
+    'InitWithOptions':   { 'parameters': { 1: { 'default_value': False, },
```
marcoskirsch: you can mark this function so code-gen is skipped. It's essentially dead code now, right?

PR #329, opened 2017-09-15
Add method table to Session doc page

### LINKED ISSUE #156: Docs for session list attributes but not functions
See http://nimi-python.readthedocs.io/en/master/nidmm/session.html

a) We should use Python terms properties and methods
b) We should list methods, similar to how we list properties

texasaggie97:
* Fixes #156 
* Add method table to Session page
  * Link to full method documentation
  * Does NOT include function prototype. Should it?
  * No header, only includes public methods:
  image

* Improves documentation
* Fixes issue

* Travis
* System tests


marcoskirsch:
I think the methods in the table should say
`niswitch.Session.connect()`
not
`niswitch.connect()`
because they are methods of Session.

BUT

We don't do the full qualified name thing for properties. We should be consistent. If documentation makes it clear that these are Session methods then no need to prefix with niswitch.Session. which would be redundant and visually noisy.

texasaggie97:
Updated so it looks like
image

PR #331, opened 2017-09-18
Support enums that aren't ViInt32

### LINKED ISSUE #330: Properties resolution_digits and input_resistance incorrect
resolution_digits and input_resistance properties are linked to enums  which then sets int32 type. 
These two functions require real64 types.
session.py line 304 should be input_resistance = AttributeViReal64(1150029)
and line 531 should be resolution_digits = AttributeViReal64(1250003)

Setting properties raises TypeErrors.

texasaggie97:
* Fixes #330 
* Changes to `AttributeEnum()`
  * Now inherits from `object` instead of `AttributeViInt32`
  * Class of appropriate underlying type is passed in to the constructor
    ``` python
    resolution_digits = attributes.AttributeEnum(attributes.AttributeViReal64, enums.DigitsResolution, 1250003)
    ```
  * Instantiates the underlying type and uses that for actual get/set
  * Enum type is still used to return the correct type and to do type checking on the incoming type
* Tests added to get/set float based enum

* Fixes issue

* Travis
* System tests

### DIFF
--- a/build/templates/attributes.py.mako
+++ b/build/templates/attributes.py.mako
@@ -61,8 +61,8 @@ class AttributeViBoolean(Attribute):
 
 class AttributeEnum(object):
 
-    def __init__(self, underlying_meta_class, enum_meta_class, attribute_id, default_channel=''):
-        self._underlying_attribute = underlying_meta_class(attribute_id, default_channel)
+    def __init__(self, underlying_attribute_meta_class, enum_meta_class, attribute_id, default_channel=''):
+        self._underlying_attribute = underlying_attribute_meta_class(attribute_id, default_channel)
         self._attribute_type = enum_meta_class
         self._attribute_id = attribute_id
         self._default_channel = default_channel
--- a/build/templates/attributes.py.mako
+++ b/build/templates/attributes.py.mako
@@ -59,7 +59,7 @@ class AttributeViBoolean(Attribute):
         session._set_attribute_vi_boolean(channel, self._attribute_id, value)
 
 
-class AttributeEnum(Attribute):
+class AttributeEnum(object):
 
     def __init__(self, underlying_attr_type, enum_meta_class):
         self._underlying_attr_type = underlying_attr_type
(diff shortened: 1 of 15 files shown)


generated/nidmm/attributes.py
```diff
@@ -52,16 +52,20 @@ def set(self, session, channel, value):
         session._set_attribute_vi_boolean(channel, self._attribute_id, value)
 
 
-class AttributeEnum(AttributeViInt32):
+class AttributeEnum(Attribute):
 
-    def __init__(self, attribute_id, enum_meta_class, channel=''):
+    def __init__(self, underlying_attr_type, enum_meta_class):
```
marcoskirsch: Underlying_attr_type is not the type, it's an actual instance of an attribute. We can rename the parameter / member, or truly pass the type and instantiate the object here.

There's some redundancy here, because AttributeEnum also derives from Attribute, as well as the thing passed in. I think we can live with it, however, even though it's a bit confusing.

texasaggie97: Changed how AttributeEnum works and is instatiated

build/templates/attributes.py.mako
```diff
@@ -59,16 +59,20 @@ class AttributeViBoolean(Attribute):
         session._set_attribute_vi_boolean(channel, self._attribute_id, value)
 
 
-class AttributeEnum(AttributeViInt32):
+class AttributeEnum(object):
 
-    def __init__(self, attribute_id, enum_meta_class, channel=''):
+    def __init__(self, underlying_attr_type, enum_meta_class, attribute_id, default_channel=''):
```
marcoskirsch: On the same parameter list client passes two meta-classes: one for the Attribute and one for the End.
Can we make the name consistent? Either underlying_attr_meta_class or enum_type

marcoskirsch:
Small change requested.
This should go into CONTRIBUTING:

* Added support for enums with types other than ViInt32 (Fixes #330)

build/templates/attributes.py.mako
texasaggie97: Changed to `underlying_attribute_meta_class`

texasaggie97:
Added entry in CHANGELOG.md

PR #333, opened 2017-09-19
Fix #320: SideEffectsHelper applies side effects in the warning case

### LINKED ISSUE #320: In niFake_Read(), when there is a warning, return value is 0, and not what we specify
Tested with the following code. Getting AssertionError, value returned in session.read(test_maximum_time) is 0.0 and not nan.

    def test_read_with_warning(self):
        test_maximum_time = 10
        test_reading = float('nan')
        test_error_code = 42
        test_error_desc = "The answer to the ultimate question, only positive"
        self.patched_library.niFake_Read.side_effect = self.side_effects_helper.niFake_Read
        self.side_effects_helper['Read']['return'] = test_error_code
        self.side_effects_helper['Read']['reading'] = test_reading
        self.patched_library.niFake_GetError.side_effect = self.side_effects_helper.niFake_GetError
        self.side_effects_helper['GetError']['errorCode'] = test_error_code
        self.side_effects_helper['GetError']['description'] = test_error_desc
            with warnings.catch_warnings(record=True) as w:
                assert test_reading == session.read(test_maximum_time)
                assert len(w) == 1
                assert issubclass(w[0].category, nifake.NifakeWarning)
                assert test_error_desc in str(w[0].message)

marcoskirsch:

Fixes #320 
In the warning case, we wouldn't apply the other side effects.

Bug fix, allows us to unit test warning scenarios.

Ran tox to make sure this broke nothing.

### DIFF
--- a/build/templates/mock_helper.py.mako
+++ b/build/templates/mock_helper.py.mako
@@ -65,7 +65,8 @@ ivi_dance_param = helper.extract_ivi_dance_parameter(params)
 ivi_dance_size_param = helper.find_size_parameter(ivi_dance_param, params)
 %>\
     def ${c_function_prefix}${func_name}(${helper.get_params_snippet(f, helper.ParamListType.LIBRARY_IMPL_METHOD)}):  # noqa: N802
-        if self._defaults['${func_name}']['return'] != 0:
+        if self._defaults['${func_name}']['return'] < 0:
+            # In the error case, skip all other side effects.
             return self._defaults['${func_name}']['return']
 %    for p in output_params:
         if self._defaults['${func_name}']['${p['name']}'] is None:
--- a/generated/nidmm/tests/mock_helper.py
+++ b/generated/nidmm/tests/mock_helper.py
@@ -237,157 +237,188 @@ class SideEffectsHelper(object):
         self._defaults[func] = val
 
     def niDMM_Abort(self, vi):  # noqa: N802
-        if self._defaults['Abort']['return'] != 0:
+        if self._defaults['Abort']['return'] < 0:
+            # In the error case, skip all other side effects.
             return self._defaults['Abort']['return']
         return self._defaults['Abort']['return']
 
     def niDMM_ClearInterchangeWarnings(self, vi):  # noqa: N802
-        if self._defaults['ClearInterchangeWarnings']['return'] != 0:
+        if self._defaults['ClearInterchangeWarnings']['return'] < 0:
+            # In the error case, skip all other side effects.
             return self._defaults['ClearInterchangeWarnings']['return']
         return self._defaults['ClearInterchangeWarnings']['return']
 
     def niDMM_ConfigureACBandwidth(self, vi, ac_minimum_frequency_hz, ac_maximum_frequency_hz):  # noqa: N802
-        if self._defaults['ConfigureACBandwidth']['return'] != 0:
+        if self._defaults['ConfigureACBandwidth']['return'] < 0:
+            # In the error case, skip all other side effects.
             return self._defaults['ConfigureACBandwidth']['return']
         return self._defaults['ConfigureACBandwidth']['return']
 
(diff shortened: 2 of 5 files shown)


build/templates/mock_helper.py.mako
```diff
@@ -65,7 +65,8 @@ ivi_dance_param = helper.extract_ivi_dance_parameter(params)
 ivi_dance_size_param = helper.find_size_parameter(ivi_dance_param, params)
 %>\
     def ${c_function_prefix}${func_name}(${helper.get_params_snippet(f, helper.ParamListType.LIBRARY_IMPL_METHOD)}):  # noqa: N802
-        if self._defaults['${func_name}']['return'] != 0:
+        if self._defaults['${func_name}']['return'] < 0:
```
texasaggie97: This means the string value will be set even during the first part of the IVI dance.

I think maybe the correct way to fix this would be to insert the original if statement, but only when the function supports the ivi dance style of string retrieval. In other cases, it would fall through to the normal return.

marcoskirsch: > This means the string value will be set even during the first part of the IVI dance.

That's right.

> I think maybe the correct way to fix this would be to insert the original if statement, but only when the function supports the ivi dance style of string retrieval. In other cases, it would fall through to the normal return.

I think that's getting too clever. Unit test code should have very few logic and no codepaths. The caller should explicitly program this behavior. Can't we set expectations for first and second call separately?

texasaggie97: But then a test could be written that works because of the incorrect behavior of the mock, but wouldn't work on the real driver.

I think the mock should mock the behavior of the driver as closely as possible.

What do drivers do for other functions that return a value when they get an error? Looking at scope, it looks like it doesn't return data if there is an error. I was looking at some of the `CalFetch*()` functions.

marcoskirsch: > But then a test could be written that works because of the incorrect behavior of the mock, but wouldn't work on the real driver.

This is a reality of unit testing. If the developer programs the mock to behave differently than the object it's mocking, then the code won't be properly tested.

> I think the mock should mock the behavior of the driver as closely as possible.

This is true. Where that responsibility falls is the question.

> What do drivers do for other functions that return a value when they get an error? Looking at scope, it looks like it doesn't return data if there is an error. I was looking at some of the CalFetch*() functions.

It varies. For example in the fetch and timeout case: NI-DMM and NI-DCPower will return the points it managed to get before the timeout.

texasaggie97: > > But then a test could be written that works because of the incorrect behavior of the mock, but wouldn't work on the real driver.

> This is a reality of unit testing. If the developer programs the mock to behave differently than the object it's mocking, then the code won't be properly tested.

Then why are we programming the mock to behave differently?

> > I think the mock should mock the behavior of the driver as closely as possible.

> This is true. Where that responsibility falls is the question.

I think the responsibility belongs here. And if that adds complexity to the generator so that the generated side effect function is "correct", then I think that is the appropriate tradeoff.

An alternative would be to not use the generated side effect function and instead write one by hand in the test file that has the appropriate behavior.

texasaggie97: I think the default side effect should stay as it is, since that is what is needed most of the time.

When you need different behavior, like you do here, then I think the correct answer is to create a custom side effect function at the place where you need to different behavior. You can see where I did this in the ModInst unit tests. I needed a side effect function for testing iterating functionality.

marcoskirsch:
I'm closing without merging.

This is why: We're trying to make mock_helper.py work for all unit test cases. In reality, we should think of mock_helper.py as a helper that sets side effects for us in a way that works in 90% of the cases. Otherwise we risk making things too complicated.

The solution for unit tests in which mock_helper.py doesn't work as-is, is not to make mock_helper.py more complicated but to simply not use it and set side effects in the unit test itself.

PR #341, opened 2017-09-20
Updating System tests to test all remaining non-scanning functions

Fladolcetta:

This adds system tests for NI-SWITCH functions that are not related to scanning.

This improves coverage

Built and ran. Ran system tests

### DIFF
--- a/src/niswitch/system_tests/test_system_niswitch.py
+++ b/src/niswitch/system_tests/test_system_niswitch.py
@@ -17,7 +17,8 @@ def test_relayclose(session):
     assert session.get_relay_position(relay_name) == niswitch.RelayPosition.OPEN
     session.relay_control(relay_name, niswitch.RelayAction.CLOSE_RELAY)
     assert session.get_relay_position(relay_name) == niswitch.RelayPosition.CLOSED
-    assert relay_name
+    relay_count = session.get_relay_count(relay_name)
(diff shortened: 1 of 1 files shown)


src/niswitch/system_tests/test_system_niswitch.py
```diff
@@ -141,3 +144,20 @@ def test_functions_get_error_description(session):
...
+def test_functions_disable(session):
+    channel1 = 'c0'
+    channel2 = 'r0'
+    session.connect(channel1, channel2)
+    session.disable()   # expect no errors
+    assert session.can_connect(channel1, channel2) == niswitch.PathCapability.PATH_AVAILABLE
+
+
+def test_functions_interchange(session):
+    session.clear_interchange_warnings()   # expect no errors
```
marcoskirsch: This applies to all our APIs. They are sort of useless. How about we stop generating?

Fladolcetta: I agree that there is little use for them. But we do support them in both LabVIEW and LabVIEW NXG for all drivers. Removing them here would be inconsistent with that. 
Thoughts?

marcoskirsch: Opened https://github.com/ni/nimi-python/issues/342 so we can discuss there.

Issue #344, opened 2017-09-20
Scanning context manager does not work for switches.


Fladolcetta:
Currently we are looking for _initiate() when using the initiate context manager. Switches uses initiate_scan() instead of _initiate and this causes errors as the context manager cannot find _initiate()

Fladolcetta:
Closed in #347

PR #347, opened 2017-09-20
Add custom initiate and abort functions for context manager.

Fladolcetta:
[ x] This contribution adheres to [CONTRIBUTING.md]

This adds the ability to add custom initiate and abort functions.
Adds system tests for scanning for switch.

Fixes bug #344 

Built and ran unit tests. Ran system tests.


build/templates/session.py.mako
```diff
@@ -18,6 +18,8 @@ ${encoding_tag}
     functions = helper.extract_codegen_functions(functions)
 
     session_context_manager = '_' + config['context_manager_name']['task'].title() if 'task' in config['context_manager_name'] else None
```
texasaggie97: I would put all these into an if statement so that if task exists, they all must since the definition of the "task' context manager will need initiate and abort, but if task doesn't exist, the others are not required. They are now.
``` python
session_context_manager = None
if 'task' in config['context_manager_name']:
    session_context_manager = '_' + config['context_manager_name']['task'].title()
    session_context_manager_initiate = '_' + config['context_manager_name']['initiate']
    session_context_manager_abort = '_' + config['context_manager_name']['abort']
```

Fladolcetta: fixed

src/niswitch/metadata/config.py
```diff
@@ -18,6 +18,8 @@
     },
     'context_manager_name': {
         'task': 'scan',
+        'initiate': 'initiate_scan',
```
marcoskirsch: This refers to a function using its python name.
In all other places we use the C names, for example init_function (see line 24) or when specifying parameters for buffer retrieval mechanisms.
We should be consistent about this and it seems like C names as found in metadata is the way to go.

PR #363, opened 2017-09-26
Add default values for parameters for NI-SWITCH

### LINKED ISSUE #299: Add default parameter metadata to NI-SWITCH

Fladolcetta:
[x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/<reponame>/blob/master/CONTRIBUTING.md).

Fixes #299 

To make the customer experience easier by reducing the number of parameters required.

Built and ran. Ran unit and system tests


docs/niswitch/functions.rst
```diff
@@ -118,7 +118,7 @@ niswitch.Session methods
 
     :type scan_mode: :py:data:`niswitch.ScanMode`
 
-.. function:: configure_scan_trigger(scan_delay, trigger_input, scan_advanced_output)
+.. function:: configure_scan_trigger(trigger_input, scan_advanced_output, scan_delay)
```
marcoskirsch: opened https://github.com/ni/nimi-python/issues/365 (missing default in documentation)

src/niswitch/metadata/functions_addon.py
```diff
@@ -89,9 +89,17 @@
...
-    'InitWithTopology':  { 'parameters': { 1: { 'default_value': 'Configured Topology', },
-                                           2: { 'default_value': False, },
-                                           3: { 'default_value': False, },
-                                           4: { 'default_value': '', }, }, },
-}
+    'InitWithTopology':         { 'parameters': { 1: { 'default_value': 'Configured Topology', },
+                                                  2: { 'default_value': False, },
+                                                  3: { 'default_value': False, },
+                                                  4: { 'default_value': '', }, }, },
+    'ConfigureScanList':        { 'parameters': { 2: { 'default_value': 'enums.ScanMode.BREAK_BEFORE_MAKE', 'print_verbatim':True},},},
```
marcoskirsch: You don't need this new piece of redundant metadata 'print_verbatim'. Rather, you can infer that you need to do this by looking at parameter['enum'] is not None.

PR #368, opened 2017-09-27
Adding nidmm and niswitch tox system test functions

Fladolcetta:

Adds driver specific system tests.

This is required for internal test suites to run only the specific driver.

Ran both nidmm and niswitch. Verified released python packages were installed (saw errors due to default parameters change).

### DIFF
--- a/tools/run_python_system_tests.py
+++ b/tools/run_python_system_tests.py
@@ -1,0 +1,63 @@
+import os
+import sys
+import pip
+import time
+import json
+import zipfile
+import argparse
+import tempfile
+import urllib.request
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-d', '--driver', default='niswitch', type=str)
+args = parser.parse_args()
+
+
+# Install Tox
+print('****Installing tox to python.****')
+pip.main(['install', 'tox'])
+
+
+# Create Directory
+print('****Creating temporary directory.****')
+temp_dir = tempfile.gettempdir()
+working_directory = os.path.join(temp_dir, str(time.time()))
+if not os.path.exists(working_directory):
+    os.makedirs(working_directory)
+print(working_directory)
+
+
+# Read url and retrieve zip file link
+print('****Parsing github releases to obtain the latest zip file url.****')
+release_url = 'https://api.github.com/repos/ni/nimi-python/releases'
+with urllib.request.urlopen(release_url) as response:
+    html = response.read()
+url_data = json.loads(html.decode('ascii'))
+zip_url = url_data[0]['zipball_url']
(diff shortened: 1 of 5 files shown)


tools/run_python_system_tests.py
```diff
@@ -0,0 +1,63 @@
...
+import pip
+import time
+import json
+import zipfile
+import argparse
+import tempfile
+import urllib.request
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-d', '--driver', default='niswitch', type=str)
```
texasaggie97: You should make this required and not have a default.

marcoskirsch: You should also add some high-level documentation to argparse about what this tool does, so the information is printed when invoked with --help, which is the expectation of any command line tool user.

tools/run_python_system_tests.py
```diff
@@ -0,0 +1,63 @@
...
+args = parser.parse_args()
+
+
+# Install Tox
```
marcoskirsch: Since you have a print, the comment is no longer necessary. This applies elsewhere.

tools/run_python_system_tests.py
```diff
@@ -0,0 +1,63 @@
...
+print('****Installing tox to python.****')
```
marcoskirsch: -> Python

tools/run_python_system_tests.py
```diff
@@ -0,0 +1,63 @@
...
+print('****Creating temporary directory.****')
+temp_dir = tempfile.gettempdir()
+working_directory = os.path.join(temp_dir, str(time.time()))
+if not os.path.exists(working_directory):
+    os.makedirs(working_directory)
+print(working_directory)
+
+
+# Read url and retrieve zip file link
+print('****Parsing github releases to obtain the latest zip file url.****')
```
marcoskirsch: -> GitHub
-> URL

marcoskirsch:
I'll let @texasaggie97 merge this since I assume it needs tweaks to nimi-bot.

tools/run_python_system_tests.py
Fladolcetta: Fixed

tools/run_python_system_tests.py
Fladolcetta: Fixed

tools/run_python_system_tests.py
Fladolcetta: Fixed

tools/run_python_system_tests.py
Fladolcetta: Fixed

PR #370, opened 2017-09-28
Adding unit test for multipoint read

bhaswath:

Adds Unit test to NI-FAKE to test proper functionality of read_multipoint()
Modifies mock_helper.py so that read_multipoint() returns an array and not a float.

Increases coverage

Unit test passes locally

### DIFF
--- a/generated/nifake/tests/mock_helper.py
+++ b/generated/nifake/tests/mock_helper.py
@@ -247,7 +247,7 @@ class SideEffectsHelper(object):
             return self._defaults['ReadMultiPoint']['return']
         if self._defaults['ReadMultiPoint']['readingArray'] is None:
             raise MockFunctionCallError("niFake_ReadMultiPoint", param='readingArray')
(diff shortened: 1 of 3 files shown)


bhaswath:
assert measurements == test_reading_array would cause an AssertionError to raise.
The value being returned is always an array of size test_actual_number_of_points,  but the values returned in the array are always 0.0
Is there any additional file that I must modify in order for this to return the values I expect it to?

bhaswath:
How do I modify the mock_helper in bin?
I have changed mock_helper in generated.

marcoskirsch:
> How do I modify the mock_helper in bin?

This file is code-generated. You should never modify it directly. It even says so in its first line:

`# This file was generated`

The file you should be modifying is:
build/templates/mock_helper.py.mako

Fladolcetta:
Validated that multi_point read is already unit tests in the latest export of master.

https://raw.githubusercontent.com/ni/nimi-python/master/src/nifake/tests/test_session.py

PR #402, opened 2017-10-03
Fix library return type

texasaggie97:
* Cleans up ctypes usage, by using the correct ctype in and out of the DLL call
* Return the .value from the ctype object
* Kept the interface to Library the same

N/A

* Travis & System tests

### DIFF
--- a/docs/nidcpower/examples.rst
+++ b/docs/nidcpower/examples.rst
@@ -1,6 +1,15 @@
 Examples
 ========
 
+nidcpower_source_delay_measure.py
+---------------------------------
+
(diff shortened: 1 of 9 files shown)


docs/nidcpower/examples.rst
```diff
@@ -1,15 +1,6 @@
 Examples
 ========
 
-nidcpower_source_delay_measure.py
```
marcoskirsch: I've noticed that on different systems, this generates in different order. I think this should be deterministic.
Opened #403.

texasaggie97: Reverted.

PR #406, opened 2017-10-03
Remove python types

texasaggie97:
* Remove python_types
* Instead we use intrinsic types

N/A

* Travis & System tests


build/helper/helper.py
```diff
@@ -47,15 +46,29 @@ def add_to_path(p):
         sys.path = old_path
 
 
+type_mapping = {
```
marcoskirsch: move this to inside the function, and rename to type_map (noun)

build/helper/metadata_add_all.py
```diff
@@ -33,21 +33,12 @@ def _add_python_parameter_name(parameter):
...
-    return parameter
-
-
 def _add_intrinsic_type(parameter):
     '''Adds a intrinsic (basic python type) key/value pair to the parameter metadata'''
     if parameter['enum'] is None:
         parameter['intrinsic_type'] = get_intrinsic_type_from_visa_type(parameter['type'])
     else:
-        parameter['intrinsic_type'] = parameter['python_type']
+        parameter['intrinsic_type'] = 'enums.' + parameter['enum']
```
marcoskirsch: Why not call this 'python_type'?
That way every parameter has 2 types associated to it: the Visa one and the Python API one. Intrinsic is less clear in my opinion.

Or is this a temporary rename pending more changes?

build/helper/metadata_add_all.py
```diff
@@ -201,11 +191,36 @@ def add_all_function_metadata(functions, config):
     return functions
 
 
+def _compare_values(actual, expected):
```
marcoskirsch: What are these changes about? Are they needed for the PR?

build/helper/metadata_add_all.py
```diff
@@ -297,7 +310,7 @@ def test_add_all_metadata_simple():
             'python_name': 'make_a_foo',
             'returns': 'ViStatus',
             'returns_ctype': 'ViStatus_ctype',
-            'returns_python': 'ViStatus'
+            'intrinsic_return_type': 'int',
```
marcoskirsch: this is the type of a return value... it is now inconsistent with 'returns_ctype'.

build/helper/helper.py
texasaggie97: Done

build/helper/metadata_add_all.py
texasaggie97: Next PR

build/helper/metadata_add_all.py
texasaggie97: They help for tracking down why the actual and expected don't match. When `assert actual == expected` there are two huge dictionaries that differ in some small way. This breaks the check down into smaller parts so it can tell you how they don't match. Such and such key is missing or something like that.

marcoskirsch: I think this should be pulled into its own PR.

build/helper/metadata_add_all.py
texasaggie97: Changed in future PR

build/helper/metadata_add_all.py
texasaggie97: Moved to #410

PR #408, opened 2017-10-03
Use error_message function for error handling

### LINKED ISSUE #154: We should use niXXX_error_message, not niXXX_GetErrorMessage

Fladolcetta:

Replaces GetErrorMessage with error_message when getting error descriptions on an invalid session. We do this because it's present in all drivers even though string length is limited.

Fixes #154 

Ran System Tests, Unit Tests, Flake8, and then tried to get error_message by doing:
    session1 = nidmm.Session(args.name)
    session2 = nidmm.Session(args.name)
    session1.read()


generated/nidmm/library.py
```diff
@@ -575,6 +576,14 @@ def niDMM_close(self, vi):  # noqa: N802
                 self.niDMM_close_cfunc.restype = ViStatus_ctype  # noqa: F405
         return self.niDMM_close_cfunc(vi).value
 
+    def niDMM_error_message(self, vi, error_code, error_message):  # noqa: N802
```
marcoskirsch: if you add this function to NI-DMM, you should remove the one we no longer use.

generated/nifake/tests/test_session.py
```diff
@@ -477,35 +477,35 @@ def test_get_attribute_error(self):
                 assert e.code == test_error_code
                 assert e.description == test_error_desc
 
-    # Error descriptions
-
-    def test_get_error_description_get_error_message(self):
+    '''
```
marcoskirsch: Why is this commented out?
Add a TODO that explains why and when it needs to be put back in.

generated/nifake/tests/test_session.py
```diff
@@ -477,35 +477,35 @@ def test_get_attribute_error(self):
                 assert e.code == test_error_code
                 assert e.description == test_error_desc
 
-    # Error descriptions
```
marcoskirsch: Put back "# Error descriptions"?
The whole thing is sort of organized with these headers.

src/nidmm/metadata/functions_addon.py
```diff
@@ -12,7 +12,7 @@
     'CheckAttribute.+':                { 'codegen_method': 'no',       },  # We do not include any Check Attribute functions
     '.etAttribute.+':                  { 'codegen_method': 'private',  },  # All Set/Get Attribute functions are private
     'init':                            { 'codegen_method': 'no',       },
-    'error_message':                   { 'codegen_method': 'no',       },
+    'error_message':                   { 'codegen_method': 'private',  },
     'GetError':                        { 'codegen_method': 'private',  },
     'GetErrorMessage':                 { 'codegen_method': 'private',  },
```
marcoskirsch: If these are now dead code, then don't generate.
Applies to all.

src/niswitch/metadata/__init__.py
```diff
@@ -10,12 +10,14 @@
...
 from metadata.functions_addon import functions_params_types
 from metadata.functions_addon import functions_buffer_info
+from metadata.functions_addon import functions_is_error_handling
 from metadata.functions_addon import function_default_value
 
 helper.merge_dicts(functions, functions_codegen_method)
 helper.merge_dicts(functions, functions_enums)
 helper.merge_dicts(functions, functions_params_types)
 helper.merge_dicts(functions, functions_buffer_info)
+helper.merge_dicts(functions, functions_is_error_handling)
```
marcoskirsch: This is an unrelated fix, isn't it?

generated/nidmm/library.py
Fladolcetta: Fixed

generated/nifake/tests/test_session.py
Fladolcetta: Fixed

generated/nifake/tests/test_session.py
Fladolcetta: Fixed

src/nidmm/metadata/functions_addon.py
Fladolcetta: fixed

src/niswitch/metadata/__init__.py
Fladolcetta: Sort of? We were not merging this from functions.addon. In order to test the fix to use error_message, this needs to be added

marcoskirsch: ok

PR #415, opened 2017-10-04
Make visatypes "typedefs" rather than subclasses

texasaggie97:
* Make visatypes "typedefs" rather than subclasses
  * This is done as setting them equal to the ctype instead of a class that inherits from ctypes

N/A

* Travis
* System tests

### DIFF
--- a/generated/nidcpower/library.py
+++ b/generated/nidcpower/library.py
@@ -76,7 +76,7 @@ class Library(object):
                 self.niDCPower_Abort_cfunc = self._library.niDCPower_Abort
                 self.niDCPower_Abort_cfunc.argtypes = [ViSession]  # noqa: F405
                 self.niDCPower_Abort_cfunc.restype = ViStatus  # noqa: F405
-        return self.niDCPower_Abort_cfunc(vi).value
+        return self.niDCPower_Abort_cfunc(vi)
 
     def niDCPower_Commit(self, vi):  # noqa: N802
         with self._func_lock:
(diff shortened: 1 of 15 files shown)


generated/nidcpower/library.py
```diff
@@ -76,372 +76,372 @@ def niDCPower_Abort(self, vi):  # noqa: N802
                 self.niDCPower_Abort_cfunc = self._library.niDCPower_Abort
                 self.niDCPower_Abort_cfunc.argtypes = [ViSession]  # noqa: F405
                 self.niDCPower_Abort_cfunc.restype = ViStatus  # noqa: F405
-        return self.niDCPower_Abort_cfunc(vi).value
+        return self.niDCPower_Abort_cfunc(vi)
```
marcoskirsch: If I understand correctly, this used to return an int. Now it returns a ViStatus/ctypes.c_long.
Is that right?

generated/nidcpower/visatype.py
```diff
@@ -6,69 +6,19 @@
...
-class ViConstString(ViString):  # noqa: N801
-    @property
-    def value(self):  # Makes 'value' readonly
-        return super(ViConstString, ViString).value
-
-
-class ViBoolean(ctypes.c_ushort):  # noqa: N801
-    @classmethod
-    def from_param(cls, param):
-        return ctypes.c_uint16(1) if bool(param) else ctypes.c_uint16(0)
```
marcoskirsch: We no longer have this.
I suppose ctypes just handled all the conversion from and to True and False?

PR #419, opened 2017-10-05
Adding system tests for uncovered dmm methods

injaleea:

[X] I've updated [CHANGELOG.md](https://github.com/ni/<reponame>/blob/master/CHANGELOG.md) if applicable.

Adding more system tests for nidmm

- test_fetch_waveform
- test_fetch_waveform_error
- test_get_measurement_peiod
- test_perform_cable_compensation
- test_read_waveform
- test_send_software_trigger
- test_reset_method

Pytest passed on dev machine

### DIFF
--- a/src/nidmm/system_tests/test_system_nidmm.py
+++ b/src/nidmm/system_tests/test_system_nidmm.py
@@ -250,8 +250,9 @@ def test_fetch_waveform(session):
     session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1000, 10)
     with session.initiate():
         number_of_points_to_read = 10
-        measurements = session.fetch_waveform(array_size=number_of_points_to_read)[1]
-        assert measurements == number_of_points_to_read
+        measurements,actual_number_of_points = session.fetch_waveform(number_of_points_to_read)
+        assert len(measurements) == number_of_points_to_read
+        assert actual_number_of_points == number_of_points_to_read
 
 
 def test_fetch_waveform_error(session):
@@ -259,45 +260,43 @@ def test_fetch_waveform_error(session):
         session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1000, 10)
         with session.initiate():
             number_of_points_to_read = 100
-            session.fetch_waveform(array_size=number_of_points_to_read)   # trying to fetch points more than configured
+            session.fetch_waveform(number_of_points_to_read)   # trying to fetch points more than configured
     except nidmm.Error as e:
         assert e.code == -1074126845  # Max Time exceeded before operation completed
 
 
-def test_get_measurement_peiod():
+def test_get_measurement_period():
         with nidmm.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:4072; BoardType:PXI') as session:
             session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 10, 5.5)
             measurement_period = session.get_measurement_period()
-            assert measurement_period == 0.0071333333333333335   # 0.0071333333333333335 is the time required for 4072 to take a DC_VOLT measurement with range 10V on Digits_resolution 5.5
+            assert measurement_period == 0.0071333333333333335  # 0.0071333333333333335 is the time required for 4072 to take a DC_VOLT measurement with range 10V on Digits_resolution 5.5
 
 
 def test_perform_cable_compensation(session):
     session.configure_measurement_digits(nidmm.Function.CAPACITANCE, 0.002, 5.5)
     conductance, susceptance = session.perform_open_cable_comp()
-    assert conductance == 0
+    assert conductance == 0   # simulated device should return conductance, susceptance as 0
     assert susceptance == 0
     resistance, reactance = session.perform_short_cable_comp()
     assert resistance == 0
-    assert reactance == 0
+    assert reactance == 0   # simulated device should return resistance,reactance as 0
 
 
 def test_read_waveform(session):
     session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1800000, 1000)
     with session.initiate():
         number_of_points_to_read = 100
-        measurements = session.read_waveform(array_size=number_of_points_to_read)[1]
-        assert measurements == number_of_points_to_read
+        measurements, actual_number_of_points = session.read_waveform(array_size=number_of_points_to_read)
+        assert len(measurements) == number_of_points_to_read
+        assert actual_number_of_points == number_of_points_to_read
 
 
 def test_send_software_trigger(session):
-    try:
-        session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 10, 5.5)
-        session.configure_multi_point(sample_count=0, sample_trigger=nidmm.SampleTrigger.SOFTWARE_TRIG, trigger_count=1)
-        with session.initiate():
-            session.send_software_trigger()    # Send_software_trigger will send triggers automatically for simulated devices. This line of code confirms there is no error while calling send_trigger function
-            session.fetch_multi_point(3)
-    except nidmm.Error as e:
-        assert False
+    session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 10, 5.5)
+    session.configure_multi_point(sample_count=0, sample_trigger=nidmm.SampleTrigger.SOFTWARE_TRIG, trigger_count=1)
+    with session.initiate():
+        session.send_software_trigger()    # Send_software_trigger will send triggers automatically for simulated devices. This line of code confirms there is no error while calling send_trigger function
+        session.fetch_multi_point(3)
 
 
 def test_reset_method(session):
--- a/src/nidmm/system_tests/test_system_nidmm.py
+++ b/src/nidmm/system_tests/test_system_nidmm.py
@@ -250,8 +250,9 @@ def test_fetch_waveform(session):
     session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1000, 10)
     with session.initiate():
         number_of_points_to_read = 10
-        measurements,actual_number_of_points = session.fetch_waveform(number_of_points_to_read)
+        measurements, actual_number_of_points = session.fetch_waveform(number_of_points_to_read)
         assert len(measurements) == number_of_points_to_read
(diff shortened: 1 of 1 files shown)


src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,65 @@ def test_configure_waveform_acquisition(session):
     assert session.range == 100
     assert session.waveform_rate == 100000
     assert session.waveform_points == 400
+
+
+def test_fetch_waveform(session):
+    session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1000, 10)
+    with session.initiate():
+        number_of_points_to_read = 10
+        measurements = session.fetch_waveform(array_size=number_of_points_to_read)[1]
```
Fladolcetta: What is the [1] for?
You dont need to call out 'array_size=number_of_points_to_read'
You can simply say 'session.fetch_waveform(number_of_points_to_read)'

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,65 @@ def test_configure_waveform_acquisition(session):
...
+        assert measurements == number_of_points_to_read
```
Fladolcetta: Shouldnt this be 'len(measurements) == number_of_points_to_read' ?

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,65 @@ def test_configure_waveform_acquisition(session):
...
+
+
+def test_fetch_waveform_error(session):
+    try:
+        session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1000, 10)
+        with session.initiate():
+            number_of_points_to_read = 100
+            session.fetch_waveform(array_size=number_of_points_to_read)   # trying to fetch points more than configured
```
Fladolcetta: No need for array_size is parameter

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,65 @@ def test_configure_waveform_acquisition(session):
...
+
+
+def test_perform_cable_compensation(session):
+    session.configure_measurement_digits(nidmm.Function.CAPACITANCE, 0.002, 5.5)
+    conductance, susceptance = session.perform_open_cable_comp()
+    assert conductance == 0
+    assert susceptance == 0
+    resistance, reactance = session.perform_short_cable_comp()
+    assert resistance == 0
+    assert reactance == 0
```
Fladolcetta: Can you add a comment on why all of these should be 0? Simulated session maybe?

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,65 @@ def test_configure_waveform_acquisition(session):
...
+
+
+def test_read_waveform(session):
+    session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1800000, 1000)
+    with session.initiate():
+        number_of_points_to_read = 100
+        measurements = session.read_waveform(array_size=number_of_points_to_read)[1]
```
Fladolcetta: Same comment about array_size and [1] as above

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,65 @@ def test_configure_waveform_acquisition(session):
...
+        assert measurements == number_of_points_to_read
```
Fladolcetta: Same comment as above. Shouldnt this be 'len(measurements)'?

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,65 @@ def test_configure_waveform_acquisition(session):
...
+
+def test_send_software_trigger(session):
+    try:
+        session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 10, 5.5)
+        session.configure_multi_point(sample_count=0, sample_trigger=nidmm.SampleTrigger.SOFTWARE_TRIG, trigger_count=1)
+        with session.initiate():
+            session.send_software_trigger()    # Send_software_trigger will send triggers automatically for simulated devices. This line of code confirms there is no error while calling send_trigger function
+            session.fetch_multi_point(3)
+    except nidmm.Error as e:
+        assert False
```
Fladolcetta: I think you are trying to assert False to ensure that the test fails if we get an exception right?

I dont think we need the try: except: block. If it errors, you will get an exception and the test will fail automatically. The try block is not useful here

src/nidmm/system_tests/test_system_nidmm.py
injaleea: fetch_waveform returns two output waveformArray and actualNumberOfPoints respectively. len(waveformArray) should be actualNumberOfPoints. inorder to reduce the number of line assigning actualNumberOfPoints  to measurements.

will change missleading parameter name  measurements

src/nidmm/system_tests/test_system_nidmm.py
injaleea: measurement is actualNumberOfPoints parameter returned by fetch_multiple. So we don't need len(measurements) here. 

To make test more clear will update test with both variables.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,64 @@ def test_configure_waveform_acquisition(session):
...
+    resistance, reactance = session.perform_short_cable_comp()
+    assert resistance == 0
+    assert reactance == 0   # simulated device should return resistance,reactance as 0
+
+
+def test_read_waveform(session):
+    session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1800000, 1000)
+    with session.initiate():
+        number_of_points_to_read = 100
+        measurements, actual_number_of_points = session.read_waveform(array_size=number_of_points_to_read)
```
Fladolcetta: Remove array_size=

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,64 @@ def test_configure_waveform_acquisition(session):
...
+
+
+def test_perform_cable_compensation(session):
+    session.configure_measurement_digits(nidmm.Function.CAPACITANCE, 0.002, 5.5)
+    conductance, susceptance = session.perform_open_cable_comp()
+    assert conductance == 0   # simulated device should return conductance, susceptance as 0
+    assert susceptance == 0
+    resistance, reactance = session.perform_short_cable_comp()
+    assert resistance == 0
+    assert reactance == 0   # simulated device should return resistance,reactance as 0
```
Fladolcetta: Move comment one line up

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,64 @@ def test_configure_waveform_acquisition(session):
...
+
+
+def test_read_waveform(session):
+    session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1800000, 1000)
+    with session.initiate():
+        number_of_points_to_read = 100
+        measurements, actual_number_of_points = session.read_waveform(array_size=number_of_points_to_read)
+        assert len(measurements) == number_of_points_to_read
```
Fladolcetta: Can we instead do:
'assert isinstance(measurements[1], float)'

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,64 @@ def test_configure_waveform_acquisition(session):
...
     assert session.waveform_rate == 100000
     assert session.waveform_points == 400
+
+
+def test_fetch_waveform(session):
+    session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1000, 10)
+    with session.initiate():
+        number_of_points_to_read = 10
+        measurements,actual_number_of_points = session.fetch_waveform(number_of_points_to_read)
+        assert len(measurements) == number_of_points_to_read
```
Fladolcetta: Can we instead do:
'assert isinstance(measurements[1], float)'

texasaggie97: I think that this should be 'in addition to' rather than 'instead of'. Verifying that the length of the array matches what was  returned for number of points read is a valid test.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,64 @@ def test_configure_waveform_acquisition(session):
...
+        assert actual_number_of_points == number_of_points_to_read
+
+
+def test_fetch_waveform_error(session):
+    try:
+        session.configure_waveform_acquisition(nidmm.Function.WAVEFORM_VOLTAGE, 10, 1000, 10)
+        with session.initiate():
+            number_of_points_to_read = 100
+            session.fetch_waveform(number_of_points_to_read)   # trying to fetch points more than configured
+    except nidmm.Error as e:
```
texasaggie97: You should 'assert False' after the fetch_waveform since you expect an error. That way you don't get a false pass if it doesn't actually error.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -244,3 +244,64 @@ def test_configure_waveform_acquisition(session):
...
+        assert e.code == -1074126845  # Max Time exceeded before operation completed
+
+
+def test_get_measurement_period():
+        with nidmm.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:4072; BoardType:PXI') as session:
+            session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 10, 5.5)
+            measurement_period = session.get_measurement_period()
+            assert measurement_period == 0.0071333333333333335  # 0.0071333333333333335 is the time required for 4072 to take a DC_VOLT measurement with range 10V on Digits_resolution 5.5
```
texasaggie97: Checks for equality with floating point numbers are brittle. See this [StackOverflow](https://stackoverflow.com/questions/5595425/what-is-the-best-way-to-compare-floats-for-almost-equality-in-python) link for a possible solution.

src/nidmm/system_tests/test_system_nidmm.py
texasaggie97: Same comment about about 'in addition to'

PR #420, opened 2017-10-05
Clean up string and buffer handling

### LINKED ISSUE #259: Investigate functions_addon.py todo for functions_params_types
There is a TODO in functions_params_types  that says to remove that section. Currently this breaks builds if it is removed. We should investigate what we need to do to remove this section.

### LINKED ISSUE #392: Fixed and Passed in ViChar functions do not correctly create mock helpers
Right now we only do IVI Dance generation for mock helpers and need to have .encode('ascii') added to value return for fixed and passed in ViChar.

Additionally ViChar functions only work with a single character. If more than one character is present it says it expects "One Character Byte"

### LINKED ISSUE #396: Parameters of type ViChar with is_buffer of True do not generate proper code
This is niDCPower_SetAttributeViInt32 from NI-DCPower's public header file:

```
   ViStatus _VI_FUNC  niDCPower_SetAttributeViInt32(
      ViSession vi,
      ViConstString channelName,
      ViAttr attribute,
      ViInt32 value);
```

This is what extracted metadata looks like, minus documentation (for readability):

```
    'SetAttributeViInt32': {
        'codegen_method': 'public',
        'returns': 'ViStatus',
        'parameters': [
            {
                'direction': 'in',
                'enum': None,
                'name': 'vi',
                'type': 'ViSession',
            },
            {
                'direction': 'in',
                'enum': None,
                'is_buffer': True,
                'name': 'channelName',
                'type': 'ViChar',
            },
            {
                'direction': 'in',
                'enum': None,
                'name': 'attributeId',
                'type': 'ViAttr',
            },
            {
                'direction': 'in',
                'enum': None,
                'name': 'attributeValue',
                'type': 'ViInt32',
            },
        ],
    },
```

This is what generated Library class says:

```
    def niDCPower_SetAttributeViInt32(self, vi, channel_name, attribute_id, attribute_value):  # noqa: N802
        with self._func_lock:
            if self.niDCPower_SetAttributeViInt32_cfunc is None:
                self.niDCPower_SetAttributeViInt32_cfunc = self._library.niDCPower_SetAttributeViInt32
                self.niDCPower_SetAttributeViInt32_cfunc.argtypes = [ViSession_ctype, ViChar_ctype, ViAttr_ctype, ViInt32_ctype]  # noqa: F405
                self.niDCPower_SetAttributeViInt32_cfunc.restype = nidcpower.python_types.ViStatus
        return self.niDCPower_SetAttributeViInt32_cfunc(vi, channel_name, attribute_id, attribute_value)
```

Specifically, when setting argtypes, argument channelName is being given a type ViChar_ctype even though the key is_buffer is set to True. That means a string cannot be passed to it at all. Note that type is mismatched with the header file (ViConstString vs ViChar) but that may be an inconsistency with the source of the extracted metadata – the CVI FP file.

The same issue likely exists in many other places.

### LINKED ISSUE #404: Multiple issues with revision_query()
```
with nidmm.Session('name') as session:
    print(session.revision_query())
```
returns (1,0) for a simulate 4080 and (1,1) for real 408x.
For 4072(both simulated and real) returns following error 
```
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python36-32\lib\sit
e-packages\nidmm\session.py", line 2757, in revision_query
    return python_types.ViChar(instrument_driver_revision_ctype.value), python_t
ypes.ViChar(firmware_revision_ctype.value)
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python36-32\lib\sit
e-packages\nidmm\python_types.py", line 20, in __new__
    return int(val)
ValueError: invalid literal for int() with base 10: b'r'
```

texasaggie97:
* All string types are consolidated to ViChar
  * Remove [] from name
  * Change ViString, ViConstString & ViRsrc to ViChar
  * Set 'is_buffer' to true
* Encode/decode page is now determined by variable on Session `self._encoding`
  * Defaults to 'windows-1251'
  * No mechanism to change in Session creation
* Add size information to revistion_query
  * Allows enabling revision_query test
* Simplify method codegen in Session
* No longer need to 'coerce' types in an array to the correct Python type
* Separate setting defaults for 'size' and 'is_buffer'
* Fix #396 
* Fix #404
* Fix #392 
* Fix #259 
* Travis
* System tests
* Hand testing


build/helper/helper.py
```diff
@@ -32,20 +31,6 @@ def sorted_attrs(a):
     return sorted(a, key=lambda k: a[k]['name'])
 
 
-# We need this to allow us to dynamically add and remove a folder to the search
```
marcoskirsch: It'd be best if this went in own PR

build/helper/metadata_add_all.py
```diff
@@ -51,12 +51,8 @@ def _add_ctypes_variable_name(parameter):
 def _add_ctypes_type(parameter):
     '''Adds a ctypes_type key/value pair to the parameter metadata for calling into the library'''
     parameter['ctypes_type'] = parameter['type']
-    if parameter['direction'] == 'out':
-        if parameter['type'] == 'ViString' or parameter['type'] == 'ViRsrc' or parameter['type'] == 'ViConstString':
-            # These are defined as c_char_p which is already a pointer!
-            parameter['ctypes_type_library_call'] = parameter['ctypes_type']
-        else:
-            parameter['ctypes_type_library_call'] = "ctypes.POINTER(" + parameter['ctypes_type'] + ")"
+    if parameter['direction'] == 'out' or parameter['is_buffer'] is True:
```
marcoskirsch: remove "is True" (pythonic)

build/helper/metadata_add_all.py
```diff
@@ -77,13 +73,27 @@ def _add_is_error_handling(f):
 
 def _add_buffer_info(parameter):
     '''Adds buffer information to the parameter metadata iff 'size' is defined else assume not a buffer'''
+
+    # We are going to treat ViChar[], ViString, ViConstString and ViRsrc the same. For simplicity in the generated
```
marcoskirsch: reword comment for clarity:

For simplicity, we are going to treat ViChar[], ViString, ViConstString, and ViRsrc the same: As ViChar and is_buffer True

build/helper/metadata_add_all.py
```diff
@@ -77,13 +73,27 @@ def _add_is_error_handling(f):
...
+    # code, we will treat them as 'ViChar' and is_buffer true
+    t = parameter['type']
+    if (t.find('[ ]') > 0) or (t.find('[]') > 0) or t == 'ViString' or t == 'ViConstString' or t == 'ViRsrc':
+        parameter['type'] = 'ViChar'
```
marcoskirsch: I wish we weren't modifying metadata, only adding to it. I do see you preserve original info in 'original_type'.

Suggestion: In metadata extraction, instead of type, use something else (c_api_type?), then add type. I can see this being a different PR however.

If you agree, let's open the Issue.

build/helper/metadata_add_all.py
```diff
@@ -77,13 +73,27 @@ def _add_is_error_handling(f):
...
+        parameter['original_type'] = t
+        parameter['is_buffer'] = True
+
     try:
         parameter['size']
```
marcoskirsch: FWIW, more pythonic is to say
if 'size' not in parameter:
    # Not populated, assume {'mechanism': 'fixed', 'value': 1}
    ....

Of course this was already that way so you can choose to not fix.

build/helper/metadata_add_all.py
```diff
@@ -300,24 +309,25 @@ def test_add_all_metadata_simple():
...
+                    'ctypes_type_library_call': 'ctypes.POINTER(ViChar)',
                     'direction': 'in',
                     'documentation': {
                         'description': 'The channel to call this on.'
                     },
                     'is_repeated_capability': True,
                     'enum': None,
-                    'python_type': 'str',
-                    'is_buffer': False,
+                    'python_type': 'int',
```
marcoskirsch: shouldn't this be str?

build/templates/mock_helper.py.mako
```diff
@@ -76,8 +72,7 @@ ivi_dance_size_param = helper.find_size_parameter(ivi_dance_param, params)
             raise MockFunctionCallError("${c_function_prefix}${func_name}", param='${ivi_dance_param['name']}')
         if ${ivi_dance_size_param['python_name']} == 0:
             return len(self._defaults['${func_name}']['${ivi_dance_param['name']}'])
-        t = ${module_name}.visatype.${ivi_dance_param['ctypes_type']}(self._defaults['${func_name}']['${ivi_dance_param['name']}'].encode('ascii'))
-        ${ivi_dance_param['python_name']}.value = ctypes.cast(t, ${module_name}.visatype.${ivi_dance_param['ctypes_type']}).value
+        ${ivi_dance_param['python_name']}.value = self._defaults['${func_name}']['${ivi_dance_param['name']}'].encode('ascii')
```
marcoskirsch: 'ascii' still hardcoded here.
I suppose we deal with this later? I ask because you made it a variable in Session... with value 'windows-1251'

docs/nidcpower/functions.rst
```diff
@@ -398,7 +398,7 @@ nidcpower.Session methods
 
         
 
-    :type sequence_name: str
+    :type sequence_name: int
```
marcoskirsch: this doesn't seem right

generated/nimodinst/session.py
```diff
@@ -165,42 +166,177 @@ def close(self):
     ''' These are code-generated '''
 
     def _close_installed_devices_session(self, handle):
+        '''_close_installed_devices_session
```
marcoskirsch: What's up with this change? Is this an independent thing?

marcoskirsch:
This is a vast improvement.

build/helper/metadata_add_all.py
texasaggie97: "[As a rule of thumb, you should always use is with the built-in constants True, False and None](https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true)"

build/helper/helper.py
texasaggie97: Put back in

build/helper/metadata_add_all.py
texasaggie97: I think that is a reasonable change, but it should be it's own PR. It would touch many places just for that.

build/helper/metadata_add_all.py
texasaggie97: I started using if 'something' not in something_else but went with this to be consistent to other things. I figured if I changed all of them to the if way, you'd tell me it goes in a different PR. :)

build/helper/metadata_add_all.py
texasaggie97: No. ViChar is int. See #422 where I handle it in documentation.

build/templates/mock_helper.py.mako
texasaggie97: I left it that way on purpose. For tests, we can guarantee that the strings are ascii. Once we actually support changing it, we can make it so mock_helper uses what session uses.

docs/nidcpower/functions.rst
texasaggie97: See #422

generated/nimodinst/session.py
texasaggie97: I just copied the part of the template from build/templates/session.py.mako that generated the method. Apparently we weren't adding the docstring before.

build/helper/metadata_add_all.py

build/helper/metadata_add_all.py
marcoskirsch: Good to know!

build/helper/metadata_add_all.py
marcoskirsch: > I figured if I changed all of them to the if way, you'd tell me it goes in a different PR. :)

You're probably right. Easy fix for own PR if someone feels like it.

build/helper/metadata_add_all.py
marcoskirsch: I thought that the 'python_type' metadata value refers to the type of the parameter passed in to the Python API. But now that I think of it, that wasn't the case with buffers before.

How would you define this metadata value?

generated/nimodinst/session.py
marcoskirsch: Shouldn't this be a separate PR?

build/helper/metadata_add_all.py
marcoskirsch: We've settled this offline, and came up with the following definitions for metadata, whose names can be improved but would happen in a separate PR.

Python type metadata:

* *type*: “Simple” type of a parameter in the driver’s C API functions.
* *python_type*: “Simple” type of a parameter for the Python API (methods of Session). Can be int, float, boolean, enum (but not lists nor str)
* *is_buffer*: Is it a list or str
* *enum*: Which enum to use in the Python API.
* *direction*: Is it an input or output parameter.

PR #424, opened 2017-10-06
Update functions_addon.py for nidcpower functions that use strings

Fladolcetta:
[x ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/<reponame>/blob/master/CONTRIBUTING.md).

* Update functions_addon.py for nidcpower functions that use strings
* Add get_channel_name to code generation

* Built and ran.
* Ran unit tests


Fladolcetta:
This WILL fail the system tests.
This is because the C function for initialize with channels will NOT allow you to have a blank string, Null string, or fake string without a real device. This is a bug in the C api. We will need to figure out how to use simulated devices for testing... Maybe use the LV version of this function?

In my system if I create a DAQmx Simulated 4110 called 'FakeDevice' this works as expected.

generated/nidcpower/ctypes_types.py
```diff
@@ -0,0 +1,69 @@
+import ctypes
```
texasaggie97: This file was deleted in a different PR, so you shouldn't be adding it back.

Fladolcetta: Pulled latest and updated generated

PR #439, opened 2017-10-09
Add nidcpower API defaults and enums

### LINKED ISSUE #300: Add default parameter metadata to NI-DCPower

Fladolcetta:
[x ] This contribution adheres to [CONTRIBUTING.md]

Adds defaults and new required enums for some functions

* Fixes #300 
* Brings API closer to our desired final API

Built and ran unit tests


docs/nidcpower/functions.rst
```diff
@@ -79,7 +79,7 @@ nidcpower.Session methods
 
         
 
-    :type aperture_time: float
+    :type aperture_time: :py:data:`nidcpower.ApertureTimeUnits`
```
marcoskirsch: This used to be float?
Hopefully that was a (now fixed) bug.

generated/nidcpower/session.py
```diff
@@ -2536,11 +2536,13 @@ def configure_aperture_time(self, aperture_time, units):
...
         '''
-        error_code = self._library.niDCPower_ConfigureApertureTime(self._vi, self._repeated_capability.encode(self._encoding), aperture_time, units)
+        if type(aperture_time) is not enums.ApertureTimeUnits:
+            raise TypeError('Parameter mode must be of type ' + str(enums.ApertureTimeUnits))
+        error_code = self._library.niDCPower_ConfigureApertureTime(self._vi, self._repeated_capability.encode(self._encoding), aperture_time.value, units)
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
-    def fetch_multiple(self, timeout, count):
+    def fetch_multiple(self, count, timeout=1):
```
marcoskirsch: If the parameter is float, you should specify the literal as 1.0 so the runtime doesn't need to convert.
Applies everywhere.

generated/nidcpower/session.py
```diff
@@ -2487,7 +2487,7 @@ def get_error_description(self, error_code):
 
     ''' These are code-generated '''
 
-    def configure_aperture_time(self, aperture_time, units):
+    def configure_aperture_time(self, units, aperture_time=enums.ApertureTimeUnits.SECONDS):
```
marcoskirsch: This is wrong.

generated/nidcpower/session.py
```diff
@@ -3407,7 +3409,7 @@ def _set_attribute_vi_string(self, attribute_id, attribute_value):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
-    def set_sequence(self, values, source_delays, size):
+    def set_sequence(self, source_delays, size, values=''):
```
marcoskirsch: values should be None or [], not blank string. values is an array!

generated/nidcpower/session.py
```diff
@@ -3556,7 +3558,7 @@ def commit(self):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
-    def configure_digital_edge_measure_trigger(self, input_terminal, edge):
+    def configure_digital_edge_measure_trigger(self, edge, input_terminal=enums.DigitalEdge.RISING):
```
marcoskirsch: You have these off by one errors in many places. The edge is rising, not the input terminal!

generated/nidcpower/session.py
Fladolcetta: Fixed

generated/nidcpower/session.py
Fladolcetta: Fixed

generated/nidcpower/session.py
Fladolcetta: Fixed

generated/nidcpower/session.py
Fladolcetta: Fixed

generated/nidcpower/session.py
```diff
@@ -3407,7 +3409,7 @@ def _set_attribute_vi_string(self, attribute_id, attribute_value):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
-    def set_sequence(self, values, source_delays, size):
+    def set_sequence(self, source_delays, size, values=None):
```
marcoskirsch: size should not be here

src/nidcpower/metadata/functions_addon.py
```diff
@@ -70,6 +78,8 @@
     'revision_query':               { 'parameters': { 1: { 'size': {'mechanism':'fixed', 'value':256}, }, 
                                                       2: { 'size': {'mechanism':'fixed', 'value':256}, }, }, }, # From documentation
     'GetChannelName':               { 'parameters': { 3: { 'size': {'mechanism':'ivi-dance', 'value':'bufferSize'}, }, }, },
+    'SetSequence':                  { 'parameters': { 1: { 'size': {'mechanism':'passed-in', 'value':'Size'}, }, }, },
```
marcoskirsch: The problem above is probably that you misspelled this parameter.
The code-generator should die if mechanism is "passed-in" _and_ value is not found in the parameter list. If you confirm this was the issue, we can look into adding that check.

Fladolcetta: This was correctly spelled, but will be resolved by the len feature @texasaggie97 is working on.

generated/nidcpower/session.py
Fladolcetta: This will be resolved by the len feature @texasaggie97 is working on.

PR #441, opened 2017-10-09
Remove revision_query methods

### LINKED ISSUE #393: Do not expose revision_query() method
This method just returns the value of two read-only attributes.
As a point of reference, NI-DCPower's .NET API decided not to expose it for that same reason.

Fladolcetta:
[ x] This contribution adheres to [CONTRIBUTING.md]

Removes revision_query methods from APIs, a function that simply returns two attribute values. It doesn't add enough value to the Python APIs so we will not include it

* Fix #393 

Ran system tests


Fladolcetta:
Conflicts resolved

PR #447, opened 2017-10-10
Handle rep caps properly in documentation

### LINKED ISSUE #371: Add information about methods that support repeated capability to documentation
Per #297 we are removing channel input from methods. This makes it non-obvious which methods support channels and which don't.

This issue is for somehow displaying this information in the generated documentation. Function metadata contains this information under the 'has_repeated_capability' key.

### LINKED ISSUE #428: Documentation not matching with method definition
**Are we not creating channel_name parameter for nidcpower ?**
Channel_name parameter is not listed in any of the method definition. But there exists documentation for the same . 

Arg section for configure_aperture_time describes about **channel_name** parameter which is not available on method definition

```
    def configure_aperture_time(self, aperture_time, units):
        '''configure_aperture_time

        Vistatus configure_aperture_time(ViSession vi, ViConstString
        channelName, ViReal64 apertureTime, ViInt32 units);

        Configures the aperture time on the specified channel(s).

        The supported values depend on the **units**. Refer to the *Aperture
        Time* topic for your device in the *NI DC Power Supplies and SMUs Help*
        for more information. In general, devices support discrete
        **apertureTime** values, and if you configure **apertureTime** to some
        unsupported value, NI-DCPower coerces it up to the next supported value.

        Refer to the *Measurement Configuration and Timing* or *DC Noise
        Rejection* topic for your device in the *NI DC Power Supplies and SMUs
        Help* for more information about how to configure your measurements.

        **Related Topics:**

        `Aperture
        Time <http://zone.ni.com/reference/en-XX/help/370384T-01/dmm/aperture/>`__

        Note:
        This function is not supported on all devices. Refer to `Supported
        Functions by
        Device <javascript:LaunchMergedHelp('NI_DC_Power_Supplies_Help.chm',%20'NIDCPowerCRef.chm',%20'supportedFunctions.html')>`__
        for more information about supported devices.

        Args:
            channel_name (string): Specifies the output channel(s) to which this configuration value
                applies. Specify multiple channels by using a channel list or a channel
                range. A channel list is a comma (,) separated sequence of channel names
                (for example, 0,2 specifies channels 0 and 2). A channel range is a
                lower bound channel followed by a hyphen (-) or colon (:) followed by an
                upper bound channel (for example, 0-2 specifies channels 0, 1, and 2).
                In the Running state, multiple output channel configurations are
                performed sequentially based on the order specified in this parameter.
            aperture_time (float): Specifies the aperture time. Refer to the *Aperture Time* topic for your
                device in the *NI DC Power Supplies and SMUs Help* for more information.
            units (int): Specifies the units for **apertureTime**.
                **Defined Values**:

                +----------------------------------------+------------------------------+
                | NIDCPOWER_VAL_SECONDS (1028)           | Specifies seconds.           |
                +----------------------------------------+------------------------------+
                | NIDCPOWER_VAL_POWER_LINE_CYCLES (1029) | Specifies Power Line Cycles. |
                +----------------------------------------+------------------------------+
        '''
        error_code = self._library.niDCPower_ConfigureApertureTime(self._vi, self._repeated_capability.encode(self._encoding), aperture_time, units)
        return
```

texasaggie97:
* Add comment in documentation when a method takes a repeated capabilities parameter
* Add comment in documentation when an attribute is repeated capabilities based
  * Needs updated metadata - fix submitted for extractor
  * Updated metadata will be a separate PR
  * Fix verified to add correct content

* Fix #428 
* Fix #371

* Travis
* System tests


build/helper/documentation_helper.py
```diff
@@ -306,6 +308,17 @@ def _format_type_for_rst_documentation(param, config):
     return p_type
 
 
+rep_cap_method_desc = '''
+This method uses repeated capabilities (usually channels). If you call this method on the base session, then
```
marcoskirsch: I think "base session" is confusing. Here's suggested edit:

```
This method uses repeated capabilities (usually channels). If called directly on the nidmm.Session object, then it will apply to the full session. You can specify what repeated capabilities to use using the Python
index notation:

.. code:: python

    session['0,1'].{0}({1})

```

I asked for feedback from @thejcannon and @MrElusive who are very familiar with NI-DCPower and Python.

generated/niswitch/session.py
```diff
@@ -1197,9 +1179,6 @@ def _set_attribute_vi_real64(self, attribute_id, attribute_value):
         the penalty of redundant instrument I/O.
 
         Args:
-            channel_name (string): Some attributes are unique per channel. For these, pass the name of the
```
marcoskirsch: docstring should say something about repeated capabilities as well

build/helper/metadata_filters.py
```diff
@@ -6,9 +6,9 @@ def filter_codegen_functions(functions):
     return {k: v for k, v in functions.items() if v['codegen_method'] != 'no'}
 
 
-def filter_input_parameters(parameters, session_name='vi'):
+def filter_input_parameters(parameters):
     '''Returns list of parameters that includes only input parameters, except the session parameter if it exists'''
```
marcoskirsch: repeated capability as well...
But... how about getting rid of this function altogether and use the all-powerful filter function?

build/templates/attributes.rst.mako
```diff
@@ -28,6 +28,21 @@ table_contents = [
          ]
 table = helper.as_rest_table(table_contents)
 
+rep_cap_attr_desc = '''
+This attribute requires repeated capabilities (usually channels). If set or get directly on the
```
marcoskirsch: Use "property" since this refers to the Python API property session.foo. Also should not say "requires".

build/helper/metadata_filters.py
texasaggie97: Which all powerful filter function? The one added when adding the 'len' mechanism? That one only looks for a specific mechanism.

build/templates/attributes.rst.mako
texasaggie97: Updated

build/helper/metadata_filters.py
marcoskirsch: filter_parameters() in build/helper/codegen_helper.py

texasaggie97: Do you want me to make that change in this PR? There would be other filter functions that should be removed as well. Do I get rid of all of them?

marcoskirsch: > Do you want me to make that change in this PR?

Up to you.

> There would be other filter functions that should be removed as well. Do I get rid of all of them?

The ones for filtering parameters, yeah, we can get rid of most if not all.

texasaggie97:
test this please

PR #448, opened 2017-10-10
Improve RST unit tests

### LINKED ISSUE #442: test_get_function_rst relies on trailing whitespace, which editors remove on save
The unit test added in #423 asserts that uses a multi-line Python constant as expected value. Problem is 
that it contains trailing whitespace and we configure our editors to remove trailing whitespace. So save the file, break the test.

@texasaggie97 says his editor (SlickEdit) appears to be clever enough to leave multiline Python string constants alone.
Mine is not that smart (Visual Studio Code).

Specifically:

```
@@ -573,7 +570,7 @@ def test_get_function_rst():
 
     Returns the **ID** of selected Turtle Type.
 
-
+    
 
     .. note:: The RAPHAEL Turtles dont have an ID.
 
@@ -604,7 +601,7 @@ def test_get_function_rst():
 
             Returns the **ID** of selected turtle.
 
-
+            
```

(you cannot see it here, you need to select in order to see where this whitespace is)

marcoskirsch:

* Makes string comparison in unit test pass when the strings differ only by trailing whitespace (RST doesn't care)
* Makes sting comparison in unit test pass when there are different blank strings (RST doesn't care)
* Improves errors in the assertions.

* Fix #442 

tox
Made a test fail, looked at output.

### DIFF
--- a/build/helper/documentation_helper.py
+++ b/build/helper/documentation_helper.py
@@ -11,12 +11,12 @@ import sys
 
 
 # Python 2/3 compatibility
-def normalize_string_type(d):
+def _normalize_string_type(d):
     '''Normalize string type between python2 & python3'''
     if sys.version_info.major < 3:
         if type(d) is dict:
             for k in d:
-                d[k] = normalize_string_type(d[k])
+                d[k] = _normalize_string_type(d[k])
         elif type(d) is str:
             d = d.decode('utf-8')
     return d
@@ -45,7 +45,7 @@ def get_indented_docstring_snippet(d, indent=4):
             ret_val += '\n'
             if len(l.rstrip()) > 0:
                 ret_val += (' ' * indent)
-        ret_val += normalize_string_type(l.rstrip())
+        ret_val += _normalize_string_type(l.rstrip())
     return ret_val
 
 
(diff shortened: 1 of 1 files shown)


marcoskirsch:
@texasaggie97 : alternatively we could trim in the function itself, perhaps that's more correct.
Thoughts?

build/helper/documentation_helper.py
```diff
@@ -631,7 +637,7 @@ def test_get_function_rst():
 
             
 ''' # noqa
-    assert actual_function_rst == expected_fuction_rst
+    assert _remove_trailing_whitespace(actual_function_rst) == _remove_trailing_whitespace(expected_fuction_rst)
```
texasaggie97: I think it would be more useful to compare line by line. That way the error can pinpoint exactly where the mismatch is. That also lets the comparison be more correct as far as what matters to rst. Like 1 empty line is the same as 42 empty lines.

marcoskirsch: ok

build/helper/documentation_helper.py
```diff
@@ -469,6 +469,22 @@ def as_rest_table(data, header=True):
 
 
 # Unit Tests
+
+
+def _remove_trailing_whitespace(s):
+    '''Removes trailing whitespace and empty lines in multi-line strings. https://stackoverflow.com/a/17350806/316875'''
+    return re.sub(r'\s+$', '', s, flags=re.M)
```
texasaggie97: Does this remove all empty lines? If it does then that means there could be breaking changes that these tests wouldn't catch. 0 empty lines does not mean the same as 1 empty line, but 1 or more empty lines do mean the same thing.

I think this is still better than just asserting on the two multiline strings being equal, so I would be okay with accepting this and making it better later, if needed.

marcoskirsch: Pain in the neck...
I will merge as is because it's useful and opened https://github.com/ni/nimi-python/issues/451 to follow up.

PR #467, opened 2017-10-12
adding Exportsignal enum for export signal method to enums_addon.py

### LINKED ISSUE #388: Populate nidcpower functions_addon.py
We are generating using untouched extracted metadata. API isn't usable without adding extra metadata via functions_addon.py.

injaleea:

~~[] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~

Adding enum signal needed by export_signal method

* Fix #388 

tox, systemtests


src/nidcpower/metadata/functions_addon.py
```diff
@@ -52,7 +52,10 @@
     'ConfigureDigitalEdgeSourceTrigger':            { 'parameters': { 2: { 'enum': 'DigitalEdge',                 }, }, },
     'ConfigureDigitalEdgeStartTrigger':             { 'parameters': { 2: { 'enum': 'DigitalEdge',                 }, }, },
     'SendSoftwareEdgeTrigger':                      { 'parameters': { 1: { 'enum': 'SendSoftwareEdgeTriggerType', }, }, },
-    'WaitForEvent':                                 { 'parameters': { 2: { 'enum': 'Event', }, }, },
+    'WaitForEvent':                                 { 'parameters': { 2: { 'enum': 'Event',                       }, }, },
+    'Measure':                                      { 'parameters': { 2: { 'enum': 'MeasurementTypes',            }, }, },
+    'QueryOutputState':                             { 'parameters': { 2: { 'enum': 'OutputStates',                }, }, },
+    'ExportSignal':                                 { 'parameters': { 1: { 'enum': 'Signals',                     }, }, },
```
Fladolcetta: Do we have system tests for these to add to this review?

src/nidcpower/metadata/functions_addon.py
```diff
@@ -53,8 +53,6 @@
     'ConfigureDigitalEdgeStartTrigger':             { 'parameters': { 2: { 'enum': 'DigitalEdge',                 }, }, },
     'SendSoftwareEdgeTrigger':                      { 'parameters': { 1: { 'enum': 'SendSoftwareEdgeTriggerType', }, }, },
     'WaitForEvent':                                 { 'parameters': { 2: { 'enum': 'Event',                       }, }, },
-    'Measure':                                      { 'parameters': { 2: { 'enum': 'MeasurementTypes',            }, }, },
-    'QueryOutputState':                             { 'parameters': { 2: { 'enum': 'OutputStates',                }, }, },
     'ExportSignal':                                 { 'parameters': { 1: { 'enum': 'Signals',                     }, }, },
```
injaleea: Removed MeasurementTypes adn QueryOutput state from this PR since there exist another PR for the same

src/nidcpower/metadata/functions_addon.py
injaleea: Tests are being created. Will submit in another PR.

src/nidcpower/metadata/enums.py
```diff
@@ -566,6 +566,87 @@
             },
         ],
     },
+    'Signals': {
```
marcoskirsch: Where did this term Signals come from? Is it pre-established for NI-DCPower or another one of the APIs for some other ADE? Or was it derived from "niDCPower_ExportSignal"?

I do see several APIs have it.

injaleea: Derived from niDCPower_ExportSignal. should it be changed to Export Signal ?

marcoskirsch: It should be derived directly from the corresponding typedef name in LabVIEW. If you find yourself having to come up with names / documentation, then we need to be careful.

Fladolcetta: All,

We need to have the names and descriptions of the enums come from the LabVIEW API rather than the C API. This is because all of our extraction for enums already comes from the LabVIEW API.

This will cause a disconnect in the documentation, which we need to look at in #502.

Please update the enums to follow the LabVIEW naming and documentation conventions.

injaleea: Updated internal files to have names following LabVIEW naming convensions, instead C.

Should we change it here since it is generated. Will extraction tool take care of this?

marcoskirsch: Extraction tool will generate the new files, but somebody needs to commit them to GitHub. @texasaggie97 usually handles it.

texasaggie97:
We have changed our policy on how to handle enums that are missing.

Instead of adding them to the source so that they get added to metadata, instead, they should be added via enum_addons. This means they will be hand maintained in nimi-python. This isn't ideal but it reduces the overhead and delay getting them added or integrated into the source.

bhaswath:
@texasaggie97 How will we be handling enums that have not yet been added for NI-Fgen?

marcoskirsch:
Enums that don't come in the extracted metadata (that is, enums that are used exclusively for functions) will need to be added to src/nifgen/metadata/enums_addon.py.

See https://github.com/ni/nimi-python/pull/582 for reference.

PR #468, opened 2017-10-12
Adding more dcpower system tests

### LINKED ISSUE #390: Write nidcpower system tests
They have not been written. Placeholder file added in #387

injaleea:

Adding systemtests for nidcpower.

- test_commit_after_initiate
- test_get_self_cal_last_date_and_time
- test_get_self_cal_last_temp
- test_read_current_temperature
- test_reset_device
- test_reset_with_default
- test_reset
- test_disable

TODO: List of issues.

Partial fix #390 

tox, systemtests.

### DIFF
--- a/src/nidcpower/system_tests/test_system_nidcpower.py
+++ b/src/nidcpower/system_tests/test_system_nidcpower.py
@@ -38,3 +38,68 @@ def test_get_error(session):
     except nidcpower.Error as e:
         assert e.code == -1074135027  # Error : Attribute is read-only.
         assert e.description.find('Attribute is read-only.') != -1
+
+
+def test_commit_after_initiate(session):
+    try:
+        with session.initiate():
+            session.commit()
+    except nidcpower.Error as e:
+        assert False
+
+
+def test_get_self_cal_last_date_and_time(session):
+    try:
+        session.get_self_cal_last_date_and_time()
+        assert False
+    except nidcpower.Error as e:
+        assert e.code == -1074118643  # This operation is invalid on a simulated device.
+
+
+def test_get_self_cal_last_temp(session):
+    try:
+        session.get_self_cal_last_temp()
+        assert False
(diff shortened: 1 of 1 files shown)


src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -40,6 +40,71 @@ def test_get_error(session):
         assert e.description.find('Attribute is read-only.') != -1
 
 
+def test_commit_after_initiate(session):
```
marcoskirsch: What is this testing? If this is testing that you can call commit after initiate was called in the NI-DCPower API, then that's outside the scope of what should be tested in this file. We aren't testing the NI-DCPower driver, just the Python bindings.

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -40,6 +40,71 @@ def test_get_error(session):
...
+    try:
+        with session.initiate():
+            session.commit()
+    except nidcpower.Error as e:
+        assert False
+
+
+def test_get_self_cal_last_date_and_time(session):
```
marcoskirsch: These tests are unfortunate because they don't really tell us much... for example, the interesting aspect of session.get_self_cal_last_date_and_time() is that it returns several values - yet that's not being tested.

I don't have a great suggestion on how to improve this, however. But it is a definite hole in our coverage.

injaleea: Probably we can add these kind of tests to another python file which will be run with our internal testsuites. 

Thoughts ?

marcoskirsch: I don't think it's worth the complexity of having different tests in NI's internal test suites (which have hardware) and the stuff we run on nimi-bot (CI system tests).

This might simply end up being a gap in our test coverage.

PR #487, opened 2017-10-18
Improve returned error for invalid properties.

### LINKED ISSUE #466: Errors when you have a typo in a property is misleading.
Running code snippet
```
with nidcpower.Session('4143', '', False, 'Simulate=1, DriverSetup=Model:4143; BoardType:PXIe') as session:
    channel = session['0']
    default_output_function = channel.output_function
    channel.function = nidcpower.OutputFunction.DC_CURRENT
```
retruning Type Error 

> E           TypeError: <nidcpower.session._RepeatedCapability object at 0x0320D3 70> is a frozen class

When setting output function for the second time to DC_CURRENT

Fladolcetta:
[x ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

[ x] I've added any unit tests that are applicable for this pull request

Gives a better error for setting an invalid attribute. This error is identical to the error one might get when getting an invalid attribute.

* Fix #466 

Tox and system tests.


build/templates/session.py.mako
```diff
@@ -126,7 +126,7 @@ init_call_params = helper.get_params_snippet(init_function, helper.ParameterUsag
 
     def __setattr__(self, key, value):
         if self._is_frozen and key not in dir(self):
-            raise TypeError("%r is a frozen class" % self)
+            raise AttributeError("'Session' object has no attribute %r" % key)
```
marcoskirsch: From the Python side of things, the proper term is "property" not attribute. Please use that.
Also, use .format() which is what we've been using everywhere rather than the old printf style syntax.

marcoskirsch: I stand corrected.

>>> class Foo:
...  a = 2
... 
>>> f = Foo()
>>> f.a
2
>>> f.b
AttributeError: Foo instance has no attribute 'b'
>>>

build/templates/session.py.mako
```diff
@@ -126,7 +126,7 @@ init_call_params = helper.get_params_snippet(init_function, helper.ParameterUsag
 
     def __setattr__(self, key, value):
         if self._is_frozen and key not in dir(self):
-            raise TypeError("%r is a frozen class" % self)
+            raise AttributeError("'Session' object has no attribute '{0}'".format(key))
```
marcoskirsch: What error do you get if you write the wrong attribute on:

    session['0'].potato

Thing is, session['0'] is not a Session. This would also break should we ever create Session subclasses. A more robust solution would be to use type(self).

generated/nifake/tests/test_session.py
```diff
@@ -765,12 +765,14 @@ def test_add_properties_to_session_error(self):
             try:
                 session.nonexistent_property = 5
```
marcoskirsch: we need to also test session['4']. session.nonexistent_property

(also, shouldn't this be session.non_existent_property?)

generated/nimodinst/attributes.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
```
marcoskirsch: This file is gone in master, don't re-add.

src/nimodinst/templates/session.py.mako
```diff
@@ -64,8 +64,8 @@ class Device(object):
             return object.__getattribute__(self, name).__getitem__(None)
 
     def __setattr__(self, name, value):
-        if self._is_frozen and name not in ['_is_frozen', 'index']:
-            raise TypeError("%s is not writable" % name)
+        if self._is_frozen:
+            raise AttributeError("Set attribute not supported.")
```
marcoskirsch: This is misleading.

* Text is hard to decifer "Set attribute not supported."
* You get the same error for a real attribute as you do for "potato"

generated/nimodinst/attributes.py
Fladolcetta: Fixed

src/nimodinst/templates/session.py.mako
Fladolcetta: Should I change the text? "Set attribute is not a supported function."

Also per our discussion yesterday, if a customer tries to set any attribute (real or other wise) they will get the above error. I feel that this makes sense and gives the customer an error that they would expect. While I agree it might be nice to say "potato is not an attribute, and you also can't set attributes", the difficulty to do this is high due to the dir() issues discussed earlier, and the value add is very minimal.

 If a customer tries to call "Get Attribute" with potato they will get an expected error that says that the attribute does not exist (which is tested in test_modinst.py).

build/templates/session.py.mako
Fladolcetta: Line 177 of test_modinst.py currently tests that case.

  >  def test_cannot_add_properties_to_device(self):
        with nimodinst.Session('') as session:
            try:
                session[0].nonexistent_property = 5
                assert False
            except AttributeError as e:
                assert str(e) == "Set attribute not supported."
            try:
                session[0].nonexistent_property
                assert False
            except AttributeError as e:
                assert str(e) == "'Device' object has no attribute 'nonexistent_property'"

Where would I do type(self)?

src/nimodinst/templates/session.py.mako
Fladolcetta: Fixed per our discussion

generated/nifake/tests/test_session.py
Fladolcetta: Fixed

build/templates/session.py.mako
Fladolcetta: Fixed

generated/nifake/tests/test_session.py
```diff
@@ -763,14 +763,31 @@ def test_set_attribute_error(self):
     def test_add_properties_to_session_error(self):
         with nifake.Session('dev1') as session:
             try:
-                session.nonexistent_property = 5
+                session.non_existent_property = 5
                 assert False
-            except TypeError as e:
+            except AttributeError as e:
+                assert str(e) == "'Session' object has no attribute 'non_existent_property'"
+                pass
```
marcoskirsch: not needed, applies elsewhere

generated/nifake/tests/test_session.py
```diff
@@ -763,14 +763,31 @@ def test_set_attribute_error(self):
...
+            try:
+                value = session.non_existent_property  # noqa: F841
```
marcoskirsch: Best to have more granular tests rather than a single test that tries several things.

generated/nifake/tests/test_session.py
Fladolcetta: fixed

generated/nifake/tests/test_session.py
Fladolcetta: fixed

PR #506, opened 2017-10-18
Fix enum error

### LINKED ISSUE #503: Bad error when passing wrong enum type
I ran into this by accident:

```
Z:\>python src\nidcpower\examples\nidcpower_advanced_sequence.py
  File "src\nidcpower\examples\nidcpower_advanced_sequence.py", line 24, in <module>
    session.current_level_autorange = nidcpower.VoltageLevelAutorange.ON
  File "C:\Python27\lib\site-packages\nidcpower\session.py", line 2905, in __setattr__
  File "C:\Python27\lib\site-packages\nidcpower\attributes.py", line 69, in __set__
    raise TypeError('must be nidcpower.' + str(self._attribute_type.__name__) + ' not ' + str(type(value).__name__))
TypeError: must be nidcpower.CurrentLevelAutorange not VoltageLevelAutorange
```

**nidcpower.CurrentLevelAutorange** is inconsistent with **VoltageLevelAutorange**. They should both include nidcpower. prefix, or neither.

Fladolcetta:

[x ] I've added any unit tests that are applicable for this pull request

- Makes error reporting more consistent to always simply return the type() of the expected enum and the type() of the passed in variable.
- Updates generated for Switch which was not updated with pull request to remove IsScanning

* Fix #503 

Ran tox.


docs/niswitch/functions.rst
```diff
@@ -87,7 +87,7 @@ niswitch.Session methods
     semi-colons. For example, the following scan list will scan the first
     three channels of a multiplexer: com0->ch0; com0->ch1; com0->ch2; Refer
     to Scan Lists for more information on scan list syntax To see the status
-    of the scan, call either :py:func:`niswitch.is_scanning` or
+    of the scan, call either :py:func:`niswitch.IsScanning` or
```
marcoskirsch: I'm confused as to why this shows up on your diff.

generated/nifake/attributes.py
```diff
@@ -66,7 +66,7 @@ def __get__(self, session, session_type):
 
     def __set__(self, session, value):
         if type(value) is not self._attribute_type:
```
marcoskirsch: I think this is wrong. @texasaggie97 shouldn't this be isinstance()? That way it works with subclasses.

docs/niswitch/functions.rst
Fladolcetta: I'm guessing that the generated code was not submitted from when we removed is_scanning from master. I made mention of it in the pull request above.

marcoskirsch: Yeah that was it. I've updated generated files in master. If you merge again this will go away.

generated/nifake/attributes.py
Fladolcetta: According to https://docs.python.org/3/library/enum.html#restricted-subclassing-of-enumerations. Subclassing an enumeration is allowed only if the enumeration does not define any members. Given that our enums have members, customers cannot create a subclass of our enums.

Issue #505, opened 2017-10-18
Input lists don't work


marcoskirsch:
Some functions (notably in the NI-FGEN API) require client to pass in an array / list. Our generator doesn't handle it correctly. We just turn around and hand it off to the Library class. 

See this error I got while writing an NI-DCPower example (attribute_ids is a Python list):

```
Z:\>python src\nidcpower\examples\nidcpower_advanced_sequence.py
Traceback (most recent call last):
  File "src\nidcpower\examples\nidcpower_advanced_sequence.py", line 31, in <module>
    session.create_advanced_sequence('my_sequence', attribute_ids)
  File "C:\Python27\lib\site-packages\nidcpower\session.py", line 4338, in create_advanced_sequence
    error_code = self._library.niDCPower_CreateAdvancedSequence(self._vi, sequence_name.encode(self._encoding), attribute_id_count, attribute_ids, set_as_active_sequence)
  File "C:\Python27\lib\site-packages\nidcpower\library.py", line 142, in niDCPower_CreateAdvancedSequence
    return self.niDCPower_CreateAdvancedSequence_cfunc(vi, sequence_name, attribute_id_count, attribute_ids, set_as_active_sequence)
ctypes.ArgumentError: argument 4: <type 'exceptions.TypeError'>: expected LP_c_long instance instead of list
```

marcoskirsch:
I have a fix for this in branch https://github.com/ni/nimi-python/tree/explicit-ctypes-variables but it's nowhere ready for merging because the unit tests require a lot of retrofits.

PR #508, opened 2017-10-19
Fix nidcpower's functions_addon.py

### LINKED ISSUE #496: Incorrect generation of nidcpower.Session. create_advanced_sequence() and SetSequence()
This is the NI-DCPower function:

```
ViStatus _VI_FUNC niDCPower_CreateAdvancedSequence(
      ViSession vi,
      ViConstString sequenceName,
      ViInt32 attributeIDCount,
      const ViInt32 attributeIDs[],
      ViBoolean setAsActiveSequence);
```

This is the signature of the corresponding Session method:

```
    def create_advanced_sequence(self, sequence_name, attribute_id_count, attribute_ids, set_as_active_sequence=True):
```

**attribute_id_count** should not be in the method, instead it should be derived from len(attribute_ids) and passed to the library that way. We actually request that in nimi-python/src/nidcpower/metadata/functions_addon.py yet it doesn't appear to be working.

### LINKED ISSUE #499: nidcpower.Session.wait_for_event() is wrong
def wait_for_event(self, timeout, event_id=10.0)

Default is in the wrong parameter.

marcoskirsch:

[X] I've added any unit tests that are applicable for this pull request

Fixes the signature of several method in nidcpower.Session:

* set_sequence
* create_advanced_sequence
* wait_for_event

* Fix #499 
* Fix #496 

* Inspected diffs in generated code.
* Wrote and ran system tests. While not all system tests passed due to #505, it made it far enough for me to determine that the signatures were correct.


generated/nimodinst/attributes.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
```
texasaggie97: This file was removed and shouldn't be added back.

PR #530, opened 2017-10-25
Add support for attributes_addon & mark obsolete attributes as don't codegen

### LINKED ISSUE #528: Do not include any obsolete properties in initial (0.9) release
We need to stop code-generating obsolete properties (aka attributes). When we extract the metadata that eventually gets sourced on GitHub as attributes.py, it includes some obsolete properties. We should mark them as 'codegen_method':'no'.

In a future release, when a property (or function) that was already included in the Python APIs is obsoleted, we will likely need to keep including it and handle it somehow else. A new metadata tag such as is_obsolete:True would need to be created at that point.

texasaggie97:
~~[ ] I've added any unit tests that are applicable for this pull request~~
* Add support for attributes_addon similar to functions_addon
* Mark attributes marked as obsolete in the C API as do not codengen

* Fix #528 

* Unit
* System


build/helper/metadata_merge_dicts.py
```diff
@@ -22,11 +22,14 @@ def merge_dicts(into, outof):
                 for item2 in outof[item]:
                     into[item][item2] = outof[item][item2]
             else:
-                # Handle regex in addon
-                for item2 in into:
-                    if re.search(item, item2):
-                        assert type(into[item2]) is dict
-                        merge_dicts(into[item2], outof[item])
+                # attributes keys are integers so they do not need the regex check (and
```
marcoskirsch: One thing I noticed about this piece of code, even though it is not directly related to this change, is that substrings match.

So if you have:
"ConfigureVoltage" in the add ons, then it will match "ConfigureVoltageLevel" which may not be desirable.
Ideally you'd be forced to write "ConfigureVoltage.*" if you want to match both.

I saw this in NI-FGEN with "Initiate" matching "InitiateGeneration". I don't think it was intentional. Something to think about.

build/templates/session.py.mako
```diff
@@ -13,7 +13,8 @@ ${encoding_tag}
 
     module_name = config['module_name']
     c_function_prefix = config['c_function_prefix']
-    attributes = template_parameters['metadata'].attributes
+
```
marcoskirsch: For correctness, consider doing this in NI-Modinst's template.

src/nidmm/metadata/attributes_addon.py
```diff
@@ -0,0 +1,36 @@
+# These dictionaries are applied to the generated attributes dictionary at build time
+# Any changes to the API should be made here. attributes.py is code generated
+
+attributes_obsolete = {
```
marcoskirsch: I think it's an odd variable name. Sure, _today_ we only have obsolete attributes and we are marking them an "no". But tomorrow we'll be adding a "is_obsolete":True flag to things we obsolete after the Python API's firs "ship".

I'd change to attributes_codegene_method and document that we are not generating attributes found to be obsolete before the _first_ "release" of the Python API.

build/helper/metadata_add_all.py
```diff
@@ -184,6 +184,25 @@ def add_all_function_metadata(functions, config):
     return functions
 
 
+def _add_codegen_method(a, attributes):
+    '''Adds a boolean that sets the attribute obsolete. Will not be code genned when true'''
```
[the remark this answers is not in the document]

build/templates/session.py.mako
texasaggie97: Done

src/nidmm/metadata/attributes_addon.py
texasaggie97: Changed

PR #549, opened 2017-10-31
 Inject ctypes library into nifoo.Library so we can better unit test it

### LINKED ISSUE #317: Properly add code coverage to library.py and library_singleton.py
In https://github.com/ni/nimi-python/pull/313 we added a bunch of 

` # pragma: no cover`

The reason we did this is because we want to check test coverage of Library when running system tests. Some codepaths are not accessible in system tests (like unsupported platform).

For library_singleton.py, we can indeed inject/swap the right calls to get 100% unit test coverage. We just haven't done it.

We need to:

* Remove pragmas.
* Refactor Library so it is given the information it needs rather than figure it out by calling the OS. That way system tests can get to 100% coverage.
* Write unit tests for library_singleton.py and don't run coverage for it during system tests.

marcoskirsch:
[ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

~~[ ] I've added any unit tests that are applicable for this pull request~~

Uses dependency injection of the ctypes library into out Library class so that we don't need to use #pragmas in the Library.py 

* Fix #317 

tox and system tests

### DIFF
--- a/build/templates/library_singleton.py.mako
+++ b/build/templates/library_singleton.py.mako
@@ -8,11 +8,14 @@ module_name = config['module_name']
 
 import platform
 
+import ctypes
 from ${module_name} import errors
 from ${module_name} import library
+import threading
 
 
 _instance = None
+_instance_lock = threading.Lock()
 
 
 def _get_library_name():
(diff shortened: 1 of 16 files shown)


build/templates/library_singleton.py.mako
```diff
@@ -35,10 +38,19 @@ def get():
...
+    with _instance_lock:
+        if _instance is None:
+            try:
+                library_type = _get_library_type()
+                if library_type == 'windll':
+                    ctypes_library = ctypes.WinDLL(_get_library_name())
+                else:  # pragma: no cover
+                    assert library_type == 'cdll'
+                    ctypes_library = ctypes.CDLL(_get_library_name())
+                    _instance = library.Library(ctypes_library)
```
texasaggie97: I think this needs to be outdented by 1 level.

texasaggie97: Also, remove TODO from PR description.

marcoskirsch: fixed

marcoskirsch:
please test this

marcoskirsch:
test this please

texasaggie97:
test this please

marcoskirsch:
> Remove TODO in PR description

What is wrong with me!?

PR #555, opened 2017-11-02
Adding system tests for NI-DCPower

bhaswath:
[X ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

[X ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.

[X ] I've added any unit tests that are applicable for this pull request

Adds system tests for nidcpower

Ran tests on local machine and it passes


src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -243,3 +249,59 @@ def test_wait_for_event_with_timeout():
...
+def test_export_signal(single_channel_session):
+    expected_trigger_terminal = "//Engine0/MeasureTrigger"
+    single_channel_session.export_signal(nidcpower.Event.SOURCE_COMPLETE, expected_trigger_terminal)
+    assert expected_trigger_terminal == single_channel_session.source_complete_event_output_terminal
+'''
+
+
+def test_configure_digital_edge_measure_trigger_error(single_channel_session):
+    try:
+        single_channel_session.configure_digital_edge_measure_trigger("invalid string")
```
marcoskirsch: Try not to write the tests so that they rely on error.
Don't have dead code in the tests. For example line 273 never executes.
You don't need to call initiate. Reading an attribute is enough to trigger a verify of the session. Calling initiate tries to verify, commit, and start the session.

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -243,3 +249,59 @@ def test_wait_for_event_with_timeout():
...
+            single_channel_session.wait_for_event(nidcpower.Event.SOURCE_COMPLETE, 0.5)
+        assert False
+    except nidcpower.Error as e:
+        assert e.code == -1074097882  # Error : Invalid property value
+        assert e.description.find('Requested value is not a supported value for this property.') != -1
+
+
+def test_configure_digital_edge_pulse_trigger_error(single_channel_session):
+    try:
+        single_channel_session.configure_digital_edge_pulse_trigger("invalid string")
```
marcoskirsch: This is suspect, since we are simulating a 4162 which is a new device. Is this a bug in the driver? Or is this truly not supposed to be supported?

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -243,3 +249,59 @@ def test_wait_for_event_with_timeout():
...
+        assert False
+    except nidcpower.Error as e:
+        assert e.code == -1074135023  # Error : Unsupported function
+        assert e.description.find('Function or method not supported.') != -1
+
+
+def test_configure_digital_edge_sequence_advance_trigger(single_channel_session):
+    expected_trigger_terminal = "//PXI_Trig0"
```
marcoskirsch: Why the double slash?
Is this truly what the driver is returning? It appears to be a bug!

injaleea:
test this please

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -243,3 +249,62 @@ def test_wait_for_event_with_timeout():
         with session.initiate():
             session.wait_for_event(nidcpower.Event.SOURCE_COMPLETE, 0.5)
 
+
+def test_commit(single_channel_session):
+    non_default_current_limit = 0.00021
+    single_channel_session.current_limit = non_default_current_limit
+    single_channel_session.commit()
```
marcoskirsch: Don't need to commit, because the act of reading back an attribute is enough to guarantee that we wrote the attribute.

PR #571, opened 2017-11-06
Add support for custom types

### LINKED ISSUE #543: Support custom types in generated code
NI-SCOPE Fetch and read functions return a struct for the waveform info. We need to support this.

texasaggie97:
[X] I've added any unit tests that are applicable for this pull requesttests were not warranted.

* Add support in code gen for custom types
  * Must be defined in separate .py file

* Fix #543 

* Unit
* System


generated/nifake/session.py
```diff
@@ -635,6 +637,20 @@ def get_array_using_ivi_dance(self):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return [float(array_out_ctype[i]) for i in range(array_size_ctype.value)]
 
+    def get_custom_type(self):
+        '''get_custom_type
+
+        This function takes a custom type.
```
marcoskirsch: takes -> returns

generated/nifake/session.py
```diff
@@ -635,6 +637,20 @@ def get_array_using_ivi_dance(self):
...
+
+        Returns:
+            cs (custom_struct): Set using custom type
+        '''
+        vi_ctype = visatype.ViSession(self._vi)  # case 1
+        cs_ctype = custom_struct.CustomStruct()  # case 13
+        error_code = self._library.niFake_GetCustomType(vi_ctype, ctypes.pointer(cs_ctype))
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
+        return custom_struct.custom_struct(cs_ctype)
```
marcoskirsch: If I understand this correctly, custom_struct.CustomStruct is the ctypes object and custom_struct.custom_struct is the object clients know about.

custom_struct.CustomStruct is an implementation detail... does it leak to the client?

src/nifake/metadata/config.py
```diff
@@ -22,5 +22,8 @@
         'abort_function': 'Abort',
     },
     'init_function': 'InitWithOptions',
+    'custom_types': [
+        {'file_name': 'custom_struct', 'python_name': 'custom_struct', 'ctypes_name': 'CustomStruct', },
```
marcoskirsch: I think more intuitive from a client point of view would be to have:
    
    {'file_name': 'custom_struct', 'python_name': 'CustomStruct', 'ctypes_name': '_CustomStruct', },

src/nifake/custom_types/custom_struct.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, data=None, struct_int=0, struct_double=0.0):
+        if data is not None:
+            self.struct_int = data.struct_int
+            self.struct_double = data.struct_double
+        else:
+            self.struct_int = struct_int
+            self.struct_double = struct_double
+
+
+custom_struct = CustomStructPython
```
marcoskirsch: This is strange... I think it's ok for the customer to reference CustomStruct (no Python) (uppercase, camelCase) because it is a type. As far as our convention goes, custom_struct is not the name of a type.

I think the type CustomStruct(ctypes.Structure) should be renamed to _CustomStruct or CustomStructCType or something because it is private.

generated/nifake/session.py
texasaggie97: I renamed the ctypes implementation to `_CustomStructCType` and the Python type to `CustomStruct`.

I also created `custom_struct` that is the same as the ctype. I needed the CamelCase name to make flake8 happy (unless we use noqa) but snake case to match the API definition. This is what NI-SCOPE will look like where the API type is `wfm_info`

The client will interact only with `CustomStruct`.

src/nifake/custom_types/custom_struct.py
texasaggie97: See above

src/nifake/metadata/config.py
texasaggie97: The 'ctypes_name' (now 'ctypes_type') needs to match what is in the API definition (function.py). For NI-SCOPE this will be `wfm_info` which is not a name that will make flake8 happy.

See comment above

build/helper/helper.py
```diff
@@ -31,7 +31,7 @@ def sorted_attrs(a):
     return sorted(a, key=lambda k: a[k]['name'])
 
 
-def get_python_type_for_visa_type(visa_type):
+def get_python_type_for_visa_type(visa_type, config):
```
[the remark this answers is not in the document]
texasaggie97: Renamed to `get_python_type_for_api_type` and updated docstring to say it can also handle custom types

generated/nifake/custom_struct.py
```diff
@@ -0,0 +1,37 @@
...
+        super(ctypes.Structure, self).__init__()
+        if data is not None:
+            self.struct_int = data.struct_int
+            self.struct_double = data.struct_double
+        else:
+            self.struct_int = struct_int
+            self.struct_double = struct_double
+
+
+# We need to match the name in the function definition
```
marcoskirsch: I don't understand why this is necessary. Can you explain just for my benefit? You mentioned something about making flake8 happy.

build/helper/codegen_helper.py
```diff
@@ -138,49 +148,55 @@ def get_ctype_variable_declaration_snippet(parameter, parameters):
        13. Output scalar or enum:                                       visatype.ViInt32()
     '''
 
+    # First we need to determine the module. If it is a custom type then the module is the file associated wit that type, otherwise 'visatype'
```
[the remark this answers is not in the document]
texasaggie97: Fixed

generated/nifake/__init__.py
```diff
@@ -5,3 +5,8 @@
 from nifake.errors import Error     # noqa: F401
 from nifake.errors import NifakeWarning   # noqa: F401
 from nifake.session import Session  # noqa: F401
+
```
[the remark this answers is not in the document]
texasaggie97: Comment added

generated/nifake/__init__.py
```diff
@@ -5,3 +5,8 @@
...
+from nifake.custom_struct import CustomStruct  # noqa: F401
+
+from nifake.custom_struct import custom_struct  # noqa: F401
```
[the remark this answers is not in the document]
texasaggie97: Changed as discussed offline.

We need the name that the API uses. In the case of nifake, yes, we could make that API say what ever we want. For niscope, the type name is wfm_info. I was making the tests close to what we will really need.

generated/nifake/custom_struct.py
texasaggie97: The part about making flake8 happy is we need `custom_struct` to be the (or at least a) name for the type. This name doesn't follow the convention of CamelCase so flake8 would complain if that was the name of the class. To make flake8 happy, I gave the class a proper name and then assigned it to the name we needed in the API.

I removed `custom_struct = _CustomStructCType` and just called said `class custom_struct`. This did require a noqa statement to tell flake8 not to complain about it.

marcoskirsch:
Merging. But do note decrease in test coverage.

PR #580, opened 2017-11-09
Add support for arrays of custom types

texasaggie97:
* Add support for arrays of custom types in the API
  * Added 14th ctype declaration
 
* Unit
* System


build/helper/codegen_helper.py
```diff
@@ -146,13 +146,14 @@ def get_ctype_variable_declaration_snippet(parameter, parameters, config):
        11. Output buffer with mechanism ivi-dance:                      None
        12. Output buffer with mechanism passed-in:                      (visatype.ViInt32 * buffer_size)()
        13. Output scalar or enum:                                       visatype.ViInt32()
+       14. Input buffer (custom type):                                  (custom_struct * len(list))(*[custom_struct(l) for l in list])
```
marcoskirsch: As I wrote this, I made sure to keep this list and the code below in the same order... so 14 should be 5... of course this means everything else needs to be renumbered. The way I would do it (it took many passes) is to add the new thing as 4.5, commit and make sure everything works, then renumber in a different commit.

build/helper/codegen_helper.py
```diff
@@ -146,13 +146,14 @@ def get_ctype_variable_declaration_snippet(parameter, parameters, config):
...
     '''
 
     # First we need to determine the module. If it is a custom type then the module is the file associated with that type, otherwise 'visatype'
     module_name = 'visatype'
-    c = find_custom_type(parameter, config)
-    if c is not None:
-        module_name = c['file_name']
+    custom_type = find_custom_type(parameter, config)
```
marcoskirsch: Much better! Thanks.

build/templates/matchers.py.mako
```diff
@@ -89,23 +89,53 @@ class ViStringMatcher(object):
...
+class CustomTypeBufferMatcher(object):
+    def __init__(self, expected_element_type, expected_value):
+        self.expected_value = expected_value
+        self.expected_size = len(expected_value)
+        self.expected_type = expected_element_type * self.expected_size
+        self.expected_element_type = expected_element_type
+
+    def __eq__(self, actual):
+        if not isinstance(actual, self.expected_type):
+            print("Unexpected array type. Expected: {0}. Received: {1}".format(self.expected_type, type(actual)))
```
marcoskirsch: I realized that ctypes calls these things "arrays" not "buffers". Are we misnaming the matchers?

build/templates/matchers.py.mako
```diff
@@ -89,23 +89,53 @@ class ViStringMatcher(object):
...
+            return False
+        if self.expected_size != len(actual):
+            print("Unexpected length. Expected: {0}. Received: {1}".format(self.expected_size, len(actual)))
+            return False
+        if self.expected_value is not None:
+            # Can't compare the objects directly because they're different types (one is list, another is ctypes array).
+            # Go element by element, which allows for reporting the first index where different values were found.
+            for i in range(0, len(self.expected_value)):
```
marcoskirsch: I think the Pythonic way of doing this is to use zip. You can give it a shot. Something like:

```
for i, j in zip(self.expected_value , actual):
```

(we've been doing it wrong all over the place, too much ancient C++ under our belts)

marcoskirsch: ... This only works after you verified len() is the same:

```
>>> zip([1,2], [1,2,3])
[(1, 1), (2, 2)]
```

build/helper/codegen_helper.py
texasaggie97: Renumbered

build/templates/matchers.py.mako
texasaggie97: Maybe? Both terms get the point across, though array is probably more correct when talking about ctypes.

marcoskirsch: Right... we have lists and arrays. Buffer is really ambiguous (you know how I feel about that).
For a different PR.

build/templates/matchers.py.mako
texasaggie97: `zip()` works, but it has memory implications. Because it copies each of the lists, if they are large, then you've just doubled the memory required.

It may be that here it is okay since we will never test with very large lists/arrays.

marcoskirsch: I don't think so. At least not in Python 3.

> Make an iterator that aggregates elements from each of the iterables.

https://docs.python.org/3/library/functions.html#zip

texasaggie97: You're right about Python 3, but we support both Python 2 as well.

marcoskirsch: IMHO, there's no need to support the _code generator_ on Python 3. Just the generated code.

texasaggie97: Do we then drop Python 2 from Travis? It first builds then tests, which means the code generator has to support Python 2.

marcoskirsch: I think it's tricky to do because of how Travis works.

Ideally we would

1. Run unit tests for build in the one version of Python that we support.
2. Build the code in the one version of Python that we support.
3. Run unit tests on all versions of Python that we support.

But given that each version of Python is in its own Docker container and that we must unit test generated code on each version of Python, we have no choice but to build in each version of Python unless we do some major restructuring of the project.

So because of this, it seems simpler to keep ensuring that the generator runs on all supported version of Python for the foreseeable future. But I don't care if zip() uses more RAM in the Python 2 builds :)

Issue #444, opened 2017-10-09
niDCPower_MeasureMultiple does not have array size input


marcoskirsch:
This is the function signature:

```
ViStatus _VI_FUNC niDCPower_MeasureMultiple(
ViSession vi,
ViConstString channelName,
ViReal64 voltageMeasurements[],
ViReal64 currentMeasurements[]);
```

No parameter tells us the size of the arrays voltageMeasurements and currentMeasurements.
How is this used?
How do we make this work in Python? these are output arrays so we need to figure out how large the buffers we allocate need to be.

marcoskirsch:
@injaleea commented in duplicate #445:

> Calling nidcpower measure_multiple causes python environment to crash when doing pytest.
> 
> ```
> import nidcpower
> import pytest
> import re
> 
> 
> @pytest.fixture(scope='function')
> def session():
>     with nidcpower.Session('FakeDevice', '', False, 'Simulate=1, DriverSetup=Model:4143; BoardType:PXIe') as simulated_session:
>         yield simulated_session
> 
> 
> def test_measure_multiple(session):
>     session.source_mode = nidcpower.SourceMode.SINGLE_POINT
>     '''session.configure_aperture_time(0, nidcpower.ApertureTimeUnits.SECONDS)'''
>     #session.voltage_level = 1
>     count = 10
>     #session.measure_when = nidcpower.MeasureWhen.ON_DEMAND
>     with session.initiate():
>         voltage_measurements, current_measurements = session.measure_multiple()
>     assert len(voltage_measurements) == 4   # measuremultiple will return a reading for all channel , since 4143 has 4 channel expecting 4 readings 
>     assert len(current_measurements) == 4 
>     assert isinstance(voltage_measurement[1], float)  
>     assert isinstance(current_measurements[1], float) 
> 
> ```
> Also noticed that assertion on test_get_attribute_string fails sometime when it is expecting 4143  session.instrument_model is returning as 4110.[Failure was consistent and disappeared after reboot]

marcoskirsch:
> How is this used?

Client needs to allocate arrays large enough for the number of channels specified in the channelName input. So "" means "all the channels in the session, based on when you opened the session". "0-3" means "4 channels". "0, 1" means 2 channels.

* **Option 1**: Don't support niDCPower_MeasureMultiple() in Python. We need to evaluate how bad this would be and what the workarounds are.
    * **Option 1b**: ... until a better alternative exists in NI-DCPower.
[...]

PR #582, opened 2017-11-09
Use enum addon for nidcpower and niscope

texasaggie97:
* Add support for adding functions/attributes/enums via addon
  * Must use dictionary name `<type>_additional_<type>`. I.e. `enums_additional_enums`
  * Move the hand added DCPower enums to enum_addon
* Add support for overwriting enums via enum_addon
  * Must use dictionary name `replacement_enums`
  * Add NI-SCOPE 'VideoSignalFormat' and 'TriggerCoupling' - Remove when CAR 675174 is fixed


build/helper/metadata_add_all.py
```diff
@@ -168,6 +168,11 @@ def add_all_function_metadata(functions, config):
         for m in dir(config['modules']['metadata.functions_addon']):
             if m.startswith('functions_'):
                 merge_dicts(functions, config['modules']['metadata.functions_addon'].__getattribute__(m))
+            # We need to explicitly copy new entries
+            if m == 'functions_additional_functions':
```
marcoskirsch: This will be useful for #444 if we decide to use the private function.

build/helper/metadata_add_all.py
```diff
@@ -259,6 +269,19 @@ def add_all_enum_metadata(enums, config):
         for m in dir(config['modules']['metadata.enums_addon']):
             if m.startswith('enums_'):
                 merge_dicts(enums, config['modules']['metadata.enums_addon'].__getattribute__(m))
+            # We need to explicitly copy new entries
```
marcoskirsch: This function is suspiciously similar to add_all_function_metadata()
I think we could probably have a single merging function, even if it means we rename the dictionaries in the metadata (i.e. this looks for things that start with "enums_"... what if we simply merged anything we find in the file?

src/niscope/metadata/enums_addon.py
```diff
@@ -8,4 +8,169 @@
 enums_codegen_method = {
 }
 
-
+# We explicitly don't start with enums_ since we don't want this merged. These will replace the existing enums
+# Once NI Internal CAR #675174 is fixed, this can be removed along with the overwrite code in __init__.py
```
marcoskirsch: I see how this can conflict with my previous suggestion.

PR #584, opened 2017-11-10
Move common prefix and suffix handling to codegen instead of extractor

texasaggie97:
* Look for and remove common prefixes and suffixes from all names within an enum'
  * Allow white listing prefix or suffix in metadata config

* Unit
* System


build/helper/metadata_merge_dicts.py
```diff
@@ -6,6 +6,31 @@
 pp = pprint.PrettyPrinter(indent=4, width=80)
 
 
+def merge_helper(metadata, metadata_type, config):
```
marcoskirsch: If you can avoid passing all of config, you should. It looks like only the 'modules' dictionary is needed, which could be optionally None. That way we have less code that depends on the structure of all of config.

build/helper/metadata_merge_dicts.py
```diff
@@ -6,6 +6,31 @@
...
+        for m in dir(config['modules'][metadata_module]):
+            if m.startswith('{0}_'.format(metadata_type)):
+                merge_dicts(metadata, config['modules'][metadata_module].__getattribute__(m))
+            # We need to explicitly copy new entries
+            if m == '{0}_additional_{0}'.format(metadata_type):
+                outof = config['modules'][metadata_module].__getattribute__(m)
+                for a in outof:
+                    metadata[a] = outof[a]
+
+    # Delete any entries that are empty
```
marcoskirsch: When would this happen, can you provide context in the comment?

build/helper/metadata_merge_dicts.py
```diff
@@ -6,6 +6,31 @@
...
+    # Have to do this in two steps. Otherwise the dictionary changes size and errors
+    to_delete = []
+    for m in metadata:
+        if len(metadata[m]) == 0:
+            to_delete.append(m)
+
```
marcoskirsch: remove this new line, since this is part of what you described in the comment

docs/nidcpower/enums.rst
```diff
@@ -576,7 +576,7 @@ Enums used in NI-DCPower
 
 
 
-    .. py:attribute:: nidcpower.TriggerType.DIGITAL
+    .. py:attribute:: nidcpower.TriggerType.DIGITAL_EDGE
```
marcoskirsch: Is this right?

Before:

TriggerType
* DIGITAL
* SOFTWARE

Now

TriggerType
* DIGITAL_EDGE
* SOFTWARE_EDGE

src/nidmm/metadata/attributes.py
```diff
@@ -292,7 +294,9 @@
         'resettable': 'No',
         'type': 'ViInt32',
 'documentation': {
-'description': 'Returns the minor version number of this instrument driver.',
+'description': '''
+The minor version number of this instrument driver.
```
marcoskirsch: Is it a problem that there's always a new line before and after the content of the documentation? Seems like it shouldn't be there.

src/nidmm/metadata/attributes.py
```diff
@@ -628,8 +634,13 @@
         'resettable': 'No',
         'type': 'ViInt32',
 'documentation': {
-'description': 'For the NI 4060 only, specifies a delay interval after a sample trigger.',
-'table_body': [['0', 'IVI compliant', 'The Sample Interval property is only used when the Sample Trigger is set to Interval.'], ['1', 'Not IVI compliant', 'The Sample Interval property is used as a delay after any type of Sample Trigger.']],
+'description': '''
+For the NI 4060 only, specifies a delay interval after an sample external trigger.
+0
+NIDMM_ATTR_SAMPLE_INTERVAL is only used when the Sample Trigger attribute is set to  INTERVAL.
```
marcoskirsch: Here's an idea that we can open an issue for, if you think it has merit:

In metadata we now have (or will have) the full original C name for the symbol. We could go through all the documentation in all the metadata looking for 'NIDMM_ATTR', 'NIDMM_VAL', 'niDmm_' and when/if found, do the appropriate replacement/linking.

src/nidmm/metadata/enums_addon.py
```diff
@@ -8,4 +8,6 @@
 enums_codegen_method = {
 }
 
-
+enums_additional_enums = {
+    'DigitsResolution': {},  # Just delete this enum. Otherwise we have to make it so that it will generate valid code
```
marcoskirsch: please add more context to this document... something like
```
# Just delete this enum, client will have to pass a float. Otherwise the constants start with a number and we have to deal with it. Not worth it.
```

src/nifgen/metadata/attributes.py
```diff
@@ -228,6 +215,20 @@
 ''',
 },
     },
+    1050322: {
```
marcoskirsch: we need `'codegen-method': 'no'` for this one.

src/niscope/metadata/enums.py
```diff
@@ -876,35 +870,42 @@
...
-                'name': 'LF_REJECT',
+                'name': 'NISCOPE_VAL_LF_REJECT',
                 'value': 3,
 'documentation': {
 'description': 'Lowpass filter coupling',
 },
             },
             {
-                'name': 'AC_PLUS_HF_REJECT',
+                'name': 'LF_REJECT',
```
marcoskirsch: Is this a known bug? Or not a bug?
NISCOPE_VAL_LF_REJECT and LF_REJECT?

src/niscope/metadata/enums.py
```diff
@@ -1079,119 +1069,224 @@
...
-                'name': 'SECAM',
+                'name': 'NISCOPE_VAL_SECAM',
                 'value': 3,
 'documentation': {
 'description': 'SECAM signal format supports line numbers from 1 to 625',
 },
             },
             {
                 'name': 'M_PAL',
+                'value': 4,
```
marcoskirsch: All these missing values in one of the source APIs is already reported right?

build/helper/metadata_add_all.py
```diff
@@ -263,17 +247,63 @@ def _add_enum_codegen_method(enums, config):
                     enums[e]['codegen_method'] = a_codegen_method
 
 
+def _cleanup_names(enum_info, config):
```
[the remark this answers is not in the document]
texasaggie97: I can use 'python_name' instead, but that will be a bigger change since we don't currently only use 'name'. Changing will touch more places.

build/helper/metadata_add_all.py
```diff
@@ -263,17 +247,63 @@ def _add_enum_codegen_method(enums, config):
...
+
+    # We are using an os.path function do find any common prefix. So that we don't
+    # get 'O' in 'ON' and 'OFF' we remove characters at the end until they are '_'
+    names = [v['name'] for v in enum_info['values']]
+    prefix = os.path.commonprefix(names)
+    while len(prefix) > 0 and prefix[-1] != '_':
+        prefix = prefix[:-1]
+
+    # If the prefix is in the whitelist, we don't want to remove it so set to empty string
+    if 'whitelist_prefix' in config and prefix in config['whitelist_prefix']:
```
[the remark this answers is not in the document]
texasaggie97: Right now it is, but I can see a use for it in other places. Would we want multiple whitelist keys?

build/helper/metadata_add_all.py
```diff
@@ -284,6 +314,9 @@ def add_all_enum_metadata(enums, config):
         pass
 
     _add_enum_codegen_method(enums, config)
+    for e in enums:
+        enums[e] = _cleanup_names(enums[e], config)
```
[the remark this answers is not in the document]
texasaggie97: I'm fine with changing the name but I prefer to pass around the config object. That way the parameter lists are consistent: what we are operating on and config.

build/helper/metadata_add_all.py
```diff
@@ -263,17 +247,63 @@ def _add_enum_codegen_method(enums, config):
                     enums[e]['codegen_method'] = a_codegen_method
 
 
+def _cleanup_names(enum_info, config):
+    # First remove <DIRVER>_VAL_ if it is there
```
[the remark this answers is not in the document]
texasaggie97: Removed comment. Would need to change nifake metadata to be able to assert that here.

build/helper/metadata_merge_dicts.py
texasaggie97: I prefer to pass all of config. Two reasons.
1. It makes the parameter lists consistent, like I mentioned above
1. When another item from config is needed, it is already available instead of having to go back and change parameter lists all the way back up to where config was last available. Which does happen fairly regularly.

build/helper/metadata_merge_dicts.py
texasaggie97: See `DigitsResolution` in nidmm's enum_addon.

build/helper/metadata_merge_dicts.py
texasaggie97: Removed

build/helper/metadata_add_all.py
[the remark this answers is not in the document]
marcoskirsch: I think it's the right thing to do / worth it.

build/helper/metadata_add_all.py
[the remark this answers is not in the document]
marcoskirsch: It's both more explicit immediately, and more future proofing going forward.

build/helper/metadata_add_all.py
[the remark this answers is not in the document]
marcoskirsch: if you think passing the config object is better, then I would keep the name as config.

docs/nidcpower/enums.rst
texasaggie97: Yes, because _EDGE is not common with all names. There is also a `NONE` name.

build/helper/metadata_merge_dicts.py
marcoskirsch: Yeah, I can see your point. I just don't like giving a function more information than it needs because it makes it less evident what it is acting on.  It goes against Single Responsibility / Separation of Concerns and it obscures things a little. The fact than in Python you cannot even tell if the thing being passed in is modified or not doesn't help.

src/nidmm/metadata/attributes.py
texasaggie97: New lines before and after are always cleaned up before formatting the documentation

src/nidmm/metadata/attributes.py
texasaggie97: Yes, that is my plan for the future. I didn't make it an issue though.

docs/nidcpower/enums.rst
marcoskirsch: So the code is doing what it is supposed to do. But do we want _EDGE in these cases? In other places I remember seeing just as DIGITAL and SOFTWARE.

src/nidmm/metadata/enums_addon.py
texasaggie97: We were already setting it to not code gen so the clients would already have to pass a float.

I am setting it to {} here so it will be deleted completely. Otherwise we'd have to set the names to something that wouldn't cause issues when we remove common suffixes. `_5` is common with all, but `_5_5` becomes empty when removing `_5`, thus breaking things before we are even able to say don't codegen it.

src/niscope/metadata/enums.py
texasaggie97: These are already reported.

src/niscope/metadata/enums.py
texasaggie97: This is a bug and has already been CARed. We overwrite this whole enum using enum_addon

docs/nidcpower/enums.rst
texasaggie97: Once this is in, we can go through all the enums and see if there are any names we want to explicitly override. There may be others in addition to this one. At that point we will have to decide if the better name is worth hand maintaining it.

src/nifgen/metadata/attributes.py
texasaggie97: Added

build/helper/metadata_add_all.py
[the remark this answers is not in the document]
texasaggie97: Changed

src/nidmm/metadata/attributes.py
marcoskirsch: Opened #589

docs/nidcpower/enums.rst
marcoskirsch: ok

build/helper/metadata_add_all.py
```diff
@@ -615,10 +647,10 @@ def test_add_enums_metadata_simple():
         'Color': {
             'codegen_method': 'no',
             'values': [
-                {'documentation': {'description': 'Like blood.'}, 'name': 'RED', 'value': 1},
-                {'documentation': {'description': 'Like the sky.'}, 'name': 'BLUE', 'value': 2},
-                {'documentation': {'description': 'Like a banana.'}, 'name': 'YELLOW', 'value': 2},
-                {'documentation': {'description': "Like this developer's conscience."}, 'name': 'BLACK', 'value': 2}
+                {'documentation': {'description': 'Like blood.'}, 'name': 'RED', 'value': 1, 'python_name': 'RED'},
```
marcoskirsch: You should change name to be like NIFAKE_VAL_RED

PR #599, opened 2017-11-14
Add size mechanism python-code

### LINKED ISSUE #590: Need a new size mechanism for NI-SCOPE and NI-DCPower
This will allow us to run arbitrary python code defined in metadata such as calling a function. This is because some functions require the size to be calculated in ways that don't fit into the current set of mechanisms and are very specific.

The proposal is to have a mechanism called "python-code".

texasaggie97:
* Add new size mechanism called `python-code`
  * The value associated with this is actual python code that will be inserted as-in into session.py
    * It should use python names and types
    * Any function calls or attribute queries should be in relation to `self.`
  * Add unit tests for array of double and array of custom type

* Fix #590 

* Unit
* System


marcoskirsch:
```
_______________________________ test_arb_script _______________________________

session = <nifgen.session.Session object at 0x000001CE4E5D0860>

    def test_arb_script(session):
        waveform_data = [0, 9630, 15582, 15582, 9630, 0, -9630, -15582, -15582, -9630]
>       session.output_mode = nifgen.OutputMode.NIFGEN_VAL_OUTPUT_SCRIPT

src\nifgen\system_tests\test_system_nifgen.py:203: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <enum 'OutputMode'>, name = 'NIFGEN_VAL_OUTPUT_SCRIPT'

    def __getattr__(cls, name):
        """Return the enum member matching `name`
    
            We use __getattr__ instead of descriptors or inserting into the enum
            class' __dict__ in order to support `name` and `value` being both
            properties for enum members (which live in the class' __dict__) and
            enum members themselves.
    
            """
        if _is_dunder(name):
            raise AttributeError(name)
        try:
            return cls._member_map_[name]
        except KeyError:
>           raise AttributeError(name) from None
E           AttributeError: NIFGEN_VAL_OUTPUT_SCRIPT

.tox\system_tests\lib\enum.py:324: AttributeError
```

I assume master is broken too.

texasaggie97:
Reduced number of cases for `python-code` mechanism. Only required cases are
* input param, single item
* output param, buffer/array

PR #620, opened 2017-11-20
Updating niscope enums

### LINKED ISSUE #616: Add NI-SCOPE enum metadata

injaleea:

~~[] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~

Updating enums_addon.py with some of the missing enums and attaching some enums to parameter on functions_addon.py for niscope.
 
Added systemtest for niscope:

- test_acquisition_status
- test_self_cal
- test_probe_compensation_signal
- test_configure_channel_characteristics
- test_waveform_processing
- test_vi_boolean_attribute
- test_vi_string_attribute

fix #616 

Systemtests, tox


src/niscope/metadata/enums_addon.py
```diff
@@ -139,38 +139,810 @@
...
+                    'description': 'Highpass and lowpass filter coupling',
+                },
             },
         ],
     },
-}
+    'ArrayMeasurement': {
+        'values': [
+            {
+                'name': 'NO_MEASUREMENT',
```
texasaggie97: All names should use the C name `NISCOPE_VAL_<>`. So this one should be `NISCOPE_VAL_NO_MEASUREMENT`

src/niscope/metadata/enums_addon.py
```diff
@@ -139,37 +140,1084 @@
...
+                'name': 'Channel_6',
+                'value': 6,
+            },
+            {
+                'name': 'Channel_7',
+                'value': 7,
+            },
+            {
+                'name': 'VAL_EXTERNAL',
+                'value': 'VAL_EXTERNAL',
```
texasaggie97: Why is this one a string when the others are numbers?

injaleea: I could not see TriggerSource defined in niscope.h. I took these values form LabVIEW, According to LabVIEW control this rest all are numbers while VAL_EXTERNAL is string. 

VAL_EXTERNAL is defined as string in iviscope header file aswell.

texasaggie97: From an API perspective, the function expects a single type. Where this particular enum is used `ConfigureTriggerWindow` it needs to be a string.

Another option is to look at the actual LV VI. While the values are the numbers you have listed, they are passed in as strings, not integers.

Looking in niScope.h, I do see a list of VAL's for trigger source, but the channels aren't in that list of defined values. So for this, there are a couple of things:
1. Make all the values strings: '1', '2', etc.
1. Make the names look like they came from the C API, even though they didn't really: 'NISCOPE_VAL_CHANNEL_1', etc

There are a lot of entries in the TRIGGER_SOURCE list in niScope.h, so I see two options here. I don't have a strong preference which way you go.
1. Rename this enum to be specific to trigger type window. `WindowTriggerSource`?
1. Leave the name generic so it can be used else where, but add all the other sources from the header file.

I see pros and cons with either solution.

Fladolcetta: I have fixed this by adding all enums from .h file, I have also fixed the names

marcoskirsch: What we had done in NI-FGEN and NI-DCPower is to **not** provide string-type enums for terminals, since these APIs participate in system-wide routing, and providing enums limit routing options.

We want customers to route between devices by using strings such as "/TheOtherDevice/MarkerEvent2"

PR #586, opened 2017-11-13
Updating enums_addon and system tests for nifgen

bhaswath:
[X ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

~~[] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~

[X] I've added tests applicable for this pull request

### What does this Pull Request accomplish?
* Adds enums for NI-Fgen
  * ByteOrder
  * HardwareState
  * RelativeTo
  * Signal
  * Trigger
  * TriggerWhen
* Adds system tests for NI-Fgen

### List issues fixed by this Pull Request below, if any.
* Fix #484  
* Fix #538 

### What testing has been done?
System tests pass

PR #622, opened 2017-11-21
System test error on adding Triggers enum to enums_addon.py (Not for merging)

bhaswath:
[ X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

[X ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.

Shows the error that is created when adding Trigger enum to enums_addon.py 

System tests fail

### DIFF
--- a/docs/nifgen/enums.rst
+++ b/docs/nifgen/enums.rst
@@ -1374,6 +1374,17 @@ Enums used in NI-FGEN
 
 
 
+.. py:data:: Trigger
+
+    .. py:attribute:: nifgen.Trigger.START
+
+
+
+    .. py:attribute:: nifgen.Trigger.SCRIPT
+
+
+
+
 .. py:data:: TriggerMode
 
     .. py:attribute:: nifgen.TriggerMode.SINGLE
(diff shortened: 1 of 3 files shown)


injaleea:
Changing dictionary_name to **replacement_enums** from **enums_additional_enums**  generates the code properly.

@marcoskirsch , @texasaggie97 what is the diff b/w these two keys in enums_addon.py . I could not find any code for extracting values for key **enums_additional_enums** in codegne. There exists the key **enums_additional_enums** for enums_addon.py for nidcpower and nidmm. How we are generating them?

texasaggie97:
The difference between all dictionaries that start with `enums_` and `replacement_enums` is the former all go through a process that uses regular expressions, and the latter does not.

This is causing problems is because 'Trigger' matches multiple entries like 'TriggerMode' and 'TriggerSource', so they all end up with the values from 'Trigger'.

I would say for PR #586 put this one enum in `replacement_enums` like you have here, but open an issue about this and reference the issue in comments say why you put the one enum in a different place and that it can be moved to be with the others once the referenced issue is fixed.

PR #627, opened 2017-11-22
Add default values to scope functions

### LINKED ISSUE #301: Add default parameter metadata to NI-SCOPE

Fladolcetta:

Adds default values to scope functions to improve usability.
Adds select enums that are required for the functions that required default values

* Fix #301 

Ran tox


generated/niscope/enums.py
```diff
@@ -317,6 +398,40 @@ class TriggerSlope(Enum):
...
+class TriggerSource(Enum):
+    Channel_0 = 0
+    Channel_1 = 1
+    Channel_2 = 2
+    Channel_3 = 3
+    Channel_4 = 4
+    Channel_5 = 5
+    Channel_6 = 6
+    Channel_7 = 7
+    VAL_EXTERNAL = 'VAL_EXTERNAL'
```
texasaggie97: Why is this one a string and the others are numbers?

Also, the names should be all upper case

Fladolcetta: As mentioned on These enums came from the https://github.com/ni/nimi-python/pull/620 review marked above. I am not sure where @injaleea got the values. https://github.com/ni/nimi-python/pull/627/commits/8c870aa934435c3a5b880ee449c1d3acb616c633

Fladolcetta: This has been updated to match Jaleels

PR #636, opened 2017-11-29
Fix version in readthedocs webpage title

texasaggie97:
* Install nidmm wheel so we can get version based on what we just built
  * We use nidmm because @Fladolcetta picked 2


docs/conf.py
```diff
@@ -63,9 +64,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.1'
+version = pkg_resources.get_distribution("nidmm").version
```
marcoskirsch: Please either document why this is the way it is or open an issue that explains it.

texasaggie97: Added comments.

Also changed `version` to be X.Y with `release` being the entire string

PR #638, opened 2017-11-29
Build docs/conf.py from template

texasaggie97:
* Previous solution didn't work in the read the docs environment since they don't run the actual build
* This requires all files be committed already updated
* Again, nidmm has been chosen to be the representative driver for the version

* Unit
* System


texasaggie97:
test this please

texasaggie97:
test this please

build/templates/conf.py.mako
```diff
@@ -0,0 +1,191 @@
+<%
+config = template_parameters['metadata'].config
+%>\
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# NI Modular Instruments Python API documentation build configuration file, created by
```
marcoskirsch: > NI Modular Instruments Python API documentation build configuration file

Right now we only have configuration for build but I think that's too narrow of a scope for this file. This file will contain configuration for the overall nimi-python project.

build/templates/conf.py.mako
```diff
@@ -0,0 +1,191 @@
...
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = 'NI Modular Instruments Python API'
+copyright = '2017, National Instruments'
```
marcoskirsch: Why not use the current year in mako instead of taking it from configuration?
One less thing that we need to manually update.

src/nidmm/nidmm.mak
```diff
@@ -8,6 +8,15 @@ MODULE_FILES_TO_COPY := $(DEFAULT_PY_FILES_TO_COPY)
 
 RST_FILES_TO_GENERATE := $(DEFAULT_RST_FILES_TO_GENERATE)
 
+# nidmm has been chosen to represent all the drivers for the purpose
```
marcoskirsch: This is no longer true, right?

marcoskirsch:
This leaves even less files being hand-maintained in docs, right? What remains?

build/templates/conf.py.mako
texasaggie97: This file is only for documentation.

build/templates/conf.py.mako
texasaggie97: Now says 2017-2017. Will change to 2017-2018 in 19 days - based on current year.

src/nidmm/nidmm.mak
texasaggie97: This file can now be built by any driver.

PR #674, opened 2017-12-07
Make NI-SWITCH init_with_topology private

### LINKED ISSUE #660: NI-SWITCH Init with topology is exposed and should be made private.

Fladolcetta:
[x ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

[x ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.

Makes the initialization function private for NI-SWITCH

* Fix #660 

Tox


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ All notable changes to this project will be documented in this file.
 * ### NI-Switch
   * #### Added
   * #### Changed
+    * Made init_with_topology a private function.
```
marcoskirsch: I think you can put this under Removed. One thing we should do as well and we haven't is better document the rationale behind some of these decisions. In most cases it's as simple as saying:

> Removed init_with_topology. Clients should use niswitch.Session.__init__(). See #660.

Of course, we need to be good about explaining things in our issues, which is not always the case. Remember this is all open to the world to see.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ All notable changes to this project will be documented in this file.
   * #### Added
   * #### Changed
   * #### Removed
+    * Removed init_with_topology. Clients should use niswitch.Session.init(). See #660.
```
marcoskirsch: There is no niswitch.Session.init()
There is `niswitch.Session.__init__()`. I think Markdown rendering screwed things up. You can change that to

> Clients should use niswitch.Session() constructor.

Issue #511, opened 2017-10-19
Support numpy in the APIs


texasaggie97:
Right now we take or return Python lists from functions where the underlying driver API takes or returns an array. Should we use numpy in our APIs?

Here's a performance comparison on NI-DMM fetch(). We use simulation so the data is probably getting generated as we fetch.

1. List: After calling the library, samples are copied onto a Python list.
2. numpy: Uses a hacked version of the Python bindings to pass a numpy array straight to the library, which is returned.
3. LabVIEW (LV): Allocation happens in the LabVIEW VI. Notably, this zeroes all the data before passing to the library.

|              | List (s) | numpy (s) | LV (s) |
|-------------:|----------|-----------|--------|
| 1M samples   | 0.2      | 0.04      | 0.04   |
| 10M samples  | 2.2      | 0.4       | 0.41   |
| 100M samples | 23       | 3.8       | 4.1    |

Questions

* Should we use numpy at all?
  * Looks like probably yes
* Should the API have methods for both numpy as well as Python list? Or numpy only?
* Do we need to somehow make the numpy dependency optional?
* Do we allocate the numpy buffer or require the user to allocate?
  * Make it optional and allocate if the user doesn't?

marcoskirsch:
Consistency with other NI Python driver APIs (primarily https://github.com/ni/nidaqmx-python and https://github.com/ni/nixnet-python) is very important here. As a client, it would be terrible to deal with this in an inconsistent manner just because I happen to have a digitizer and an MIO board.

marcoskirsch:
@texasaggie97, @Fladolcetta, myself met with @epage from [nixnet](https://github.com/ni/nixnet-python) and @neilvana from [nidaqmx](https://github.com/ni/nidaqmx-python/) to discuss this topic, and how to best do this.

Conclusions:

* If we use numpy, it will be as an optional dependency.
* Only use numpy arrays for waveforms / samples. Not for sequence steps, filter coefficients, etc.
* For APIs such as NI-SCOPE that have different "fetch" functions, for different types:
    * Expose only the scaled double version, and have it return list
    * Create fetch_into that take in numpy arrays. These will call the right underlying function based on datatype
* For APIs such as NI-FGEN that have different "write" function for different types:
    * Expose only write, then based on parameter type call appropriate underlying function

epage:
For reference, here is the xnet issue: https://github.com/ni/nixnet-python/issues/200

epage:
[...]

PR #680, opened 2017-12-10
Allow specifying multiple templates for a single function

marcoskirsch:

Allows to specify via metadata multiple, rather than just one, templates to apply to a single function.
This way we can generate twice for one function. This is necessary for numpy support: for example, we want both nidmm.Session.fetch_waveform() and nidmm.Session.fetch_waveform_into().

* Work towards #511 

Verified that by default generated code is the same. Feature will be leveraged in future PRs.


marcoskirsch:
Merged more than intended. Closing.

PR #681, opened 2017-12-10
Allow specifying multiple templates for a single function

marcoskirsch:

Allows to specify via metadata multiple, rather than just one, templates to apply to a single function.
This way we can generate twice for one function. This is necessary for numpy support: for example, we want both nidmm.Session.fetch_waveform() and nidmm.Session.fetch_waveform_into().

* Work towards #511 

Verified that by default generated code is the same. Feature will be leveraged in future PRs.

### DIFF
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -156,7 +156,7 @@ def _add_method_template_filenames(f):
     # Prefix the templates with a / so mako can find them. Not sure mako it works this way.
     prefixed_filenames = []
     for filename in f['method_template_filenames']:
-        prefixed_filenames.append('/' + filename if filename != '/' else filename)
+        prefixed_filenames.append('/' + filename if filename[0] != '/' else filename)
     f['method_template_filenames'] = prefixed_filenames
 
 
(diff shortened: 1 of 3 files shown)


build/helper/metadata_add_all.py
```diff
@@ -149,12 +149,15 @@ def _add_default_value_name_for_docs(parameter, module_name):
...
-    if f['method_template_filename'][0] != '/':
-        f['method_template_filename'] = '/' + f['method_template_filename']
+def _add_method_template_filenames(f):
+    '''Adds a list of 'method_template_filenames' value to function metadata if not found. This are the mako templates that will be used to render the method.'''
+    if 'method_template_filenames' not in f:
+        f['method_template_filenames'] = ['session_default_method.py.mako']
+    # Prefix the templates with a / so mako can find them. Not sure mako it works this way.
+    prefixed_filenames = []
+    for filename in f['method_template_filenames']:
+        prefixed_filenames.append('/' + filename if filename != '/' else filename)
```
texasaggie97: Shouldn't the statement in the append be `'/' + filename if filename[0] != '/' else filename`

Note the `[0]`

marcoskirsch:
I think we're going to have to do something equivalent for documentation (RST) files.

build/helper/metadata_add_all.py
marcoskirsch: fixed

PR #682, opened 2017-12-10
Small metadata fix to NI-FAKE

marcoskirsch:

Fixes a small metadata inconsistency in NI-FAKE. A string was not marked as a buffer. The generator didn't catch it due to how the logic is laid out.

I caught it in an attempt to refactor parts of the code generator. This will come in a future PR.

N/A

Generated Python doesn't change so all is the same. The documentation got fixed though.

### DIFF
--- a/src/nifake/metadata/functions.py
+++ b/src/nifake/metadata/functions.py
@@ -916,6 +916,7 @@ functions = {
             {
                 'direction': 'in',
                 'enum': None,
+                'is_buffer': True,
                 'name': 'aString',
                 'type': 'ViChar',
                 'documentation': {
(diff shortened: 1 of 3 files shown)


src/nifake/metadata/functions.py
```diff
@@ -916,6 +916,7 @@
             {
                 'direction': 'in',
                 'enum': None,
+                'is_buffer': True,
                 'name': 'aString',
                 'type': 'ViChar',
```
texasaggie97: Wouldn't the fix be to set the type to `ViChar[]`?

marcoskirsch: fixed

PR #685, opened 2017-12-11
Make the abort function public on all drivers

### LINKED ISSUE #655: abort() is not a public method

Fladolcetta:

[ x] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.

The existing system and unit tests will call the abort() function via the context manager.

This makes the abort function a public function

* Fix #655 

tox


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 * ### ALL
   * #### Added
   * #### Changed
+    * Changed abort() to be a public function.
```
marcoskirsch: Clients don't care about public/private. That's an implementation detail and we typically only list things here that affect clients.
So you didn't change abort. You added it.

generated/nifgen/session.py
```diff
@@ -2581,8 +2581,8 @@ def close(self):
 
     ''' These are code-generated '''
 
-    def _abort_generation(self):
-        '''_abort_generation
+    def abort_generation(self):
```
marcoskirsch: This is the counterpart to a public method called initiate(), not initiate_generation().
So the public name should be abort().

PR #690, opened 2017-12-15
nifgen.Session.create_waveform() supports numpy arrays

marcoskirsch:

Allows for nifgen.Session.create_waveform() to be called with numpy.array of float64 or int16, or simply with a list of float.

* Part of #511 

System tests.


generated/nifgen/session.py
```diff
@@ -1427,6 +1427,47 @@ def configure_standard_waveform(self, waveform, amplitude, frequency, dc_offset=
...
+
+        Args:
+            waveform_data_array (list of float): Array of data for the new arbitrary waveform. This may be an iterable of float, or for best performance a numpy.ndarray of dtype int16 or float64.
+
+        Returns:
+            waveform_handle (int): The handle that identifies the new waveform. This handle is used in other methods when referring to this waveform.
+        '''
+        try:
+            import numpy
+            numpy_imported = True
```
texasaggie97: Should we cache whether we can import numpy or not?

That way we don't take the hit of trying to import and throwing every time this function is called and numpy isn't installed.

marcoskirsch: Good idea. We can do this:

* Move the import to Session.py (with try/except).
* When needed, check using `if 'numpy' in sys.modules:`

I'll make a note for its own PR.

texasaggie97: Wouldn't it be faster to set a global variable or even a class variable that is just a boolean? That way we don't even have to do a string lookup in a list each time.

PR #699, opened 2017-12-18
Add numpy based read and fetch functions

texasaggie97:
* Adds numpy support for niscope
  * numpy based functions have _into suffix and are marked private
  * Access using public `fetch_into()` function with an already allocated numpy array of the type to acquire. Types supported
    * float64
    * int8
    * int16
    * int32
  * New function FetchDouble
    * Public name set to `fetch()`
    * Internally calls _fetch
  * New function called FetchDispatcher
    * Public name set to `fetch_into()`
    * Uses type of waveform array to "dispatch" to the appropriate internal numpy based function
* Step towards #511

* Unit
* System


src/niscope/metadata/functions_addon.py
```diff
@@ -186,16 +194,301 @@
...
+                    'description': 'The time to wait in seconds for data to be acquired; using 0 for this parameter tells NI-SCOPE to fetch whatever is currently available. Using -1 for this parameter implies infinite timeout.',
+                },
+            },
+            {
+                'direction': 'in',
+                'enum': None,
+                'name': 'numSamples',
+                'type': 'ViInt32',
+                'documentation': {
+                    'description': '''
```
marcoskirsch: Did you copy all this? If you did, why not assign from the other function and DRY?
This may apply to other places. It's a lot of text to keep out own copy.

src/niscope/metadata/functions_addon.py
```diff
@@ -186,16 +194,301 @@
...
+Device <REPLACE_DRIVER_SPECIFIC_URL_1(features_supported_main)>`__ for
+more information.
+''',
+        },
     },
 }
 
+# Override the 'python' name for some functions.
+functions_python_name = {
+    'Fetch':                      { 'python_name': '_fetch_double',                   },
```
marcoskirsch: In nifgen I kept the python names the same when private. Can you keep this as _fetch?

texasaggie97: That means we would have `fetch`, `fetch_into`, `_fetch` and `_fetch_into`. The first two are hand maintained but different templates and the second two come from Fetch.

Is that okay?

src/niscope/metadata/functions_addon.py
texasaggie97: You mean something like
```python
functions_additional_functions['FetchDouble']['parameters'][3]['documentation'] = functions_additional_functions['FetchDispatcher']['parameters'][3]['documentation']
```

??

src/niscope/metadata/functions_addon.py
marcoskirsch: I think it's fine.

src/niscope/metadata/functions_addon.py
marcoskirsch: I meant from the extracted metadata itself. (I would use `deepcopy()`)

docs/niscope/functions.rst
```diff
@@ -1290,19 +1282,20 @@ niscope.Session methods
 
     :type timeout: float
 
-    :rtype: tuple (wfm, wfm_info)
+    :rtype: tuple (meas_wfm, wfm_info)
 
         WHERE
 
-        wfm (list of float): 
+        meas_wfm (list of float):
```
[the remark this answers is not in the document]
texasaggie97: This is `fetch_array_measurement()` not `fetch()`

The diff is kinda messed up.

src/niscope/metadata/functions_addon.py
texasaggie97: I have made changes to the documentation that was extracted to be more correct for Python.

src/niscope/metadata/functions_addon.py
texasaggie97: Updated

PR #701, opened 2017-12-19
Add default value for array size params in `_into` functions

texasaggie97:
* Requires #699 
* Add default value of None for array_size params for `_into` functions
* If `array_size is None` then get the size from the numpy array

* Unit
* System


texasaggie97:
That is what I am doing with `array_size = len(waveform_array)` or `len(wfm) / self._actual_num_wfms()`

Or are you saying we don't take the parameter at all?

That would get complicated with scope when the size of the array is `number_of_samples_per_wfm * self._actual_num_wfms()`

texasaggie97:
test this please

marcoskirsch:
> Or are you saying we don't take the parameter at all?

Yeah, that's what I'm saying. For the last call you can simply pass a slice. Say you need to fetch 350 samples and your array is 100 elements long.

```
wfm_data = numpy.empty(100)
session.fetch_into(wfm_data)
session.fetch_into(wfm_data)
session.fetch_into(wfm_data)
session.fetch_into(wfm_data[:50])
```

This is more pythonic than having the last line be:

```
session.fetch_into(wfm_data, 50)
```

texasaggie97:
You would have to say `wfm_data[:50*actual_num_wfms()]` for scope, if that will even work.

marcoskirsch:
> You would have to say wfm_data[:50*actual_num_wfms()] for scope, if that will even work.

Should work.

> Edit: Also, there's nothing stopping you from using wfm_data[:50*actual_num_wfms()] now, instead of using the num_samples parameter.

Yeah, but why provide two ways of doing the same thing? We have one way that is both simpler and more idiomatic.

texasaggie97:
We are not going to use this method

Issue #659, opened 2017-12-06
Use Python timedelta and datetime whenever appropriate


marcoskirsch:
Example:

    def get_cal_date_and_time(self, cal_type)

...should return a datetime instead of "(month, day, year, hour, minute)

    def get_ext_cal_recommended_interval(self)

...should return a timedelta rather than a "month"

Any functions that take a timeout... etc.

PR #704, opened 2017-12-19
Add support for custom methods

texasaggie97:
- [x] I've ~~added~~updated tests applicable for this pull request
* Add support for custom, hand maintained methods, including documentation
* Example of how it works for `nifake.get_cal_date_and_time()`

* Show how #659 can be fixed

* Unit
* System


build/templates/functions.rst.mako
```diff
@@ -9,17 +9,26 @@
...
     functions = helper.filter_codegen_functions(functions)
 
+    doc_list = {}
+    for fname in sorted(functions):
+        if functions[fname]['codegen_method'] == 'public':
+            for method_template in functions[fname]['method_templates']:
+                name =  functions[fname]['python_name'] + method_template['method_python_name_suffix']
+                doc_list[name] = { 'filename': method_template['documentation_filename'], 'method_template': method_template, 'function': functions[fname], }
+
+    for custom_method in config['custom_methods']:
```
marcoskirsch: Consider `handcoded_methods' rather than 'custom_methods'.

build/templates/session.py.mako
```diff
@@ -124,6 +124,11 @@ init_call_params = helper.get_params_snippet(init_function, helper.ParameterUsag
 <%include file="${'/session.py' + method_template['session_filename'] + '.py.mako'}" args="f=functions[func_name], config=config, method_template=method_template" />\
 % endfor
 % endfor
+% for custom_method in config['custom_methods']:
+% if custom_method['channel_based']:
```
marcoskirsch: We already have a key `render_in_session_base` to denote "generate this in the base class". You should use that. I can see us needing to put hand coded methods in the base class regardless of whether they are channel based.

src/nidmm/templates/session.py/get_cal_date_and_time.py.mako
```diff
@@ -0,0 +1,27 @@
+<%page args="f, config"/>\
+    def get_cal_date_and_time(self, cal_type):
+        '''get_cal_date_and_time
+
+        Returns the date and time of the last calibration performed.
```
marcoskirsch: it's annoying that the same info is present here and in get_cal_date_and_time.rst.mako. Is there no better way to single source it?

src/nifake/metadata/functions.py
```diff
@@ -1980,5 +1980,80 @@
...
+        'codegen_method': 'public',
+        'returns': 'ViStatus',
+        'parameters': [
+            {
+                'direction': 'in',
+                'enum': None,
+                'name': 'vi',
+                'type': 'ViSession',
+                'documentation': {
+                    'description': 'Identifies a particular instrument session. You obtain the **vi** parameter from niDMM\_init or niDMM\_InitWithOptions. The default is None.',
```
marcoskirsch: I've written shorter descriptions for all nifake documentation strings, rather than copying from real drivers. We should be consistent.

src/nifake/templates/session.py/datetime_wrappers.py.mako
```diff
@@ -0,0 +1,27 @@
+<%page args="f, config"/>\
+    def get_cal_date_and_time(self, cal_type):
+        '''get_cal_date_and_time
+
+        Returns the date and time of the last calibration performed.
```
marcoskirsch: I would simplify this and not make it NI-DMM specific.

texasaggie97: How? This function is very NI-DMM specific. No one else returns all these parameters.

src/nifake/metadata/functions.py
texasaggie97: Shortened

src/nidmm/templates/session.py/get_cal_date_and_time.py.mako
texasaggie97: An alternative would be to add new functions in function_addons and then use method_templates. You would still need specialized templates though because these handwritten functions are so specific it's not work torquing around the code generator to support them that way.

marcoskirsch: I think it's worth it, especially if you can inherit some information from the original function you are wrapping (by assigning metadata from the original dictionary).

src/nifake/templates/session.py/datetime_wrappers.py.mako
texasaggie97: Okay, I''ll make it different from the nidmm function.

Issue #106, opened 2017-07-28
Some public enum values start with _


marcoskirsch:
NI-DMM has a few enum values that start with a number: i.e. NIDMM_ATTR_ID_2_WIRE_RESISTANCE
Currently in Python this is nidmm.Function._2_WIRE_RESISTANCE
But leading _ is reserved for private which is not the case here. We should look for alternatives.

Other MI drivers have similar problems:

NIFGEN_ATTR_5450_OSCILLATOR_PHASE_DAC_VALUE
NISCOPE_ATTR_5102_ADJUST_PRETRIGGER_SAMPLES
NISCOPE_ATTR_5V_OUT_OUTPUT_TERMINAL
NISCOPE_ATTR_5102_ADJUST_PRETRIGGER_SAMPLES
...

DAQmx API spells out the numbers. This would work for NIDMM_ATTR_ID_2_WIRE_RESISTANCE but be really awkward for things such as NIFGEN_ATTR_5450_OSCILLATOR_PHASE_DAC_VALUE.

Fladolcetta:
Is this an issue with the Meta-data or is there another issue? Couldn't we just change the meta data in enum.py to remove the underscore? 
This would change "nidmm.Function._2_WIRE_RESISTANCE" to "nidmm.Function.2_WIRE_RESISTANCE"

marcoskirsch:
I propose we modify the generator to spit out a warning whenever it encounters these. That will at least allow us to compile a full list of the offending attribute names. Once we have that, we can think of a good solution.

marcoskirsch:
From [documentation](https://docs.python.org/3.6/library/enum.html?highlight=enum#allowed-members-and-attributes-of-enumerations)

> The rules for what is allowed are as follows: names that start and end with a single underscore are reserved by enum and cannot be used; all other attributes defined within an enumeration will become members of this enumeration, with the exception of special methods (__str__(), __add__(), etc.) and descriptors (methods are also descriptors).

Fladolcetta:
Lets see if nidaqmx-python and nixnet-python ran into this, and see how they dealt with it.

epage:
Yes, in nidaqmx.  I can't remember for nixnet.  They mirrored our C# API which had the same problem  In most cases, that meant using a word ("FOUR" instead of "4")  I don't think we add leading model numbers which complicates things

As for alternative solutions, I strongly recommend against a leading `_`.  A leading `_` indicates it is private / protected / implementation detail and anyone using it might break

[...]

PR #706, opened 2017-12-19
Remove enums that should have been numerics

### LINKED ISSUE #708: Several properties / functions use enums as constants
For example, property `powerline_freq` uses enum `PowerlineFrequency` which actually contains constants for 50.0 and 60.0. These aren't actually enumerated values and arguably make the API harder to use because there are more types to know about.

* These were inherited from CVI APIs.
* In many cases in LabVIEW we don't use enums for the equivalent properties.
* In C, we do often provide these values as constants, but not enums.

Fladolcetta:
[x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

[x] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.

* This removes enums that simply provide constants.
* Reduces severity of #106.

* Fix #708 

Tox


Fladolcetta:
@marcoskirsch Are you recommending that I also remove SampleClockTimebaseSource, SampleClockSource, and ReferenceClockSource from NI-FGEN? If so, I can do that in this same pull request.

marcoskirsch:
> @marcoskirsch Are you recommending that I also remove SampleClockTimebaseSource, SampleClockSource, and ReferenceClockSource from NI-FGEN? If so, I can do that in this same pull request.

I'm not sure all of these are hooked up to the routing subsystem under the hood.
If they are, then we should remove them but it's ok to do it in a separate PR.
If they aren't, then we should keep things as is, I think.

src/niscope/metadata/enums_addon.py
```diff
@@ -1014,3 +1002,8 @@
         ],
     },
 }
+
+
+enums_additional_enums = {
+    'InputImpedance': {},  # Enum metadata actually contains constants.
```
texasaggie97: If `InputImpedance` is not in the extracted enums.py then you don't need to remove it here.

PR #713, opened 2018-01-02
Add framework for input parameter converters

texasaggie97:
* Adds the ability to define and use converters for input parameters
* Example is timedelta for timeouts
  * Only added to nifake along with tests
  * Remaining drivers will be handled in a separate PR

* Unit
* System


marcoskirsch:
FYI conflicts.

build/helper/codegen_helper.py
```diff
@@ -188,6 +188,7 @@ def _get_ctype_variable_definition_snippet_for_scalar(parameter, parameters, ivi
       7.5. Input is size of output buffer with mechanism ivi-dance, GET_DATA:   visatype.ViInt32(error_code)
         8. Input is size of output buffer with mechanism passed-in:             visatype.ViInt32(buffer_size)
        14. Output scalar or enum:                                               visatype.ViInt32()
+       15. Input uses converter                                                 _TimedeltaConverter(timeout, visatype.ViReal64, 'second')
```
marcoskirsch: converter -> custom converter
_TimedeltaConverter -> _TimeDeltaConverter

build/helper/codegen_helper.py
```diff
@@ -201,6 +202,9 @@ def _get_ctype_variable_definition_snippet_for_scalar(parameter, parameters, ivi
             definition = '{0}.{1}({2})  # case 0.0'.format(module_name, parameter['ctypes_type'], parameter['size']['value'])
         elif parameter['enum'] is not None:
             definition = '{0}.{1}({2}.value)  # case 10'.format(module_name, parameter['ctypes_type'], parameter['python_name'])
+        elif 'python_api_converter_name' in parameter:
+            converter_params = [parameter['python_name'], module_name + '.' + parameter['ctypes_type']] + parameter['python_api_converter_additional_params']
```
marcoskirsch: I think it's simpler / clearer to get rid of additional_params if possible.
In this case, I think you could have subclasses for seconds, milliseconds, etc.

build/templates/session.py.mako
```diff
@@ -33,6 +33,41 @@ from ${module_name} import visatype
 from ${module_name} import ${c['file_name']}  # noqa: F401
 % endfor
 
+import pprint
+pp = pprint.PrettyPrinter(indent=4)
+
+
+class _TimedeltaConverter(object):
```
marcoskirsch: We should try to put converters in their own files, it'd be cleaner.
It means that this template would need to figure out what to import, which can be done by iterating over parameters for all functions and sticking all needed converters in a set. Then expanding the import statements.

docs/conf.py
```diff
@@ -55,15 +55,15 @@
...
+copyright = '2017-2018, National Instruments'
 author = 'National Instruments'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The full version, including alpha/beta/rc tags.
-release = '0.6.1.dev0'
+release = '0.6.0'
```
marcoskirsch: how come this was bumped _down_ ?

generated/nidcpower/session.py
```diff
@@ -8,6 +8,41 @@
...
+                scaling = 1
+            elif library_units == 'milliseconds':
+                scaling = 1000
+            elif library_units == 'microseconds':
+                scaling = 1000000
+            else:
+                raise TypeError("units must be 'seconds', 'milliseconds', or 'microseconds'. Actual {0}".format(library_units))
+
+            self.value = value.total_seconds() * scaling
+        else:
```
marcoskirsch: This allows customer to pass either the original type or the timedelta... I don't think it's worth the flexibility at this point. I'd rather accept just the type.

generated/nidcpower/session.py
```diff
@@ -8,6 +8,41 @@
...
+        self._value = value
+        self._library_type = library_type
+        self._library_units = library_units
+
+        if str(type(value)).find("'datetime.timedelta'") != -1:
+            if library_units == 'seconds':
+                scaling = 1
+            elif library_units == 'milliseconds':
+                scaling = 1000
+            elif library_units == 'microseconds':
```
marcoskirsch: If we create a type for each of these units (seconds, milliseconds, microseconds) the string comparison goes away. I like that.

src/nifake/metadata/functions_addon.py
```diff
@@ -69,6 +69,16 @@
...
 }
 
+# Converted parameters
+functions_converters = {
+    'Read':                             { 'parameters': { 1: { 'python_api_converter_name': '_TimedeltaConverter', 
+                                                               'python_api_converter_type': 'datetime.timedelta',
+                                                               'python_api_converter_additional_params': ["'seconds'"], }, }, },
+    'ReadFromChannel':                  { 'parameters': { 2: { 'python_api_converter_name': '_TimedeltaConverter', 
+                                                               'python_api_converter_type': 'datetime.timedelta',
+                                                               'python_api_converter_additional_params': ["'microseconds'"], }, }, },
```
marcoskirsch: we should avoid this by introducing more converters

src/nifake/metadata/functions_addon.py
```diff
@@ -69,6 +69,16 @@
     'EnumInputFunctionWithDefaults':    { 'parameters': { 1: { 'default_value': 'Turtle.LEONARDO', }, }, },
 }
 
+# Converted parameters
+functions_converters = {
+    'Read':                             { 'parameters': { 1: { 'python_api_converter_name': '_TimedeltaConverter', 
+                                                               'python_api_converter_type': 'datetime.timedelta',
+                                                               'python_api_converter_additional_params': ["'seconds'"], }, }, },
+    'ReadFromChannel':                  { 'parameters': { 2: { 'python_api_converter_name': '_TimedeltaConverter', 
+                                                               'python_api_converter_type': 'datetime.timedelta',
```
marcoskirsch: isn't there a python_type metadata already?

build/helper/codegen_helper.py
texasaggie97: N/A now that they are in a separate file as functions.

build/helper/codegen_helper.py
texasaggie97: N/A now that they are in a separate file as functions. Functions are based on seconds, etc.

build/templates/session.py.mako
texasaggie97: Done.

docs/conf.py
texasaggie97: Didn't make clean first

generated/nidcpower/session.py
texasaggie97: Done, but not as type but as separate functions.

generated/nidcpower/session.py
texasaggie97: I disagree. It seems like it is overkill to force them to create a timedelta if they don't plan on using it.

src/nifake/metadata/functions_addon.py
texasaggie97: I need this for documentation. The python_type is the underlying intrinsic numeric type.

src/nifake/metadata/functions_addon.py
texasaggie97: Done, but as functions.

build/templates/_converters.py.mako
```diff
@@ -0,0 +1,32 @@
+<%
+    config = template_parameters['metadata'].config
+    module_name = config['module_name']
+%>\
+from ${module_name} import visatype
+
+
+def _timedelta_converter(value, library_type, scaling):
```
marcoskirsch: These are functions, so their names should be verbs, not nouns.

* _convert_timedelta
* convert_timedelta_to_seconds
* convert_timedelta_to_milliseconds
* convert_timedelta_to_microseconds

build/templates/_converters.py.mako
```diff
@@ -0,0 +1,32 @@
...
+    if str(type(value)).find("'datetime.timedelta'") != -1:
```
marcoskirsch: What is this if statement for? Is it for backwards compatilibilty / allow users to pass a number in the public API?
If that's the case, I think we should remove it. Does that make a mess of things with defaults?

build/templates/_converters.py.mako
```diff
@@ -0,0 +1,32 @@
...
+        scaled_value = value.total_seconds() * scaling
+    else:
+        scaled_value = value
+
+    if not library_type == visatype.ViReal64:  # ctype integer types don't convert to int from float so we need to
+        scaled_value = int(scaled_value)
```
marcoskirsch: With timeouts, it's better to round up rather than floor the value. Of course it just happens that this conversion function is used with timeouts, but there's nothing about the generator's design that implies this is only for timeouts.

Do you see any problem with rounding down the scaled_value blindly?

marcoskirsch: Also... what about ViReal32?
Maybe we should do the explicit cast no matter what (meaning, remove the if statement).

generated/niscope/session.py
```diff
@@ -2,6 +2,7 @@
 # This file was generated
 import ctypes
 
+from niscope import _converters  # noqa: F401
```
marcoskirsch: We'll have to remember to remove `# noqa: F401`. Can you put a TODO on the same line or will flake8 whine?

build/helper/codegen_helper.py
```diff
@@ -616,6 +619,30 @@ def get_dictionary_snippet(d, indent=4):
         'type': 'ViChar',
         'use_in_python_api': True,
     },
+    {  # 14
+        'ctypes_type': 'ViReal64',
+        'ctypes_type_library_call': 'ViReal64',
+        'ctypes_variable_name': 'timeout_ctype',
+        'default_value': 1.0,
+        'direction': 'in',
+        'documentation': {'description': 'Timeout in seconds'},
```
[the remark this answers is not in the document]
texasaggie97: Are you talking about the documentation? This is only for a test.

build/templates/_converters.py.mako
texasaggie97: Fixed

build/templates/_converters.py.mako
texasaggie97: I believe we should support a bare number as well as a timedelta object.

Removing it is what makes a mess of defaults.
* Right now we only import datetime if we need it
  * We would have to import it before the function so we could do something like `timeout=datetime.timedelta(seconds=5.0)`
* We would have to add support for doing something like `timeout=datetime.timedelta(seconds=5.0)`
  * The code generator as is I don't think would support that.

build/templates/_converters.py.mako
texasaggie97: None of our APIs use ViReal32.

We can't do this all the time. If the library type is a double, we need to leave it as is.

generated/niscope/session.py
texasaggie97: Added since flake8 didn't care.

build/templates/_converters.py.mako
marcoskirsch: > Right now we only import datetime if we need it

We think all the APIs have at least one function with a timeout, so we could add the import to session.py.mako and be done with it.

> The code generator as is I don't think would support that.
 
You're right. This is a bummer. I tried it and it generates (note extra quotes):

     def wait_for_event(self, event_id, timeout='datetime.timedelta(10.0)'):

It seems we added logic to the generator to handle enums. Maybe we got too clever and we make it so that the default must be specified as a string (not a number) in the metadata and is copied verbatim (for strings we'd have to make sure the string literal includes quotes).

I don't think this is too hard to fix and may even be a simplification. It's weird that some APIs take in numbers as float, others as int, different units... in addition to timedelta.

From [Zen of Python](https://www.python.org/dev/peps/pep-0020/#id3)
_"There should be one-- and preferably only one --obvious way to do it."_

Allowing numbers breaks this.

docs/niscope/functions.rst
```diff
@@ -110,7 +110,7 @@ niscope.Session methods
 
 
 
-.. function:: cal_self_calibrate(option=niscope.Option.SELF_CALIBRATE_ALL_CHANNELS)
+.. function:: cal_self_calibrate(option=niscope.enums.Option.SELF_CALIBRATE_ALL_CHANNELS)
```
marcoskirsch: Is this a fix to the documentation?
I thought when you import niscope all the enums would be imported inside the niscope namespace.

Is that not the case? And if not, isn't that what we'd want?
I couldn't find "enums" in any of the examples.

```
Kirsch0:nimi-python kirsch$ find  * | grep examples.*\.py | xargs grep -ir enums
```

generated/nifake/session.py
```diff
@@ -272,14 +272,14 @@ def read_from_channel(self, maximum_time):
             session['0,1'].read_from_channel(maximum_time)
 
         Args:
-            maximum_time (int): Specifies the **maximum_time** allowed in years.
+            maximum_time (datetime.timedelta or int): Specifies the **maximum_time** allowed in microseconds.
```
marcoskirsch: No longer `or int`.

docs/niscope/functions.rst
texasaggie97: I had simplified add_all too much. I can't put the 'enum.' in functions_addon and instead have to go back to enums being a special case for default values.

generated/nifake/session.py
texasaggie97: Removed or clause

docs/nidmm/functions.rst
```diff
@@ -1036,7 +1036,7 @@ nidmm.Session methods
 
 
 
-.. function:: get_dev_temp(options='')
+.. function:: get_dev_temp(options='""')
```
marcoskirsch: Do these extra quotes cause any trouble?

texasaggie97: No, they are required since we use this string exactly. Other wise it would end up being option_string= instead of option_string=""

texasaggie97:
test this please

texasaggie97:
test this please

PR #714, opened 2018-01-02
Create object and return iter instead of implementing iter on nimodinst session

### LINKED ISSUE #669: Iterating over devices of an nimodinst.Session changes the state of the Session object
This is incorrect behavior, and means that using two iterators is impossible.

The solution is to create a new object and return that rather than self in __iter__().

texasaggie97:
[X] I've ~~added~~removed tests applicable for this pull request
* Create a iterator of list of Devices when iterating on session
* This allows multiple simultaneous iterators.

* Fix #669 

* Unit
* System

### DIFF
--- a/generated/nimodinst/session.py
+++ b/generated/nimodinst/session.py
@@ -140,22 +140,10 @@ class Session(object):
         return self._item_count
 
     def __iter__(self):
-        self._current_item = 0
-        return self
-
-    def get_next(self):
-        if self._current_item + 1 > self._item_count:
-            raise StopIteration
-        else:
-            result = Device(self, self._current_item)
-            self._current_item += 1
-            return result
-
-    def next(self):
-        return self.get_next()
-
-    def __next__(self):
-        return self.get_next()
+        devices = []
+        for i in range(self._item_count):
+            devices.append(Device(self, i))
+        return iter(devices)
 
     def close(self):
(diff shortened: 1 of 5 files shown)


generated/nimodinst/session.py
```diff
@@ -140,22 +140,10 @@ def __len__(self):
         return self._item_count
 
     def __iter__(self):
```
marcoskirsch: This implementation is functionally correct but i think it's not the proper way of doing iterators and goes against the reason why they exist.

This front loads performance and memory hit. I think a more correct implementation would be to have an iterator class which takes a Session and sets its _current_item to 0. __get_item__() constructs the Device. __next__() increments _current_item.

texasaggie97: If it goes into another class, then we either:
1. open a new session, which is a much larger performance and memory hit
1. pass in the modinst session and somehow keep it open even if the session is explicitly closed.

The only memory and performance is in the Device creation. ModInst has internally already gotten all that information.

marcoskirsch: You would do #2: pass the already open session object. You don't need to keep it open more than you do today... as you have it, if the session is explicitly closed the iterator will stop working because even though the Device instances exist, the calls to the driver will fail.

texasaggie97: Changed

PR #717, opened 2018-01-03
[2]Remove enums whose values can/should be represented by a boolean

### LINKED ISSUE #666: Get rid of enums that could be a boolean
We should check properties and functions.

We would need to add a mechanism to replace attributes that use enums with a boolean.

texasaggie97:
* Removes enums that are or should be represented by a intrinsic boolean
  * I manually went through the generated enums.py to inspect
  * There were some that could have been represented by a boolean, but I felt they shouldn't since the intent wouldn't be clear from True/False. I.e. Positive/Negative polarity
* Disconnect enum from the associated attribute

* Fix #666 

* System
* Unit


docs/nidcpower/attributes.rst
```diff
@@ -433,17 +433,17 @@ nidcpower.Session properties
...
-    | Permissions    | read-write                       |
-    +----------------+----------------------------------+
-    | Channel Based  | True                             |
-    +----------------+----------------------------------+
-    | Resettable     | No                               |
-    +----------------+----------------------------------+
+    +----------------+------------+
+    | Characteristic | Value      |
+    +================+============+
+    | Datatype       | int        |
```
marcoskirsch: Shouldn't these be booleans?

generated/nidcpower/session.py
```diff
@@ -185,7 +185,7 @@ class _SessionBase(object):
         session['0,1'].current_level = var
         var = session['0,1'].current_level
     '''
-    current_level_autorange = attributes.AttributeEnum(attributes.AttributeViInt32, enums.CurrentLevelAutorange, 1150017)
+    current_level_autorange = attributes.AttributeViInt32(1150017)
```
marcoskirsch: Does this work correctly? We want clients to pass a bool, and then we turn around and pass a ViInt32 to the C API.

src/nidcpower/metadata/enums_addon.py
```diff
@@ -179,6 +179,10 @@
         ],
     },
     'PowerLineFrequency': {}, # Enum metadata actually contains constants.
```
marcoskirsch: It'd be better to keep the comments consistent. For example, in src/nidcpower/metadata/attributes_addon.py you have "# POWER_LINE_FREQUENCY" but a comment like this one could apply perfectly fine.

Also:
"# Disconnecting enum since it is a boolean" vs " # Delete because boolean values only"

texasaggie97: Made comments consistent across drivers

docs/nidmm/session.rst
```diff
@@ -36,7 +36,7 @@ nidmm.Session
    +-----------------------------------------------------+----------------------------------------------+
    | :py:attr:`current_source`                           | float                                        |
    +-----------------------------------------------------+----------------------------------------------+
-   | :py:attr:`dc_bias`                                  | :py:data:`DCBias`                            |
+   | :py:attr:`dc_bias`                                  | int                                          |
```
marcoskirsch: in attributes.rst this is listed as bool, same for some others.

generated/nidcpower/session.py
```diff
@@ -546,7 +596,8 @@ class _SessionBase(object):
     This attribute is not available in a session involving multiple channels.
     '''
     measure_record_length_is_finite = attributes.AttributeViBoolean(1150064)
-    '''
+    '''Data Type: bool
```
marcoskirsch: Very minor and personal preference, I think "Type" suffices.

generated/nidcpower/session.py
```diff
@@ -705,19 +767,22 @@ class _SessionBase(object):
...
+
     Specifies the power source to use. NI-DCPower switches the power source used by the  device to the specified value.
     Default Value: NIDCPOWER_VAL_AUTOMATIC
     is set to NIDCPOWER_VAL_AUTOMATIC. However, if the session is in the Committed or Uncommitted state  when you set this attribute, the power source selection only occurs after you call the  niDCPower_Initiate function.
 
     Note: Automatic selection is not persistent and occurs only at the time this attribute
     '''
     power_source_in_use = attributes.AttributeEnum(attributes.AttributeViInt32, enums.PowerSourceInUse, 1150001)
-    '''
+    '''Data Type: PowerSourceInUse
```
marcoskirsch: I'm not sure about this, but for enums, would it make sense to type them as `nidcpower.PowerSourceInUse`? This is how a user would type it by default.

docs/nidmm/session.rst
texasaggie97: That's because it is one of the ones disconnected from the underlying enum. The values are ON (1) and OFF (0).

This is fixing the issue you brought up earlier in the PR that the type should be bool and not int. 

Which do you want?

generated/nidcpower/session.py
texasaggie97: Fixed

generated/nidcpower/session.py
texasaggie97: Fixed

docs/nidmm/session.rst
marcoskirsch: I'm a little confused.

Doesn't the client need to pass a bool for this property? Why would the documentation say int?

texasaggie97: Because there were multiple places this needed to be fixed. Added 'python_type' to attributes and used that everywhere instead of triplicating the logic.

nidmm_sfp.py
```diff
@@ -0,0 +1,20 @@
+import nimodinst
```
marcoskirsch: I'm guessing this was added by accident

src/nidmm/metadata/enums_addon.py
```diff
@@ -6,12 +6,14 @@
...
 # are supported in Python
 enums_codegen_method = {
-    'PowerlineFrequency': { "codegen_method": "no" }, # Enum metadata actually contains constants.
-    'CurrentSource': { "codegen_method": "no" }, # Enum metadata actually contains constants.
-    'InputResistance': { "codegen_method": "no" }, # Enum metadata actually contains constants.
 }
 
 enums_additional_enums = {
-    'DigitsResolution': {},  # Just delete. Otherwise we have to make it so that it will generate valid code
+    'DigitsResolution': {},  # Just delete this enum. Otherwise we have to make it so that it will generate valid code
```
marcoskirsch: Enum metadata actually contains constants.

texasaggie97: That too. But we were deleting it even before because the values of it caused the code generator to break.

nidmm_sfp.py
texasaggie97: Removed

src/nidmm/metadata/enums_addon.py
texasaggie97: Updated comments but still mention that invalid code would be generated by this enum

PR #718, opened 2018-01-04
Change repeated capabilities

texasaggie97:
* Moves repeated capabilities away from session object and on to specific instances of the different types
* Repeated capability can be integer or string, single item or list

* System
* Unit


build/templates/session.py.mako
```diff
@@ -48,7 +48,23 @@ class ${session_context_manager}(object):
...
-class _SessionBase(object):
+% for rep_cap in config['repeated_capabilities']:
+class _${rep_cap['python_class_name']}(object):
+    def __init__(self, ${config['session_handle_parameter_name']}):
+        self._${config['session_handle_parameter_name']} = ${config['session_handle_parameter_name']}
+
+    def __getitem__(self, repeated_capability):
+        '''Set/get properties or call methods with a repeated capability (i.e. channels)'''
+        if isinstance(repeated_capability, list):
+            rep_cap_list = [str(r) if str(r).startswith('${rep_cap['prefix']}') else '${rep_cap['prefix']}' + str(r) for r in repeated_capability]
```
auchter: Hmmm... IVI repeated capability strings are case insensitive, but the prefix check here is case sensitive.

texasaggie97: Made check case insensitive.

CHANGELOG.md
```diff
@@ -14,6 +14,26 @@ All notable changes to this project will be documented in this file.
...
+    * Repeated capabilities are no longer accessed directly on the session object. Instead there is now a repeated capabilies object for each type of repeated capability that the driver supports. Some drivers do not use repeated capabilities so there is no change for them.
+    * The repeated capabilities object can take integers or strings, a single item or a list.
+    * The repeated capabilities object knows the proper prefix and will add it if needed.
+    * The following are all legal
+        ```python
+        import niscope
+        session = niscope.Session('PXI1Slot2')
+
+        # Channel repeated capabilities
+        session.channel['0'].channel_enabled = True
```
marcoskirsch: since this is a container that you're indexing, shouldn't it be in plural form ("channels")?

CHANGELOG.md
```diff
@@ -14,6 +14,26 @@ All notable changes to this project will be documented in this file.
...
+        session.channel[0].channel_enabled = True
+        session.channel[[0, 1, 3]].channel_enabled = True
+        wfm = session.channel[[0, 1, 3]].fetch(5000)
+
+        # P2P repeated capabilities
```
marcoskirsch: P2P is not supported, right? Maybe use NI-FGEN MarkerEvents for the example?

CHANGELOG.md
```diff
@@ -29,14 +49,19 @@ All notable changes to this project will be documented in this file.
   * #### Removed
 * ### NI-DCPower
   * #### Added
+    * Channel repeated capability
   * #### Changed
   * #### Removed
 * ### NI-FGEN
   * #### Added
+    * Channel repeated capability
+    * P2P repeated capability
```
marcoskirsch: I thought we took all P2P stuff out of the Python bindings.

CHANGELOG.md
```diff
@@ -29,14 +49,19 @@ All notable changes to this project will be documented in this file.
   * #### Removed
 * ### NI-DCPower
   * #### Added
+    * Channel repeated capability
   * #### Changed
   * #### Removed
 * ### NI-FGEN
   * #### Added
```
marcoskirsch: Script Triggers, Marker Events are two NI-FGEN and NI-HSDIO repeated capabilities. Don't know of others.

build/helper/documentation_helper.py
```diff
@@ -402,7 +402,7 @@ def _format_type_for_docstring(param, numpy, config):
 
 
 rep_cap_method_desc_docstring = rep_cap_method_desc + '''
-    session['0,1'].{1}({2})
+    session.channel[[0, 1]].{1}({2})
```
marcoskirsch: Question for the Python crowd... is indexing using an array "Pythonic"? Or is it strange?

build/templates/session.py.mako
```diff
@@ -48,7 +48,25 @@ class ${session_context_manager}(object):
...
+        self._encoding = encoding
+
+    def __getitem__(self, repeated_capability):
+        '''Set/get properties or call methods with a repeated capability (i.e. channels)'''
+        if isinstance(repeated_capability, list):
+            rep_cap_list = [str(r) if str(r).lower().startswith('${rep_cap['prefix'].lower()}') else '${rep_cap['prefix']}' + str(r) for r in repeated_capability]
+        else:
+            rep_cap_list = [str(repeated_capability) if str(repeated_capability).lower().startswith('${rep_cap['prefix'].lower()}') else '${rep_cap['prefix']}' + str(repeated_capability)]
+
+        return _RepeatedCapbilities(${config['session_handle_parameter_name']}=self._${config['session_handle_parameter_name']}, repeated_capability=','.join(rep_cap_list), library=self._library, encoding=self._encoding, freeze_it=True)
```
marcoskirsch: > _RepeatedCapbilities

typo

generated/nidcpower/session.py
```diff
@@ -21,7 +21,23 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        if isinstance(repeated_capability, list):
+            rep_cap_list = [str(r) if str(r).lower().startswith('') else '' + str(r) for r in repeated_capability]
+        else:
+            rep_cap_list = [str(repeated_capability) if str(repeated_capability).lower().startswith('') else '' + str(repeated_capability)]
+
+        return _RepeatedCapbilities(vi=self._vi, repeated_capability=','.join(rep_cap_list), library=self._library, encoding=self._encoding, freeze_it=True)
+
+
+class _RepeatedCapbilities(object):
     '''Base class for all NI-DCPower sessions.'''
```
marcoskirsch: Seems odd that a class named _RepeatedCapbilities is 'Base class for all NI-DCPower sessions.'

build/templates/session.py.mako
```diff
@@ -48,7 +48,25 @@ class ${session_context_manager}(object):
...
+% for rep_cap in config['repeated_capabilities']:
+class _${rep_cap['python_class_name']}(object):
+    def __init__(self, ${config['session_handle_parameter_name']}, library, encoding):
+        self._${config['session_handle_parameter_name']} = ${config['session_handle_parameter_name']}
+        self._library = library
+        self._encoding = encoding
+
+    def __getitem__(self, repeated_capability):
+        '''Set/get properties or call methods with a repeated capability (i.e. channels)'''
+        if isinstance(repeated_capability, list):
```
marcoskirsch: This seems restrictive. Clients should be able to pass anything iterable even if not a list. For example, range() doesn't return a list.

texasaggie97: Changed to use duck typing. Anything iterable will now work.

texasaggie97:
There is a scope system test failing because it sets `test_channels = '0,1'`. Should I
1. Fix the test to say `test_channels = range(2)`
1. Fix the code to allow it?

build/helper/documentation_helper.py
texasaggie97: Channels 0, 1, 4, 12 definitely needs to be supported. Something like `my_channels = session.channels[0:1] + session.channels[4] + session.channels[12]` to do it though seems like a step backwards.

thejcannon: If you wanted a list that contains elements at indices 0, 1, 4, and 12, how would you construct that list?

MrElusive: Indexing using arrays is not strange at all. See Numpy's advanced indexing feature: https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.indexing.html#advanced-indexing

texasaggie97: We don't want 4 object in a list out of this. We need one object where we can internally put together a string that looks like '0,1,4,12'.

MrElusive: @texasaggie97 So do that: index with an array and return whatever object you need.

thejcannon: After seeing numpy's "advanced-indexing" I think allowing an array makes sense. I think as long as the user can use integers and slices then users will find themselves successful. Allowing arrays seems like a "nice-to-have" but if it is easy to support, and the user can easily understand what happens when an array is used, then I say go for it.

> We need one object where we can internally put together a string that looks like '0,1,4,12'.
The string is an implementation detail. What you want is for the user to be able to say "on channels x do y". Since we're exposing the "seuqence" version of  `__getitem__` on the `channels` attribute to access specific channels, I see it the same as accessing indices of a sequence type to operate on.

texasaggie97: Yes, it is an implementation detail, but it is an important detail. We need the call to condense down to a single call with '0,1,4,12' not a call with '0,1', then a call with '4', then a call with '12'.

It is very possible that I don't see something that would allow `my_channels = session.channels[0:1] + session.channels[4] + session.channels[12]` to do this. But also, this seems  very verbose.

`(session.channels[0:1] + session.channels[4] + session.channels[12]).fetch(1000)` seems unwieldy.

marcoskirsch: This can probably be achieved, if we decide that's the way we want to go, by implementing __add__().

marcoskirsch: > The string is an implementation detail.

Under the hood, this will touch hardware. If a user sets the voltage on several channels "all at once" but the Python code somewhat serializes this, DUT can be damaged. Mark is right that we need to make this work.

generated/nidcpower/session.py
marcoskirsch: I haven't come up with a perfect name for it but _SessionBase is probably better than _RepeatedCapabilities.

When I read _RepeatedCapabilities, I think of a collection of repeated capabilities such as "0" or "ScriptTrigger2".

marcoskirsch:
> There is a scope system test failing because it sets test_channels = '0,1'. Should I

Why does it fail? Which test?

build/helper/documentation_helper.py
thejcannon: I agree serializing the calls is not an option. I just want to make sure the implementation details don't bleed out into the interface. After seeing that numpy uses arrays to mean a list of indices, this is starting to look less weird.

I also don't think `add()` should be added to the type that `__getitem__()` returns just yet (mainly because the list-of-indices should cover *most* of the use cases). I just wanted to bring it up to show the similarities between two sequence types (yours and lists), to help us think about how we should design the interface,

Just make sure the normal use case isn't broken (integers and slices), and the array-of-indices has good documentation and I think you're good to go. 👍

texasaggie97:
The test no longer fails because I change channels to be `range(2)` instead of '0,1'. 

The code that generates the string doesn't handle that correctly, nor does it handle '0-3'.  

Should it?

marcoskirsch:
The code that generates the string doesn't handle that correctly, nor does it handle '0-3'.
    
    Should it?

So for channels, we can no longer use the syntax that the C/LabVIEW APIs take?
I suppose that's ok, because https://www.python.org/dev/peps/pep-0020/#id3

texasaggie97:
Right now we don't accept that syntax. I think I could fix it so we do, but should I?

generated/nifake/session.py
```diff
@@ -113,10 +140,16 @@ class _SessionBase(object):
     An attribute of type string with read/write access.
     '''
 
-    def __init__(self, repeated_capability):
-        self._library = library_singleton.get()
+    def __init__(self, repeated_capability, vi=None, library=None, encoding=None, freeze_it=False):
```
marcoskirsch: Why are vi, library, encoding optional?

generated/nifake/unit_tests/test_session.py
```diff
@@ -658,14 +658,98 @@ def test_repeated_capability_method_on_specific_channel(self):
         self.patched_library.niFake_ReadFromChannel.side_effect = self.side_effects_helper.niFake_ReadFromChannel
         self.side_effects_helper['ReadFromChannel']['reading'] = test_reading
         with nifake.Session('dev1') as session:
-            value = session['3'].read_from_channel(test_maximum_time)
+            value = session.channels['3'].read_from_channel(test_maximum_time)
         self.patched_library.niFake_ReadFromChannel.assert_called_once_with(matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), matchers.ViStringMatcher('3'), matchers.ViInt32Matcher(test_maximum_time_us), matchers.ViReal64PointerMatcher())
         assert value == test_reading
 
+    def test_repeated_capability_method_on_specific_channel_list_int(self):
```
marcoskirsch: I think that if we test the (not yet created) function that creates the repeated capability string very thoroughly, then the tests from Session can be a lot simpler, just verify that that one function is called correctly. Would make things easier to follow: you wouldn't need so many verbose unit tests here.

generated/nifgen/session.py
```diff
@@ -59,6 +59,114 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                # Add prefix to each entry
+                rep_cap_list = ['' + str(r) for r in rep_cap_list]
+            # Otherwise it must be a single item
+            except (TypeError, AttributeError):
+                rep_cap_list = [str(repeated_capability) if str(repeated_capability).lower().startswith('') else '' + str(repeated_capability)]
+
+        return _SessionBase(vi=self._vi, repeated_capability=','.join(rep_cap_list), library=self._library, encoding=self._encoding, freeze_it=True)
+
+
+class _P2PStreams(object):
```
marcoskirsch: I don't think you need to have a _Channels class, a _P2PStreams class, etc.
A single class would suffice, the only thing that changes is the prefix which can be passed in when the container is instantiated inside Session.

generated/nifgen/session.py
```diff
@@ -3255,22 +3369,24 @@ def _error_message(self, error_code):
...
         super(Session, self).__init__(repeated_capability='')
+        self._library = library_singleton.get()
+        self._encoding = 'windows-1251'
         self._vi = 0  # This must be set before calling _initialize_with_channels().
         self._vi = self._initialize_with_channels(resource_name, reset_device, option_string)
+        self.channels = _Channels(self._vi, self._library, self._encoding)
+        self.p2p_streams = _P2PStreams(self._vi, self._library, self._encoding)
+        self.script_triggers = _ScriptTriggers(self._vi, self._library, self._encoding)
+        self.markers = _Markers(self._vi, self._library, self._encoding)
+        param_list = []
```
marcoskirsch: This param_list stuff is just for __repr__ right? (this just for clarity)
Regardless consider moving the code that creates that script to a separate function or at least add a comment.

generated/nifgen/session.py
```diff
@@ -3255,22 +3369,24 @@ def _error_message(self, error_code):
...
 class Session(_SessionBase):
     '''An NI-FGEN session to a National Instruments Signal Generator.'''
 
     def __init__(self, resource_name, reset_device=False, option_string=""):
         super(Session, self).__init__(repeated_capability='')
+        self._library = library_singleton.get()
+        self._encoding = 'windows-1251'
         self._vi = 0  # This must be set before calling _initialize_with_channels().
         self._vi = self._initialize_with_channels(resource_name, reset_device, option_string)
+        self.channels = _Channels(self._vi, self._library, self._encoding)
```
marcoskirsch: add a comment before you unroll this

`# Instantiate repeated capabilities`

Although if you go with the single class approach and the name is self-explanatory then the comment may become unnecessary.

src/nidcpower/metadata/config.py
```diff
@@ -23,5 +23,8 @@
     },
     'init_function': 'InitializeWithChannels',
     'custom_types': [],
+    'repeated_capabilities': [
+        {'python_name': 'channels', 'python_class_name': 'Channels', 'prefix': '', },
```
marcoskirsch: `python_class_name` not needed.

generated/nifgen/session.py
texasaggie97: Done

generated/nifgen/session.py
texasaggie97: Added this plus some more comments in \_\_init\_\_

generated/nifgen/session.py
texasaggie97: Added comment since that is easier that code genning another function and code genning the call to the function.

src/nidcpower/metadata/config.py
texasaggie97: Removed

build/templates/session.py.mako
```diff
@@ -86,6 +86,35 @@ class ${session_context_manager}(object):
 
 
 % endif
+% for rep_cap in config['repeated_capabilities']:
+class _${rep_cap['python_class_name']}(object):
+    def __init__(self, ${config['session_handle_parameter_name']}, library, encoding):
```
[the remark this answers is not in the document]
marcoskirsch: I think it's ok to do within our own module, just like we import files that are private (start with a _).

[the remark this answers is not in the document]
texasaggie97: `_RepeatedCapabilitied()` now takes `Session()` and uses values from it instead of taking the individual values.

CHANGELOG.md
```diff
@@ -34,6 +35,17 @@ All notable changes to this project will be documented in this file.
...
     * #### Added
+        * `channel` repeated capability - See #737 for discussion
+    * #### Changed
+        * Metadata updated to NI-DCPower 17.6.1
+    * #### Removed
+* ### NI-FGEN
+    * #### Added
+        * Repeated capablilites - See #737 for discussion:
+            * `channel` repeated capability
+            * `p2p_streams` repeated capability
```
marcoskirsch: Since #724 is not merged yet, nor is it decided upon, you should remove p2p_streams from this PR.

README.rst
```diff
@@ -83,6 +83,33 @@ The following is a basic example of using the **nidmm** module to open a session
...
+
+* Repeated capabilities are accessed using a repeated capabilies object for each type of repeated capability that the driver supports. Some drivers do not use repeated capabilities.
+* The repeated capabilities object can take integers or strings, a single item or a list, or a slice.
+* The repeated capabilities object knows the proper prefix and will add it if needed.
+* The following are all legal
+
+    .. code-block:: python
+
+        import nifgen
+        session = nifgen.Session('PXI1Slot2')
```
marcoskirsch: how come you didn't use `with`? If you'd still rather not, then maybe show how you close the session?

README.rst
```diff
@@ -83,6 +83,33 @@ The following is a basic example of using the **nidmm** module to open a session
...
+
+        # Channel repeated capabilities
+        session.channels['0'].channel_enabled = True
+        session.channels[0].channel_enabled = True
+        session.channels[[0, 1, 3]].channel_enabled = True
+        session.channels[range(8)].channel_enabled = True  # channels 0, 1, 2, 3, 4, 5, 6, 7
+        session.channels[:8].channel_enabled = True  # channels 0, 1, 2, 3, 4, 5, 6, 7
+        wfm = session.channels[[0, 1, 3]].fetch(5000)
+
+        # P2P repeated capabilities
```
marcoskirsch: bad comment

README.rst
```diff
@@ -83,6 +83,33 @@ The following is a basic example of using the **nidmm** module to open a session
...
+        i = session.script_triggers['0'].SCRIPT_TRIGGERS_COUNT
```
marcoskirsch: use python_names for the attributes.

README.rst
```diff
@@ -83,6 +83,33 @@ The following is a basic example of using the **nidmm** module to open a session
...
+        i = session.script_triggers[0].SCRIPT_TRIGGERS_COUNT
+        i = session.script_triggers[[0, 1, 3]].SCRIPT_TRIGGERS_COUNT
+        i = session.script_triggers['ScriptTrigger0'].SCRIPT_TRIGGERS_COUNT
```
marcoskirsch: Does this actually work? you don't get "ScriptTriggerScriptTrigger0" passed into the C function?
If it works, I'd consider it an implementation fluke and would not document it.

build/helper/parameter_usage_options.py
```diff
@@ -5,32 +5,36 @@ class ParameterUsageOptions(Enum):
     '''Different usage options for parameter lists.'''
 
     SESSION_METHOD_DECLARATION = 1
```
marcoskirsch: We should [just use auto](https://docs.python.org/3/library/enum.html#using-automatic-values) here.

build/templates/_converters.py.mako
```diff
@@ -7,6 +7,26 @@ from ${module_name} import visatype
 import datetime
 
 
+def convert_repeated_capabilities(repeated_capability, prefix=''):
+    # First try it as a list
+    if isinstance(repeated_capability, list):
```
marcoskirsch: This might be too restrictive. What if the structure is something other than a list, like array.array? Python user would expect it to work because duck typing.
Consider implementing with try/except pattern instead of looking for specific type.

build/templates/_converters.py.mako
```diff
@@ -7,6 +7,26 @@ from ${module_name} import visatype
...
+        rep_cap_list = [str(r) if str(r).lower().startswith(prefix.lower()) else prefix + str(r) for r in repeated_capability]
+    else:
+        # Then try it as a slice
+        try:
+            def ifnone(a, b):
+                return b if a is None else a
+            # Turn the slice into a list so we can iterate over it
```
marcoskirsch: If you turn the slice into a list, can't you just call this function with the list recursively?

generated/nifake/session.py
marcoskirsch: I think it's less error prone if they weren't optional and Session passes them when it chains up to its parent class' `__init__()` method.

texasaggie97: Which is preferred?
```python
super().__init__(rep_cap, None, None, None)
self._vi = ...
self._library = ...
self._encoding = ...
```

or
```python
self._vi = ...
self._library = ...
self._encoding = ...
super().__init__(rep_cap, self._vi, self._library, self._encoding)
```

marcoskirsch: Normally it's good practice to chain up the constructor first, then add additional code.

texasaggie97: That is why I had made them optional. So don't make them optional, but explicitly pass in `None`?

build/helper/parameter_usage_options.py
texasaggie97: `auto()`  is only available in 3.6 and the enum backport. I used the solution given in the 3.5 documentation:
```python
class AutoNumber(Enum):
    def __new__(cls):
        value = len(cls.__members__) + 1
        obj = object.__new__(cls)
        obj._value_ = value
        return obj


class ParameterUsageOptions(AutoNumber):
    '''Different usage options for parameter lists.'''

    SESSION_METHOD_DECLARATION = ()
    '''For declaring a regular method in Session'''
    SESSION_INIT_DECLARATION = ()```

build/templates/_converters.py.mako
texasaggie97: I did that at first. This is fixing the issue that broke using string. Strings are iterable so it was iterating over the string and that is what would turn '0,1' into '0,,,1' and 'ScriptTrigger0' into 'S,c,r,i,p,t,T,r,i,g,g,e,r,0'.

I would need to explicitly look for a string and handle that first, and then try to iterate.

build/templates/_converters.py.mako
texasaggie97: Done

README.rst
marcoskirsch: > Do you think we should never accept strings that already have the prefix?

Well, the whole point of these containers is for the client to not have to remember magic strings. I don't think it needs to be documented, even if we still accept them.

build/templates/_converters.py.mako
marcoskirsch: I think that's reasonable.

texasaggie97: Changed the logic to first look for six.string_type or six.text_type. This allows us to accept any iterable.

This adds a dependency to six for all drivers.

README.rst
texasaggie97: Okay, removed from readme.

PR #720, opened 2018-01-09
[NOT TO BE MERGED]Arrays prototype

texasaggie97:
* Prototype using the Python array type
* Added fetch_array method, did not remove any others
* Testing with USB-5133
    * Python 2.7.7, 32bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00400 | 0.00400 | 0.00400
        100000 | 0.03300 | 0.01600 | 0.01500
        1000000 | 0.30000 | 0.12500 | 0.11800
        5000000 | 1.48100 | 0.61400 | 0.57300
        8300000 | 2.44300 | 1.01200 | 0.94500
        15000000 | 4.42700 | 1.83200 | 1.70900
        20000000 | 0.00000 | 2.36600 | 2.27600

    * Python 3.4.3, 64bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00519 | 0.00424 | 0.00397
        100000 | 0.03394 | 0.01587 | 0.01500
        1000000 | 0.31080 | 0.12530 | 0.11717
        5000000 | 1.54739 | 0.61432 | 0.57180
        8300000 | 2.60150 | 1.01512 | 0.94128
        15000000 | 4.65273 | 1.83666 | 1.70207
        20000000 | 6.21035 | 2.44924 | 2.26922

    * Python 3.5.0, 64bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00460 | 0.00409 | 0.00405
        100000 | 0.03342 | 0.01610 | 0.01537
        1000000 | 0.30525 | 0.12506 | 0.11702
        5000000 | 1.52203 | 0.61116 | 0.57092
        8300000 | 2.53468 | 1.00989 | 0.94101
        15000000 | 4.56597 | 1.82732 | 1.70366
        20000000 | 6.08748 | 2.43403 | 2.26965

    * Python 3.6.3, 64bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00464 | 0.00387 | 0.00443
        100000 | 0.03337 | 0.01548 | 0.01599
        1000000 | 0.30236 | 0.12489 | 0.11719
        5000000 | 1.50744 | 0.61145 | 0.57100
        8300000 | 2.50758 | 1.01020 | 0.94058
        15000000 | 4.52244 | 1.82764 | 1.70259
        20000000 | 6.02093 | 2.43612 | 2.26899

* Testing with PXIe-5122
    * Python 2.7.7, 32bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00100 | 0.00000 | 0.00100
        100000 | 0.02600 | 0.01000 | 0.00800
        1000000 | 0.26500 | 0.09000 | 0.08100
        5000000 | 1.32400 | 0.44800 | 0.40800
        8300000 | 2.18700 | 0.74100 | 0.67000

    * Python 3.4.3, 64bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00100 | 0.00100 | 0.00000
        100000 | 0.02703 | 0.01000 | 0.00900
        1000000 | 0.27808 | 0.09402 | 0.08502
        5000000 | 1.37935 | 0.46612 | 0.41912
        8300000 | 2.31263 | 0.77321 | 0.71819

    * Python 3.5.0, 64bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00000 | 0.00000 | 0.00000
        100000 | 0.02701 | 0.01000 | 0.00900
        1000000 | 0.26908 | 0.09302 | 0.08502
        5000000 | 1.35436 | 0.46411 | 0.42511
        8300000 | 2.26361 | 0.76720 | 0.72419

    * Python 3.6.3, 64bit

        type | list | array | numpy array
        --- | --- | --- | ---
        1000 | 0.00000 | 0.00101 | 0.00000
        100000 | 0.02601 | 0.00999 | 0.00801
        1000000 | 0.26006 | 0.09501 | 0.08503
        5000000 | 1.30532 | 0.47913 | 0.42111
        8300000 | 2.18660 | 0.77821 | 0.71719

* Time in seconds
* Time of 0 indicates out of memory
* Code used to test moved to tools directory


marcoskirsch:
MD tables are so much simpler than RST one!

docs/niscope/session.rst
```diff
@@ -410,6 +410,8 @@ niscope.Session
    +-------------------------------------------------------+
    | :py:func:`fetch`                                      |
    +-------------------------------------------------------+
+   | :py:func:`fetch`                                      |
```
marcoskirsch: Seems like the name rendered wrong, should be fetch_array. Something to look into.
Should I open an issue?

generated/niscope/session.py
```diff
@@ -2302,6 +2302,21 @@ def fetch(self, num_samples, timeout=5.0):
...
         return self._fetch(num_samples, timeout)
 
+    def fetch_array(self, num_samples, timeout=5.0):
+        import array
+
+        vi_ctype = visatype.ViSession(self._vi)  # case 1
+        channel_list_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case 2
+        timeout_ctype = visatype.ViReal64(timeout)  # case 9
+        num_samples_ctype = visatype.ViInt32(num_samples)  # case 9
+        wfm_array = array.array("d", range(num_samples * self._actual_num_wfms()))
```
marcoskirsch: This initializes the array as [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0,... which surely isn't what you intended and may explain the crappy performance.

So... one way to initialize to zeroes that may be marginally faster than actually using the iterator returned by range which returns an int that also needs to be cast is to write our own (untested):

'''
class Zeroes:
    def __init__(self):
        pass
    	
    def __iter__(self):
        return self

    def next(self): # Python 3: def __next__(self)
        return 0.0
 '''

I'm looking for a cheaper way, but I'm guessing you looked as well?

generated/niscope/session.py
```diff
@@ -2302,6 +2302,21 @@ def fetch(self, num_samples, timeout=5.0):
...
+        addr, count = wfm_array.buffer_info()
```
marcoskirsch: According to [documentation](https://docs.python.org/3/library/array.html#array.array.buffer_info) this is to be avoided.

generated/niscope/session.py
marcoskirsch: I'm guessing that a combination of bytearray, memoryview, etc. might do the trick. Need to play with it though.

docs/niscope/session.rst
texasaggie97: This doesn't do anything with the documentation. It will be wrong for now.

generated/niscope/session.py
texasaggie97: Changed to `wfm_array = array.array("d", [0]) * (num_samples * self._actual_num_wfms())`

texasaggie97:
Updated how the array is allocated, which makes arrays almost as fast as numpy arrays in all versions
Added tests using simulated PXIe-5162

generated/niscope/session.py
```diff
@@ -2311,6 +2311,19 @@ def fetch(self, num_samples, timeout=5.0):
...
         return self._fetch(num_samples, timeout)
 
+    def fetch_array(self, num_samples, timeout=5.0):
+        import array
+
+        vi_ctype = visatype.ViSession(self._vi)  # case 1
+        channel_list_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case 2
+        timeout_ctype = visatype.ViReal64(timeout)  # case 9
+        num_samples_ctype = visatype.ViInt32(num_samples)  # case 9
+        wfm_array = array.array("d", [0]) * (num_samples * self._actual_num_wfms())
```
marcoskirsch: minor: instead of 0, it should be 0.0, for correctness.
This may complicate generator enough that it's not worth it once we do for real.

marcoskirsch:
One thing to note is that this shows performance differences, but it would also be interesting to see the memory use differences. I'm going to guess that the underlying implementation of Python lists is not completely naive and optimizes when lots of elements are of the same type. But I don't know specifics or how it varies across Python versions. With array we know memory use should be optimized, especially for smaller types (i.e. ViInt16 used by NI-FGEN).

marcoskirsch:
In the test code, we initiate and fetch right away. We should wait until the data is available before fetching so that we don't block waiting for data, thus affecting the metrics.

epage:
For context @texasaggie97  and I talked about some pros / cons of this
- For read/fetch functions
  - User is restricted on how they can modify the data (can't append or delete elements).  We both felt this wasn't too bad
  - I wish there was a good way to document / enforce that array vs list vs custom type is an implementation detail and we can change it later.  In theory, people could do type checks or use the array-specific functions and we'd break them by changing
- For write functions
  - I'd be concerned about the usability.  If the usability is bad, then what are we gaining by offering it?  They might as well use numpy which interops with their existing numpy code.
  - Would read/fetch use memory views or arrays?  What about write?  Would an inconsistency be a problem?
  - If we just don't offer this for write, would the inconsistency with read be a problem?

Note on inconsistency: this mostly applies to APIs that have both input/output like XNET and DAQmx.  How much consistency do we prefer within nimi-python

tools/scope_perf_test.py
```diff
@@ -0,0 +1,71 @@
...
+            continue
+        except MemoryError:
+            continue
+
+        with session.initiate():
+            while session.acquisition_status() != niscope.AcquisitionStatus.COMPLETE:
+                pass
+            try:
+                list_time_start = time.time()
+                wfm, wfm_info = session[args.channels].fetch(num_samples=rec_len)
```
marcoskirsch: We shouldn't fetch until the data is available, in order to minimize waiting for the device.

marcoskirsch:
> For read/fetch functions
> User is restricted on how they can modify the data (can't append or delete elements). We both felt this wasn't too bad

Agreed. The tradeoff seems clear to me.

> For write functions
> I'd be concerned about the usability. If the usability is bad, then what are we gaining by offering it? ?
> They might as well use numpy which interops with their existing numpy code.

Write functions today accept any iterable, and we have an optimized codepath for numpy.array. We'd add one for array.array. It's transparent to customer. As far as how the customer creates their waveform data as one of these things... not sure.

texasaggie97:
This was just a prototype and is no longer needed

PR #722, opened 2018-01-12
[4]Start adding driver specific information to documentation

texasaggie97:

* Modularizes readme content so it can be used elsewhere
* Adds driver specific installation instruction, status and usage
* Soon to be driver specific readme files for wheel
* Documentation build enabled: http://nimi-python.readthedocs.io/en/docs-session/
* Project README and driver specific README files are pieced together from generated and static rst files.
  * Driver specific is used for wheels that are uploaded to PyPI.

* Unit
* System
* Visual inspection of content


README.rst
```diff
@@ -1,28 +1,42 @@
...
-|                      | NI-FGEN    | 17.0.0                            | |nifgenLatestVersion|      | |nifgenPythonVersion|        |
-|                      +------------+-----------------------------------+----------------------------+------------------------------+
-|                      | NI-SCOPE   | 17.0.2                            | |niscopeLatestVersion|     | |niscopePythonVersion|       |
-|                      +------------+-----------------------------------+----------------------------+------------------------------+
-|                      | NI-SWITCH  | 17.0.0                            | |niswitchLatestVersion|    | |niswitchPythonVersion|      |
-|                      +------------+-----------------------------------+----------------------------+------------------------------+
-|                      | NI-ModInst | 17.0.0                            | |nimodinstLatestVersion|   | |nimodinstPythonVersion|     |
-+----------------------+------------+-----------------------------------+----------------------------+------------------------------+
+Overall Status
```
marcoskirsch: Looking at http://nimi-python.readthedocs.io/en/docs-session/

I would do the following:

* Under [Documentation](http://nimi-python.readthedocs.io/en/docs-session/#documentation), have links to the top level per-driver documentation, rather than expanding the whole tree at the end.
* Make [additional documentation](http://nimi-python.readthedocs.io/en/docs-session/#additional-documentation) a smaller heading under Documentation, rather than a peer.

build/templates/installation.inc.mako
```diff
@@ -0,0 +1,28 @@
...
+    attributes    = helper.filter_codegen_attributes(config['attributes'])
+    functions     = helper.filter_codegen_functions(config['functions'])
+    module_name   = config['module_name']
+    driver_name   = config['driver_name']
+    c_function_prefix = config['c_function_prefix']
+%>\
+
+.. _${module_name}_installation-section:
+
+Installation
```
marcoskirsch: I like having per driver top level readme.
I think the nimi-python project readme shouldn't have installation instructions. Seems overly repetitive.

build/templates/installation.inc.mako
```diff
@@ -0,0 +1,28 @@
...
+------------
+
+As a prerequisite to using the ${module_name} module, you must install the ${driver_name} runtime on your system. Visit `ni.com/downloads <http://www.ni.com/downloads/>`_ to download the driver runtime for your devices.
```
marcoskirsch: Since we are using mako, we could mention the specific driver and even add a link to the installer download (although I can see why we wouldn't want to add that last piece of maintenance burden).

build/templates/status.inc.mako
```diff
@@ -0,0 +1,35 @@
...
+
+
+${helper.get_rst_picture_reference('{0}LatestVersion'.format(module_name), 'http://img.shields.io/pypi/v/{0}.svg'.format(module_name), 'Latest {0} Version'.format(driver_name), 'http://pypi.python.org/pypi/{0}'.format(module_name), indent=0)}
+
+${helper.get_rst_picture_reference('{0}PythonVersion'.format(module_name), 'http://img.shields.io/pypi/pyversions/{0}.svg'.format(module_name), '{0} supported Python versions'.format(driver_name), 'http://pypi.python.org/pypi/{0}'.format(module_name), indent=0)}
+
+${helper.get_rst_picture_reference('{0}OpenIssues'.format(module_name), 'https://img.shields.io/github/issues/ni/nimi-python/{0}.svg'.format(module_name), 'Open Issues + Pull Requests for {0}'.format(driver_name), 'https://github.com/ni/nimi-python/issues?q=is%3Aopen+is%3Aissue+label%3A{0}'.format(module_name), indent=0)}
+
+${helper.get_rst_picture_reference('{0}OpenPRs'.format(module_name), 'https://img.shields.io/github/issues-pr/ni/nimi-python/{0}.svg'.format(module_name), 'Pull Requests for {0}'.format(driver_name), 'https://github.com/ni/nimi-python/pulls?q=is%3Aopen+is%3Aissue+label%3A{0}'.format(module_name), indent=0)}
+
```
marcoskirsch: Is the extension .inc you used standard?
We used .py.mako for snippets that get inlined in the session.py file. If this is similar we should be consistent and use an rst suffix.

build/tools.mak
```diff
@@ -27,7 +27,4 @@ define make_with_tracking_file
 	$(_hide_cmds)$(call log_command,touch $1)
 endef
 
-# Delete the file whose absence will trigger tox to run
```
marcoskirsch: why did you get rid of this?

README.rst
texasaggie97: Updated per our conversation

build/templates/installation.inc.mako
texasaggie97: Add links from project readme to read the docs installation procedure per driver

build/templates/installation.inc.mako
texasaggie97: This does mention the specific driver. For NI-DMM this becomes:
As a prerequisite to using the nidmm module, you must install the NI-DMM runtime on your system.

build/templates/status.inc.mako
texasaggie97: We can't use rst. Sphinx finds and processes all rst files in the directory hierarchy and each one becomes an html file. These files are not meant to be standalone html files, so there were two options given to prevent it. Don't let Sphinx find all rst files or use a different extension. We rely too much on Sphinx finding the rst files so that would have been difficult, tedious and error prone to list all the files explicitly. .inc was the recommended extension that other people were using.

build/tools.mak
texasaggie97: We don't use it any more. This was from when make ran tox to test. We no longer do that. Instead, we use tox to run make and then we test.

This should have been removed when we made those other changes.

PR #723, opened 2018-01-15
Session.__init__() can take dict for the option parameter

### LINKED ISSUE #661: Improve Session constructor API by replacing option_string with something more Pythonic
Proposed by Michael Auchter:

re: InitWithOptions option string
See page 83, http://www.ivifoundation.org/downloads/Architecture%20Specifications/IVI-3.2_Inherent_Capabilities_2017-02-07.pdf
 
There are a limited number of valid option string attributes:
   RangeCheck (bool)
   QueryInstrStatus (bool)
   Cache (bool)
   Simulate (bool)
   RecordCoercions (bool)
   InterchangeCheck (bool)
   DriverSetup (string)
 
and these are expected to be specified as comma-separated "Name=Value" pairs. DriverSetup is special: it needs to be the last thing in the option string, and everything after the equals sign is passed to the driver. In practice, most (all?) MI drivers treat the DriverSetup as a semi-colon separated list of "Name:Value" pairs.
 
That's kind of a mess, and it's something I frequently get wrong.
 
I'd prefer to see these (well, specifically "Simulate" and "DriverSetup"; I doubt the others see much use) handled explicitly.
So. instead of:

    niscope.Session(rsrc, option_string='Simulate=1,DriverSetup=Model:5162 (4CH);Bitfile:CustomProcessing')

Something like:

    niscope.Session(rsrc, simulate=True, driver_setup={'Model': '5162 (4CH)', 'Bitfile': 'CustomProcessing'})
 
and have __init__ take care of building up the correctly formatted option string for the driver's consumption.

texasaggie97:
* Allows the options string to be passed in as a dictionary and the string will be generated from that.

* Fix #661 

* Unit
* System


CHANGELOG.md
```diff
@@ -12,34 +12,62 @@ All notable changes to this project will be documented in this file.
...
+    * #### Changed
+    * #### Removed
 * ### NI-SCOPE
-  * #### Added
-  * #### Changed
-  * #### Removed
+    * #### Added
+    * #### Changed
+        * `niscope.Session()` no longer takes id_query parameter
+        * Option string can now be a python dictionary instead of a string. It will be converted as needed
```
marcoskirsch: Why is the option string change only in two drivers and not under ALL ?

build/templates/_converters.py.mako
```diff
@@ -29,6 +29,52 @@ def convert_timedelta_to_microseconds(value, library_type):
     return _convert_timedelta(value, library_type, 1000000)
 
 
+# This converter is not called from the normal codegen path for function. Instead it is
+# call from init and is a special case. Also, it just returns a string rather than a ctype object
+def init_with_options_converter(value, encoding):
```
marcoskirsch: convert_init_with_options_dictionary()

build/templates/_converters.py.mako
```diff
@@ -29,6 +29,52 @@ def convert_timedelta_to_microseconds(value, library_type):
...
+    if type(value) is str:
+        init_with_options_string = value
+    else:
+        good_keys = {
+            'rangecheck': 'RangeCheck',
```
marcoskirsch: Why allow multiple spellings? We don't typically do that elsewhere.

build/templates/_converters.py.mako
```diff
@@ -29,6 +29,52 @@ def convert_timedelta_to_microseconds(value, library_type):
...
+            'record_coercions': 'RecordCoercions',
+            'interchange_check': 'InterchangeCheck',
+            'driver_setup': 'DriverSetup',
+        }
+        # First we validate that only allowed keys are in the incoming dictionary
+        init_with_options = []
+        for k in sorted(value.keys()):
+            assert k.lower() in good_keys
+            if good_keys[k.lower()] == 'DriverSetup':
+                assert isinstance(value[k], dict)
```
marcoskirsch: I think raise is more appropriate than assert for parameters that came from the user.

generated/nidmm/session.py
```diff
@@ -1943,7 +1943,7 @@ def _init_with_options(self, resource_name, id_query=False, reset_device=False,
                 | VI_FALSE          | 0 | Don't Reset  |
                 +-------------------+---+--------------+
 
-            option_string (str): | Sets the initial value of certain attributes for the session. The
+            option_string (dict or str): | Sets the initial value of certain attributes for the session. The
```
marcoskirsch: python_name should not be option_string anymore...
options?
option_dict?

CHANGELOG.md
texasaggie97: Only two drivers take option strings.

build/templates/_converters.py.mako
texasaggie97: I wanted to support either option strings they already have, or an option string that matches Python naming conventions.

build/templates/_converters.py.mako
texasaggie97: Changed to raise.

build/templates/_converters.py.mako
texasaggie97: Renamed

generated/nidmm/session.py
texasaggie97: Renamed to options

CHANGELOG.md
```diff
@@ -19,6 +19,20 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
         * `nidmm.Session()` no longer takes id_query parameter
+        * Option string can now be a python dictionary instead of a string. It will be converted as needed
+            * Valid options:
+                * RangeCheck (bool)
```
marcoskirsch: We should document only the 'pythonic' spellings for these; even if things will work with the driver spellings.
Alternatively, we could support only the original spellings which allows us to remove knowledge* of valid options from nimi-python, which I think would be a good thing.

* with the exception of DriverSetup / driver_setup

I also don't think it's a good idea to document in such detail in changeling. What we should do is
* Document in the right place
* Link to the issue that got fixed
* In that issue we could document what we did as well

build/templates/_converters.py.mako
marcoskirsch: "option strings they already have" would be given by customers as str and passed right along by the Python bindings, no?

build/templates/_converters.py.mako
```diff
@@ -29,6 +29,54 @@ def convert_timedelta_to_microseconds(value, library_type):
...
+            'recordcoercions': 'RecordCoercions',
+            'interchangecheck': 'InterchangeCheck',
+            'driversetup': 'DriverSetup',
+            'range_check': 'RangeCheck',
+            'query_instr_status': 'QueryInstrStatus',
+            'record_coercions': 'RecordCoercions',
+            'interchange_check': 'InterchangeCheck',
+            'driver_setup': 'DriverSetup',
+        }
+        # First we validate that only allowed keys are in the incoming dictionary
```
marcoskirsch: Is this validation needed? What if you just pass it along and let the driver do the validation?

build/templates/_converters.py.mako
```diff
@@ -29,6 +29,54 @@ def convert_timedelta_to_microseconds(value, library_type):
...
+        init_with_options = []
+        for k in sorted(value.keys()):
+            if k.lower() not in good_keys:
+                raise TypeError('Invalid key: {0}'.format(k))
+            if good_keys[k.lower()] == 'DriverSetup':
+                if not isinstance(value[k], dict):
+                    raise TypeError('DriverSetup must be a dictionary')
```
marcoskirsch: We should prefer the pythonic spellings, I think, so the errors should reflect those.

build/templates/_converters.py.mako
```diff
@@ -29,6 +29,54 @@ def convert_timedelta_to_microseconds(value, library_type):
...
+
+
+# Let's run some tests
+def test_convert_init_with_options_dictionary():
+    assert convert_init_with_options_dictionary('', 'ascii') == ''
+    assert convert_init_with_options_dictionary('Simulate=1', 'ascii') == 'Simulate=1'
+    assert convert_init_with_options_dictionary({'Simulate': True, }, 'ascii') == 'Simulate=1'
+    assert convert_init_with_options_dictionary({'Simulate': False, }, 'ascii') == 'Simulate=0'
+    assert convert_init_with_options_dictionary({'Simulate': True, 'Cache': False}, 'ascii') == 'Cache=0,Simulate=1'
+    assert convert_init_with_options_dictionary({'DriverSetup': {'Model': '5162 (4CH)', 'Bitfile': 'CustomProcessing'}}, 'ascii') == 'DriverSetup=Bitfile:CustomProcessing;Model:5162 (4CH)'
```
marcoskirsch: these tests don't seem to cover the "pythonic" spellings for the keys.

src/nidmm/examples/nidmm_multi_point_measurement.py
```diff
@@ -16,7 +16,7 @@ def example(argsv):
     parser.add_argument('-op', '--option_string', default='', type=str, help='Option string')
     args = parser.parse_args(argsv)
 
-    with nidmm.Session(resource_name=args.resource_name, option_string=args.option_string) as session:
+    with nidmm.Session(resource_name=args.resource_name, options=args.option_string) as session:
```
marcoskirsch: 1. It's a shame our examples don't show the "preferred" way of using this.
2. Also, when running these from the system tests, we won't be exercising the dictionary->string code.

The second one can be fixed with a little bit more example refactoring. Specifically, the example() method should take specific parameters rather than argsv. But then we aren't testing the command line argument parsing... so we would have to invoke both ways? Not sure what the best way to proceed is.

texasaggie97: I have modified the nidcpower examples to use and test both an option string and an option dictionary. Let me know if you think the added complexity and noise in an example is worth it to show both ways of passing in options.

build/templates/_converters.py.mako
marcoskirsch: I think so, the thinner the wrappers the better off we will be.

CHANGELOG.md
```diff
@@ -14,11 +14,17 @@ All notable changes to this project will be documented in this file.
 * ### ALL
     * #### Added
     * #### Changed
+        * Option string can now be a python dictionary instead of a string. It will be converted as needed (Fix #661)
```
marcoskirsch: * Use markdown to link to #661 (we should make that standard practice)
* We should move detailed documentation to the appropriate places (RST / docstring)

CHANGELOG.md
```diff
@@ -14,11 +14,17 @@ All notable changes to this project will be documented in this file.
...
+            * Key/Value pairs approporiate for desired behavior
+                ``` python
+                session = nidmm.Session('Dev1', False, {'Simulate': True, 'DriverSetup': {'Model': '4071', 'BoardType': 'PXI'}})
```
marcoskirsch: Prefer Python convention for dictionary terms. (we settled on that right?)

CHANGELOG.md
```diff
@@ -14,11 +14,17 @@ All notable changes to this project will be documented in this file.
...
+                ```
     * #### Removed
 * ### NI-DMM
     * #### Added
     * #### Changed
         * `nidmm.Session()` no longer takes id_query parameter
+        * Metadata updated to NI-DMM 17.5
```
marcoskirsch: Why does this show up in diff? Is this not updated with latest master?

.gitignore
```diff
@@ -8,5 +8,7 @@ bin/
 commands.log
 .vscode/launch.json
 .vscode/settings.json
+.pytest_cache/
```
marcoskirsch: integration mistake, this is now in here twice.

build/helper/documentation_snippets.py
```diff
@@ -6,6 +6,17 @@
...
 {0}.Session instance, and calling this method on the result.:
 '''
+
+rep_cap_attr_desc = '''
+This property can use repeated capabilities (usually channels). If set or get directly on the
+{0}.Session object, then the set/get will use all repeated capabilities in the session.
+You can specify a subset of repeated capabilities using the Python index notation on an
+{0}.Session instance, and calling set/get value on the result.:
+
+    session['0,1'].{0} = var
```
marcoskirsch: This is not how things are handled anymore in master.

.gitignore
texasaggie97: One copy removed

build/helper/documentation_snippets.py
texasaggie97: Fixed

build/templates/_converters.py.mako
```diff
@@ -54,6 +54,59 @@ def convert_timedelta_to_microseconds(value, library_type):
     return _convert_timedelta(value, library_type, 1000000)
 
 
+# This converter is not called from the normal codegen path for function. Instead it is
```
marcoskirsch: Why not put this as part of the docstring?

texasaggie97: Because I didn't think the information in the comment was appropriate for a docstring.

I can move into one if you think it is.

texasaggie97:
test this please

texasaggie97:
test this please

PR #726, opened 2018-01-18
Use datetime objects

### LINKED ISSUE #659: Use Python timedelta and datetime whenever appropriate

texasaggie97:
* Use `datetime.timedelta()` for all input parameters that represent a relative time:
  * timeout
  * delay
  * holdoff
* Use `datetime.timedelta()` for all attributes that represent a relative time:
* Use `datetime.datetime()` for return parameters that represent an absolute time such as last cal date and time
* Fix #659 

* Unit
* System


texasaggie97:
test this please

CHANGELOG.md
```diff
@@ -25,6 +25,17 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
         * `nidmm.Session()` no longer takes id_query parameter
+        * The following functions timeout or delay parameter now is required to be a `datetime.timedelta()` object:
+            * configure_multi_point
```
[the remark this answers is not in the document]
texasaggie97: Done

generated/nidcpower/session.py
```diff
@@ -4467,7 +4504,7 @@ def wait_for_event(self, event_id, timeout=10.0):
                 | ExportSignal.READY_FOR_PULSE_TRIGGER_EVENT (1052)     | Waits for the Ready for Pulse Trigger event.     |
                 +-------------------------------------------------------+--------------------------------------------------+
 
-            timeout (float): Specifies the maximum time allowed for this function to complete, in
+            timeout (datetime.timedelta or float): Specifies the maximum time allowed for this function to complete, in
```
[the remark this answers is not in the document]
texasaggie97: Fixed

src/nidcpower/examples/nidcpower_source_delay_measure.py
```diff
@@ -22,16 +23,16 @@ def example(resource_name, channels, options, voltage1, voltage2, delay):
         session.current_limit = .06
         session.voltage_level_range = 5.0
         session.current_limit_range = .06
-        session.source_delay = delay
+        session.source_delay = datetime.timedelta(seconds=delay)
```
marcoskirsch: Instead of declaring timeout as a float on line 16, declare it as a timedelta down here. Will clean up the code later.
Alternatively, you could rely on default timeout, which is 10 s. If the user of the example uses a source delay longer than 10 s (why would they do that!!!!) then they get a timeout error. Cleaner code in my opinion.

src/nidcpower/metadata/attributes_addon.py
```diff
@@ -4,6 +4,19 @@
 # We are not code genning attributes that have been marked as obsolete prior to the initial
 # Python API bindings release
 
+attributes_converters = {
+    1150046: { 'attribute_class': 'AttributeViReal64TimeDeltaSeconds',
```
marcoskirsch: This might be a dumb question, but why do you need to specify both the class and the converter? Doesn't the class suffice?

src/nidcpower/metadata/functions_addon.py
```diff
@@ -154,4 +150,96 @@
...
+        'returns': 'ViStatus',
+        'python_name': 'get_ext_cal_last_date_and_time',
+        'real_datetime_call': 'GetExtCalLastDateAndTime',
+        'method_templates': [
+            { 'session_filename': 'datetime_wrappers', 'documentation_filename': 'default_method', 'method_python_name_suffix': '', },
+        ],
+        'parameters': [
+            {
+                'direction': 'in',
+                'enum': None,
```
marcoskirsch: Does our generator still force us to specify enum None? Should be the default...

src/nidcpower/metadata/functions_addon.py
```diff
@@ -154,4 +150,96 @@
             'description': 'Returns the number of channels.',
         },
     },
+    # Public function that wraps driver function but returns datetime object instead of individual items
+    'GetLastExtCalLastDateAndTime': {
+        'codegen_method': 'public',
+        'returns': 'ViStatus',
+        'python_name': 'get_ext_cal_last_date_and_time',
+        'real_datetime_call': 'GetExtCalLastDateAndTime',
```
marcoskirsch: who is this `real_datetime_call` used?

src/nidcpower/metadata/functions_addon.py
```diff
@@ -154,4 +150,96 @@
...
+                'enum': None,
+                'name': 'Month',
+                'type': 'datetime.datetime',
+                'documentation': {
+                    'description': 'Indicates date and time of the last calibration.',
+                },
+            },
+        ],
+        'documentation': {
+            'description': 'Returns the date and time of the last successful calibration. The time returned is 24-hour (military) local time; for example, if the device was calibrated at 2:30 PM, this function returns 14 for **hours** and 30 for **minutes**.',
```
marcoskirsch: Since we are in control of the documentation, you can get rid of everything about military time which doesn't apply.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -154,4 +150,96 @@
...
+                    'description': 'Returns the date and time the device was last calibrated.',
+                },
+            },
+        ],
+'documentation': {
+'description': '''
+Returns the date and time of the oldest successful self-calibration from 
+among the channels in the session.
+
+The time returned is 24-hour (military) local time; for example, if you
```
marcoskirsch: Remove this bit

src/nidmm/metadata/functions_addon.py
```diff
@@ -101,16 +102,16 @@
                                                    2: { 'default_value': False, },
                                                    3: { 'default_value': '""', }, }, },
     'ConfigureMultiPoint':       { 'parameters': { 3: { 'default_value': 'SampleTrigger.IMMEDIATE', },
-                                                   4: { 'default_value': -1, }, }, },
+                                                   4: { 'default_value': 'datetime.timedelta(seconds=-1)', }, }, },
```
marcoskirsch: Does this work as expected, meaning, the DLL receives a -1? I suggest adding a unit test to nifake to make sure.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -60,7 +61,7 @@ def test_fetch_binary8_into(session):
     session.configure_vertical(test_voltage, niscope.VerticalCoupling.AC)
     session.configure_horizontal_timing(50000000, test_record_length, 50.0, 1, True)
     with session.initiate():
-        wfm_infos = session.channels[test_channels].fetch_into(timeout=5.0, wfm=wfm)
+        wfm_infos = session.channels[test_channels].fetch_into(timeout=datetime.timedelta(seconds=5.0), wfm=wfm)
```
marcoskirsch: I feel like we should rely on default timeouts more in the system tests, so that those get exercised.

src/nidcpower/examples/nidcpower_source_delay_measure.py
texasaggie97: The default timeout for `fetch_multiple()` is one second: `fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0))`

Changed to declare `timeout` as `datetime.timdelta`

src/nidcpower/metadata/attributes_addon.py
texasaggie97: The `'attribute_class'` specifies both the data type to convert to as well as the units that that value should be in. `'python_api_converter_type'` says what the datatype in the API is. This is what is used in documentation.

src/nidcpower/metadata/functions_addon.py
texasaggie97: It is used in datetime_wrappers.py,mako to know what the actual function name to call is.

src/nidcpower/metadata/functions_addon.py
texasaggie97: No, removed.

src/nidcpower/metadata/functions_addon.py
texasaggie97: Cleaned up multiple documentation items.

src/nidcpower/metadata/functions_addon.py
texasaggie97: Removed

src/nidmm/metadata/functions_addon.py
texasaggie97: Yes, this works correctly.

Added explicit -1 tests.

src/niscope/system_tests/test_system_niscope.py
texasaggie97: Removed

src/nidcpower/metadata/attributes_addon.py
marcoskirsch: Hmmm... seems redundant to have to specify both since the python type can be inferred from the attribute class. But since it's needed at codegen time, I don't have a better solution in mind. Something to keep in the back of our minds.

How come it's `'python_api_converter_type'` and not `'python_type'`?

src/nidcpower/metadata/functions_addon.py
marcoskirsch: But doesn't this wrapper call the private python method `_get_ext_cal_last_date_and_time()` anyway?

texasaggie97: I wanted one template that would work for all drivers. Of course they are not consistent. Not in names and not in parameter order.

src/nidcpower/metadata/attributes_addon.py
texasaggie97: Needed something different when we could take either a datetime object or the underlying type. Now that we only accept datetime, `'python_type'` works.

PR #727, opened 2018-01-18
[0.5]Use converters for attributes

texasaggie97:
* Add support in the code generator to allow converters for attributes.
  * New attribute class per converter configuration needed. I.e. `AttributeViInt32TimeDeltaMilliseconds`, `AttributeViReal64TimeDeltaSeconds`, etc.
* Needed to finish #726 

* Unit
* System

### DIFF
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -305,7 +305,11 @@ def _add_converter_defaults(a, attributes):
 
 
 def _add_default_attribute_class(a, attributes):
-    '''Set 'attribute_class' if not set'''
+    '''Set 'attribute_class' if not set.
+
+    By default, the 'attribute_class' is only based on the 'type'.
+    It can be set in attributes_addon if we want to convert to/from a different datatype, such as datetime.timedelta
+    '''
     if 'attribute_class' not in attributes[a]:
         attributes[a]['attribute_class'] = 'Attribute' + attributes[a]['type']
 
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -279,6 +279,19 @@ def _add_python_name(a, attributes):
     attributes[a]['python_name'] = n
 
 
+def _add_converter_defaults(a, attributes):
+    '''Add 'python_api_converter_name_to_driver' and 'python_api_converter_name_from_driver' as None if they do not exist'''
+    if 'python_api_converter_name_to_driver' not in attributes[a]:
+        attributes[a]['python_api_converter_name_to_driver'] = None
+    else:
+        attributes[a]['python_api_converter_name_to_driver'] = '_converters.' + attributes[a]['python_api_converter_name_to_driver']
+
+    if 'python_api_converter_name_from_driver' not in attributes[a]:
+        attributes[a]['python_api_converter_name_from_driver'] = None
+    else:
+        attributes[a]['python_api_converter_name_from_driver'] = '_converters.' + attributes[a]['python_api_converter_name_from_driver']
+
+
 def add_all_attribute_metadata(attributes, config):
     '''Merges and Adds all codegen-specific metada to the function metadata list'''
     attributes = merge_helper(attributes, 'attributes', config)
@@ -286,6 +299,7 @@ def add_all_attribute_metadata(attributes, config):
     for a in attributes:
         _add_attr_codegen_method(a, attributes)
         _add_python_name(a, attributes)
+        _add_converter_defaults(a, attributes)
 
     return attributes
 
@@ -704,7 +718,9 @@ def test_add_attributes_metadata_simple():
             'name': 'READ_WRITE_BOOL',
             'python_name': 'read_write_bool',
             'resettable': 'No',
-            'type': 'ViBoolean'
+            'type': 'ViBoolean',
+            'python_api_converter_name_to_driver': None,
+            'python_api_converter_name_from_driver': None,
         },
     }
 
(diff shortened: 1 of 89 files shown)


texasaggie97:
We also have to take into account the units to convert to/from. I've seen seconds and milliseconds so far. So ctype_type and python_type alone isn't enough information.

build/templates/attributes.py.mako
```diff
@@ -12,52 +12,79 @@ ${encoding_tag}
...
+        self._converter_to_driver = converter_to_driver
+        self._converter_from_driver = converter_from_driver
 
 
 class AttributeViInt32(Attribute):
 
     def __get__(self, session, session_type):
-        return session._get_attribute_vi_int32(self._attribute_id)
+        value = session._get_attribute_vi_int32(self._attribute_id)
+        if self._converter_from_driver is not None:
```
marcoskirsch: Instead of duplicating this conditional logic everywhere, you could have the default converter be NullConverter.

build/templates/attributes.py.mako
```diff
@@ -12,52 +12,79 @@ ${encoding_tag}
...
+            value = self._converter_from_driver(value)
```
marcoskirsch: Is this a function? The name (noun) makes it seem as if it is a class.

generated/nidcpower/session.py
```diff
@@ -32,7 +32,7 @@ class _SessionBase(object):
     # This is needed during __init__. Without it, __setattr__ raises an exception
     _is_frozen = False
 
-    active_advanced_sequence = attributes.AttributeViString(1150074)
+    active_advanced_sequence = attributes.AttributeViString(1150074, converter_to_driver=None, converter_from_driver=None)
```
marcoskirsch: If we had specific types for conversions, then all these ugly changes wouldn't be needed.

src/nifake/metadata/attributes_addon.py
```diff
@@ -6,4 +6,12 @@
 attributes_codegen_method = {
 }
 
+attributes_converters = {
+    1000007: { 'python_api_converter_name_to_driver': 'convert_timedelta_to_seconds',
```
marcoskirsch: We could store this as "attribute_class". At add_all time, we use the regular attributes (such as AttributeViInt32), but here you can overload to specify the attribute class you need (such as AttributeTimeDeltaToSeconds)

build/templates/attributes.py.mako
texasaggie97: Switched to additional attribute classes

generated/nidcpower/session.py
texasaggie97: Switched to additional attribute classes

src/nifake/metadata/attributes_addon.py
texasaggie97: Switched to additional attribute classes

build/helper/metadata_add_all.py
```diff
@@ -291,13 +291,34 @@ def _add_python_name(a, attributes):
     attributes[a]['python_name'] = n
 
 
+def _add_converter_defaults(a, attributes):
```
marcoskirsch: Why is this still needed? I would think the attribute class is enough.

build/helper/metadata_add_all.py
```diff
@@ -291,13 +291,34 @@ def _add_python_name(a, attributes):
...
+        attributes[a]['python_api_converter_name_to_driver'] = '_converters.' + attributes[a]['python_api_converter_name_to_driver']
+
+    if 'python_api_converter_name_from_driver' not in attributes[a]:
+        attributes[a]['python_api_converter_name_from_driver'] = None
+    else:
+        attributes[a]['python_api_converter_name_from_driver'] = '_converters.' + attributes[a]['python_api_converter_name_from_driver']
+
+
+def _add_default_attribute_class(a, attributes):
+    '''Set 'attribute_class' if not set'''
```
marcoskirsch: Expand on what 'attribute_class' is, and/or add to the metadata wiki (that thing needs to be looked over)

build/templates/_converters.py.mako
```diff
@@ -10,8 +10,9 @@ import datetime
 def _convert_timedelta(value, library_type, scaling):
     scaled_value = value.total_seconds() * scaling
 
-    if not library_type == visatype.ViReal64 and not library_type == visatype.ViReal32:  # ctype integer types don't convert to int from float so we need to
-        scaled_value = int(scaled_value + 0.5)
+    # ctype integer types don't convert to int from float so we need to
```
marcoskirsch: watch out, the rounding thing got reverted in the diff

build/templates/_converters.py.mako
```diff
@@ -10,8 +10,9 @@ import datetime
...
+    if library_type in [visatype.ViInt64, visatype.ViInt32, visatype.ViUInt32, visatype.ViInt16, visatype.ViUInt16, visatype.ViInt8]:
```
marcoskirsch: I worry that once we add, say, ViUInt8 support to the generator, we won't update this list.
It'd be great if this was single sourced. Could you look at all the types and filter only those that are mapped to an integer type?

marcoskirsch: I'm ok if this becomes a TODO and is solved as a separate PR.

build/helper/metadata_add_all.py
texasaggie97: It's not. Removed

build/helper/metadata_add_all.py
texasaggie97: Expanded docstring

build/templates/_converters.py.mako
texasaggie97: That was on purpose. There are some instances where we do not want to round.

build/templates/_converters.py.mako
texasaggie97: I think it will need to be a separate PR. I don't see an simple and quick solution. Doing the `str(type())` trick doesn't work since all they all return `<type '_ctypes.PyCSimpleType'>`, including float and double.

PR #731, opened 2018-01-23
[0]Use ViString instead of ViChar + is_buffer

texasaggie97:
* During code generation, instead of signalling strings with ViChar + is_buffer, is_string instead
* This is more correct, since we don't treat strings as buffers.

* Unit
* System


build/helper/codegen_helper.py
```diff
@@ -134,7 +134,7 @@ def get_enum_type_check_snippet(parameter, indent):
 def _get_buffer_parameter_for_size_parameter(parameter, parameters):
     '''If parameter represents the size of another parameter in the C API, returns that other parameter. Otherwise None.'''
     for p in parameters:
-        if p['is_buffer'] and p['size']['value'] == parameter['name']:
+        if (p['is_buffer'] or p['type'] == 'ViString') and p['size']['value'] == parameter['name']:
```
marcoskirsch: instead of `p['type'] == 'ViString'` it may be a better fit to have another piece of metadata `'is_string'` which is really closer to the intent of the if statement and clean up the code.

marcoskirsch: I also think that `'is_buffer'` is no longer accurate. It now means `'is_array'`, right? Buffer is more about memory, and array is about the data structure.

If I understand right, then we should rename `is_buffer` (in a separate PR is fine), although `is_buffer` may still be valuable to the code-generator for some cases. `is_buffer` would be equal to `is_array or is_string`.

build/helper/metadata_add_all.py
```diff
@@ -482,6 +485,7 @@ def _do_the_test_add_all_metadata(functions, expected):
     _compare_dicts(actual, expected)
 
 
+'''
```
marcoskirsch: commented out?

generated/nidcpower/session.py
```diff
@@ -1622,14 +1622,14 @@ def fetch_multiple(self, count, timeout=1.0):
...
-        channel_name_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case B520
+        channel_name_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case C010
         timeout_ctype = visatype.ViReal64(timeout)  # case S150
         count_ctype = visatype.ViInt32(count)  # case S190
         voltage_measurements_ctype = (visatype.ViReal64 * count)()  # case B600
         current_measurements_ctype = (visatype.ViReal64 * count)()  # case B600
         in_compliance_ctype = (visatype.ViBoolean * count)()  # case B600
         actual_count_ctype = visatype.ViInt32()  # case S200
-        error_code = self._library.niDCPower_FetchMultiple(vi_ctype, channel_name_ctype, timeout_ctype, count_ctype, voltage_measurements_ctype, current_measurements_ctype, in_compliance_ctype, ctypes.pointer(actual_count_ctype))
+        error_code = self._library.niDCPower_FetchMultiple(vi_ctype, channel_name_ctype, timeout_ctype, count_ctype, voltage_measurements_ctype, current_measurements_ctype, in_compliance_ctype, None if actual_count_ctype is None else (ctypes.pointer(actual_count_ctype)))
```
marcoskirsch: Is this change intentional? It appears like it's harmless because actual_count_ctype is declared right above.

generated/nimodinst/session.py
```diff
@@ -185,13 +185,13 @@ def _get_error_description(self, error_code):
...
-        error_info_ctype = None
+        error_info_buffer_size_ctype = visatype.ViInt32()  # case S170
+        error_info_ctype = None  # case C050
         error_code = self._library.niModInst_GetExtendedErrorInfo(error_info_buffer_size_ctype, error_info_ctype)
         if error_code <= 0:
             return "Failed to retrieve error description."
-        error_info_buffer_size_ctype = visatype.ViInt32(error_code)
-        error_info_ctype = ctypes.create_string_buffer(error_info_buffer_size_ctype.value)
+        error_info_buffer_size_ctype = visatype.ViInt32(error_code)  # case S180
+        error_info_ctype = (visatype.ViChar * error_info_buffer_size_ctype.value)()  # case C060
```
marcoskirsch: comment says this is hand-maintained yet you have the `# case` comments. How come?

generated/niscope/unit_tests/mock_helper.py
```diff
@@ -273,17 +273,11 @@ def niScope_Fetch(self, vi, channel_list, timeout, num_samples, wfm, wfm_info):
         if self._defaults['Fetch']['Wfm'] is None:
             raise MockFunctionCallError("niScope_Fetch", param='Wfm')
         a = self._defaults['Fetch']['Wfm']
-        import sys
```
marcoskirsch: we should just import this at the top of mock_helper.py.

generated/niscope/unit_tests/mock_helper.py
```diff
@@ -273,17 +273,11 @@ def niScope_Fetch(self, vi, channel_list, timeout, num_samples, wfm, wfm_info):
...
-        if sys.version_info.major > 2 and type(a) is str:
-            a = a.encode('ascii')
         for i in range(min(len(wfm), len(a))):
             wfm[i] = a[i]
         if self._defaults['Fetch']['wfmInfo'] is None:
             raise MockFunctionCallError("niScope_Fetch", param='wfmInfo')
         a = self._defaults['Fetch']['wfmInfo']
-        import sys
-        if sys.version_info.major > 2 and type(a) is str:
-            a = a.encode('ascii')
```
marcoskirsch: I wonder if the six library can take care of this for us, where we make a single function call and it does the right thing based on Python version without us needing to add conditional logic. Worth looking into (can go into separate PR).

generated/niswitch/unit_tests/mock_helper.py
```diff
@@ -356,8 +356,8 @@ def niSwitch_error_message(self, vi, error_code, error_message):  # noqa: N802
             return self._defaults['error_message']['return']
         if self._defaults['error_message']['errorMessage'] is None:
             raise MockFunctionCallError("niSwitch_error_message", param='errorMessage')
-        a = self._defaults['error_message']['errorMessage']
         import sys
```
marcoskirsch: should just import at the top

build/helper/codegen_helper.py
texasaggie97: Switched to using `'is_string'`

I thought about that but some are still lists and are not `array.array` or `numpy.array`. Because of that I didn't think that `'is_array'` was any better.

build/helper/metadata_add_all.py
texasaggie97: Not any more

generated/nidcpower/session.py
texasaggie97: Yes. It is needed for the `'ivi-dance'` case where we set it to None the first time.

generated/nimodinst/session.py
texasaggie97: I copied those pieces. I can remove the case XXXX if you want me to.

generated/niscope/unit_tests/mock_helper.py
texasaggie97: Done

generated/niswitch/unit_tests/mock_helper.py
texasaggie97: Done

generated/niscope/unit_tests/mock_helper.py
texasaggie97: I'd rather make it a separate PR

build/helper/codegen_helper.py
marcoskirsch: I see your point. I think that in terms of metadata representation (not C or Python specific) then having something be a string makes sense. Having something be an array makes sense - this is independent of the `array.array` or `numpy.ndarray` optimizations.

The piece of metadata that would inform the generator to generate `array.array` or `numpy.ndarray` optimizations would be something like `is_waveform`.

generated/nimodinst/session.py
marcoskirsch: I don't feel strongly about it, simply was confused by it.

build/helper/codegen_helper.py
texasaggie97: We need to differentiate between lists and arrays.

I may need to make them `'is_list'` and `'is_array'` when I get further along with the array PR.

PR #732, opened 2018-01-24
[1]Use arrays instead of lists where we can

texasaggie97:
* Switch any list that we use in function calls to or from the Python API and use `array.array` instead, when we can
* The exceptions to using `array.array` are
  * custom types
  * strings
  * booleans
  * enums
* Unit
* System


marcoskirsch:
At add_all time, I don't think it's necessary to change original metadata related to strings (hacky!) because we now have `is_string`. That should be enough.

texasaggie97:
No longer overwrite `'type'` for strings. We still do for arrays/lists.

build/helper/codegen_helper.py
```diff
@@ -295,31 +298,73 @@ def _get_ctype_variable_definition_snippet_for_buffers(parameter, parameters, iv
...
-            definition = '({0}.{1} * {2})()  # case B600'.format(module_name, parameter['ctypes_type'], size_parameter['python_name'])
+            if parameter['is_array']:
+                size_declaration = '{0}_size = {1}  # case B600'.format(parameter['python_name'], size_parameter['python_name'])
+                array_declaration = '{0}_array = array.array("{1}", [0] * {0}_size)  # case B600'.format(parameter['python_name'], get_array_type_for_api_type(parameter['ctypes_type']))
+                ctypes_declaration = '{0} = _converters.convert_iterable_to_ctypes({2}_array, ({1}.{3}))  # case B600'.format(parameter['ctypes_variable_name'], module_name, parameter['python_name'], parameter['ctypes_type'])
+                definitions.append(size_declaration)
+                definitions.append(array_declaration)
+                definitions.append(ctypes_declaration)
+            elif parameter['is_list']:
+                definition = '({0}.{1} * {2})()  # case B600'.format(module_name, parameter['ctypes_type'], size_parameter['python_name'])
```
marcoskirsch: This function is pretty weirdly written.

Specifically, you exist the nested if/else structure of death and sometimes `definition` is defined in which case we still need to append something to the left of it and stick it inside `definitions`... or sometimes `definitions` is ready to go.

I think it should be done consistently, which is to say: whenever you exit the nested if/else structure of death definitions is already populated and ready to go.

build/helper/helper.py
```diff
@@ -97,3 +98,27 @@ def get_numpy_type_for_api_type(api_type, config):
         # We didn't find it so assert
         assert False, 'Unknown value for api_type: {0}'.format(api_type)
 
+
+def get_array_type_for_api_type(api_type):
+    '''Returns the array type to use in the Python API from the original visa or custom type used in the C API
+
+    Do not use this with enums.
```
marcoskirsch: can you assert this?

build/helper/metadata_add_all.py
```diff
@@ -109,36 +109,57 @@ def _add_is_error_handling(f):
...
+
+    assert 'is_buffer' not in parameter or not parameter['is_buffer'], "if 'is_buffer' is in metadata then it must be set to False"
+    assert 'is_string' not in parameter, "'is_string' should not be set by metadata or in addons"
+    assert 'is_array' not in parameter, "'is_array' should not be set by metadata or in addons"
+    assert 'is_list' not in parameter, "'is_list' should not be set by metadata or in addons"
+
+    is_string = False
+    is_array = False
+    is_list = False
+    ot = parameter['type']
```
marcoskirsch: what is ot?

build/helper/metadata_add_all.py
```diff
@@ -109,36 +109,57 @@ def _add_is_error_handling(f):
...
-
-    # For simplicity, we are going to treat ViChar[], ViString, ViConstString, and ViRsrc the same: As ViChar
-    # and is_buffer True
-    t = parameter['type']
-    if (t.find('[ ]') > 0) or (t.find('[]') > 0):
-        assert 'is_buffer' not in parameter or parameter['is_buffer'] is True, 'Conflicting metadata - [] found but is_buffer already set to False.'
-        parameter['type'] = t.replace('[ ]', '').replace('[]', '')
-        parameter['original_type'] = t
-        parameter['is_buffer'] = True
+def _add_buffer_info(parameter, config):
```
marcoskirsch: I think is_string should be set in a separate `_add_is_string` function.

build/templates/_converters.py.mako
```diff
@@ -29,6 +31,18 @@ def convert_timedelta_to_microseconds(value, library_type):
     return _convert_timedelta(value, library_type, 1000000)
 
 
+def convert_iterable_to_ctypes(value, library_type=None):
```
marcoskirsch: This is misleading.

Conversion implies that something comes in, and something else comes out. But this function returns a ctypes pointer that references the original data.

Another question: Why not use this for lists too?

build/templates/_converters.py.mako
```diff
@@ -29,6 +31,18 @@ def convert_timedelta_to_microseconds(value, library_type):
...
+    ctypes_type = None
+    if isinstance(value, array.array):
+        addr, _ = value.buffer_info()
+        ctypes_type = ctypes.cast(addr, ctypes.POINTER(library_type))
+    elif str(type(value)).find("'numpy.ndarray'") != -1:
+        import numpy
+        ctypes_type = numpy.ctypeslib.as_ctypes(value)
+
+    return ctypes_type
```
marcoskirsch: This doesn't return a type, it returns the pointer itself.
Consider not even declaring this variable and returning from the if/else structure.

build/helper/metadata_add_all.py
marcoskirsch: I see... original type. Might as well spell it out.

build/helper/metadata_add_all.py
```diff
@@ -109,36 +109,57 @@ def _add_is_error_handling(f):
...
-    if t == 'ViConstString' or t == 'ViRsrc' or t == 'ViString' or (parameter['type'] == 'ViChar' and parameter['is_buffer']):
-        parameter['type'] = 'ViString'
-        parameter['original_type'] = t
-        parameter['is_buffer'] = False
-        parameter['is_string'] = True
+    string_types = ['ViConstString', 'ViRsrc', 'ViString', 'ViChar[]', ]
+    if ot in string_types:
+        is_string = True
+    elif ot.find('[]') > 0:
+        parameter['type'] = ot.replace('[]', '')
```
marcoskirsch: We should try to remove this "modify original metadata" hack in a separate PR.

generated/nidcpower/session.py
```diff
@@ -1625,13 +1627,17 @@ def fetch_multiple(self, count, timeout=1.0):
         channel_name_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case C010
         timeout_ctype = visatype.ViReal64(timeout)  # case S150
         count_ctype = visatype.ViInt32(count)  # case S190
-        voltage_measurements_ctype = (visatype.ViReal64 * count)()  # case B600
-        current_measurements_ctype = (visatype.ViReal64 * count)()  # case B600
+        voltage_measurements_size = count  # case B600
```
marcoskirsch: I did a little bit of experimentation that made me question our use of array.array.

Specifically: The main thing we get by using array.array is that we no longer copy elements into a list. How would that be different than returning the object we get from `(visatype.ViReal64 * count)()`, which would be of type `<class '__main__.c_double_Array_10'>` rather than `<type 'array.array'>`?

Very little.

* In both cases the data is zeroed out so that's a shame and the advantage of numpy.
* In both cases the object the user gets back can be indexed.
* In both cases the object's size cannot be changed.
* array.array.__str__() is more useful.

Is that it?
Does it warrant the increased complexity in the generator?

generated/nidcpower/unit_tests/mock_helper.py
```diff
@@ -231,21 +231,37 @@ def niDCPower_ExportSignal(self, vi, signal, signal_identifier, output_terminal)
...
         if self._defaults['FetchMultiple']['voltageMeasurements'] is None:
             raise MockFunctionCallError("niDCPower_FetchMultiple", param='voltageMeasurements')
         a = self._defaults['FetchMultiple']['voltageMeasurements']
-        for i in range(min(len(voltage_measurements), len(a))):
-            voltage_measurements[i] = a[i]
+        try:
+            voltage_measurements_ref = voltage_measurements.contents
+        except AttributeError:
+            voltage_measurements_ref = voltage_measurements
+        for i in range(min(len(voltage_measurements_ref), len(a))):
```
marcoskirsch: Why pick the minimum of the two? Wouldn't you want an assertion if the sizes are different?

generated/nidcpower/unit_tests/mock_helper.py
```diff
@@ -231,21 +231,37 @@ def niDCPower_ExportSignal(self, vi, signal, signal_identifier, output_terminal)
     def niDCPower_FetchMultiple(self, vi, channel_name, timeout, count, voltage_measurements, current_measurements, in_compliance, actual_count):  # noqa: N802
         if self._defaults['FetchMultiple']['return'] != 0:
             return self._defaults['FetchMultiple']['return']
+        # voltage_measurements
         if self._defaults['FetchMultiple']['voltageMeasurements'] is None:
             raise MockFunctionCallError("niDCPower_FetchMultiple", param='voltageMeasurements')
         a = self._defaults['FetchMultiple']['voltageMeasurements']
```
marcoskirsch: Looking at this code made me wish variable `a` had a more descriptive name.

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -227,10 +228,10 @@ def test_create_advanced_arb_sequence():
         loop_counts_array = [10, 20, 30]
         session.output_mode = nifgen.OutputMode.SEQ
         # Test relies on value of sequence handles starting at a known value and incrementing sequentially. Hardly ideal.
-        assert ([], seq_handle_base + 0) == session.create_advanced_arb_sequence(waveform_handles_array, loop_counts_array=loop_counts_array)
-        assert ([], seq_handle_base + 1) == session.create_advanced_arb_sequence(waveform_handles_array, loop_counts_array=loop_counts_array, sample_counts_array=sample_counts_array)
-        assert (marker_location_array, seq_handle_base + 2) == session.create_advanced_arb_sequence(waveform_handles_array, loop_counts_array=loop_counts_array, marker_location_array=marker_location_array)
-        assert (marker_location_array, seq_handle_base + 3) == session.create_advanced_arb_sequence(waveform_handles_array, loop_counts_array=loop_counts_array, sample_counts_array=sample_counts_array, marker_location_array=marker_location_array)
+        assert (array.array('l'), seq_handle_base + 0) == session.create_advanced_arb_sequence(waveform_handles_array, loop_counts_array=loop_counts_array)
```
marcoskirsch: I must say this is a little strange. What does documentation say this returns?
http://nimi-python.readthedocs.io/en/master/nifgen/functions.html#nifgen.Session.create_advanced_arb_sequence

```
coerced_markers_array (list of int):

Returns an array of all given markers that are coerced (rounded) to the nearest marker quantum. Not all devices coerce markers.

Default Value: None
```

marcoskirsch:
Missing from this PR is input waveforms: we have optimizations for numpy.ndarray but you didn't add for array. Is that intentional, targeted for separate PR, or something you overlooked?

generated/nidcpower/unit_tests/mock_helper.py
texasaggie97: Renamed to `test_value`

generated/nidcpower/unit_tests/mock_helper.py
texasaggie97: They will be different.
Changed to `assert len(voltage_measurements_ref) >= len(test_value)`

src/nifgen/system_tests/test_system_nifgen.py
texasaggie97: Fixed documentation. Now says `coerced_markers_array (array.array("l")):`

generated/nidcpower/session.py
texasaggie97: ctype arrays don't support memoryview like an `array.array` does. You can create the memoryview but then you can't slice it.

build/templates/_converters.py.mako
texasaggie97: Not all code paths return.

Removed ctypes_type and explicitly return None or appropriate value

build/helper/metadata_add_all.py
texasaggie97: I can do that but then in `_add_buffer_info()` I have to filter out `ViChar[]` since string doesn't change `['type']`. The logic here handles that.

build/helper/helper.py
texasaggie97: No, not more than it already does. I don't have the information to know that this is an enum. It will end up asserting on the final one saying 'Unknown type'

build/helper/codegen_helper.py
texasaggie97: Every case exits with `definition` set. Some cases will have added more lines.

build/helper/metadata_add_all.py
texasaggie97: Done

generated/nifake/custom_struct.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, data=None, struct_int=0, struct_double=0.0):
             self.struct_double = struct_double
 
     def __repr__(self):
-        return '{0}(data=None, struct_int={1}, struct_double={2}'.format(self.__class__.__name__, self.struct_int, self.struct_double)
+        return '{0}(data=None, struct_int={1}, struct_double={2})'.format(self.__class__.__name__, self.struct_int, self.struct_double)
```
marcoskirsch: I'm guessing this was just a bug in unrelated, unexercised code?

generated/nifake/unit_tests/test_session.py
```diff
@@ -387,6 +415,47 @@ def test_multiple_array_types(self):
                 matchers.ViInt16BufferMatcher(input_array_of_integers)
             )
 
+    '''
```
marcoskirsch: why is this commented out?

generated/nifgen/session.py
```diff
@@ -2696,15 +2713,22 @@ def _write_named_waveform_f64(self, waveform_name, data):
...
+            data (array.array("d")): Specifies the array of data to load into the waveform. The array must
                 have at least as many elements as the value in **size**.
 
         '''
         vi_ctype = visatype.ViSession(self._vi)  # case S110
         channel_name_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case C010
         waveform_name_ctype = ctypes.create_string_buffer(waveform_name.encode(self._encoding))  # case C020
         size_ctype = visatype.ViInt32(0 if data is None else len(data))  # case S160
-        data_ctype = None if data is None else (visatype.ViReal64 * len(data))(*data)  # case B550
+        if data is not None:  # case B550
```
marcoskirsch: this whole if/else structure could be a helper function

generated/niscope/session.py
```diff
@@ -2509,12 +2519,18 @@ def fetch_measurement_stats(self, scalar_meas_function, timeout=5.0):
         channel_list_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case C010
         timeout_ctype = visatype.ViReal64(timeout)  # case S150
         scalar_meas_function_ctype = visatype.ViInt32(scalar_meas_function.value)  # case S130
-        result_ctype = (visatype.ViReal64 * self._actual_num_wfms())()  # case B560
-        mean_ctype = (visatype.ViReal64 * self._actual_num_wfms())()  # case B560
-        stdev_ctype = (visatype.ViReal64 * self._actual_num_wfms())()  # case B560
-        min_ctype = (visatype.ViReal64 * self._actual_num_wfms())()  # case B560
-        max_ctype = (visatype.ViReal64 * self._actual_num_wfms())()  # case B560
-        num_in_stats_ctype = (visatype.ViInt32 * self._actual_num_wfms())()  # case B560
+        result_size = self._actual_num_wfms()  # case B560
```
marcoskirsch: I guess we still call _actual_num_wfms() a bunch of times :(

src/nifgen/templates/session.py/create_waveform.py.mako
```diff
@@ -17,6 +17,13 @@
                 return self._create_waveform_i16_numpy(waveform_data_array)
             else:
                 raise TypeError("Unsupported dtype. Is {0}, expected {1} or {2}".format(waveform_data_array.dtype, numpy.float64, numpy.int16))
+        elif str(type(waveform_data_array)).find("'array.array'") != -1:
```
marcoskirsch: use isinstance()

src/nifgen/templates/session.py/write_waveform.py.mako
```diff
@@ -18,6 +18,13 @@
                 return self._write_named_waveform_i16_numpy(waveform_name_or_handle, data) if use_named else self._write_binary16_waveform_numpy(waveform_name_or_handle, data)
             else:
                 raise TypeError("Unsupported dtype. Is {0}, expected {1} or {2}".format(data.dtype, numpy.float64, numpy.int16))
+        elif str(type(data)).find("'array.array'") != -1:
```
marcoskirsch: isinstance

build/helper/helper.py
```diff
@@ -97,3 +98,27 @@ def get_numpy_type_for_api_type(api_type, config):
...
+    '''
+    type_map = {
```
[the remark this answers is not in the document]
texasaggie97: Consolidated into a single map. All functions now use that map

build/helper/metadata_add_all.py
```diff
@@ -109,36 +109,49 @@ def _add_is_error_handling(f):
...
-    t = parameter['type']
-    if (t.find('[ ]') > 0) or (t.find('[]') > 0):
-        assert 'is_buffer' not in parameter or parameter['is_buffer'] is True, 'Conflicting metadata - [] found but is_buffer already set to False.'
-        parameter['type'] = t.replace('[ ]', '').replace('[]', '')
-        parameter['original_type'] = t
-        parameter['is_buffer'] = True
+def _add_buffer_info(parameter, config):
+    '''Adds buffer information to the parameter metadata
+
+    There are 4 different 'is_*' variables we will set
```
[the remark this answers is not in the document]
texasaggie97: Updated docstring

build/helper/metadata_add_all.py
```diff
@@ -109,36 +109,49 @@ def _add_is_error_handling(f):
...
+        'is_string' - Used for any string type - see below for complete list
+        'use_list'  - Used for an array of custom types. We are not putting custom types into array.array or numpy.array
+        'use_array' - Used for arrays of simple types
+        'is_buffer' - True when either 'use_list' or 'use_array' is True
+    '''
```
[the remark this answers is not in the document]
texasaggie97: Updated docstring

build/helper/metadata_add_all.py
```diff
@@ -109,36 +109,49 @@ def _add_is_error_handling(f):
...
+
+    assert 'is_buffer' not in parameter or not parameter['is_buffer'], "if 'is_buffer' is in metadata then it must be set to False"
+    assert 'is_string' not in parameter, "'is_string' should not be set by metadata or in addons"
+    assert 'use_list' not in parameter, "'use_list' should not be set by metadata or in addons"
+
+    is_string = False
+    use_array = False
+    use_list = False
+    oiginal_type = parameter['type']
```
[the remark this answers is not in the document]

build/helper/metadata_add_all.py
```diff
@@ -109,36 +109,49 @@ def _add_is_error_handling(f):
...
-    if t == 'ViConstString' or t == 'ViRsrc' or t == 'ViString' or (parameter['type'] == 'ViChar' and parameter['is_buffer']):
-        parameter['type'] = 'ViString'
-        parameter['original_type'] = t
-        parameter['is_buffer'] = False
-        parameter['is_string'] = True
+    string_types = ['ViConstString', 'ViRsrc', 'ViString', 'ViChar[]', ]
+    if oiginal_type in string_types:
+        is_string = True
+    elif oiginal_type.find('[]') > 0:
+        parameter['type'] = oiginal_type.replace('[]', '')
```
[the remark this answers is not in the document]
texasaggie97: Added `# TODO(marcoskirsch) Don't change metadata, add new key`

build/templates/_converters.py.mako
```diff
@@ -29,6 +31,24 @@ def convert_timedelta_to_microseconds(value, library_type):
     return _convert_timedelta(value, library_type, 1000000)
 
 
+def convert_buffer_to_ctypes(value=None, library_type=None, size=None):
```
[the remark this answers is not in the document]
texasaggie97: Renamed.

Should it be moved out of `_converters`?

generated/nifake/custom_struct.py
texasaggie97: Yes, though it is cosmetic rather and doesn't cause an error.

generated/nifake/unit_tests/test_session.py
texasaggie97: Removed. It was left over from when I was fixing the test

generated/nifgen/session.py
texasaggie97: It could but since this is generated code, the actual source is not duplicated, so what would it gain us? A few bytes in size for session.py?

generated/niscope/session.py
texasaggie97: Yes, those are all separate variables and there is no lookahead or lookbehind when coming up with those declarations.

src/nifgen/templates/session.py/create_waveform.py.mako
texasaggie97: Done

src/nifgen/templates/session.py/write_waveform.py.mako
texasaggie97: Done

generated/nifgen/session.py
marcoskirsch: Improved readability in both the generated code and the generator.

build/templates/_converters.py.mako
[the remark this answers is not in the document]
marcoskirsch: I think so.

generated/nifgen/session.py
texasaggie97: Moved into a helper function in _converters.py

Still do this in two lines, one for the array, one for the ctypes. I didn't make a single function that returned a tuple because then this case would not be able to use the common "var_ctype = definition" that you had commented about earlier.

build/templates/_converters.py.mako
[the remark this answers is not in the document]
texasaggie97: To where? session.py?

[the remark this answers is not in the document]
marcoskirsch: Yeah, I think free-standing function in session.py is the best we have right now. If we end up with a handful of similar functions we can move to a new file in the future.

marcoskirsch:
FYI: 

> This branch has conflicts that must be resolved

build/templates/_converters.py.mako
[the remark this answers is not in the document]
texasaggie97: Moved to `session.py`

Issue #695, opened 2017-12-15
Dead code in Library and mock_helper for pseudo functions


marcoskirsch:
We add a pseudo function to API metadata in order to generate methods for "Dispatcher". For example, src/nifgen/functions_add_on.py adds WriteWaveformDispatcher. This doesn't correspond to any functions in the DLL, rather it's used in order to inject our own templates that take care of dispatching the call to the real function.

However, code for calling the DLL and for mocking that call is still generated. This is non-functional confusing dead code. We should skip generating them in Library and mock_helper.

PR #742, opened 2018-02-15
Create "Fancy Fetch" for niscope

texasaggie97:
* Change `fetch()` to be more user friendly
  * `fetch(timeout=5.0, num_samples=None, fetch_relative_to=None, fetch_offet=None, fetch_record_number=None, fetch_num_records=None, wfm=None)`
  * `fetch_relative_to`, `fetch_offet`, `fetch_record_number`, & `fetch_num_records`: If not `None` will set the corresponding attribute
  * If wfm is not `None` calls the appropriate `fetch_into()` function
  * In Python 3, a memoryview of each record is created and added to the corresponding `wfm_info`
  * In Python 2, an array.array is created and the points are copied.  `memoryview` in Python 2 only supports string and bytearray, not array.array or numpy.ndarray of arbitrary types. :(
* Additional documentation requires #722 
* Instead of returning all of the captured data in a single 1D array that the user has to then correlate different parts to each record, the waveform for a given record is put into the corresponding wfm_info
  before:
  ``` python
  # Assume num records = 2
  wfm, wfm_info = session.channel[[0,1]].fetch(num_samples=1000)
  # Channel 0, record 0
  # wfm_info[0].<absoluteX, etc.> and wfm[0:999]
  # Channel 1, record 0
  # wfm_info[1].<absoluteX, etc.> and wfm[1000:1999]
  # Channel 0, record 1
  # wfm_info[2].<absoluteX, etc.> and wfm[2000:2999]
  # Channel 1, record 1
  # wfm_info[3].<absoluteX, etc.> and wfm[3000:3999]
  ```
  after:
  ``` python
  wfm_info = session.channel[[0,1]].fetch(num_samples=1000, fetch_num_records=2)
  # Channel 0, record 0
  # wfm_info[0].<absoluteX, etc.> and wfm_info[0].wfm
  # Channel 1, record 0
  # wfm_info[1].<absoluteX, etc.> and wfm_info[1].wfm
  # Channel 0, record 1
  # wfm_info[2].<absoluteX, etc.> and wfm_info[1].wfm
  # Channel 1, record 1
  # wfm_info[3].<absoluteX, etc.> and wfm_info[1].wfm
  # Basically, the waveform becomes just one more attribute on the waveform info object
  ```

* Unit
* System


src/niscope/system_tests/test_system_niscope.py
```diff
@@ -66,6 +67,23 @@ def test_fetch_binary8_into(session):
...
+    test_record_length = 2000
+    test_channels = range(2)
+    test_num_channels = 2
+    wfm = numpy.ndarray(test_num_channels * test_record_length, dtype=numpy.int8)
+    # Initialize with NaN so we can later verify all samples were overwritten by the driver.
+    wfm.fill(float('nan'))
+    session.configure_vertical(test_voltage, niscope.VerticalCoupling.AC)
+    session.configure_horizontal_timing(50000000, test_record_length, 50.0, 1, True)
+    with session.initiate():
+        wfm_infos = session.channels[test_channels].fetch(timeout=5.0, wfm=wfm)
```
epage: > If wfm is not None calls the appropriate fetch_into() function

I thought we had decided against doing that kind of stuff ie having functions that both mutate their parameter and return it?

texasaggie97: It doesn't directly return the wfm. It puts slices/memoryviews of it into the wfm_info object.

texasaggie97:
I updated the PR description with before and after.
The changelog lists the change. Until #722 is merged, I can't add driver specific, hand maintained documentation.

epage:
> The changelog lists the change. Until #722 is merged, I can't add driver specific, hand maintained documentation.

What do you mean by "the changelog lists the change" and why is driver-specific, hand maintained documentation relevant?

epage:
> In Python 3, a memoryview of each record is created and added to the corresponding wfm_info
> In Python 2, an array.array is created and the points are copied. :(

I assume we talked about this but can't remember the results.  Why is [`memoryview`](https://docs.python.org/2/library/stdtypes.html#memoryview) not used in Python2?

texasaggie97:
> I assume we talked about this but can't remember the results. Why is memoryview not used in Python2?

`memoryview` in Python 2 only supports string and bytearray, not array.array or numpy.ndarray of arbitrary types.

I updated the PR description and added a comment to the source about why.

docs/niscope/functions.rst
```diff
@@ -1212,9 +1212,18 @@ niscope.Session methods
...
         .. code:: python
 
-            session.channels['0,1'].fetch(num_samples, timeout='datetime.timedelta(seconds=5.0)')
+            session.channels['0,1'].fetch(timeout='datetime.timedelta(seconds=5.0)', num_samples=None, fetch_relative_to=None, fetch_offet=None, fetch_record_number=None, fetch_num_records=None, wfm=None)
 
 
+    :param timeout:
+
+
+        The time to wait in seconds for data to be acquired; using 0 for this parameter tells NI-SCOPE to fetch whatever is currently available. Using -1 for this parameter implies infinite timeout.
```
marcoskirsch: if you have control over this text, remove "in seconds" (in general make sure it works for timedelta)

docs/niscope/functions.rst
```diff
@@ -1224,29 +1233,54 @@ niscope.Session methods
...
+        
 
 
-            Returns an array whose length is the **numSamples** times number of waveforms. Call :py:meth:`niscope.Session._actual_num_wfms` to determine the number of waveforms.
+    :type fetch_offet: int
+    :param fetch_record_number:
 
-            
 
+        Zero-based index of the first record to fetch.  Use fetch__num_records to set the number of records to fetch. If not set, use value of NISCOPE_ATTR_RECORD_NUMBER
```
marcoskirsch: two underscores (type)
If you call out the attribute, use its python name. Applies elsewhere.
"If not set, use value of record_number property"

generated/niscope/session.py
```diff
@@ -1839,6 +1839,127 @@ def configure_vertical(self, range, coupling, offset=0.0, probe_attenuation=1.0,
...
+
+            fetch_record_number (int): Zero-based index of the first record to fetch.  Use fetch__num_records to set the number of records to fetch. If not set, use value of NISCOPE_ATTR_RECORD_NUMBER
+
+            fetch_num_records (int): Number of records to fetch. Use -1 to fetch all configured records. If not set, use value of NISCOPE_ATTR_NUM_RECORDS
+
+            wfm (list of float): Returns an array whose length is the **numSamples** times number of waveforms. Call _actual_num_wfms to determine the number of waveforms.
+
+
+        Returns:
+            wfm_info (list of WaveformInfo): Returns an array of classed with the following timing and scaling information about each waveform:
```
marcoskirsch: In the documentation for this thing there is no `wfm` member.

generated/niscope/session.py
```diff
@@ -1839,6 +1839,127 @@ def configure_vertical(self, range, coupling, offset=0.0, probe_attenuation=1.0,
...
+        num_wfms = self._actual_num_wfms()
+
+        if num_samples is None:
+            num_samples_to_use = int(len(wfm) / num_wfms)
+        else:
+            num_samples_to_use = num_samples
+
+        total_wfm_size = num_samples_to_use * num_wfms
+
+        # If the waveform is not passed in, we will use the builtin python array with element type 'double'
```
marcoskirsch: I don't think wfm should *ever* be passed in to fancy fetch.

generated/niscope/session.py
```diff
@@ -1839,6 +1839,127 @@ def configure_vertical(self, range, coupling, offset=0.0, probe_attenuation=1.0,
...
+        else:
+            if len(wfm) < total_wfm_size:
+                raise TypeError('The size of the input array is too small for the acquisition. Needs to be {0}, was {1}'.format(total_wfm_size, len(wfm_to_use)))
+
+            wfm_to_use = wfm
+            wfm_info = self.fetch_into(wfm_to_use, timeout)
+
+        # memoryview in Python 2 doesn't support numeric types, so we copy into an array.array to put in the wfm. :( You should be using Python 3!
+        # Or use the _into version. memoryview in Python 2 only supports string and bytearray, not array.array or numpy.ndarray of arbitrary types.
+        if sys.version_info.major < 3:
```
marcoskirsch: can six help in any way for this?

generated/niscope/session.py
```diff
@@ -1839,6 +1839,127 @@ def configure_vertical(self, range, coupling, offset=0.0, probe_attenuation=1.0,
...
+                        numpy.dtype('int16'): 'h',
+                        numpy.dtype('int32'): 'l',
+                        numpy.dtype('float64'): 'd',
+                    }[wfm_to_use.dtype]
+
+                start = i * num_samples_to_use
+                end = start + num_samples_to_use
+                wfm_info[i].wfm = array.array(typecode, wfm_to_use[start:end])
+        else:
+            # In Python 3 we can use memoryview objects to give us pieces of the underlying array. This is much faster
```
marcoskirsch: "Python 3 and newer"

generated/niscope/session.py
```diff
@@ -1839,6 +1839,127 @@ def configure_vertical(self, range, coupling, offset=0.0, probe_attenuation=1.0,
...
+        if wfm is None:
+            wfm_to_use, wfm_info = self._fetch(num_samples, timeout)
+        else:
+            if len(wfm) < total_wfm_size:
+                raise TypeError('The size of the input array is too small for the acquisition. Needs to be {0}, was {1}'.format(total_wfm_size, len(wfm_to_use)))
+
+            wfm_to_use = wfm
+            wfm_info = self.fetch_into(wfm_to_use, timeout)
+
+        # memoryview in Python 2 doesn't support numeric types, so we copy into an array.array to put in the wfm. :( You should be using Python 3!
```
marcoskirsch: move the comment inside the if statement since it only applies to the `sys.version_info.major < 3` case.

generated/niscope/session.py
```diff
@@ -1839,6 +1839,127 @@ def configure_vertical(self, range, coupling, offset=0.0, probe_attenuation=1.0,
...
+        import sys
+
+        if num_samples is None and wfm is None:
+            raise TypeError("Either 'num_samples' or 'wfm' must be set")
+
+        # Check optional parameters
+        if fetch_relative_to is not None:
+            self.fetch_relative_to = fetch_relative_to
+        if fetch_offet is not None:
+            self.fetch_offet = fetch_offet
```
marcoskirsch: This is a hidden side effect. It won't be obvious that passing the parameter will change the state of the session.

Ideally, I would get rid of the (public) attributes so that there's only one way to do this. But that cannot be done because of _into flavor. Or can it? We'd have to add them as parameters as well. Then maybe they don't need to all be optional parameters.

Alternatively, the correct solution is for the parameters to not have a visible effect on the Session state. How? I think we'd have to use some sort of property context manager that on __exit__ restores the old value.

What do you think?

marcoskirsch: Also, I don't see where this gets used later on. Do these not have an effect on how you interpret the waveform data?

generated/niscope/unit_tests/mock_helper.py
```diff
@@ -59,6 +59,9 @@ def __init__(self):
         self._defaults['Disable']['return'] = 0
         self._defaults['ExportSignal'] = {}
         self._defaults['ExportSignal']['return'] = 0
+        self._defaults['FancyFetch'] = {}
```
marcoskirsch: #695 keeps nagging at me.

src/niscope/examples/niscope_fetch.py
```diff
@@ -13,9 +13,9 @@ def example(resource_name, channels, options, length, voltage):
         session.configure_vertical(range=voltage, coupling=niscope.VerticalCoupling.AC)
         session.configure_horizontal_timing(min_sample_rate=50000000, min_num_pts=length, ref_position=50.0, num_records=1, enforce_realtime=True)
         with session.initiate():
-            wfm, wfm_infos = session.channels[channels].fetch(num_samples=length)
-        print('Number of samples acquired: {:,}\n'.format(len(wfm)))
+            wfm_infos = session.channels[channels].fetch(num_samples=length)
```
marcoskirsch: How about `wfm_infos` -> `waveforms` ?

src/niscope/metadata/functions_addon.py
```diff
@@ -270,14 +272,52 @@
...
             },
             {
-                'direction': 'out',
+                'direction': 'in',
+                'enum': 'FetchRelativeTo',
+                'default_value': None,
+                'name': 'fetchRelativeTo',
+                'type': 'ViInt32',
+                'documentation': {
+                    'description': 'Position to start fetching within one record. If not set, use value of NISCOPE_ATTR_FETCH_RELATIVE_TO',
```
marcoskirsch: NISCOPE_ATTR_FETCH_RELATIVE_TO didn't get converted to python_name by the generator.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -66,6 +67,23 @@ def test_fetch_binary8_into(session):
...
+    test_record_length = 2000
+    test_channels = range(2)
+    test_num_channels = 2
+    wfm = numpy.ndarray(test_num_channels * test_record_length, dtype=numpy.int8)
+    # Initialize with NaN so we can later verify all samples were overwritten by the driver.
+    wfm.fill(float('nan'))
+    session.configure_vertical(test_voltage, niscope.VerticalCoupling.AC)
+    session.configure_horizontal_timing(50000000, test_record_length, 50.0, 1, True)
+    with session.initiate():
+        wfm_infos = session.channels[test_channels].fetch(wfm=wfm)
```
marcoskirsch: Interesting.
You're using type of wfm to determine which binary flavor to call into? I was under the impression fancy fetch was ease of use so it should only support float. In the name of KISS I would remove this flexibility. If you think we should keep it, I'd like to use a different way of selecting which underlying thing to call because it's odd to modify the input param _and_ also return it (or memory views to it?).

Client wants fast? Call the appropriate fetch_into. Or use C, really.

src/niscope/templates/session.py/fancy_fetch.py.mako
```diff
@@ -0,0 +1,76 @@
...
+%>\
+    def ${f['python_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.SESSION_METHOD_DECLARATION)}):
+        '''${f['python_name']}
+
+        ${helper.get_function_docstring(f, False, config, indent=8)}
+        '''
+        import sys
+
+        if num_samples is None and wfm is None:
+            raise TypeError("Either 'num_samples' or 'wfm' must be set")
```
marcoskirsch: I vote: remove wfm parameter, make num_samples not optional.

src/niscope/templates/session.py/fetch_waveform.py.mako
```diff
@@ -15,13 +15,13 @@
         num_samples = int(len(wfm) / self._actual_num_wfms())
 
         if wfm.dtype == numpy.float64:
-            return self._fetch_into(num_samples=num_samples, wfm=wfm, timeout=timeout)
+            return self._fetch_into_numpy(num_samples=num_samples, wfm=wfm, timeout=timeout)
         elif wfm.dtype == numpy.int8:
-            return self._fetch_binary8_into(num_samples=num_samples, wfm=wfm, timeout=timeout)
+            return self._fetch_binary8_into_numpy(num_samples=num_samples, wfm=wfm, timeout=timeout)
```
marcoskirsch: why rename these?
won't they work with array.array anymore?

docs/niscope/functions.rst
texasaggie97: Removed, but added seconds after -1:
'The time to wait for data to be acquired; using 0 for this parameter tells NI-SCOPE to fetch whatever is currently available. Using -1 seconds for this parameter implies infinite timeout.'

docs/niscope/functions.rst
texasaggie97: Two underscores fixed.

Anything with <DRIVER>_ATTR_<C_NAME> should be replaced with a link to session.<python_name>. I need to figure out why these aren't.

generated/niscope/session.py
texasaggie97: Added with text 'waveform array whose length is the **numSamples**'

generated/niscope/session.py
texasaggie97: These attributes control how fetch works. I don't think it is a problem that we are setting the attributes on the session since the purpose of this is to simplify the fetch.

Even if we added them to the _into dispatcher function, they would still be optional, it's just they would always be set since the default value wouldn't be `None`, but whatever the default value is, if there is a single value that works. There may not be if there are device specific defaults.

These do not control how I interpret the waveform data, they are used by the driver to determine exactly which samples to fetch.

generated/niscope/session.py
texasaggie97: Why? You don't want to support ease of use with the advantages of numpy?

generated/niscope/session.py
texasaggie97: Moved

generated/niscope/session.py
texasaggie97: I don't see anything in `six` for `memoryview`'s

generated/niscope/session.py
texasaggie97: Changed, even though we don't know what Python 4 will bring.

generated/niscope/unit_tests/mock_helper.py
texasaggie97: Yeah, this'll get fixed eventually, but since it won't affect the API, it has a lower priority.

src/niscope/examples/niscope_fetch.py
texasaggie97: Changed

src/niscope/system_tests/test_system_niscope.py
texasaggie97: If you have to call _into, you don't get the ease of use benefits. 

We don't really return the same parameter that was passed in. We return either new `array.array`'s or `memoryview`'s of the array passed in, but these correspond to the acquired records and are fairly transformed from what was originally passed in.

It seems like the benefits of the fancy fetch should be available along with the speed of numpy and _into.

src/niscope/templates/session.py/fancy_fetch.py.mako
texasaggie97: I vote: keep wfm parameter, num_samples optional.

src/niscope/templates/session.py/fetch_waveform.py.mako
texasaggie97: They don't work with `array.array` now.

I renamed these because I wanted to add `array.array` but saw it was going to be more involved that I wanted for this PR. I decided to leave the '_numpy' suffix because I would like to get '_into_array' supported but in a different PR.

src/niscope/metadata/functions_addon.py
texasaggie97: Yeah, I need to figure out why. I'll keep looking into it.

src/niscope/system_tests/test_system_niscope.py
marcoskirsch: "fancy fetch" is also "easy to use fetch"

I expect +90% of users to want scaled float results. They will use "fancy fetch".
I expect only a very small share of users to want to extract absolute maximum performance and don't need things scaled for some rare reason. They will use something other than float and the "fetch_into" methods.
Users who want "easy to use fetch" _and_ unscaled data will be extremely rare if any. I don't think adding complexity to "fancy fetch"'s usage that impact +90% of the users is the right tradeoff.

generated/niscope/session.py
marcoskirsch: Because it makes the API more complex for the vast majority of users.

generated/niscope/session.py
marcoskirsch: Ok, they'd still be optional.

But the root issue remains, that there is a hidden side effect / WTH factor. A reasonable user wouldn't expect that passing an attribute value would change the state of the Session object.

We can live with it, but I think it's unexpected and thus an issue.

texasaggie97: I think we are beyond where I have a strong opinion on which way to go with this. I will defer to you or @auchter.

I do agree if we make them private that that should be in a separate PR.

generated/niscope/session.py
texasaggie97: The vast majority of users won't deal with it and will leave it with the default of `None`.

src/niscope/system_tests/test_system_niscope.py
texasaggie97: I don't think it add much complexity at all. Not at the API level (one additional, optional parameter) or in the implementation. 
I do think the additional benefits of encapsulating the property calls as well as bundling the actual waveform with the rest of the waveform information for each individual record is worth it.

generated/niscope/session.py
marcoskirsch: > The vast majority of users won't deal with it and will leave it with the default of None.

They will have to understand that the right thing to do for them is to rely on the default. So they are dealing with it.

src/niscope/metadata/functions_addon.py
texasaggie97: See #749

generated/niscope/session.py
auchter: I agree with Marcos: I don't like that hidden side effect either.

I lean pretty strongly towards deprecating the fetch attributes in favor of making them parameters. So I think what you have here is a good start, but instead of having these parameters be optional and None, they should be optional and specified to the same defaults that the driver uses. So probably like `def fetch(self, timeout=5.0, samples=1000, records=1, relative_to=enums.FetchRelativeTo.PRETRIGGER, offset=0)`.

I think the goal should be that it's possible to look at a call to fetch and understand, in isolation, what data it's going to retrieve.

Also, looking at that signature... why is `channels` not a parameter to fetch?

marcoskirsch: > ...but instead of having these parameters be optional and None, they should be optional and specified to the same defaults that the driver uses.

What worries me about this is that there is an assumption that the driver will use the same defaults for all devices. This isn't guaranteed, is it? Or perhaps if the assumption is broken and the API has different defaults, it's not a huge deal?

> Also, looking at that signature... why is channels not a parameter to fetch?

Check [the green tip](http://nimi-python.readthedocs.io/en/0.7.0/niscope/functions.html#niscope.Session.fetch). :)
Which I just noticed could use some updating in the text.

marcoskirsch: @texasaggie97 I think, summarizing @auchter's suggestion:

1. All fetch methods should take the parameters.
2. We should make the properties private. This is not something the generator supports yet so it'd be a different PR.
3. The parameters should default to explicit values rather than None for discoverability.
4. The defaults should match driver defaults and hope those never change 🤞

I like that this is an overall simplification of the usage API. And it eliminates the side effect because it's no longer visible to clients. I believe all public fetch methods have some hand-coding so it should not be too bad.

texasaggie97:
@auchter 

> Definitely misspoke (typed?) yesterday when I suggested 1000 for num samples. I think this should default to -1, to fetch the entire configured record.

`num_samples` now has a default of `None` (not -1) since it is more Pythonic. I need the actual number so if it is `None`, I go query the actual number in order to allocate the array.

texasaggie97:
> I'd also suggest removing the "fetch_" prefix from these params since it's a but redundant.

Does this include `fetch_offset`? Would just `offset` be too generic?

texasaggie97:
Made fetch attributes private and added as parameters to `fetch_into()`.

marcoskirsch:
Changelog needs updating for the now hidden attributes.

docs/niscope/functions.rst
```diff
@@ -1212,9 +1212,18 @@ niscope.Session methods
 
         .. code:: python
 
-            session.channels['0,1'].fetch(num_samples, timeout='datetime.timedelta(seconds=5.0)')
+            session.channels['0,1'].fetch(timeout='datetime.timedelta(seconds=5.0)', num_samples=None, relative_to=niscope.FetchRelativeTo.PRETRIGGER, offset=0, record_number=0, num_records=-1)
```
marcoskirsch: Since num_samples=None is the default and means "you choose for me", why not make num_records the same?

docs/niscope/session.rst
```diff
@@ -108,13 +108,13 @@ niscope.Session
    +-----------------------------------------------------+---------------------------------------+
    | :py:attr:`fetch_meas_num_samples`                   | int                                   |
    +-----------------------------------------------------+---------------------------------------+
-   | :py:attr:`fetch_num_records`                        | int                                   |
+   | :py:attr:`_fetch_num_records`                       | int                                   |
```
marcoskirsch: Private attributes are still in the RST, they should be removed.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 * ### NI-SCOPE
     * #### Added
     * #### Changed
+        * `niscope.fetch()` now takes additional parameters for associated attributes and add resulting record as part of the waveform info
```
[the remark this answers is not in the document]
texasaggie97: Added

CHANGELOG.md
```diff
@@ -111,7 +112,7 @@ All notable changes to this project will be documented in this file.
         * Repeated capablilites - See [#737](https://github.com/ni/nimi-python/issues/737) for discussion:
             * `channel` repeated capability
     * #### Changed
-        * `niscope.Session()` no longer takes id_query parameter (Fix [#670](https://github.com/ni/nimi-python/issues/670))
+        * `niscope.Session()` no longer takes id_query parameter
```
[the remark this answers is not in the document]
texasaggie97: Fixed

docs/niscope/functions.rst
texasaggie97: The difference is that `None` means go to the driver and get a number that I use to allocate memory, and the -1 is passes into the driver as is.

It would make the logic look like this:
```python
self._fetch_num_records = -1 if num_records is None else num_records
if num_samples is None:
    num_samples = self.horz_record_length
```

Is that okay?

docs/niscope/session.rst
texasaggie97: Removed private attributes.

docs/niscope/functions.rst
marcoskirsch: I think that's ok. The important thing is the interface.

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ All notable changes to this project will be documented in this file.
 * ### NI-SCOPE
     * #### Added
     * #### Changed
+        * `niscope.fetch()` and `niscope.fetch_into()` now takes additional parameters for associated attributes and add resulting record as part of the waveform info
+        * Following properties are now private (added as parameters to fetch calls):
```
marcoskirsch: Minor: The fact that they are private is an implementation detail. From the API point of view, the properties were removed.

`Following properties are removed. Use new parameters to `Session.fetch()` and `Session.fetch_into()`

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ All notable changes to this project will be documented in this file.
 * ### NI-SCOPE
     * #### Added
     * #### Changed
+        * `niscope.fetch()` and `niscope.fetch_into()` now takes additional parameters for associated attributes and add resulting record as part of the waveform info
```
marcoskirsch: Should be `niscope.Session.fetch()` or since you are inside the NI-SCOPE section `Session.fetch()`

docs/niscope/functions.rst
```diff
@@ -1191,7 +1191,7 @@ niscope.Session methods
 
     :type signal_identifier: str
 
-.. py:method:: fetch(num_samples, timeout='datetime.timedelta(seconds=5.0)')
+.. py:method:: fetch(timeout='datetime.timedelta(seconds=5.0)', num_samples=None, relative_to=niscope.FetchRelativeTo.PRETRIGGER, offset=0, record_number=0, num_records=-1)
```
marcoskirsch: I think timeout is (almost?) always the last parameter. Let's do that here too.

docs/niscope/functions.rst
marcoskirsch: On one hand, I think public API is what matters. On the other hand, we have fully generated functions in which -1 means "automatic". I don't have a strong opinion either way, as long as docstring is helpful.

generated/niscope/session.py
```diff
@@ -1899,6 +1914,97 @@ def configure_vertical(self, range, coupling, offset=0.0, probe_attenuation=1.0,
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
+    def fetch(self, timeout=datetime.timedelta(seconds=5.0), num_samples=None, relative_to=enums.FetchRelativeTo.PRETRIGGER, offset=0, record_number=0, num_records=-1):
+        '''fetch
+
+        Returns the waveform from a previously initiated acquisition that the
+                        digitizer acquires for the specified channel. This method returns
```
marcoskirsch: indentation is odd here. How does this affect things?

src/niscope/examples/niscope_fetch.py
```diff
@@ -13,11 +13,11 @@ def example(resource_name, channels, options, length, voltage):
         session.configure_vertical(range=voltage, coupling=niscope.VerticalCoupling.AC)
         session.configure_horizontal_timing(min_sample_rate=50000000, min_num_pts=length, ref_position=50.0, num_records=1, enforce_realtime=True)
         with session.initiate():
-            wfm, wfm_infos = session.channels[channels].fetch(num_samples=length)
-        print('Number of samples acquired: {:,}\n'.format(len(wfm)))
-        for i in range(len(wfm_infos)):
+            waveforms = session.channels[channels].fetch(num_samples=length)
+        for i in range(len(waveforms)):
+            print('Number of samples acquired: {:,}\n'.format(len(waveforms[i].wfm)))
```
marcoskirsch: I think this line needs to go after the header printed in line 19. Right?

src/niscope/metadata/functions_addon.py
```diff
@@ -425,29 +495,29 @@
...
-    'FetchBinary32':                                 { 'method_templates': [
-        { 'session_filename': 'numpy_read_method', 'method_python_name_suffix': '_into', },
+    'FetchBinary32':   { 'method_templates': [
+        { 'session_filename': 'numpy_read_method', 'method_python_name_suffix': '_into_numpy', },
     ], },
-    'Fetch':                { 'method_templates': [
+    'Fetch':           { 'method_templates': [
         { 'session_filename': 'default_method', 'method_python_name_suffix': '', },
-        { 'session_filename': 'numpy_read_method', 'method_python_name_suffix': '_into', },
+        { 'session_filename': 'numpy_read_method', 'method_python_name_suffix': '_into_numpy', },
```
marcoskirsch: The reason these were renamed is because they don't work as is with array.array, right?
And the plan is to fix that and change the name back? If so, I don't think the rename is necessary, just an issue.

src/niscope/templates/session.py/fancy_fetch.py.mako
```diff
@@ -0,0 +1,43 @@
...
+%>\
+    def ${f['python_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.SESSION_METHOD_DECLARATION)}):
+        '''${f['python_name']}
+
+        ${helper.get_function_docstring(f, False, config, indent=8)}
+        '''
+        import sys
+
+        # Set the fetch attributes
+        with _NoChannel(session=self):
```
marcoskirsch: I think this looks very good, very self-documented.

src/niscope/templates/session.py/fancy_fetch.py.mako
```diff
@@ -0,0 +1,43 @@
...
+                end = start + num_samples
+                wfm_info[i].wfm = array.array('d', wfm[start:end])
+        else:
+            # In Python 3 and newer we can use memoryview objects to give us pieces of the underlying array. This is much faster
+            mv = memoryview(wfm)
+
+            for i in range(len(wfm_info)):
+                start = i * num_samples
+                end = start + num_samples
+                wfm_info[i].wfm = mv[start:end]
```
marcoskirsch: It seems most of the code in the the if / else blocks is the same except for this line and the creation of the memory view. If you think you can reduce repetition without making it dirtier, that would be preferable.

src/niscope/templates/session.py/fancy_fetch.py.mako
```diff
@@ -0,0 +1,43 @@
+<%page args="f, config, method_template"/>\
+<%
+    '''Forwards to _fetch_double()'''
```
marcoskirsch: Is this still accurate?

CHANGELOG.md
texasaggie97: Changed

CHANGELOG.md
texasaggie97: Changed

generated/niscope/session.py
texasaggie97: It doesn't. It is cosmetic, but now fixed.

src/niscope/templates/session.py/fancy_fetch.py.mako
texasaggie97: Actually, forwards to _fetch(). Fixed

marcoskirsch:
Note flake8 failed.

Issue #728, opened 2018-01-19
Numpy support is difficult to find for first time users


Fladolcetta:

Fladolcetta:
#693 #694 and #755 should help

PR #755, opened 2018-03-01
Add documentation for niscope fetch() and fetch_into() on usage page

texasaggie97:
* Add usage documentation and example code for `fetch()` and `fetch_into()`
* Partial for #728 


docs/_static/niscope_usage.inc
```diff
@@ -10,8 +10,98 @@ The following is a basic example of using the **niscope** module to open a sessi
         session.configure_vertical(range=1.0, coupling=niscope.VerticalCoupling.AC)
         session.configure_horizontal_timing(min_sample_rate=50000000, min_num_pts=1000, ref_position=50.0, num_records=1, enforce_realtime=True)
         with session.initiate():
-            wfm, wfm_infos = session[args.channels].fetch(num_samples=args.length)
-        print('Number of samples acquired: {:,}\n'.format(len(wfm)))
+            waveforms = session[[0,1]].fetch()
+        print('Number of samples acquired: {:,}\n'.format(len(waveforms.wfm)))
+
+The waveform returned from `fetch <http://nimi-python.readthedocs.io/en/master/niscope/functions.html#niscope.Session.fetch>`_ is a 1D list of Python objects
```
marcoskirsch: * I don't think there are 2D lists, so 1D is redundant.
* Instead of Python objects, can you refer to the type and maybe make it so it links to it?

docs/_static/niscope_usage.inc
```diff
@@ -10,8 +10,98 @@ The following is a basic example of using the **niscope** module to open a sessi
...
+
+        -  **offset** the offset factor of the given channel; useful for scaling binary data with the following formula:
+
+            .. math::
+
+                voltage = binary data * gain factor + offset
+
+        - **wfm** floating point array of samples
+
+    - Such that all record 0 waveforms are first. For example, with a channel list of 0,1, you would have the following index values:
```
marcoskirsch: API question - 
Instead of a list of WaveformInfo, why not return list that contains a dictionary where channels are the key and WaveformInfo the value?

docs/_static/niscope_usage.inc
```diff
@@ -10,8 +10,98 @@ The following is a basic example of using the **niscope** module to open a sessi
...
+
+        - index 0 = record 0, channel 0
+        - index 1 = record 0, channel 1
+        - index 2 = record 1, channel 0
+        - index 3 = record 1, channel 1
+        - etc.
+
+
+If you need more performance, need to work with `SciPy <https://www.scipy.org/>`_, or are concerned about memory usage, you can use the `fetch_into()` method instead of `fetch()`. This
```
marcoskirsch: Don't go into so much detail about fetch_into here, just let the client click through and figure it out.

marcoskirsch: "For minimum overhead at the expense of ease of use, or to best interface with SciPy, use fetch_into() ..."
Since the memory thing is not always accurate (Python 3).

docs/_static/niscope_usage.inc
```diff
@@ -10,8 +10,98 @@ The following is a basic example of using the **niscope** module to open a sessi
...
+method takes an already allocated `numpy <http://www.numpy.org/>`_ array and puts the acquired samples in it. Data types supported:
+
+    - `numpy.float64`
```
marcoskirsch: Right now the method is called fetch_into_numpy. Eventually it will be fetch_into and work with either numpy or array.array.

Do you want to push this as is, or fix array.array first?

docs/_static/niscope_usage.inc
```diff
@@ -10,8 +10,98 @@ The following is a basic example of using the **niscope** module to open a sessi
...
+    - Such that all record 0 waveforms are first. For example, with a channel list of 0,1, you would have the following index values:
+
+        - index 0 = record 0, channel 0
+        - index 1 = record 0, channel 1
+        - index 2 = record 1, channel 0
+        - index 3 = record 1, channel 1
+        - etc.
+
+
+.. note:: Unlike the normal `fetch()` the waveform_infos objects do not include the waveform for that record. Instead, samples are in the waveform passed into the function using the following layout:
```
marcoskirsch: Should this be a Python 2 only limitation, because you want to avoid copies?

docs/niscope/functions.rst
```diff
@@ -1280,7 +1280,8 @@ niscope.Session methods
 
             -  **relative_initial_x** the time (in seconds) from the trigger to the first sample in the fetched waveform
             -  **absolute_initial_x** timestamp (in seconds) of the first fetched sample. This timestamp is comparable between records and acquisitions; devices that do not support this parameter use 0 for this output.
-            -  **x_increment** the time between points in the acquired waveform in seconds -  **actual_samples** the actual number of samples fetched and placed in the waveform array
+            -  **x_increment** the time between points in the acquired waveform in seconds
+            -  **actual_samples** the actual number of samples fetched and placed in the waveform array
```
marcoskirsch: I'm a little lost. Is this supposed to be here?

docs/_static/niscope_usage.inc
texasaggie97: I was trying to convey that it is `[0, 1, 2, 3, 4, 5]` rather than `[[0, 1, 2,], [3, 4, 5]]`

How should I say that if not 1D? Flat?

docs/_static/niscope_usage.inc
texasaggie97: This would require parsing the channel list to know what the channels are.

Using `session.channel['0,2-4'].fetch()` is valid, but would have to be parsed to know that the list is `[0,2,3,4]`.

I considered doing this so I could reshape the returned list to be 2D (or whatever you would want to call it), but figured you would push back about re-implementing that parsing.

docs/_static/niscope_usage.inc
texasaggie97: The function they would use is still called `fetch_into()`. The `_numpy` suffix is only used on the internal, private functions. Like `_fetch_binary8_into_numpy()`

docs/_static/niscope_usage.inc
texasaggie97: Removed memory usage mention. Though I curious how it doesn't help since we aren't creating huge waveforms over and over that need to be garbage collected.

docs/_static/niscope_usage.inc
texasaggie97: Yes. I could add the `memoryview` to the waveform info when using Python 3, but I wanted to maintain compatibility between Python 2 & 3.

docs/niscope/functions.rst
texasaggie97: Yes, it was formatted incorrectly before and the actual_samples line ended up on the same line as x_increment.

docs/_static/niscope_usage.inc
marcoskirsch: Interesting.
We had a similar conundrum in NI-DCPower, and ended up calling a private function they expose.
It's weird to go this far to provide fancy fetch, but stop short with this. Honestly I'm not sure what the right tradeoff is.

docs/_static/niscope_usage.inc
marcoskirsch: That's fair.
But for the Python 3 user, it makes fetch more different from fetch_into than it needs to be. And it makes fetch_into harder to use than it needs to be. I think it's more likely for a client of niscope to stick to one version of Python and call both methods in client code than to switch between versions of Python and stick to one fetch flavor.

The difference between Python versions would need to be documented of course.

docs/_static/niscope_usage.inc
texasaggie97: Added in #759

docs/_static/niscope_usage.inc
texasaggie97: Added

texasaggie97:
I am closing this PR. #761 has these changes plus others that make more sense to be merged together

PR #759, opened 2018-03-01
Add repeated capability list to _SessionBase

texasaggie97:
* Add a list representation of the repeated capabilities to `_SessionBase()`
* Allows adding channel information later

* Unit
* System


build/templates/_converters.py.mako
```diff
@@ -8,6 +8,32 @@ import datetime
 import six
 
 
+def _string_to_list(string, prefix):
```
marcoskirsch: It might sense to reference public documentation on the allowable syntax. I haven't looked for something generic but found [this](http://zone.ni.com/reference/en-XX/help/370736U-01/nidcpowerviref/nidcpower_initialize_with_channels/):

> Specify multiple channels by using a channel list or a channel range. A channel list is a comma (,) separated sequence of channel names (for example, 0,2 specifies channels 0 and 2). A channel range is a lower bound channel followed by a hyphen (-) or colon (:) followed by an upper bound channel (for example, 0-2 specifies channels 0, 1, and 2). In the Running state, multiple output channel configurations are performed sequentially based on the order specified in this parameter.

So at the very least, you are missing `:` as a delimiter. Also make sure you have tests that validate that ordering is preserved based on the last part.

build/templates/_converters.py.mako
```diff
@@ -8,6 +8,32 @@ import datetime
...
+    '''Convert a string to a list of repeated capabilities numbers I.e. no prefix
+
+    '0,1,3' becomes [0, 1, 3]
+    '0-2' becomes [0, 1, 2]
+    'ScriptTrigger0,ScriptTrigger1' becomes [0,1]
```
marcoskirsch: Add example '0,1,5:8,2' becomes [0, 1, 5, 6, 7, 8, 2]

build/templates/_converters.py.mako
```diff
@@ -164,55 +191,85 @@ def test_convert_timedelta_to_microseconds_int():
...
+    assert test_result == ['0', '1']
+    test_result = _string_to_list('ScriptTrigger1, ScriptTrigger0', 'ScriptTrigger')
+    assert test_result == ['1', '0']
+    test_result = _string_to_list('ScriptTrigger0-ScriptTrigger2', 'ScriptTrigger')
+    assert test_result == ['0', '1', '2']
+    test_result = _string_to_list('ScriptTrigger0, ScriptTrigger3-ScriptTrigger7', 'ScriptTrigger')
+    assert test_result == ['0', '3', '4', '5', '6', '7']
+    test_result = _string_to_list('ScriptTrigger2-ScriptTrigger0', 'ScriptTrigger')
+    assert test_result == ['2', '1', '0']
+    test_result = _string_to_list('ScriptTrigger2-ScriptTrigger0, ScriptTrigger4', 'ScriptTrigger')
```
marcoskirsch: wow, this works in our drivers?

build/templates/_converters.py.mako
```diff
@@ -8,6 +8,32 @@ import datetime
 import six
 
 
+def _string_to_list(string, prefix):
+    '''Convert a string to a list of repeated capabilities numbers I.e. no prefix
+
```
marcoskirsch: I think you should add a big fat note about this function duplicating logic in our drivers, and having to keep up with changes and deal with versioning issues should any of that change or be inconsistent across MI drivers.

build/templates/_converters.py.mako
```diff
@@ -8,6 +8,32 @@ import datetime
...
+            else:
+                for i in range(int(start), int(end) + 1):
+                    channels_list.append(str(i))
+        else:
+            channels_list.append(c)
+
+    return channels_list
+
+
 def convert_repeated_capabilities(repeated_capability, prefix=''):
```
marcoskirsch: add docstring on what this function does and returns.

build/templates/session.py.mako
```diff
@@ -93,9 +93,9 @@ class _RepeatedCapabilities(object):
 
     def __getitem__(self, repeated_capability):
         '''Set/get properties or call methods with a repeated capability (i.e. channels)'''
-        rep_caps = _converters.convert_repeated_capabilities(repeated_capability, self._prefix)
+        rep_caps, rep_caps_list = _converters.convert_repeated_capabilities(repeated_capability, self._prefix)
 
-        return _SessionBase(${config['session_handle_parameter_name']}=self._session._${config['session_handle_parameter_name']}, repeated_capability=rep_caps, library=self._session._library, encoding=self._session._encoding, freeze_it=True)
+        return _SessionBase(${config['session_handle_parameter_name']}=self._session._${config['session_handle_parameter_name']}, repeated_capability=rep_caps, repeated_capability_list=rep_caps_list, library=self._session._library, encoding=self._session._encoding, freeze_it=True)
```
marcoskirsch: If both rep_caps, rep_caps_list came from the same repeated_capability, why do you need to pass both to _SessionBase.__init__()?

Shouldn't you pass just one and let _SessionBase process it?

generated/nidcpower/_converters.py
```diff
@@ -4,6 +4,32 @@
 import six
 
 
+def _string_to_list(string, prefix):
```
marcoskirsch: Rename: _repeated_capability_string_to_list()
it's a very specific string syntax...

build/templates/_converters.py.mako
texasaggie97: Added

build/templates/_converters.py.mako
```diff
@@ -8,14 +8,70 @@ import datetime
...
 
+def _repeated_capability_string_to_list(channels, prefix):
+    '''Convert a string to a list of repeated capabilities numbers I.e. no prefix
+
+    This duplicates parsing in the driver, so if changes are made there, they will need to be replicated here.
+
+    '0,1,3' becomes [0, 1, 3]
+    '0-2' becomes [0, 1, 2]
+    'ScriptTrigger0,ScriptTrigger1' becomes [0,1]
+    '0,1,5-8,2' becomes [0, 1, 5, 6, 7, 8, 2]
```
marcoskirsch: add an example that uses `:` to the documentation.

build/templates/_converters.py.mako
```diff
@@ -8,14 +8,70 @@ import datetime
 import six
 
 
+def _repeated_capability_string_to_list(channels, prefix):
```
marcoskirsch: function says "repeated capability string" but parameter name is channels. In our heads it's more or less the same thing, but we should use consistent terminology. Please rename parameter.

build/templates/_converters.py.mako
```diff
@@ -8,14 +8,70 @@ import datetime
...
+    '''
+    # If we recieved a tuple, then call ourselves with each item
+    if isinstance(repeated_capability, tuple):
+        rep_cap_list = []
+        rep_cap = []
+        for r in repeated_capability:
+            rc, rcl = convert_repeated_capabilities(r, prefix)
+            rep_cap.append(rc)
+            rep_cap_list += rcl
+        return ','.join(rep_cap), rep_cap_list
```
marcoskirsch: I think you should return only one thing from this function, rep_cap_list.
`','.join()` is trivial to do on the client side. It adds little value to do it in the function and obscures the code.

build/templates/_converters.py.mako
```diff
@@ -164,55 +223,137 @@ def test_convert_timedelta_to_microseconds_int():
...
+    test_result, test_result_list = convert_repeated_capabilities(slice(0, 1))
+    assert test_result_list == ['0']
     assert test_result == '0'
-    test_result = convert_repeated_capabilities(slice(0, 2))
+    test_result, test_result_list = convert_repeated_capabilities(slice(0, 2))
+    assert test_result_list == ['0', '1']
     assert test_result == '0,1'
-    test_result = convert_repeated_capabilities(slice(None, 2))
+    test_result, test_result_list = convert_repeated_capabilities(slice(None, 2))
+    assert test_result_list == ['0', '1']
```
marcoskirsch: In all these tests, it's weird that you test the second thing returned by the function first.

build/templates/session.py.mako
marcoskirsch: So you pass `repeated_capability = 'ScriptTrigger0,ScriptTrigger1', rep_caps_list =  ['ScriptTrigger0', 'ScriptTrigger1']`? Still seems very redundant. Maybe I misunderstand, but there's no documentation in __init__().

build/templates/_converters.py.mako
texasaggie97: Changed terminology

build/templates/_converters.py.mako
texasaggie97: Updated docstring with new behavior

build/templates/_converters.py.mako
texasaggie97: I have changed the logic for both `convert_repeated_capabilities()` and `_repeated_capability_string_to_list()`, including only returning the list

build/templates/_converters.py.mako
texasaggie97: Only returning the list so only test that.

build/templates/session.py.mako
texasaggie97: I was. I fixed the logic in `convert_repeated_capabilities()` and now it only returns the list.

generated/nifake/unit_tests/test_session.py
```diff
@@ -680,6 +680,16 @@ def test_device_method_not_exist_on_repeated_capability_error(self):
             except AttributeError:
                 pass
 
+    def test_repeated_capabilities_list(self):
```
marcoskirsch: Missing: test with invalid values passed for repeated capabilities.

generated/nifake/unit_tests/test_session.py
```diff
@@ -680,6 +680,16 @@ def test_device_method_not_exist_on_repeated_capability_error(self):
...
+        with nifake.Session('dev1') as session:
+            assert session.channels['0']._repeated_capability_list == ['0']
+            assert session.channels['r0']._repeated_capability_list == ['r0']
+            assert session.channels['0,1']._repeated_capability_list == ['0', '1']
+            assert session.channels['0-2']._repeated_capability_list == ['0', '1', '2']
+            assert session.channels['2-0']._repeated_capability_list == ['2', '1', '0']
```
marcoskirsch: Missing: test for colon as well.

I will say, however, that a lot of this is redundant testing because the conversion function is already tested. You could simply mock the conversion function and make sure that:

* The parameters make it correctly to the function
* They are used correctly by our client code
* Errors are handled properly by our client code

You end up with 3 tests.

generated/nifake/unit_tests/test_session.py
texasaggie97: We actually test all the possibilities more comprehensively in the _converters unit tests, so remove all but one here. Just test that the repeated capabilities are passed along correctly.

generated/nifake/unit_tests/test_session.py
texasaggie97: Same as above. We have one test here now.

build/templates/_converters.py.mako
```diff
@@ -8,28 +8,87 @@ import datetime
 import six
 
 
+def _repeated_capability_string_to_list(repeated_capability, prefix):
+    '''Convert a IVI string format range into a list of repeated capabilities numbers I.e. no prefix
+
+    This duplicates parsing in the driver, so if changes are made there, they will need to be replicated here.
```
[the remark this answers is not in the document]
marcoskirsch: Is the new diff up?

PR #764, opened 2018-03-06
Update public names that start with _

### LINKED ISSUE #756: No function for 2 and 4 wire resistance measurement in nidmm.Function
Measuring resistance is a pretty basic capability of the dmm modules, but the nidmm.Function library doesn't have a 2 or 4 wire resistance measurement function.

texasaggie97:
* Update names that started with a digit and therefore had a `'_'` added

* Fix #756

* Unit
* System
* Visual inspection


CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ All notable changes to this project will be documented in this file.
 * ### NI-DMM
     * #### Added
     * #### Changed
+        * Enum values that start with an underscore + digit have been renamed
+            * `Function._2_WIRE_RES` --> `Function.TWO_WIRE_RES`
+            * `Function._4_WIRE_RES` --> `Function.FOUR_WIRE_RES`
+            * `ThermistorType._44004` --> `ThermistorType._44004`
```
marcoskirsch: Not changed?

CHANGELOG.md
```diff
@@ -40,6 +48,27 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
         * `Session.fetch()` and `Session.fetch_into()` now takes additional parameters for associated attributes and add resulting record as part of the waveform info
+        * Enum values and attribute names that start with an underscore + digit have been renamed
+            * `Session._5102_adjust_pretrigger_samples` --> `Session.adjust_pretrigger_samples_5102`
+            * `Session._5v_out_output_terminal` --> `Session.five_v_out_output_terminal`
+            * `ExportableSignals._5V_OUT` --> `ExportableSignals.FIVE_V_OUT`
```
marcoskirsch: Are these based on the corresponding .NET API changes?

build/helper/metadata_add_all.py
```diff
@@ -348,13 +348,15 @@ def add_all_function_metadata(functions, config):
 
 def _add_python_name(a, attributes):
     '''Adds 'python_name' - lower case + leading '_' if first character is a digit'''
```
marcoskirsch: I think we should now assert if first character is digit.

docs/niscope/enums.rst
```diff
@@ -557,7 +557,7 @@ Enums used in NI-SCOPE
 
 .. py:data:: FlexFIRAntialiasFilterType
 
-    .. py:attribute:: FlexFIRAntialiasFilterType._48_TAP_STANDARD
+    .. py:attribute:: FlexFIRAntialiasFilterType.FOURTYEIGHT_TAP_STANDARD
```
marcoskirsch: Looked at NI-SCOPE .NET API. Names are: Tap48Standard , Tap16Hanning and so forth. I think that if we cannot base the Python names directly from the C names, then we should fall back to basing them off the .NET API rather than invent another transform.

This would be, then, `tap_48_standard`.

See http://digital.ni.com/manuals.nsf/websearch/9F213B8EDEB716B986257EBA0077C1FA

src/nidmm/metadata/enums_addon.py
```diff
@@ -17,3 +17,19 @@
     'OffsetCompensatedOhms': {},  # Delete because boolean values only
 }
 
+enums_override_values = {
+    'Function': { 'values': {  # Write out number
```
marcoskirsch: `# Override names that can't be directly converted from C names into valid Python names`

src/niscope/metadata/attributes_addon.py
```diff
@@ -12,6 +12,11 @@
     1150128: { 'enum': None, 'python_type': 'bool', }, # ENABLE_TIME_INTERLEAVED_SAMPLING, Don't use the enum because a bool will do
 }
 
+attributes_override_values = {
+    1150085: { 'python_name': 'ADJUST_PRETRIGGER_SAMPLES_5102'.lower(), },
```
marcoskirsch: Seems strange to call lower() rather than to spell out the final python_name. What was the reasoning for this?

src/niscope/metadata/enums_addon.py
```diff
@@ -52,91 +60,91 @@
                 },
             },
             {
-                'name': 'NISCOPE_VAL_480I_59_94_FIELDS_PER_SECOND',
+                'name': 'NISCOPE_VAL_VIDEO_480I_59_94_FIELDS_PER_SECOND',
```
marcoskirsch: Are these here because they're missing from CVI file?

CHANGELOG.md
texasaggie97: No.

However,
`NISCOPE_ATTR_5V_OUT_OUTPUT_TERMINAL` ==> `FiveVoltOutputTerminal`
`NISCOPE_ATTR_5102_ADJUST_PRETRIGGER_SAMPLES` ==> not found
`NISCOPE_VAL_5V_OUT` ==> `FiveVoltOutputTerminal` (usage doesn't match up though)

build/helper/metadata_add_all.py
texasaggie97: Done

docs/niscope/enums.rst
texasaggie97: Does that mean we also want to follow their lead on the video triggering enum?

`NISCOPE_VAL_1080I_50_FIELDS_PER_SECOND` == `I1080Lines50FieldsPerSecond`

I don't really like moving the 'P' and 'I' to the beginning, because it is not `I1080` video, it is `1080I` video. Also it isn't a tap 48 filter, it is a 48 tap filter. To me, this is not the right solution to this issue, since this changes the familiar name of these things.

src/nidmm/metadata/enums_addon.py
texasaggie97: Comments changed

src/niscope/metadata/attributes_addon.py
texasaggie97: Less error prone. Copy/paste, then call lower.

Changed.

src/niscope/metadata/enums_addon.py
texasaggie97: No, not missing. Since we use both LV & CVI and they differ we get a superset of them. This is to limit to the portion we want.

docs/niscope/enums.rst
marcoskirsch: Ok. So let it be recorded then in this PR that we're making a conscious decision to ignore the precedent in naming set by .NET.

PR #766, opened 2018-03-08
Rename wfm to waveform

texasaggie97:
* Rename `wfm` parameter to `waveform` in all fetch and read functions

* Unit
* System


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
         * `Session.fetch()` and `Session.fetch_into()` now takes additional parameters for associated attributes and add resulting record as part of the waveform info
+        * Rename `wfm` parameter to `waveform` in `fetch()` and `fetch_into()`
```
marcoskirsch: niscope.Session.fetch() doesn't have a waveform parameter... I don't think it matters that the private stuff changed or not. It isn't user visible.

`fetch(num_samples=None, relative_to=niscope.FetchRelativeTo.PRETRIGGER, offset=0, record_number=0, num_records=None, timeout='datetime.timedelta(seconds=5.0)')`

src/niscope/metadata/functions_addon.py
```diff
@@ -491,6 +491,16 @@
     'FancyFetch':                 { 'python_name': 'fetch', },
 }
 
+functions_parameter_names = {
```
marcoskirsch: Most (all?) here are private. I think it's unnecessary to override, it's just more code to maintain. If you prefer to do it regardless for consistency that's ok but it should then be noted in a comment.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -31,9 +31,9 @@ def test_read(session):
     test_num_channels = 2
     session.configure_vertical(test_voltage, niscope.VerticalCoupling.AC)
     session.configure_horizontal_timing(50000000, test_record_length, 50.0, 1, True)
-    wfm, wfm_infos = session.channels[test_channels].read(test_record_length)
-    assert len(wfm) == test_num_channels * test_record_length
-    assert len(wfm_infos) == test_num_channels
+    waveform, waveforms = session.channels[test_channels].read(test_record_length)
```
marcoskirsch: Do we need to do Fancy Read to match Fancy Fetch (eventually after Fancy Fetch is done)?

src/niscope/metadata/functions_addon.py
texasaggie97: They are and I didn't originally have them, but then there were things like `self._fetch_into_numpy(num_samples=num_samples, wfm=waveform, timeout=timeout)` so I went for consistency.

Comment added.

src/niscope/system_tests/test_system_niscope.py
texasaggie97: I think we do. Though once Fancy Fetch is done, I think the same template will be able to be reused with very minor changes.

I think in general, once Fancy Fetch for Scope is done and Fancy Fetch Multiple is done for DCPower, we should sit down, possibly with driver API experts and see if there are any other low effort improvements we can make to the APIs.

PR #767, opened 2018-03-08
DCPower Fancy Fetch Multiple

### LINKED ISSUE #748: nidcpower.Session.fetch_multiple() API is not Pythonic
See [documentation](http://nimi-python.readthedocs.io/en/0.6.0/nidcpower/functions.html#nidcpower.fetch_multiple)

> Return type: | tuple (voltage_measurements, current_measurements, in_compliance, actual_count)
> WHERE
> voltage_measurements (list of float):Returns an array of voltage measurements. Ensure that sufficient space has been allocated for the returned array.
> current_measurements (list of float):Returns an array of current measurements. Ensure that sufficient space has been allocated for the returned array.
> in_compliance (list of bool):Returns an array of Boolean values indicating whether the output was in compliance at the time the measurement was taken. Ensure that sufficient space has been allocated for the returned array.
> actual_count (int):Indicates the number of measured values actually retrieved from the device.

Instead of returning 3 lists and an int, this should return a single list of dict, each with a voltage, current and in_compliance (exact names TBD).

texasaggie97:
* Add ease of use wrapper for fetch_multiple

* Fix #748 

* Unit
* System


docs/_static/nidcpower_usage.inc
```diff
@@ -22,10 +22,10 @@ The following is a basic example of using the **nidcpower** module to open a ses
...
         with session.initiate():
             while samples_acquired < 20:
-                voltage_measurements, current_measurements, in_compliance = session.fetch_multiple(count=session.fetch_backlog)
-                samples_acquired += len(voltage_measurements)
-                for i in zip(range(len(voltage_measurements)), voltage_measurements, current_measurements, in_compliance):
-                    print(row_format.format(i[0], i[1], i[2], i[3]))
+                measurements = session.fetch_multiple(count=session.fetch_backlog)
+                samples_acquired += len(measurements)
+                for i in range(len(measurements)):
+                    print(row_format.format(i, measurements[i].voltage, measurements[i].current, measurements[i].in_compliance))
```
marcoskirsch: named tuple has a nice repr() / str() so you can simplify a bit and just print out measurements[i]

docs/_static/nidcpower_usage.inc
```diff
@@ -22,10 +22,10 @@ The following is a basic example of using the **nidcpower** module to open a ses
         row_format = '{0:3d}:   {1:8.6f}   {2:8.6f}   {3}'
         with session.initiate():
             while samples_acquired < 20:
```
marcoskirsch: I think you can simplify the example and just fetch all `session.measure_record_length` values. The way we have it is a little odd because if fetch is called twice, the count will start from 0 somewhere in the middle.

So I would get rid of the while loop (assume timeout is large enough).

docs/nidcpower/attributes.rst
```diff
@@ -1876,7 +1876,7 @@ nidcpower.Session properties
...
+    Default Value: If the :py:data:`nidcpower.Session.source_mode` property is set to :py:data:`~nidcpower.SourceMode.SINGLE_POINT`, the default value is  :py:data:`~nidcpower.MeasureWhen.ON_DEMAND`. This value supports only the :py:meth:`nidcpower.Session.measure` method and :py:meth:`nidcpower.Session.measure_multiple`  method. If the :py:data:`nidcpower.Session.source_mode` property is set to :py:data:`~nidcpower.SourceMode.SEQUENCE`, the default value is  :py:data:`~nidcpower.MeasureWhen.AUTOMATICALLY_AFTER_SOURCE_COMPLETE`. This value supports only the :py:meth:`nidcpower.Session._fetch_multiple` method.
```
marcoskirsch: observation: this documentation now references a private method :(

generated/nidcpower/session.py
```diff
@@ -2378,6 +2378,55 @@ def fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
...
+        Args:
+            count (int): Specifies the number of measurements to fetch.
+
+            timeout (datetime.timedelta): Specifies the maximum time allowed for this method to complete, in seconds. If the method does not complete within this time interval, NI-DCPower returns an error.
+
+                Note: When setting the timeout interval, ensure you take into account any triggers so that the timeout interval is long enough for your application.
+
+
+        Returns:
+            measurements (list of float): Returns an array of named tuples.
```
marcoskirsch: This says `(list of float)`

generated/nidcpower/session.py
```diff
@@ -2378,6 +2378,55 @@ def fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
...
+
+        '''
+        import collections
+        Measurements = collections.namedtuple('Measurements', ['voltage', 'current', 'in_compliance'])
```
marcoskirsch: I think the name of the tuple should be singular `Measurement`, right?

generated/nidcpower/session.py
```diff
@@ -2378,6 +2378,55 @@ def fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
...
+
+        voltage_measurements, current_measurements, in_compliance = self._fetch_multiple(count, timeout)
+
+        measurements = []
+        for i in range(count):
```
marcoskirsch: Better:

    for i in zip(self._fetch_multiple(count, timeout)):
        measurements.append(Measurement(i))

I bet this can be a clean one-liner too:

    return [ ... ]

src/nidcpower/examples/nidcpower_source_delay_measure.py
```diff
@@ -6,10 +6,10 @@
 import sys
 
 
-def print_fetched_measurements(voltage_measurements, current_measurements, in_compliance):
-    print('             Voltage : {:f} V'.format(voltage_measurements[0]))
-    print('              Current: {:f} A'.format(current_measurements[0]))
-    print('        In compliance: {0}'.format(in_compliance[0]))
+def print_fetched_measurements(measurements):
+    print('             Voltage : {:f} V'.format(measurements[0].voltage))
```
marcoskirsch: consider leveraging the repr() of the namedtuple Measurement. Less code.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -147,6 +148,82 @@
...
+                'documentation': {
+                    'description': 'Specifies the number of measurements to fetch.',
+                },
+            },
+            {
+                'direction': 'in',
+                'name': 'Timeout',
+                'type': 'ViReal64',
+                'documentation': {
+                    'description': 'Specifies the maximum time allowed for this function to complete, in seconds. If the function does not complete within this time interval, NI-DCPower returns an error.',
```
marcoskirsch: remove `in seconds`?

src/nidcpower/metadata/functions_addon.py
```diff
@@ -147,6 +148,82 @@
...
+                'name': 'measurements',
+                'type': 'ViReal64[]',
+                'documentation': {
+                    'description': 'Returns an array of named tuples.',
+                },
+            },
+        ],
+        'documentation': {
+            'description': '''
+Returns an array of voltage measurements, an array of current
```
marcoskirsch: Taylor this better to fancy fetch.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -147,6 +148,82 @@
...
+measurements, and an array of compliance measurements that were
+previously taken and are stored in the NI-DCPower buffer. This function
+should not be used when the NIDCPOWER_ATTR_MEASURE_WHEN attribute is
+set to NIDCPOWER_VAL_ON_DEMAND. You must first call
+niDCPower_Initiate before calling this function.
+
+Refer to the `Acquiring
+Measurements <REPLACE_DRIVER_SPECIFIC_URL_1(acquiringmeasurements)>`__
```
marcoskirsch: Since this is hand-maintained, don't put `<REPLACE_DRIVER_SPECIFIC_URL_1(acquiringmeasurements)>` in there.

src/nidcpower/templates/session.py/fancy_fetch.py.mako
```diff
@@ -0,0 +1,23 @@
+<%page args="f, config, method_template"/>\
+<%
+    '''Forwards to _fetch() with a nicer interface'''
```
marcoskirsch: Actually `_fetch_multiple`

generated/nidcpower/session.py
```diff
@@ -2378,6 +2378,55 @@ def fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
...
+            measurements.append(Measurements(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=in_compliance[i]))
+
+        return measurements
+
+    def _fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
```
MrElusive: Couldn't count have a default as well? It's 1 in the LV API. 
(I would have added this comment to fetch_multiple above, but apparently github doesn't allow me to comment on unchanged code)

generated/nidcpower/library.py
```diff
@@ -175,6 +176,14 @@ def niDCPower_ExportSignal(self, vi, signal, signal_identifier, output_terminal)
                 self.niDCPower_ExportSignal_cfunc.restype = ViStatus  # noqa: F405
         return self.niDCPower_ExportSignal_cfunc(vi, signal, signal_identifier, output_terminal)
 
+    def niDCPower_FancyFetchMultiple(self, vi, channel_name, count, timeout, measurements):  # noqa: N802
```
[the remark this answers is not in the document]
marcoskirsch: It doesn't?!

This is actually dead code. The way the generator works is that we add a FancyFetchMultiple function to the metadata with a custom template. But the generator doesn't yet have a way to know that the function doesn't exist in the library and it shouldn't generate this. See #695

generated/nidcpower/session.py
marcoskirsch: I was thinking along the same lines, but would make the default be the backlog.
I didn't make a comment because of our policy: https://github.com/ni/nimi-python/wiki/API-design-guidelines#default-values-for-methods

src/nidcpower/metadata/functions_addon.py
texasaggie97: Rewritten

src/nidcpower/metadata/functions_addon.py
texasaggie97: Paragraph removed

src/nidcpower/templates/session.py/fancy_fetch.py.mako
texasaggie97: Changed

generated/nidcpower/session.py
marcoskirsch: Odd. This worked for me:

    >>> a = array.array('i',[1,2,3])
    >>> b = [1,2,3]
    >>> zip (a,b)
    [(1, 1), (2, 2), (3, 3)]

generated/nidcpower/session.py
marcoskirsch: It contains the voltage and the current measurement, but I'm pretty sure they are considered one thing. The count is 1 for each of these.

docs/nidcpower/attributes.rst
marcoskirsch: Definitely separate. We have this problem with a lot of functions we made private such as "configure" functions that simply wrap attributes.

docs/_static/nidcpower_usage.inc
marcoskirsch: ok

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 * ### NI-DCPower
     * #### Added
     * #### Changed
+        * `Session.fetch_multiple()` now returns array of named tuples instead of multiple arrays. See [documentation](http://nimi-python.readthedocs.io/en/master/nidcpower/functions.html#nidcpower.Session.fetch_multiple)
```
marcoskirsch: array -> list ?

docs/_static/nidcpower_usage.inc
marcoskirsch: ok

docs/nidcpower/functions.rst
```diff
@@ -704,27 +704,23 @@ nidcpower.Session methods
 
     :type signal_identifier: str
 
-.. py:method:: fetch_multiple(count, timeout=datetime.timedelta(seconds=1.0))
+.. py:method:: fetch_multiple(count=None, timeout=datetime.timedelta(seconds=1.0))
 
-    Returns an array of voltage measurements, an array of current
-    measurements, and an array of compliance measurements that were
+    Returns an array of named tuples that were
```
marcoskirsch: Why not call out the name of the named tuple?

docs/_static/nidcpower_usage.inc
```diff
@@ -22,10 +22,10 @@ The following is a basic example of using the **nidcpower** module to open a ses
...
-                voltage_measurements, current_measurements, in_compliance = session.fetch_multiple(count=session.fetch_backlog)
-                samples_acquired += len(voltage_measurements)
-                for i in zip(range(len(voltage_measurements)), voltage_measurements, current_measurements, in_compliance):
-                    print(row_format.format(i[0], i[1], i[2], i[3]))
+                measurements = session.fetch_multiple(count=session.fetch_backlog)
```
marcoskirsch: might as well leverage your new default

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -143,15 +143,33 @@ def test_fetch_multiple(single_channel_session):
...
+        assert len(measurements) == count
+        assert isinstance(measurements[1].voltage, float)
+        assert isinstance(measurements[1].current, float)
+        assert measurements[1].in_compliance in [True, False]
+        assert measurements[1].voltage == 1.0
+        assert measurements[1].current == 0.00001
+
+
+'''
+Enable once fetch_backlog is figured out
```
marcoskirsch: Is your plan to figure out in this PR? If not, you should open an issue.

docs/nidcpower/functions.rst
texasaggie97: Added, though I'm not sure it any clearer.

generated/nidcpower/library.py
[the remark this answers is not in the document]
texasaggie97: And see #770 for a fix

generated/nidcpower/session.py
texasaggie97: When I use the named tuple I get this:
```
>   return [Measurement(*m) for m in zip(self._fetch_multiple(count, timeout))]
E   TypeError: __new__() missing 2 required positional arguments: 'current' and 'in_compliance'
```

It doesn't matter if I use `Measurement(m)` or `Measurement(*m)`

src/nidcpower/system_tests/test_system_nidcpower.py
texasaggie97: If it is something I am doing wrong then yes, I would want to fix it here. If this is how the driver works then I would just remove the `None` default value and remove this test.

Do you know how it is supposed to work? The issue is that backlog is always 1 no matter how long I wait.

generated/nidcpower/session.py
MrElusive: We do consider them one thing, and we try to refer to the constituents as "measurement components."

generated/nidcpower/session.py
MrElusive: Was that code executed in Python 2 or 3?

generated/nidcpower/session.py
MrElusive: Are you testing with a simulated device? On some implementations the backlog will always return that one measurement is available.

texasaggie97: Yes, we do our system tests with simulated devices. This particular test is simulating a PXIe-4162.

I can simulate a different one for this test if it will help. What would be an alternative?

generated/nidcpower/session.py
```diff
@@ -2368,9 +2368,63 @@ def configure_aperture_time(self, aperture_time, units=enums.ApertureTimeUnits.S
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
-    def fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
+    def fetch_multiple(self, count=None, timeout=datetime.timedelta(seconds=1.0)):
```
marcoskirsch: I've been thinking about the default `count=None`. I don't think it's the right thing to do.

The most common use case would be for the client to:

* Configure N measurements
* Initiate
* Fetch 

...with the intent of fetching N measurements. Given the default of "whatever is available", most of the time client will get 0 or no measurements.

I'm leaning towards no default for `count`.

texasaggie97: I agree. Default value for count removed.

PR #770, opened 2018-03-09
Remove unused code

### LINKED ISSUE #695: Dead code in Library and mock_helper for pseudo functions

texasaggie97:
* Removes functions that don't exist in the driver from library.py and mock_helper.py
  * They are dead code and will never be used
* Support additional `'codegen_method'` value of `'api'`
  * Added filter method that filters out `'api'` and `'no'` values
* Updated nidcpower, nidmm, nifake, nifgen, and niscope metadata
* No user visible change
* Fix #695 
* Unit
* System
* Inspection of generated files


build/helper/metadata_filters.py
```diff
@@ -301,6 +301,11 @@ def filter_codegen_functions(functions):
     return {k: v for k, v in functions.items() if v['codegen_method'] != 'no'}
 
 
+def filter_library_functions(functions):
+    '''Returns function metadata only for those functions to be included in codegen'''
```
marcoskirsch: A little confusing, because we do code generate for the 'api' only functions. Just not in the library.
Maybe reword: Returns function metadata only for those functions we want to include in the Library class. ?

docs/nidcpower/session.rst
```diff
@@ -328,14 +328,10 @@ nidcpower.Session
    +------------------------------------------------------------+
    | :py:func:`get_channel_name`                                |
    +------------------------------------------------------------+
-   | :py:func:`get_ext_cal_last_date_and_time`                  |
```
marcoskirsch: We shouldn't remove these, they work without cal session and they are useful in user programs.

docs/niscope/functions.rst
```diff
@@ -1191,238 +1191,6 @@ niscope.Session methods
 
     :type signal_identifier: str
 
-.. py:method:: fetch(num_samples=None, relative_to=niscope.FetchRelativeTo.PRETRIGGER, offset=0, record_number=0, num_records=None, timeout=datetime.timedelta(seconds=5.0))
```
marcoskirsch: Why is this disappearing from docs?

generated/nidcpower/library.py
```diff
@@ -263,22 +261,6 @@ def niDCPower_GetExtCalRecommendedInterval(self, vi, months):  # noqa: N802
                 self.niDCPower_GetExtCalRecommendedInterval_cfunc.restype = ViStatus  # noqa: F405
         return self.niDCPower_GetExtCalRecommendedInterval_cfunc(vi, months)
 
-    def niDCPower_GetLastExtCalLastDateAndTime(self, vi, month):  # noqa: N802
```
marcoskirsch: I wonder if we should call all these functions "fancy whatever" so it's obvious they are only fancy python wrapping functions.

generated/nifgen/library.py
```diff
@@ -37,7 +37,6 @@ def __init__(self, ctypes_library):
         self.niFgen_CreateAdvancedArbSequence_cfunc = None
         self.niFgen_CreateArbSequence_cfunc = None
         self.niFgen_CreateFreqList_cfunc = None
-        self.niFgen_CreateWaveformDispatcher_cfunc = None
```
marcoskirsch: ... for example we could call this niFgen_FancyCreateWaveform

src/nidcpower/metadata/functions_addon.py
```diff
@@ -149,7 +149,7 @@
     },
     # Public function that wraps driver function but returns datetime object instead of individual items
     'GetLastExtCalLastDateAndTime': {
-        'codegen_method': 'public',
+        'codegen_method': 'api',
```
marcoskirsch: I think `python-only` would be more self-explanatory. `api` doesn't make it clear if it's the Python bindings or the C API.

Also: Don't forget to update the wiki once this is merged!

build/helper/metadata_filters.py
texasaggie97: Add clarification text: `Returns function metadata only for those functions to included the library layer (library.py and mock_helper.py)`

docs/nidcpower/session.rst
texasaggie97: Missed a conditional. Added back.

docs/niscope/functions.rst
texasaggie97: Missed conditional. Not purposeful.

generated/nidcpower/library.py
texasaggie97: In this PR?

src/nidcpower/metadata/functions_addon.py
texasaggie97: Changed to `'python-only'`

generated/nidcpower/library.py
marcoskirsch: no way

docs/nidcpower/session.rst
marcoskirsch: The diff still doesn't show them...

texasaggie97: There was yet another place where filtering was happening.

Added `filter_public_functions()` and used that for documentation.

PR #773, opened 2018-03-14
Rename all python files to start with '_', other than session

### LINKED ISSUE #672: Modules that are installed should be marked as private
We dont want or expect customers to import nidmm.library and actively want to discourage this to enable use to make changes as necessary. We should make these modules private.

texasaggie97:
* Mark all modules other than `session.py` as private by adding '_' to the beginning of the name.

* Fix #672 

* Unit
* System


epage:
My high level thoughts:
- Any type you publicly expose through parameters, return values, or exceptions should be publicly exposed.
- [Flat is better than nested.](https://www.python.org/dev/peps/pep-0020/)
- For large API, the 90% case should be exposed at the top level (`import dcpower`) and the 10% case should be in one-level deeper imports (`from nixnet import system`).

Huh, the [api checklist](http://python.apichecklist.com/) doesn't seem to have a section on this.  We need to get around to publishing our copy of that that these repos are meant to follow.

texasaggie97:
I can see how enums might need to be public. Suppose they have all the code that uses niscope in a separate package but they may want to pass in scope enums to that package:
``` python
import myscopewrapper
from niscope import enums

def contrived_method():
    return myscopewrapper.simple_read('dev1, relative_to=enums.FetchRelativeTo.PRETRIGGER)
```
Similar type thing for errors. They may want to catch our driver specific exceptions at a higher level
``` python
import myscopewrapper
from niscope import errors

def contrived_method():
    try:
        return myscopewrapper.simple_read('dev1)
    except errors.DriverNotInstalledError:
        print('Someone forgot to install NI-SCOPE')
```
At least that is how I understand why we might want to keep enums and errors public.

What about custom types? I think that they should stay private. The only one we have is the waveform info class we return from niscope read and fetch functions. Because it is returned only, someone using niscope would never need to instantiate one of their own

epage:
> What about custom types? I think that they should stay private. The only one we have is the waveform info class we return from niscope read and fetch functions. Because it is returned only, someone using niscope would never need to instantiate one of their own

Some use cases:
- mypy type annotations.
- `isinstance` checks (less likely)

marcoskirsch:
> The only one we have is the waveform info class we return from niscope read and fetch functions.

I think it probably needs to be public, since it's a type that the customer interacts directly with even if they don't instantiate it. They may also call `__repr__()` on it.

texasaggie97:
`enums`, `errors` and custom types are now back to being public.

In the generated code I kept the `as _enums` and `as _errors`.

PR #776, opened 2018-03-19
Add wrapper for niscope.Session.get_equalization_filter_coefficients to remove size from parameters

### LINKED ISSUE #614: niscope.Session.get_equalization_filter_coefficients() should not have any input parameters
This is the signature of the method:

```
    def get_equalization_filter_coefficients(self, number_of_coefficients)
```

This is what it should be:

```
    def get_equalization_filter_coefficients(self):
```

This is the C function:

```
ViStatus _VI_FUNC niScope_GetEqualizationFilterCoefficients (ViSession vi,
                                                             ViConstString channel,
                                                             ViInt32 numberOfCoefficients,
                                                             ViReal64* coefficients);
```

The problem is that the C function takes in numberOfCoefficients, which user determines by reading attribute NISCOPE_ATTR_EQUALIZATION_NUM_COEFFICIENTS.

We want to make this work but the code-generator doesn't provide a way to do this. It's not clear how we would express this in metadata. Perhaps a new size mechanism? A combination of size mechanism for `coefficients` parameter and something else for `numberOfCoefficients` parameter?

texasaggie97:
* Add wrapper for get_equalization_filter_coefficients to remove size parameter
  * all wrapper does is pass in attribute `equalization_num_coefficients` as the size

* Fix #614 

* Unit
* System


docs/conf.py
```diff
@@ -63,7 +63,7 @@
 # built documents.
 #
 # The full version, including alpha/beta/rc tags.
-release = '0.8.0.dev1'
+release = '0.8.0.dev0'
```
marcoskirsch: merge problem?

src/niscope/metadata/functions_addon.py
```diff
@@ -487,12 +488,38 @@
...
+                'direction': 'in',
+                'name': 'channel',
+                'type': 'ViChar[]',
+                'documentation': {
+                    'description': 'The channel to configure.',
+                },
+            },
+        ],
+        'documentation': {
+            'description': 'Retrieves the custom coefficients for the equalization FIR filter on the device. This filter is designed to compensate the input signal for artifacts introduced to the signal outside of the digitizer. Because this filter is a generic FIR filter, any coefficients are valid. Coefficient values should be between +1 and –1.',
```
marcoskirsch: Isn't this string identical to the one in the extracted metadata? It'd be great if we could just reference that rather than copy it.

docs/conf.py
texasaggie97: Not a merge problem. The issue was that the build_release script didn't update the nifake version when it updates. I fixed the script and manually changed nifake's version to 0.8.0.dev1.

src/niscope/metadata/functions_addon.py
texasaggie97: The problem is we don't have access to the the function dictionary in functions_addon. It is in \_\_init\_\_ that we load functions.py and then load functions_addon.py and then call merge.

I could add a line in \_\_init\_\_.py that looks like 
``` python
functions['funcA']['documentation']['description'] = functions['funcB']['documentation']['description']
```
but that means it will be separated from all the other addon metadata. It seems less bad to duplicate the description in the addon file.

PR #777, opened 2018-03-20
Raise an exception when self test returns non zero code

### LINKED ISSUE #668: self_test() should raise an exception in the event of a self-test failure
Session.self_test() returns an int and a tuple. This is directly derived from how the C API works.

During API reviews it was pointed out that for a Python program, it's very easy to miss the return values by calling:

    session.self_test()

rather than

    code, msg = session.self_test()

It was suggested that self_test() should not return anything, instead it should raise a new exception type (SelfTestFailure or something like that).

texasaggie97:
- [X] I've added & updated tests applicable for this pull request
* Add wrapper function around generated self test
* Raise new self test exception if non zero code returned
* Fixes tests
* Add self_test to nifake to allow unit testing

* Fix #668 

* Unit
* System


src/nifake/unit_tests/test_session.py
```diff
@@ -157,6 +157,28 @@ def test_simple_function(self):
...
+        self.patched_library.niFake_self_test.side_effect = self.side_effects_helper.niFake_self_test
+        test_error_code = 1
+        test_error_message = 'error message'
+        self.side_effects_helper['self_test']['selfTestResult'] = test_error_code
+        self.side_effects_helper['self_test']['selfTestMessage'] = test_error_message
+        with nifake.Session('dev1') as session:
+            try:
+                session.self_test()
+                assert False
+            except Exception as e:
```
texasaggie97: I am not sure what to do here. We do not export the individual exceptions from errors, and the other exceptions inherit from `Exception`, not the driver `Error`. I see two ways to fix this:
* Export all the different exceptions
* All driver specific exceptions inhering from the base driver exception called `Error`

What do others think?

texasaggie97:
``` python
import nifake.errors

try:
    blah()
except nifake.errors.SelfTestFailureError as e:
    print(e)
```
works.

Is this acceptable for customers?

build/templates/errors.py.mako
```diff
@@ -73,6 +73,15 @@ class InvalidRepeatedCapabilityError(Exception):
         super(InvalidRepeatedCapabilityError, self).__init__('An invalid character ({0}) was found in repeated capability string ({1})'.format(invalid_character, invalid_string))
 
 
+class SelfTestFailureError(Exception):
+    '''An error due to a failed self test'''
```
marcoskirsch: I believe tech comm says to use "self-test".

build/templates/session.py/self_test.py.mako
```diff
@@ -0,0 +1,15 @@
+<%page args="f, config, method_template"/>\
+<%
+    '''Call self test and throw if there is a failure'''
```
marcoskirsch: self-test

build/templates/errors.py.mako
```diff
@@ -73,6 +73,15 @@ class InvalidRepeatedCapabilityError(Exception):
         super(InvalidRepeatedCapabilityError, self).__init__('An invalid character ({0}) was found in repeated capability string ({1})'.format(invalid_character, invalid_string))
 
 
+class SelfTestFailureError(Exception):
```
marcoskirsch: I think having both "failure" and "error" (double negative) in the name of the class is redundant. One should be enough. The irony that the second sentence I typed is redundant is not lost on me.

generated/nifake/unit_tests/test_session.py
```diff
@@ -158,6 +159,28 @@ def test_simple_function(self):
...
+        test_error_message = 'error message'
+        self.side_effects_helper['self_test']['selfTestResult'] = test_error_code
+        self.side_effects_helper['self_test']['selfTestMessage'] = test_error_message
+        with nifake.Session('dev1') as session:
+            try:
+                session.self_test()
+                assert False
+            except nifake.errors.SelfTestFailureError as e:
+                assert e.code == test_error_code
+                assert e.message == test_error_message
```
marcoskirsch: You need to fail the test when an exception other than nifake.errors.SelfTestFailureError is thrown as well. As written we would get a false positive.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -148,8 +155,46 @@
             'description': 'Returns the number of channels.',
         },
     },
+    # Public function that wraps self_test and will raise on self test failure
+    'self_test_wrapper': {
```
marcoskirsch: I truly feel this should be Fancy Self Test.

texasaggie97: Changed

PR #779, opened 2018-03-20
Fix codegen for multiple arrays using ivi-dance

### LINKED ISSUE #606: Codegenerator doesn't support multiple output arrays whose size is determined by "IVI dance"
Codegenerator doesn't support multiple output arrays whose size is determined by "IVI dance".
Imagine the following function, which is actually not in my imagination at all:

```
ViStatus _VI_FUNC niScope_GetFrequencyResponse (
    ViSession vi,
    ViConstString channelName,
    ViInt32 bufferSize,
    ViReal64 frequencies[],
    ViReal64 amplitudes[],
    ViReal64 phases[],
    ViInt32* numberOfFrequencies);
```

...where frequencies, amplitudes, phases are output arrays of the same size, and size is determined by IVI-dance.

Turns out our generator doesn't support it (there's an assert). What we need is logic that says "it's ok to have multiple output arrays whose size is determined via IVI-dance as long as they are all the same size (aka use the same parameter to pass the size).

texasaggie97:
* Add support for multiple array parameters that use ivi-dance as a size mechanism
* Enable niscope `get_frequency_response()`
* Fix #606 

* Unit
* System


build/helper/metadata_filters.py
```diff
@@ -216,9 +216,9 @@ def filter_parameters(function, parameter_usage_options):
     #  not call back into ourselves, to avoid infinite recursion
     if parameter_usage_options not in [ParameterUsageOptions.IVI_DANCE_PARAMETER, ParameterUsageOptions.LEN_PARAMETER]:
         # Find the size parameter - we are assuming there can only be one
```
marcoskirsch: This comment appears to be outdated.

build/helper/metadata_filters.py
```diff
@@ -258,7 +258,7 @@ def filter_parameters(function, parameter_usage_options):
     return parameters_to_use
 
 
-def filter_ivi_dance_parameter(function):
+def filter_ivi_dance_parameters(function):
     '''Returns the ivi-dance parameter of a session method if there is one. This is the parameter whose size is determined at runtime.
```
marcoskirsch: Update to plural

build/helper/metadata_filters.py
```diff
@@ -258,7 +258,7 @@ def filter_parameters(function, parameter_usage_options):
...
 
     asserts if more than one parameter found
```
marcoskirsch: no longer true

build/helper/metadata_filters.py
```diff
@@ -270,13 +270,13 @@ def filter_ivi_dance_parameter(function):
         Parameter dict if one is found
     '''
     params = filter_parameters(function, ParameterUsageOptions.IVI_DANCE_PARAMETER)
-    if len(params) == 0:
-        return None
-    assert len(params) == 1, 'Found several ivi-dance parameters, expected only one. Found: {0}'.format([x['name'] for x in params])
-    return params[0]
+    if len(params) > 0:
+        size_param = params[0]['size']['value']
+        assert all(x['size']['value'] == size_param for x in params)
```
marcoskirsch: It's worth explaining this, either here or on the function documentation. It can be done in the assert itself.

Something like "if we have multiple ivi-dance parameters, they all must be of the same size."

build/helper/metadata_filters.py
```diff
@@ -270,13 +270,13 @@ def filter_ivi_dance_parameter(function):
...
+    return params
 
 
-def filter_len_parameter(function):
+def filter_len_parameters(function):
     '''Returns the len parameter of a session method if there is one. This is the parameter whose size is determined at runtime.
```
marcoskirsch: make plural

build/helper/metadata_find.py
```diff
@@ -5,11 +5,17 @@ def find_parameter(name, parameters):
     return parameter[0]
 
 
-def find_size_parameter(parameter, parameters):
+def find_size_parameter(parameter_list, parameters):
     '''Returns the parameter that is used to specify the size of another parameter. Applies to 'ivi-dance' and 'passed-in'.'''
```
marcoskirsch: reword to plural, explain that parameter_list are the parameters for which we are going to find the parameter that is used to specify their size.

build/helper/metadata_find.py
```diff
@@ -5,11 +5,17 @@ def find_parameter(name, parameters):
...
-    if not parameter:
+    if len(parameter_list) == 0:
         return None
-    return find_parameter(parameter['size']['value'], parameters)
+    # We verified earlier that all paramers have the same size parameter, so we only need to use the first one
```
marcoskirsch: "earlier" implies that you know the order in which this is called. You should say simply that it's an assumption.

build/helper/metadata_find.py
```diff
@@ -5,11 +5,17 @@ def find_parameter(name, parameters):
...
+    try:
+        # Try first as a list
+        return find_parameter(parameter_list[0]['size']['value'], parameters)
+    except KeyError:
+        # Not a list, so must be a single parameter
+        return find_parameter(parameter_list['size']['value'], parameters)
```
marcoskirsch: This is weird... why is parameter_list sometimes a list and sometimes not?

build/helper/metadata_find.py
texasaggie97: Changed to:
`Assumption: all paramers have the same size parameter, so we only need to use the first one`

build/helper/metadata_find.py
texasaggie97: Some places we call this function from we only have a single item, others multiple. 

I could have made the change that the places that only have a single parameter to turn it into a list in the function call. `find_size_parameter([param], parameters)` There are more than half a dozen locations that would need this and that seems odd looking.

Instead I chose to handle either in the function. Both solutions would work and I didn't feel either option was better than the other.

build/helper/metadata_find.py
```diff
@@ -5,11 +5,17 @@ def find_parameter(name, parameters):
...
 
-def find_size_parameter(parameter, parameters):
-    '''Returns the parameter that is used to specify the size of another parameter. Applies to 'ivi-dance' and 'passed-in'.'''
-    if not parameter:
+def find_size_parameter(parameter_list, parameters):
+    '''Returns the parameter that is used to specify the size other parameters. Applies to 'ivi-dance' and 'passed-in'.'''
+    if len(parameter_list) == 0:
         return None
-    return find_parameter(parameter['size']['value'], parameters)
+    # Assumption: all paramers have the same size parameter, so we only need to use the first one
```
marcoskirsch: typo

build/helper/metadata_find.py
marcoskirsch: ok

PR #781, opened 2018-03-22
Remove routing enums

### LINKED ISSUE #710: Remove FGEN string enums if they use the routing subsystem
Per discussion on #706 
SampleClockTimebaseSource, SampleClockSource, and ReferenceClockSource are all NI-FGEN Enums that are strings. If these use the routing subsystem, we should remove them from the API

texasaggie97:
* Remove enums whose values are routing strings
  * Using an enum prevents interdevice routing

* Fix #710 

* Unit
* System


src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -376,6 +376,11 @@ def test_user_standard_waveform(session):
     session.clear_user_standard_waveform()
 
 
+def test_routing_attribute(session):
```
texasaggie97: @marcoskirsch Based on what you said, I expected this to fail?

marcoskirsch: I'm not sure. NI-FGEN devices had special circuitry for clock signals. This may be one of those.
Can you try with one of the triggers? Maybe [`nifgen.Session.digital_edge_start_trigger_source`](http://nimi-python.readthedocs.io/en/master/nifgen/attributes.html#nifgen.Session.digital_edge_start_trigger_source)

marcoskirsch:
I did testing using an NI-5441 and spoke to an NI-FGEN developer that worked on the latest PXIe-5433 family, which uses a new driver stack.

* The older DAQmx-based like 5441 devices do indeed use Big-I routing subsystem which allows inter device routing for clocking signals.
* The newer devices like 5433 do not use Big-I routing subsystem for clocking signals; only for triggers and events. This makes sense as the PXI Trigger bus is not suitable for fast, low jitter clocking signals that our devices require.

I've asked him to add more details, but my recommendation is to close this PR without merging, and #710 as not a bug.

texasaggie97:
Closed based on @marcoskirsch recommendation

PR #789, opened 2018-04-02
Scope fancy read

### LINKED ISSUE #790: Add Fancy Read to niscope

texasaggie97:
- [x] I've ~~added~~updated tests applicable for this pull request
* Add ease of use wrapper for read that is almost identical to the fetch ease of use wrapper 'FancyFetch'

* Fix #790 

* Unit
* System


docs/_static/niscope_usage.inc
```diff
@@ -13,7 +13,7 @@ The following is a basic example of using the **niscope** module to open a sessi
         with session.initiate():
             waveforms = session.channels[0,1].fetch(num_records=5)
         for wfm in waveforms:
-            print('Channel {0}, record {1} samples acquired: {2:,}\n'.format(wfm.channel, wfm.record, len(wfm.wfm)))
+            print('Channel {0}, record {1} samples acquired: {2:,}\n'.format(wfm.channel, wfm.record, len(wfm.waveform)))
```
marcoskirsch: based on this... 
Do you think the member should be renamed `waveform` -> `data` or something like that?

docs/niscope/functions.rst
```diff
@@ -1669,12 +1669,12 @@ niscope.Session methods
...
 
 
-.. py:method:: read(num_samples, timeout=datetime.timedelta(seconds=5.0))
+.. py:method:: read(num_samples=None, relative_to=niscope.FetchRelativeTo.PRETRIGGER, offset=0, record_number=0, num_records=None, timeout=datetime.timedelta(seconds=5.0))
 
     Initiates an acquisition, waits for it to complete, and retrieves the
     data. The process is similar to calling :py:meth:`niscope.Session._initiate_acquisition`,
     :py:meth:`niscope.Session.acquisition_status`, and :py:meth:`niscope.Session._fetch`. The only difference is
-    that with :py:meth:`niscope.Session.read`, you enable all channels specified with
+    that with :py:meth:`niscope.Session._read`, you enable all channels specified with
```
marcoskirsch: :(
Question: does this generate a build-time warning, like the ones we generate when the function isn't found at all?
I know we won't fix this for 1.0 but an automated way (i.e. build time warning) to get the full list would be a good thing.

docs/niscope/functions.rst
```diff
@@ -1683,10 +1683,7 @@ niscope.Session methods
 
     
 
-    .. note:: Some functionality is not supported in all digitizers. Refer to
-        `Features Supported by
-        Device <REPLACE_DRIVER_SPECIFIC_URL_1(features_supported_main)>`__ for
-        more information.
+    .. note:: Some functionality, such as time stamping, is not supported in all digitizers.
```
marcoskirsch: Is this change due to new metadata? I'd prefer those to go through own PR.

docs/niscope/functions.rst
```diff
@@ -1696,91 +1693,90 @@ niscope.Session methods
...
+
+
+        The time to wait for data to be acquired; using 0 for this parameter tells NI-SCOPE to fetch whatever is currently available. Using -1 seconds for this parameter implies infinite timeout.
+
+        
 
-            
 
-            .. note:: One or more of the referenced methods are not in the Python API for this driver.
+    :type timeout: float
```
marcoskirsch: should this not be a timedelta?

generated/niscope/session.py
```diff
@@ -2039,6 +2039,113 @@ def get_equalization_filter_coefficients(self):
...
+            self._fetch_relative_to = relative_to
+            self._fetch_offset = offset
+            self._fetch_record_number = record_number
+            self._fetch_num_records = -1 if num_records is None else num_records
+            if num_samples is None:
+                num_samples = self.horz_record_length
+
+        wfm, wfm_info = self._read(num_samples, timeout)
+
+        if sys.version_info.major >= 3:
```
marcoskirsch: From this line to the end of the function, the code is the same as in fetch.
Can it be moved to its own function?

I think something we don't have the ability to do today easily is driver-specific helper functions.

One idea on how to do this is to have a <module>_helpers.py file, and import that automatically form session.py. What do you think?

src/niscope/metadata/functions_addon.py
```diff
@@ -384,6 +387,125 @@
...
+                'documentation': {
+                    'description': 'The instrument handle you obtain from niScope_init that identifies a particular instrument session.',
+                },
+            },
+            {
+                'direction': 'in',
+                'name': 'channelList',
+                'type': 'ViChar[]',
+                'documentation': {
+                    'description': 'The channel to configure.',
```
marcoskirsch: Not that it matters to the public aspects of the project, but in this case we don't actually configure the channel.

src/niscope/metadata/functions_addon.py
```diff
@@ -384,6 +387,125 @@
...
+                },
+            },
+            {
+                'direction': 'in',
+                'default_value': None,
+                'name': 'numSamples',
+                'type': 'ViInt32',
+                'documentation': {
+                    'description': 'The maximum number of samples to fetch for each waveform. If the acquisition finishes with fewer points than requested, some devices return partial data if the acquisition finished, was aborted, or a timeout of 0 was used. If it fails to complete within the timeout period, the function throws an exception.',
```
marcoskirsch: throws an exception -> raises

since we are writing Python-specific documentation.

src/niscope/metadata/functions_addon.py
```diff
@@ -384,6 +387,125 @@
...
+                    'description': 'Position to start fetching within one record.',
+                },
+            },
+            {
+                'direction': 'in',
+                'default_value': 0,
+                'name': 'offset',
+                'type': 'ViInt32',
+                'documentation': {
+                    'description': 'Offset in samples to start fetching data within each record. The offset is applied relative to NISCOPE_ATTR_FETCH_RELATIVE_TO. The offset can be positive or negative.',
```
marcoskirsch: Since this is specific to Python API, didn't we make NISCOPE_ATTR_FETCH_RELATIVE_TO private?
Applies elsewhere.

generated/niscope/session.py
marcoskirsch: I see that you reused the template for the function, so this is less relevant.
We can revisit the idea if it makes sense in the future.

src/niscope/metadata/functions_addon.py
texasaggie97: Fixed.

src/niscope/metadata/functions_addon.py
texasaggie97: Fixed

src/niscope/metadata/functions_addon.py
texasaggie97: Removed references to attributes since they are now private

docs/_static/niscope_usage.inc
marcoskirsch: Ok, so we have `WaveformInfo` and `samples`. We don't have anything called `Waveform`.
Question:

* Should `WaveformInfo` be called `Waveform`? It would be further deviation from C API.

Introducing a `Waveform` type is a big commitment although this would be actually scoped to `niscope.Waveform`.

What do you think?

docs/niscope/functions.rst
marcoskirsch: Separate.

docs/_static/niscope_usage.inc
texasaggie97: I think we should not use Waveform. It is too generic for this. What we return is very specific to niscope and if we ever were to create a common waveform data type, then we wouldn't be able to use the name Waveform without a lot of confusion. "Which Waveform type are talking about?"

marcoskirsch: ok

PR #803, opened 2018-04-05
Allow numeric timeout values

### LINKED ISSUE #796: Anywhere a timedelta is used for parameter, also allow float in seconds
This makes the function call less verbose.
Both will work:
* `fetch(timeout=datetime.timedelta(seconds=1.5))`
* `fetch(timeout=1.5)`

texasaggie97:
* Change timedelta converter to also accept a simple numeric value in seconds
* Standardize on wording `'float in seconds or datetime.timedelta'`

* Fix #796 

* Unit
* System


texasaggie97:
Regarding 3. A more precise `timedelta` won't make a difference to us. We will rely on duck typing so we won't need to make any changes.

One reason for keeping it is discoverability. By using it in our examples we show that the timeout, delay, holdoff type parameters are different that in C. If you're used to the nidmm `fetch()` taking an integer number of milliseconds, then if the only obvious difference is in Python it takes a float, you have to know the somehow that the units is in seconds. If it takes a float or `timedelta`, then it becomes more obvious that something is different.

marcoskirsch:
> By using it in our examples we show that the timeout, delay, holdoff type parameters are different that in C.

Agreed. We need to choose:

* What's more important: The Python bindings to be consistent with timeouts between each other VS Python bindings to be consistent with timeouts across ADEs.

and

* Is the added complexity, code paths, test burden, documentation, eventual support for more precise `timedelta` (maybe our code doesn't need to change, what about discoverability then? Regardless we need to test it) all worth it VS the simpler but potentially error prone (because user must look up units) float?

Fladolcetta:
Please make sure to add documentation to indicate that you can use either.

src/niswitch/metadata/functions_addon.py
```diff
@@ -85,11 +85,11 @@
 
 functions_converters = {
     'ConfigureScanTrigger':              { 'parameters': { 1: { 'python_api_converter_name': 'convert_timedelta_to_seconds',
-                                                                'python_api_converter_type': 'datetime.timedelta', }, }, },
+                                                                'python_api_converter_type': 'float in seconds or datetime.timedelta', }, }, },
```
marcoskirsch: I think it's misleading to call this piece of metadata `'python_api_converter_type'` when it does not contain a valid type. All it contains is a snippet used by documentation.

Since we have `'python_api_converter_name'` which is the name of the converter function used, this could be called something like `'python_api_converter_type_doc'` or something similar.

texasaggie97: I am calling it 'type_in_documentation'. Those are left over and didn't get changed to 'python_type' when the other did. There was some inconsistency between drivers that this will fix.

generated/nifake/session.py
```diff
@@ -115,7 +115,7 @@ class _SessionBase(object):
     An property of type float with read/write access.
```
marcoskirsch: All these say "An property", should say "A property". I bet this happened due to attribute->property replacement. If that's right, please open a separate issue about it.

generated/nifake/session.py
```diff
@@ -130,7 +130,7 @@ class _SessionBase(object):
     An property of type integer with read/write access.
     '''
     read_write_integer_with_converter = _attributes.AttributeViInt32TimeDeltaMilliseconds(1000008)
-    '''Type: datetime.timedelta
+    '''Type: int
```
marcoskirsch: Can you correctly override the type_in_documentation on these so that it matches what a public API would say?

generated/nifake/session.py
```diff
@@ -600,7 +600,7 @@ def __init__(self, resource_name, options={}, id_query=False, reset_device=False
 
 
         Returns:
-            session (nifake.Session): A session object representing the device.
+            session (int): A session object representing the device.
```
marcoskirsch: This got reverted, appears to be wrong now.

generated/nidmm/session.py
```diff
@@ -369,7 +369,7 @@ class _SessionBase(object):
     Specifies the number of measurements the DMM takes each time it receives a  trigger in a multiple point acquisition.
     '''
     sample_interval = _attributes.AttributeViReal64TimeDeltaSeconds(1250303)
-    '''Type: datetime.timedelta
+    '''Type: float
```
marcoskirsch: Inconsistent, doesn't mention datetime.timedelta anymore.
There are others with the same problem.

generated/niscope/session.py
```diff
@@ -1448,28 +1448,28 @@ class _SessionBase(object):
...
     trigger_delay_time = _attributes.AttributeViReal64TimeDeltaSeconds(1250015)
-    '''Type: datetime.timedelta
+    '''Type: float
 
     Specifies the trigger delay time in seconds. The trigger delay time is the length of time the digitizer waits  after it receives the trigger. The event that occurs when the trigger delay elapses is the Reference Event.
     Valid Values: 0.0 - 171.8
     '''
     trigger_from_pfi_delay = _attributes.AttributeViReal64TimeDeltaSeconds(1150052)
-    '''Type: datetime.timedelta
+    '''Type: float
```
marcoskirsch: A lot of these got changed. If it means a lot of duplicate metadata, then we should chat about an alternative on what this is generated from.

generated/nidmm/session.py
texasaggie97: Another place needed to use `'type_in_documentation'`. Fixed

generated/nifake/session.py
texasaggie97: It is. #808

generated/nifake/session.py
texasaggie97: They were overridden, but the template still used `'python_type'`. Fixed

generated/nifake/session.py
texasaggie97: Yet another place that was still using `'python_type'`. Fixed.

docs/nidmm/functions.rst
```diff
@@ -703,7 +703,7 @@ nidmm.Session methods
         .. note:: One or more of the referenced values are not in the Python API for this driver. Enums that only define values, or represent True/False, have been removed.
 
 
-    :type maximum_time: int
+    :type maximum_time: float in seconds or datetime.timedelta
```
marcoskirsch: These are significant changes worth noting in CHANGELOG because they are silent source breakers.
Or is this simply a documentation bug getting addressed?

docs/niswitch/functions.rst
```diff
@@ -782,7 +782,7 @@ niswitch.Session methods
         
 
 
-    :type maximum_time_ms: int
+    :type maximum_time_ms: float in seconds or datetime.timedelta
 
 .. py:method:: wait_for_scan_complete(maximum_time_ms=datetime.timedelta(milliseconds=5000))
```
marcoskirsch: Observation: Based on our defaults, we are "suggesting" clients use timedelta.

generated/niswitch/session.py
```diff
@@ -2447,7 +2447,7 @@ def wait_for_scan_complete(self, maximum_time_ms=datetime.timedelta(milliseconds
         error.
 
         Args:
-            maximum_time_ms (int): Specifies the maximum length of time to wait for the switch module to
+            maximum_time_ms (float in seconds or datetime.timedelta): Specifies the maximum length of time to wait for the switch module to
```
marcoskirsch: If this is a source breaker and not just a documentation fix please call out in CHANGELOG.

docs/nidmm/functions.rst
texasaggie97: This was a metadata bug being addressed. nidmm & niswitch were still using `'python_api_converter_type'` instead of `'python_type'`. Now all drivers are using `'type_in_documentation'`

generated/niswitch/session.py
texasaggie97: This was a metadata bug being addressed. nidmm & niswitch were still using `'python_api_converter_type'` instead of `'python_type'`. Now all drivers are using `'type_in_documentation'`

docs/niswitch/functions.rst
texasaggie97: Yes. Should we somehow document this somewhere? If so, where?

marcoskirsch: I think this is enough. It's just an observation. If we are nudging one way or another, I think we should nudge consistently. This means to keep an eye open for things such as using timeouts in examples in the form of timedeltas.

PR #805, opened 2018-04-12
Add fancy measure multiple

### LINKED ISSUE #792: Add Fancy Measure Multiple to nidcpower
* Use same namedtuple used in fancy fetch multiple
* Use `None` for `in_compliance`

texasaggie97:
* Add a wrapper around `measure_multiple()` that is similar to `fetch_multiple()`

* Fix #792 

* Unit
* System


CHANGELOG.md
```diff
@@ -42,7 +42,7 @@ All notable changes to this project will be documented in this file.
 * ### NI-DCPower
     * #### Added
     * #### Changed
-        * `Session.fetch_multiple()` now returns list of named tuples instead of multiple arrays. See [documentation](http://nimi-python.readthedocs.io/en/master/nidcpower/functions.html#nidcpower.Session.fetch_multiple)
+        * `Session.fetch_multiple()` and `Session.measure_multiple()` now returns list of named tuples instead of multiple arrays. See [documentation](http://nimi-python.readthedocs.io/en/master/nidcpower/functions.html#nidcpower.Session.fetch_multiple)
```
marcoskirsch: "now return"
Since it's two things now.

docs/nidcpower/functions.rst
```diff
@@ -721,14 +721,22 @@ nidcpower.Session methods
 
 .. py:method:: measure_multiple()
 
-    Returns arrays of the measured voltage and current values on the
-    specified output channel(s). Each call to this method blocks other
-    method calls until the measurements are returned from the device. The
-    order of the measurements returned in the array corresponds to the order
-    on the specified output channel(s).
+    Returns an list of named tuples (Measurement) that were
```
marcoskirsch: a list

marcoskirsch: This documentation seems wrong.
measure_multiple doesn't retrieve stuff from a buffer. It actually performs the measurements.
The note I think applies to fetch. Copy / paste error?

generated/nidcpower/session.py
```diff
@@ -2462,6 +2462,50 @@ def fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
...
+                - **current** (float)
+                - **in_compliance** (bool) - Always None
+
+        '''
+        import collections
+        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance'])
+
+        voltage_measurements, current_measurements = self._measure_multiple()
+
+        measurements = []
```
marcoskirsch: Can't you more Pythonic, probably more performant one line list comprehension notation?

generated/nidcpower/session.py
```diff
@@ -2462,6 +2462,50 @@ def fetch_multiple(self, count, timeout=datetime.timedelta(seconds=1.0)):
...
+
+        Returns:
+            measurements (list of Measurement): List of named tuples with fields:
+
+                - **voltage** (float)
+                - **current** (float)
+                - **in_compliance** (bool) - Always None
+
+        '''
+        import collections
```
marcoskirsch: Is this import and creation of Measurement tuple type repeated? Seems like a problem to do this inside the function, performance-wise.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -42,6 +42,7 @@
     'GetExtCalLastDateAndTime':        { 'codegen_method': 'private',  },  # Public wrapper to allow datetime
     'GetSelfCalLastDateAndTime':       { 'codegen_method': 'private',  },  # Public wrapper to allow datetime
     'FetchMultiple':                   { 'codegen_method': 'private',  },  # Fancy Fetch Public wrapper
+    'MeasureMultiple':                 { 'codegen_method': 'private',  },  # Fancy Fetch Public wrapper
```
marcoskirsch: Fancy Measure

docs/nidcpower/functions.rst
texasaggie97: Updated

generated/nidcpower/session.py
texasaggie97: Multiple imports are cheap. It caches what has already been imported.

We don't have a way to have global code that is specific to a driver, so it would be a major change to the code generator to add that.

generated/nidcpower/session.py
texasaggie97: Changed

src/nidcpower/metadata/functions_addon.py
texasaggie97: Changed

generated/nidcpower/session.py
marcoskirsch: It's not just the import, but also the creation of the namedtuple.
I think we can live with it, but something to keep in the back of our minds.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -42,6 +42,7 @@
     'GetExtCalLastDateAndTime':        { 'codegen_method': 'private',  },  # Public wrapper to allow datetime
     'GetSelfCalLastDateAndTime':       { 'codegen_method': 'private',  },  # Public wrapper to allow datetime
     'FetchMultiple':                   { 'codegen_method': 'private',  },  # Fancy Fetch Public wrapper
+    'MeasureMultiple':                 { 'codegen_method': 'private',  },  # Fancy Measure Public wrapper
```
marcoskirsch: Minor: `FancyMeasureMultiple`
I think we'd want search / grep to find the mention so we should use the precise name. Also decreases ambiguity.

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -158,11 +158,11 @@ def test_fetch_multiple(single_channel_session):
 def test_measure_multiple(session):
     with session.initiate():
         # session is open to all 12 channels on the device
-        voltage_measurements, current_measurements = session.measure_multiple()
-        assert len(voltage_measurements) == len(current_measurements) == 12
+        measurements = session.measure_multiple()
+        assert len(measurements) == 12
```
marcoskirsch: should also test that the tuple has voltage / current members.

docs/nidcpower/functions.rst
```diff
@@ -470,11 +470,11 @@ nidcpower.Session methods
 
 .. py:method:: fetch_multiple(count, timeout=datetime.timedelta(seconds=1.0))
 
-    Returns an list of named tuples (Measurement) that were
-    previously taken and are stored in the NI-DCPower buffer. This method
-    should not be used when the :py:data:`nidcpower.Session.measure_when` property is
-    set to :py:data:`~nidcpower.MeasureWhen.ON_DEMAND`. You must first call
-    :py:meth:`nidcpower.Session._initiate` before calling this method.
+    Returns arrays of the measured voltage and current values on the
```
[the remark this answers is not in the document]
texasaggie97: My understanding is they are not similar. Measure multiple takes the measurements on demand. Fetch multiple returns already taken measurements.

Given that, what should they say?

src/nidcpower/metadata/functions_addon.py
texasaggie97: Does this apply to all fancy wrappers in all drivers?

src/nidcpower/system_tests/test_system_nidcpower.py
texasaggie97: Added

src/nidcpower/metadata/functions_addon.py
marcoskirsch: I think that'd be best. Better done in own PR.

docs/nidcpower/enums.rst
```diff
@@ -288,7 +288,7 @@ Enums used in NI-DCPower
 
 
 
-        Acquires a measurement when the :py:meth:`nidcpower.Session.measure` method or :py:meth:`nidcpower.Session.measure_multiple` method is called.
+        Acquires a measurement when the :py:meth:`nidcpower.Session.measure` method or :py:meth:`nidcpower.Session._measure_multiple` method is called.
```
marcoskirsch: Note: Fix when #815 is in.

docs/nidcpower/functions.rst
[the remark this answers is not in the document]
marcoskirsch: Unless I'm crazy, I think it's backwards right now.

```
.. py:method:: fetch_multiple(count, timeout=datetime.timedelta(seconds=1.0))

    Returns arrays of the measured voltage and current values on the
    specified output channel(s). Each call to this method blocks other
    method calls until the measurements are returned from the device. The
    order of the measurements returned in the array corresponds to the order
    on the specified output channel(s).
```


```
.. py:method:: measure_multiple()

    Returns an list of named tuples (Measurement) that were
    previously taken and are stored in the NI-DCPower buffer. This method
    should not be used when the :py:data:`nidcpower.Session.measure_when` property is
    set to :py:data:`~nidcpower.MeasureWhen.ON_DEMAND`. You must first call
    :py:meth:`nidcpower.Session._initiate` before calling this method.
```

Also, "an list" -> "a list".

src/nidcpower/metadata/functions_addon.py
texasaggie97: #819

docs/nidcpower/functions.rst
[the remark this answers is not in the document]
texasaggie97: Reversed, reworded, 'an' -> 'a'

PR #837, opened 2018-04-30
Additional checking when multiple parameters use same size parameter

### LINKED ISSUE #515: nidcpower.Session.set_sequence() can crash you
Method:

    set_sequence(self, values, source_delays=None)

corresponds to function:

    ViStatus _VI_FUNC niDCPower_SetSequence(
        ViSession vi,
        ViConstString channelName,
        const ViReal64 values[],
        const ViReal64 sourceDelays[],
        ViUInt32 size);
 
The way the function works, sourceDelays array is an optional parameter. If you pass NULL (in Python the default is None), then the driver uses whatever default for sourceDelay it has. This doesn't work in our generated code:

    TypeError: object of type 'NoneType' has no len()

So the default doesn't work. That's bad. It gets worse...

Suppose the Python client does pass two lists, but they are of different size. The generated code will grab the size from one of the two, and pass that to the Library. If the one it grabbed it from happens to be larger, then you will end up passing a bunch of garbage from the smaller one (or perhaps zeroes, if ctypes zeroes out array element on allocation).

There are a few system tests, but it's hard to verify that all the right values make it down correctly.

* test_set_sequence_default_source_delays()
* test_set_sequence_no_source_delays()
* test_set_sequence_with_source_delays()
* test_set_sequence_with_too_many_source_delays()
* test_set_sequence_with_too_few_source_delays()

texasaggie97:
* Additional checking when multiple parameters use same size parameter
  * Assumes the first of the listed parameter will be valid if any are
  * 2nd and remaining parameters checked against first's size
* Fix #515 

* Unit
* System


build/helper/codegen_helper.py
```diff
@@ -148,10 +148,11 @@ def get_enum_type_check_snippet(parameter, indent):
 
 def _get_buffer_parameter_for_size_parameter(parameter, parameters):
```
marcoskirsch: Rename to `_get_buffer_parameters_for_size_parameter`
Update docstring.

build/helper/codegen_helper.py
```diff
@@ -148,10 +148,11 @@ def get_enum_type_check_snippet(parameter, indent):
...
     '''If parameter represents the size of another parameter in the C API, returns that other parameter. Otherwise None.'''
+    buffer_params = []
```
marcoskirsch: I think 151 through the end of the function can be written as a one-line list comprehension... if you're feeling pythonic:
https://www.python.org/dev/peps/pep-0202/

generated/nidcpower/session.py
```diff
@@ -3486,6 +3486,8 @@ def set_sequence(self, values, source_delays):
         values_ctype = get_ctypes_pointer_for_buffer(value=values, library_type=_visatype.ViReal64)  # case B550
         source_delays_ctype = get_ctypes_pointer_for_buffer(value=source_delays, library_type=_visatype.ViReal64)  # case B550
         size_ctype = _visatype.ViUInt32(0 if values is None else len(values))  # case S160
+        if source_delays is not None and len(source_delays) != len(values):  # case S160
+            raise ValueError("source_delays length not equal to values length")  # case S160
```
marcoskirsch: I wonder if it's nicer for the string to be "in python" rather than paraphrasing the python above it.
Something like:

``` python
raise ValueError("len(source_delays) != len(values)")  
```

Is that useful? Is paraphrasing useful? Or should we give a more digested message...
"Length of source_delays and values parameters don't match."

generated/nifake/_library.py
```diff
@@ -275,13 +275,13 @@ def niFake_Initiate(self, vi):  # noqa: N802
                 self.niFake_Initiate_cfunc.restype = ViStatus  # noqa: F405
         return self.niFake_Initiate_cfunc(vi)
 
-    def niFake_MultipleArrayTypes(self, vi, output_array_size, output_array, output_array_of_fixed_length, input_array_sizes, input_array_of_floats, input_array_of_integers):  # noqa: N802
+    def niFake_MultipleArrayTypes(self, vi, output_array_size, output_array, output_array_of_fixed_length, input_array_sizes, input_array_of_integers, input_array_of_floats):  # noqa: N802
```
marcoskirsch: For testing, could you have 3 or 4, rather than 2 input arrays whose sizes must match?

src/nifake/metadata/functions.py
```diff
@@ -1510,21 +1510,22 @@
...
+                    'description': 'Array of integers. Optional. If passed in then size must match that of inputArrayOfFloats.',
                 },
-            },
+            },        
             {
                 'direction': 'in',
                 'enum': None,
-                'name': 'inputArrayOfIntegers',
-                'type': 'ViInt16[]',
+                'name': 'inputArrayOfFloats',
```
marcoskirsch: Why did you flip the parameters? Is there some reason I'm missing where only the first one can be optional?

marcoskirsch:
There are other methods throughout the APIs with this problem. At the very least I can think of 
[nifgen.Session.create_advanced_arb_sequence](https://github.com/ni/nimi-python/blob/master/generated/nifgen/session.py#L3697)

``` C
ViStatus _VI_FUNC  niFgen_CreateAdvancedArbSequence(
   ViSession vi,
   ViInt32 seqLength,
   ViInt32 wfmHandles[],
   ViInt32 loopCounts[],
   ViInt32 sampleCounts[],
   ViInt32 markers[],
   ViInt32 coercedMarkers[],
   ViInt32 *seqHandle);
```

My preference is to

a) add support to generator in one PR
b) make the metadata changes to specific drivers in a separate PR

Issue #624, opened 2017-11-22
The values of an enum in enums_addon.py is being applied to all enums whose name contains this enum.


bhaswath:
In enums_addon.py, all dictionaries that start with enums_ go through a process that uses regular expressions.
This is causing problems is because 'Trigger' matches multiple entries like 'TriggerMode' and 'TriggerSource', so they all end up with the values from 'Trigger'.

Issue #534, opened 2017-10-27
Bad NI-FGEN metadata for GetFIRFilterCoefficients


marcoskirsch:
1. Parameter coefficientsArray should have "direction":"out"
2. Parameter numberOfCoefficientsRead should have "direction":"out"
3. Parameter numberOfCoefficientsRead should not be a buffer

```
    'GetFIRFilterCoefficients': {
        'codegen_method': 'public',
        'returns': 'ViStatus',
        'parameters': [
            {
                'direction': 'in',
                'enum': None,
                'name': 'vi',
                'type': 'ViSession',
                'documentation': {
                'description': '''
                Identifies your instrument session. **vi** is obtained from the
                niFgen\_init, nifgen\_InitWithOptions, or nifgen\_InitializeWithChannels
                functions and identifies a particular instrument session.
                ''',
                },
            },
            {
                'direction': 'in',
                'enum': None,
                'name': 'channelName',
                'type': 'ViConstString',
                'documentation': {
                'description': '''
                Specifies the channel name for which you want to configure the operation
                mode.

                **Defined Value**: "0"
                ''',
                },
            },
            {
                'direction': 'in',
                'enum': None,
                'name': 'arraySize',
                'type': 'ViInt32',
                'documentation': {
                'description': 'Specifies the size of the coefficient array',
                },
            },
            {
                'direction': 'in',
                'enum': None,
                'name': 'coefficientsArray',
                'type': 'ViReal64[]',
                'documentation': {
                'description': '''
                Specifies the array of data the onboard signal processor uses for the
                FIR filter coefficients. For the NI 5441, provide a symmetric array of
                95 coefficients to this parameter. The array must have at least as many
                elements as the value that you specify in the **numberOfCoefficients**
                parameter in this function.
                The coefficients should range between –1.00 and +1.00.
                ''',
                },
            },
            {
                'direction': 'in',
                'enum': None,
                'name': 'numberOfCoefficientsRead',
                'type': 'ViInt32[]',
                'documentation': {
                'description': '''
                Specifies the array of data containing the number of coefficients you
                want to read.
                ''',
                },
            },
        ],
        'documentation': {
        'description': '''
        | Returns the FIR filter coefficients used by the onboard signal
        processing block. These coefficients are determined by NI-FGEN and
        based on the FIR filter type and corresponding attribute (Alpha,
        Passband, BT) unless you are using the custom filter. If you are using
        a custom filter, the coefficients returned are those set with the
[...]

PR #838, opened 2018-05-01
Some functions that use script triggers were not properly being set as having a repeated capability

### LINKED ISSUE #831: Some functions in nifgen that have repeated capabilities aren't generated properly
nifgen has script triggers and marker events, and the generator doesn't by default catch those inputs as repeated capabilities, so the methods are generated incorrectly.

Example: `configure_digital_level_script_trigger(self, trigger_id, source, trigger_when)`... trigger_id is repeated capability so the usage should be:

``` python
session.ScriptTriggers[0].configure_digital_level_script_trigger(self, source, trigger_when)
```

We didn't catch this in the system test... it is:

src/nifgen/system_tests/test_system_nifgen.py:    
``` python
session.configure_digital_level_script_trigger('ScriptTrigger0', 'PXI_Trig0', nifgen.TriggerWhen.HIGH)
```

This may go away depending what we do about #827

texasaggie97:
- [X] I've ~~added~~fixed tests applicable for this pull request
* Explicitly set `'is_repeated_capability'` for parameters that aren't names with some variant of channel
* Fix nidcpower documentation to refer to public `initiate` rather than `_initiate`.

* Fix #831 

* Unit
* System


src/nifgen/metadata/functions_addon.py
```diff
@@ -84,8 +84,17 @@
     'ConfigureDigitalLevelScriptTrigger':       { 'parameters': { 3: { 'enum': 'TriggerWhen',               }, }, },
 }
 
+# Need to manually set some repeated capabilities
+functions_rep_caps = {
+    'ConfigureDigitalEdgeScriptTrigger':        { 'parameters': { 1: { 'is_repeated_capability': True, }, }, },
+    'ConfigureDigitalLevelScriptTrigger':       { 'parameters': { 1: { 'is_repeated_capability': True, }, }, },
+    'ConfigureSoftwareEdgeScriptTrigger':       { 'parameters': { 1: { 'is_repeated_capability': True, }, }, },
+    'DisableScriptTrigger':                     { 'parameters': { 1: { 'is_repeated_capability': True, }, }, },
+    'SendSoftwareEdgeTrigger':                  { 'parameters': { 2: { 'is_repeated_capability': True, }, }, },
```
texasaggie97: Should `'SendSoftwareEdgeTrigger'` change this? Is it used differently?

marcoskirsch: This is the C signature:

``` C
ViStatus niFgen_SendSoftwareEdgeTrigger (ViSession vi, viInt32 trigger, viConstString triggerID);
```

Looking at [documentation](http://zone.ni.com/reference/en-XX/help/370524T-01/fgencref/cvinifgen_sendsoftwareedgetrigger/), the `trigger` parameter should be an enum and the `triggerID` parameter should allow for string `"None"` as well. Unless the driver treats `""` and `VI_NULL` to mean the same as `"None"`.

src/nifgen/metadata/functions_addon.py
```diff
@@ -84,8 +84,17 @@
...
+}
+
 functions_issues = {
-    'GetFIRFilterCoefficients':             { 'parameters': { 3: {'direction':'out'},  # TODO(marcoskirsch): Remove when #534 solved
+    'GetFIRFilterCoefficients':             { 'parameters': { 3: { 'direction':'out'},  # TODO(marcoskirsch): Remove when #534 solved
```
marcoskirsch: #534 is closed so it appears the comment is stale.

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -389,14 +389,14 @@ def test_fir_filter_coefficients():
...
-    session.configure_digital_level_script_trigger('ScriptTrigger0', 'PXI_Trig0', nifgen.TriggerWhen.HIGH)
+    session.script_triggers[0].configure_digital_level_script_trigger('PXI_Trig0', nifgen.TriggerWhen.HIGH)
 
 
 def test_send_software_edge_trigger(session):
     waveform_data = [x * (1.0 / 256.0) for x in range(256)]
     session.create_waveform(waveform_data)
     with session.initiate():
-        session.send_software_edge_trigger(nifgen.Trigger.SCRIPT, 'ScriptTrigger0')
+        session.script_triggers[0].send_software_edge_trigger(nifgen.Trigger.SCRIPT)
```
marcoskirsch: Gotta add a test case for "None".

src/nifgen/metadata/functions_addon.py
texasaggie97: There is a `Trigger` enum in enums_addon. Should I use that? It was waiting on #624 which is closed.

I don't think repeated capabilities as currently working will support "None". It will become "ScriptTriggerNone" which is not what we want for this particular use case.

Should maybe this function not use the repeated capability? It seems like a different use.

marcoskirsch: > Should maybe this function not use the repeated capability? It seems like a different use.

I definitely don't think it works as it is in the PR right now:

``` python
send_software_edge_trigger(trigger)
```

Valid use cases would be:

``` python
session.script_triggers[1].send_software_edge_trigger(nifgen.Trigger.SCRIPT)
session["None"].send_software_edge_trigger(nifgen.Trigger.START)
```

If you use the `script_triggers container` then trigger **must be** `nifgen.Trigger.SCRIPT `. That's silly and redundant.

User has no good way of knowing `session.script_triggers["None"]` is what myst be passed if you want to send a start software trigger.

If we didn't make `viConstString triggerID` a repeated capability:

``` python
session.send_software_edge_trigger(nifgen.Trigger.SCRIPT, "ScriptTrigger1")
session.send_software_edge_trigger(nifgen.Trigger.START, "None")
```

User would need to know the magic strings. This is weird but we can live with it.

Ideally, arguably, usage should be:

``` python
session.script_triggers[1].send_software_edge_trigger()
session.send_software_edge_trigger() # sends start trigger
```

No enum needed. I'm not sure this is the best way to go, and I think it would require at the very least a fancy wrapper.

Do other APIs have the same signature for sending software triggers?

marcoskirsch: For expediency, I suggest removing the repeated capability from this function and opening an issue to track it / decide on it separately.

CHANGELOG.md
```diff
@@ -60,6 +60,9 @@ All notable changes to this project will be documented in this file.
 * ### NI-FGEN
     * #### Added
     * #### Changed
+        * Some functions missed setting repeated capabilities, leaving these as parameters instead of using the repeated capabilites object.
+            * `session.configure_digital_edge_script_trigger('0', source, ...)` becomes `session.script_triggers[0].configure_digital_edge_script_trigger(source, ...)`
+            * `session.configure_digital_level_script_trigger('0', source, ...)` becomes `session.script_triggers[0].configure_digital_level_script_trigger(source, ...)`
```
marcoskirsch: The "before" example is wrong. It should be  `session.configure_digital_level_script_trigger('ScriptTrigger0', source, ...)`

generated/nifgen/session.py
```diff
@@ -1496,6 +1496,168 @@ def configure_custom_fir_filter_coefficients(self, coefficients_array):
...
+
+        Configures the specified Script Trigger for digital level triggering.
+
+        Tip:
+        This method requires repeated capabilities (usually channels). If called directly on the
+        nifgen.Session object, then the method will use all repeated capabilities in the session.
+        You can specify a subset of repeated capabilities using the Python index notation on an
+        nifgen.Session instance, and calling this method on the result.:
+
+            session.channels['0,1'].configure_digital_level_script_trigger(source, trigger_when)
```
marcoskirsch: #851 refers to this. :(

src/nifgen/metadata/functions_addon.py
```diff
@@ -85,8 +85,17 @@
     'ConfigureDigitalLevelScriptTrigger':       { 'parameters': { 3: { 'enum': 'TriggerWhen',               }, }, },
 }
 
+# Need to manually set some repeated capabilities
+functions_rep_caps = {
+    'ConfigureDigitalEdgeScriptTrigger':        { 'parameters': { 1: { 'is_repeated_capability': True, }, }, },
```
marcoskirsch: I want to confirm that we decided to keep these functions and also the corresponding properties?
In a review of `nidcpower` we had said to remove the duplicate functionality.

CHANGELOG.md
texasaggie97: Fixed

docs/nifgen/functions.rst
```diff
@@ -474,35 +474,23 @@ nifgen.Session methods
...
 
-.. py:method:: configure_digital_edge_script_trigger(trigger_id, source, edge=nifgen.ScriptTriggerDigitalEdgeEdge.RISING)
+.. py:method:: configure_digital_edge_script_trigger(source, edge=nifgen.ScriptTriggerDigitalEdgeEdge.RISING)
 
     Configures the specified Script Trigger for digital edge triggering.
 
     
 
 
+    .. tip:: This method requires repeated capabilities (usually channels). If called directly on the
```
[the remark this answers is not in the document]
texasaggie97: We will need to either add or infer more information than we do now. At this point, all we know is that there is a repeated capability used in the function, not which one. It is up to the use to know whether to use `.channel[]` or `.script_trigger[]`.

Because we don't have this additional information (yet?) we have no way to specialize the tip.

src/nifgen/metadata/functions_addon.py
texasaggie97: This PR may become moot once we review fgen. 

However, `nidcpower` doesn't have ScriptTriggers so the functions here may have more functionality other than just setting attributes.

src/nifgen/metadata/functions_addon.py
texasaggie97: Looking at #534, I am not sure it is really closed. Looking at the code that closed it, I am not sure how it was supposed to.

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -398,7 +398,7 @@ def test_fir_filter_coefficients():
 
 def test_configure_triggers(session):
     session.configure_digital_edge_start_trigger('PFI0', nifgen.StartTriggerDigitalEdgeEdge.FALLING)
-    session.configure_digital_level_script_trigger('ScriptTrigger0', 'PXI_Trig0', nifgen.TriggerWhen.HIGH)
+    session.script_triggers[0].configure_digital_level_script_trigger('PXI_Trig0', nifgen.TriggerWhen.HIGH)
```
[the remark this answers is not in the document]
texasaggie97: `configure_software_edge_script_trigger()` and `disable_script_trigger()` are not called in the system test, therefore there is nothing to change.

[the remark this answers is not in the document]
texasaggie97: These methods are not in the public API, so I have removed both `configure_software_edge_script_trigger()` and `disable_script_trigger()` from function_addons.py

PR #841, opened 2018-05-01
Change Warning class name to be PEP8 compliant

### LINKED ISSUE #658: NidcpowerWarning name violates PEP8
@epage:

> At best, shouldn't that be "NiDcPowerWarning" or "NIDCPowerWarning"
>  
> Note that PEP8 says that acronyms should not be treated as a word (HTTP instead of Http)
> See https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles
> (I might have made mistakes with this in the past, granted, this is more of a corner case I'm not too worried about)

texasaggie97:

* Renamed warning class name to `<driver>.DriverWarning` for all drivers

* Fix #658 

* Unit
* System

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
         * `SelfTestError` now inherits from `<driver>.Error` rather than `Exception` - [#830](https://github.com/ni/nimi-python/issues/830)
-        * Warning class name is now PEP8 compliant. This means the case has changed. I.e. `NidcpowerWarning` --> `NIDCPowerWarning`, or `NimodinstWarning` --> `NIModInstWarning` - [#658](https://github.com/ni/nimi-python/issues/658)
+        * Warning class name changed to `<driver>.DriverWarning` for all drivers - [#658](https://github.com/ni/nimi-python/issues/658)
     * #### Removed
         * IVI properties as applicable - some where already removed from some drivers [#824](https://github.com/ni/nimi-python/issues/824)
             * `engine_major_version`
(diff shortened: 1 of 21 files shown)


generated/nidmm/__init__.py
```diff
@@ -3,6 +3,6 @@
 
 from nidmm.enums import *          # noqa: F403,F401,H303
 from nidmm.errors import Error     # noqa: F401
-from nidmm.errors import NidmmWarning   # noqa: F401
+from nidmm.errors import NIDMMWarning   # noqa: F401
```
marcoskirsch: I gotta say, I hate the fact that we have a class called `nidmm.NIDMMWarning` and so forth. I wish we could get away with `nidmm.Warning`.

But taking a step back... Having warning class adds a bunch of complexity. Do we even need our own warning classes at all? We never raise these things so clients can't `except` them. What is the advantage to a client in having per-driver Warning classes defined?

marcoskirsch:
> I gotta say, I hate the fact that we have a class called nidmm.NIDMMWarning

I know! Let's call it `DriverWarning`. Consistent with `DriverError` which we already have and define as:

``` python
 class DriverError(Error):
     '''An error originating from the ${driver_name} driver'''		     '''An error originating from the ${driver_name} driver'''
```

generated/nidmm/__init__.py
texasaggie97: Yes, I think we want to keep it for filtering.

I like DriverWarning. Changed.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
         * `SelfTestError` now inherits from `<driver>.Error` rather than `Exception` - [#830](https://github.com/ni/nimi-python/issues/830)
+        * Warning class name is now PEP8 compliant. This means the case has changed. I.e. `NidcpowerWarning` --> `NIDCPowerWarning`, or `NimodinstWarning` --> `NIModInstWarning` - [#658](https://github.com/ni/nimi-python/issues/658)
```
marcoskirsch: This is now outdated.

PR #846, opened 2018-05-03
Add session lock/unlock

### LINKED ISSUE #141: Provide way to lock a session
Drivers have LockSession and UnlockSession functions. It allows clients to do multiple calls to the session that are guaranteed to be serialized.

We aren't exposing that in the Python APIs. Perhaps we should. Something like:

with nidmm.Session('dev1') as session:
    spawn_thread_that_uses(session)
    with session.lock()
        # Guaranteed that no other thread can call session in between do_stuff() and do_more_stuff()
        session.do_stuff()
        session.do_more_stuff()

Of course clients have workarounds for this and can do their own synchronization in Python at a higher layer, so I am not sure we should add this functionality.

texasaggie97:
* Add `session.lock()` and `session.unlock()`
* `session.lock()` return a context manager that can be used to unlock
* Fix #141 

* Unit
* System


build/helper/metadata_add_all.py
```diff
@@ -264,6 +264,11 @@ def _add_render_in_session_base(f):
         f['render_in_session_base'] = f['has_repeated_capability'] or f['is_error_handling']
 
 
+def _add_use_session_lock(f, config):
+    '''Set 'use_session_lock' based on config['use_session_lock'] unless it already exitst'''
```
marcoskirsch: exitst typo

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 * ### ALL
     * #### Added
+        * Add `session.lock_session()` and `session.unlock_session()` to all drivers except nimodinst - [#846](https://github.com/ni/nimi-python/issues/846)
```
marcoskirsch: If we grab session lock, then it should be called out here?

> * Fix thread-safety issues by using IVI session lock

or something like that.

build/templates/functions.rst/lock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,64 @@
...
+
+    | Obtains a multithread lock on the device session. Before doing so, the
+      software waits until all other execution threads release their locks
+      on the device session.
+    | Other threads may have obtained a lock on this session for the
+      following reasons:
+
+    -  The application called the :py:meth:`nidcpower.Session.lock_session` method.
+    -  A call to NI-DCPower locked the session.
+    -  A call to the IVI engine locked the session.
```
marcoskirsch: This line doesn't make sense in the context of Python APIs. If it was extracted, you can drop.

build/templates/functions.rst/lock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,64 @@
...
+
+
+
+    :type caller_has_lock: bool
+
+    :rtype: bool
+    :return:
+
+
+            This parameter is optional. If you do not want to use this parameter, pass None.
```
marcoskirsch: This text appears to be duplicated.

build/templates/session.py.mako
```diff
@@ -87,6 +87,20 @@ class ${session_context_manager}(object):
...
+class _Lock(object):
+    def __init__(self, session):
+        self._session = session
+
+    def __enter__(self):
+        self._caller_has_lock = self._session.lock_session(caller_has_lock=False)
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self._caller_has_lock = self._session.unlock_session(self._caller_has_lock)
```
marcoskirsch: If, as suggested, we get rid of the optional input parameter, then this becomes

``` python
if self._caller_has_lock:
    self._session._unlock_session()
```

build/templates/session.py.mako
```diff
@@ -169,6 +183,11 @@ constructor_params = helper.filter_parameters(init_function, helper.ParameterUsa
             raise AttributeError("'{0}' object has no attribute '{1}'".format(type(self).__name__, key))
         object.__setattr__(self, key, value)
 
+% if config['use_session_lock']:
+    def lock(self):  # TODO(texasaggie97) Need to figure out how to document this
```
marcoskirsch: how come there's no public `unlock()`?

build/templates/session.py.mako
```diff
@@ -244,6 +263,8 @@ class Session(_SessionBase):
         return ${session_context_manager}(self)
 
     def close(self):
+        # We do not acquire a lock for close. There is a very small race condition where _close() is called but before
+        # self._vi is set to 0, another function is called. This will return a driver error, so we feel this is acceptable
```
marcoskirsch: `This will rase a DriverError`

build/templates/session.py/fancy_self_test.py.mako
```diff
@@ -2,14 +2,19 @@
 <%
     '''Call self-test and throw if there is a failure'''
     import build.helper as helper
+    lock_indent = '    ' if f['use_session_lock'] else ''
```
marcoskirsch: This seems pretty repetitive and dirty.
I would think we can use Python decorators for this stuff. Did you look into that?

build/templates/session.py/lock_session_wrapper.py.mako
```diff
@@ -0,0 +1,60 @@
...
+                lock_session call and store the return value into a variable. Pass in the variable as well
+                as putting the return value into the same variable for each call to lock_session or
+                unlock_session.
+        '''
+        vi_ctype = _visatype.ViSession(self._vi)  # case S110
+        caller_has_lock_ctype = _visatype.ViBoolean(caller_has_lock) if caller_has_lock is not None else None
+        error_code = self._library.${c_function_prefix}LockSession(vi_ctype, None if caller_has_lock_ctype is None else (ctypes.pointer(caller_has_lock_ctype)))
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=True)
+        return bool(caller_has_lock_ctype.value) if caller_has_lock else True
+
```
marcoskirsch: Why weren't you able to use the generator for this?

docs/nidcpower/attributes.rst
```diff
@@ -1824,7 +1824,7 @@ nidcpower.Session properties
 
     Specifies whether the output is enabled (True) or disabled (False).
     Depending on the value you specify for the :py:data:`nidcpower.Session.output_function` property, you also must set the  voltage level or current level in addition to  enabling the output
-    the :py:meth:`nidcpower.Session.initiate` method. Refer to the Programming States topic in the NI DC Power Supplies and SMUs Help for  more information about NI-DCPower programming states.
+    the :py:meth:`nidcpower.Session._initiate` method. Refer to the Programming States topic in the NI DC Power Supplies and SMUs Help for  more information about NI-DCPower programming states.
```
marcoskirsch: merge problem

src/nidmm/templates/functions.rst/get_cal_date_and_time.rst.mako
```diff
@@ -1,33 +0,0 @@
-<%page args="function, config, method_template, indent"/>\
```
marcoskirsch: How is this related to session lock? Merge issue?

src/niscope/templates/session.py/fetch_waveform.py.mako
```diff
@@ -13,47 +13,48 @@
...
-                if sys.version_info.major >= 3:
-                    start = i * num_samples
-                    end = start + wfm_info[i].actual_samples
-                    del wfm_info[i].actual_samples
-                    wfm_info[i].samples = mv[start:end]
-
-                i += 1
-
-        return wfm_info
+        with self.lock():
```
marcoskirsch: While this is cleaner than looking at config, it divorces it from the metadata setting of whether to use the lock or not.
Hopefully this can be all fixed through the use of decorators instead so it becomes a non-issue.

generated/nidmm/session.py
```diff
@@ -569,6 +581,9 @@ def __setattr__(self, key, value):
             raise AttributeError("'{0}' object has no attribute '{1}'".format(type(self).__name__, key))
         object.__setattr__(self, key, value)
 
+    def lock(self):  # TODO(texasaggie97) Need to figure out how to document this
```
marcoskirsch: If it doesn't go in the PR, please open an issue.

CHANGELOG.md
texasaggie97: Added

build/helper/metadata_add_all.py

build/templates/functions.rst/lock_session_wrapper.rst.mako
texasaggie97: It came from the extracted metadata, but copied here. How does it not make sense? How should it be changed?

build/templates/functions.rst/lock_session_wrapper.rst.mako
texasaggie97: It is duplicated. It is an in/out parameter, so the documentation is listed as both the input and the output.

build/templates/session.py.mako
texasaggie97: Why would we get rid of it? It still makes sense to allow using it for the same use cases as in the documentation.

build/templates/session.py.mako
texasaggie97: `lock()` is the context manager. `lock_session()` and `unlock_session()` are the functions.

build/templates/session.py.mako
texasaggie97: Fixed

build/templates/session.py/fancy_self_test.py.mako
texasaggie97: Decorators can't access the function parameters, which is what we would need to be able to call lock_session and unlock_session from the decorator.
At least not with out relying on parameters being consistent with vi always first and assuming that.
Is that acceptable? If so I can try something based on https://stackoverflow.com/questions/5929107/decorators-with-parameters

build/templates/session.py/lock_session_wrapper.py.mako
texasaggie97: Because the parameter is in/out and the generator doesn't support that. And I felt this was a better solution that contorting the generator to support it.

docs/nidcpower/attributes.rst
texasaggie97: Fixed

src/nidmm/templates/functions.rst/get_cal_date_and_time.rst.mako
texasaggie97: Unused file. Should I not clean up when I see an issue? Seems overkill to create a PR just to delete unused files

build/templates/functions.rst/lock_session_wrapper.rst.mako
marcoskirsch: >   -  A call to the IVI engine locked the session.
 
How can there be a call to the IVI engine from Python?

build/templates/session.py.mako
marcoskirsch: With `initiate()` we return the context manager and we have an `abort()` and that's it.
This is different, we have `lock()`which returns the context manager, plus `lock_session()` and `unlock_session()`. I'm guessing this has to do with returning a bool for `caller_has_lock`.

I'd like to see if this can be simplified in a way that makes sense.

build/templates/session.py/fancy_self_test.py.mako
marcoskirsch: Can you rely on the first parameter always being `self` always being a `Session`?
Since this may take a little while, I'd like to suggest adding the lock/unlock functions and context manager in one PR, then using it in a separate PR.

build/templates/session.py.mako
marcoskirsch: I need to read more on it.

src/nidmm/templates/functions.rst/get_cal_date_and_time.rst.mako
marcoskirsch: > Unused file. Should I not clean up when I see an issue? Seems overkill to create a PR just to delete unused files

The reviewer of a 67 file PR might disagree. Regardless, it wasn't clear to me if this was accidental or not. The PR's description didn't call it out either.

generated/nidcpower/session.py
```diff
@@ -3638,6 +3738,8 @@ def initiate(self):
         return _Acquisition(self)
 
     def close(self):
+        # We do not acquire a lock for close. There is a very small race condition where _close() is called but before
```
marcoskirsch: This comment no longer belongs in this PR.

generated/nidcpower/session.py
```diff
@@ -2855,6 +2870,60 @@ def _get_error(self):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=True)
         return int(code_ctype.value), description_ctype.value.decode(self._encoding)
 
+    def lock_session(self, caller_has_lock=None):
```
marcoskirsch: I don't think this `caller_has_lock` business is useful to Python users. My understanding from the documentation is that it's useful to a C user that has no access to things like RAII or context managers and is using `checkErr` macros with their fun gotos.

My vote is to expose only `def lock_session(self)` that returns the context manager and doesn't have `caller_has_lock` and `def unlock_session(self)` with no parameters.

Another thing to consider is whether we want to make these functions match, for duck-typing reasons, the naming used in rlock. Need to think about that a little bit more.

build/templates/session.py.mako
texasaggie97: * Made `lock_session()` private
* Change `unlock_session()` to `unlock()`
* `_Lock()` context manager only unlocks
* `lock()` will call `_lock_session()` and then return `_Lock()` context manager

These changes are to support both
``` python
with session.lock():
   do_something(session)
```
and
``` python
session.lock()
do_something(session)
session.unlock()
```

build/templates/_mock_helper.py.mako
```diff
@@ -95,7 +95,8 @@ if p['use_array']:
         for i in range(len(test_value)):
             ${param_name}[i] = test_value[i]
 %           else:
-        ${p['python_name']}.contents.value = self._defaults['${func_name}']['${p['name']}']
+        if ${p['python_name']} is not None:
```
marcoskirsch: Why would `python_name` ever be None? Is this still a valid codepath?

build/templates/functions.rst/lock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,26 @@
...
+.. py:method:: ${function['python_name']}()
+
+    | Obtains a multithread lock on the device session. Before doing so, the
+      software waits until all other execution threads release their locks
+      on the device session.
+    | Other threads may have obtained a lock on this session for the
+      following reasons:
+
+    -  The application called the :py:meth:`nidcpower.Session.lock_session` method.
+    -  A call to NI-DCPower locked the session.
```
marcoskirsch: Why is NI-DCPower mentioned here? Seems like you should use the mako variables.

There's a few more places with same issue.

build/templates/functions.rst/unlock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,11 @@
+<%page args="function, config, method_template, indent"/>\
+<%
+    import build.helper as helper
+%>\
+.. py:method:: ${function['python_name']}()
+
+    Releases a lock that you acquired on an device session using
+    :py:meth:`nidcpower.Session.lock_session`. Refer to :py:meth:`nidcpower.Session.lock_session` for additional
```
marcoskirsch: This is not an `nidcpower` specific template.

build/templates/functions.rst/unlock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,11 @@
+<%page args="function, config, method_template, indent"/>\
+<%
+    import build.helper as helper
+%>\
+.. py:method:: ${function['python_name']}()
```
marcoskirsch: If unlock_session became private, then this file is not needed, right? Same for lock_session_wrapper.rst.mako... ?

build/templates/functions.rst/lock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,26 @@
...
+%>\
+.. py:method:: ${function['python_name']}()
+
+    | Obtains a multithread lock on the device session. Before doing so, the
+      software waits until all other execution threads release their locks
+      on the device session.
+    | Other threads may have obtained a lock on this session for the
+      following reasons:
+
+    -  The application called the :py:meth:`nidcpower.Session.lock_session` method.
```
marcoskirsch: You say you made lock_session private, unlock_session now unlock... outdated stuff here.

build/templates/session.py/lock_session_wrapper.py.mako
```diff
@@ -0,0 +1,35 @@
+<%page args="f, config, method_template"/>\
+<%
+    import build.helper as helper
+
+    c_function_prefix = config['c_function_prefix']
+%>\
+    def ${f['python_name']}(self):
+        '''lock_session
```
marcoskirsch: lock? use mako variable.
Make sure file name matches public function.

build/templates/session.py/lock_session_wrapper.py.mako
```diff
@@ -0,0 +1,35 @@
...
+
+        | Obtains a multithread lock on the device session. Before doing so, the
+          software waits until all other execution threads release their locks
+          on the device session.
+        | Other threads may have obtained a lock on this session for the
+          following reasons:
+
+        -  The application called the lock_session method.
```
marcoskirsch: should be lock() everywhere.

build/templates/session.py/lock_session_wrapper.py.mako
```diff
@@ -0,0 +1,35 @@
...
+        -  A call to NI-DCPower locked the session.
```
marcoskirsch: no NI-DCPower, use mako variable.

build/templates/session.py/lock_session_wrapper.py.mako
```diff
@@ -0,0 +1,35 @@
...
+        -  After a call to the lock_session method returns
+           successfully, no other threads can access the device session until
+           you call the unlock_session method.
```
marcoskirsch: ... or context manager?

build/templates/session.py/unlock_session_wrapper.py.mako
```diff
@@ -0,0 +1,18 @@
+<%page args="f, config, method_template"/>\
+<%
+    import build.helper as helper
+
+    c_function_prefix = config['c_function_prefix']
+%>\
+    def ${f['python_name']}(self):
+        '''unlock_session
```
marcoskirsch: unlock()

docs/nidcpower/session.rst
```diff
@@ -344,6 +344,8 @@ nidcpower.Session
    +------------------------------------------------------------+
    | :py:func:`set_sequence`                                    |
    +------------------------------------------------------------+
+   | :py:func:`unlock`                                          |
```
marcoskirsch: where is lock()?

generated/nidcpower/session.py
```diff
@@ -2855,6 +2873,35 @@ def _get_error(self):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=True)
         return int(code_ctype.value), description_ctype.value.decode(self._encoding)
 
+    def _lock_session(self):
```
marcoskirsch: It's up to you, but if you're hand-coding both _lock_session() and lock(), I think it just complicates things to have them be separate.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -49,6 +47,15 @@
     'DeleteAdvancedSequence':          { 'codegen_method': 'private',  },  # Advanced sequence private until #504 has a fix
 }
 
+functions_locking = {
```
marcoskirsch: functions_locking seems odd... function_locking?

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -203,6 +203,12 @@ def test_create_advanced_sequence(single_channel_session):
     single_channel_session._create_advanced_sequence(sequence_name='my_sequence', attribute_ids=ids, set_as_active_sequence=True)
 
 
+def test_lock_context_manager(single_channel_session):
```
marcoskirsch: What's special about nidcpower that you added only to its system test?
How does it test that the lock gets released?

Should be enough to put this in nifake unit tests, I would think.

generated/nidcpower/session.py
texasaggie97: Moved lock into lock_session template

generated/nidmm/session.py
texasaggie97: Figured out.

src/nidcpower/metadata/functions_addon.py
texasaggie97: Has to begin with 'functions_' for add_all to find it.

src/nidcpower/system_tests/test_system_nidcpower.py
texasaggie97: Originally I was unsure if unit tests would be enough. I am now convinced that they are so removed this test

src/nidmm/templates/functions.rst/get_cal_date_and_time.rst.mako
texasaggie97: We're down to 55 files changed!

build/templates/functions.rst/lock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,47 @@
...
+    One method for ensuring there are the same number of unlock method calls as there is lock calls
+    is to use lock as a context manager
+
+        .. code:: python
+
+            with ${config['module_name']}.Session('dev1') as session:
+                with session.lock():
+                    # Calls to session within a single lock context
+
+        The first with block ensures the session is closed regardless of any exceptions raised
```
marcoskirsch: `with` (use markdown for the with keyword) consistently.

build/templates/functions.rst/lock_session_wrapper.rst.mako
```diff
@@ -0,0 +1,47 @@
+<%page args="function, config, method_template, indent"/>\
```
marcoskirsch: Since the method this is used for is `lock`, I think the filename should be `lock.rst.mako` (I'm not sure what "wrapper" buys us).

This applies to the rest of the new mako templates.

generated/nifake/unit_tests/test_session.py
```diff
@@ -151,8 +157,25 @@ def test_session_context_manager_close_with_error(self):
...
+        with nifake.Session('dev1') as session:
+            session.lock()
+            self.patched_library.niFake_LockSession.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), None)
+
+    def test_unlock_session_none(self):
+        with nifake.Session('dev1') as session:
+            session.unlock()
+            self.patched_library.niFake_UnlockSession.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), None)
 
+    def test_lock_context_manager(self):
```
marcoskirsch: consider adding test in which you exit `with` block by raising.

src/nidcpower/metadata/config.py
```diff
@@ -27,5 +27,6 @@
     'repeated_capabilities': [
         {'python_name': 'channels', 'prefix': '', },
     ],
+    'use_session_lock': True,
```
marcoskirsch: I don't think this is being used. Would be better not to have an extra piece of global metadata unless there's no better way.

src/nidmm/metadata/functions_addon.py
```diff
@@ -53,6 +51,15 @@
     'self_test':                       { 'codegen_method': 'private', 'method_name_for_documentation': 'self_test',             },  # 'fancy_self_test' Public wrapper that raises
 }
 
+functions_locking = {
```
marcoskirsch: Food for thought:

This is identical across drivers.
It'd be nice if we could single source it.

generated/nifake/unit_tests/test_session.py
texasaggie97: Added

src/nidcpower/metadata/config.py
texasaggie97: It is being used during add_all to know what the default should be. I.e. if it isn't already set for a function, the value from config is uses.

src/nidmm/metadata/functions_addon.py
texasaggie97: It would be. There's been a couple of things now that would be nice to since source, but not always add.

src/nidcpower/metadata/config.py
marcoskirsch: But if it doesn't need to be configurable, then we should remove it and default to True.

Or are you concerned that processed metadata for nimodinst will say True everywhere which is misleading? Is this enough reason to add config metadata?

Alternatively, and assuming there are potential race conditions in nimodinst bindings because several functions in the library are called within a public method call,  we could fix nimodinst race conditions by adding our own recursive lock at the Python level (I don't favor this but we could at least document it as an issue).

texasaggie97: Saying `True` for nimodinst isn't just misleading, but will generate incorrect code. It will then try to add the lock decorator for each function in #857

PR #848, opened 2018-05-04
Remove export signal

### LINKED ISSUE #828: export_signal() functions are redundant and incomplete
nifgen, niscope, nidcpower have `export_signal(self, signal, output_terminal, signal_identifier="None")`.

From [Zen of Python](https://www.python.org/dev/peps/pep-0020/#id3):

> There should be one-- and preferably only one --obvious way to do it.

The same functionality is available through properties. Additionally, there is extra functionality only available via properties, such as pulse width and polarity in some cases.

Given this and the fact that we want to nudge our customers to learn to pull signals from source terminal rather than to export to a PXI Trigger line so that NI's routing subsystem can manage those for you, we want to remove these functions and leave only the properties.

texasaggie97:
- [X] I've ~~added~~removed tests applicable for this pull request
* Remove export_signal from nidcpower, nifgen, & niscope
  * We want clients to "Pull" the signal from the device instead of exporting. I.e. The scope's trigger source would be `'fgen5433\sometrigger'`
  * `export_signal()` does not provide all the functionality that properties do
* Fix #828 

* Unit
* System


generated/niscope/enums.py
```diff
@@ -205,20 +205,6 @@ class _ClearableMeasurement(Enum):
     TIME_HISTOGRAM_NEW_HITS = 3011
 
 
-class ExportableSignals(Enum):
```
marcoskirsch: In NI-FGEN this was called `Signal`. Needless inconsistency. Glad it's gone.

mitchellkelley:
@marcoskirsch Can you give an example for what the replacement of the Export Signal function would look like using the properties as mentioned at the top. For example, how would you export a trigger to PFI 0 on a FGEN card?

marcoskirsch:
> For example, how would you export a trigger to PFI 0 on a FGEN card?

```
my_session.exported_start_trigger_output_terminal = "PFI0"
```

gfisher-NI:
@marcoskirsch @mitchellkelley I think this is a case where we may want to have something in the documentation explaining how to do this configuration, as this is a place where the Python API differs from the C API (where niFgen_ExportSignal exists). Usually, if documentation doesn't exist for something in Python, users (and supporters) can reference the C or LabVIEW API and generally figure out what they need to do, but that's not the case here.

sbethur:
> I think this is a case where we may want to have something in the documentation explaining how to do this configuration, as this is a place where the Python API differs from the C API (where niFgen_ExportSignal exists). Usually, if documentation doesn't exist for something in Python, users (and supporters) can reference the C or LabVIEW API and generally figure out what they need to do, but that's not the case here.

@gfisher-NI, can you please open an issue on GitHub for this.

gfisher-NI:
@sbethur Done!

gfisher-NI:
@marcoskirsch @sbethur I'm trying to put together an example of this, and where I am stuck is that there is no indication anywhere of what values digital_edge_start_trigger_source and exported_start_trigger_output_terminal can take. I would expect there to be an enum for it and there isn't.

marcoskirsch:
@gfisher-NI, we made the explicit the decision to not include constants in the API for terminals on the same instrument. The optimal way of using our APIs is to leverage NI's routing subsystem and "pull" signals from other instruments. In other ADEs we do declare constants, but this is mostly a carryover practice from before NI's routing subsystem (~2002).

See https://github.com/ni/nimi-python/wiki/API-design-guidelines#encourage-use-of-cross-device-trigger-routing

Unfortunately, it is not obvious to customers what the valid strings are for a given system configuration. Some instruments have a Routing tab in MAX and this can be referrenced. Many others don't. This is a widespread documentation gap throughout NI driver software.

gfisher-NI:
@marcoskirsch Okay, that does make sense, and also pointed me at the mistake in my code. This code works:

`with nifgen.Session('PXI1Slot12', [0,1]) as dev1, nifgen.Session('PXI1Slot16', [0,1]) as dev2:
    dev_list = [dev1, dev2]
    for dev in dev_list:
        dev.start_trigger_type = nifgen.StartTriggerType.TRIG_NONE
        dev.output_mode = nifgen.OutputMode.FUNC
        dev.configure_standard_waveform(waveform=nifgen.Waveform.SINE, amplitude=1.0, frequency=100000, dc_offset=0.0, start_phase=0.0)
        
    dev1.start_trigger_type = nifgen.StartTriggerType.SOFTWARE_EDGE
    
    dev2.start_trigger_type = nifgen.StartTriggerType.DIGITAL_EDGE
    dev2.digital_edge_start_trigger_edge = nifgen.StartTriggerDigitalEdgeEdge.RISING
    dev2.digital_edge_start_trigger_source = '/PXI1Slot12/0/StartTrigger'
    
    with dev2.initiate():   
        with dev1.initiate():
            dev1.send_software_edge_trigger(nifgen.Trigger.START)
            time.sleep(1)`

@mitchellkelley

PR #857, opened 2018-05-16
Add and use decorator to use IVI session locking

texasaggie97:
* Requires #846 
* Add ivi_synchronized decorator that will wrap the method call with the session lock 
* Use ivi session lock for all methods except init and close

* Unit
* System


src/nidcpower/templates/session.py/fancy_fetch.py.mako
```diff
@@ -19,6 +19,9 @@
     else:
         raise ValueError('Only fetch_multiple and measure_multiple are supported. Got {0}'.format(f['python_name']))
 %>\
+% if f['use_session_lock']:
```
marcoskirsch: I dislike the fact that we use two different terms for the same thing.
Sometimes we call it "session lock" and sometimes "IVI lock". I don't think it's going to be obvious to all contributors that these refer to the same thing.

Probably not worth changing right now, and I don't have a better name for the decorator.

src/niscope/templates/session.py/get_equalization_filter_coefficients.py.mako
```diff
@@ -5,6 +5,9 @@
 
     suffix = method_template['method_python_name_suffix']
 %>\
+% if f['use_session_lock']:
```
marcoskirsch: These three lines are in (almost?) every method template. Can't we do this at the session level, before we render the method template?

texasaggie97: Moved check into session.py.mako (non nimodinst)

This also allowed removing per driver `use_session_lock`. Always set to `True` unless already set. nimodinst session.py.mako doesn't use it and never add locking.

build/helper/metadata_add_all.py
```diff
@@ -270,6 +270,14 @@ def _add_is_repeated_capability(parameter):
         parameter['is_repeated_capability'] = parameter['name'] in _repeated_capability_parameter_names
 
 
+def _add_use_session_lock(f, config):
```
marcoskirsch: `config` not used.

PR #871, opened 2018-05-22
nifgen combine named and non-named waveform methods

### LINKED ISSUE #862: nifgen methods that have "waveform handle" and "waveform name" should use fancy wrapper
For `nifgen.Session.write_waveform`, we created a _Fancy Wrapper_ in which a parameter is either the waveform handle or the waveform name.

We found two instances that could and should do the same for consistency.

* `set_waveform_next_write_position` and `set_named_waveform_next_write_position`
* `clear_arb_waveform` and `delete_named_waveform`

The `next_write_position` methods are very niche, so we could just stop code-generating until a compelling use case arises. If that happens, we should make sure it's tracked via an issue.

texasaggie97:
* Combines methods that use `waveform_handle` or `waveform_name` into a single method
  * Internally calls the correct method
* `set_waveform_next_write_position()` and `set_named_waveform_next_write_position()` becomes `set_next_write_position()`
* `clear_arb_waveform()` and `delete_named_waveform()` becomes `delete_waveform()`
* Fix #862 

* Unit
* System

### DIFF
--- a/docs/nifgen/class.rst
+++ b/docs/nifgen/class.rst
@@ -396,8 +396,6 @@ nifgen.Session
     +-----------------------------------------------------+
     | :py:func:`clear_arb_sequence`                       |
     +-----------------------------------------------------+
-    | :py:func:`clear_arb_waveform`                       |
-    +-----------------------------------------------------+
     | :py:func:`clear_freq_list`                          |
     +-----------------------------------------------------+
     | :py:func:`clear_user_standard_waveform`             |
@@ -434,9 +432,9 @@ nifgen.Session
     +-----------------------------------------------------+
     | :py:func:`define_user_standard_waveform`            |
     +-----------------------------------------------------+
-    | :py:func:`delete_named_waveform`                    |
-    +-----------------------------------------------------+
     | :py:func:`delete_script`                            |
+    +-----------------------------------------------------+
+    | :py:func:`delete_waveform`                          |
     +-----------------------------------------------------+
     | :py:func:`disable`                                  |
     +-----------------------------------------------------+
@@ -480,9 +478,7 @@ nifgen.Session
     +-----------------------------------------------------+
     | :py:func:`send_software_edge_trigger`               |
     +-----------------------------------------------------+
-    | :py:func:`set_named_waveform_next_write_position`   |
-    +-----------------------------------------------------+
-    | :py:func:`set_waveform_next_write_position`         |
+    | :py:func:`set_next_write_position`                  |
     +-----------------------------------------------------+
     | :py:func:`unlock`                                   |
     +-----------------------------------------------------+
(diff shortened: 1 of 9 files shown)


docs/nifgen/class.rst
```diff
@@ -5788,30 +5732,33 @@ delete_named_waveform
...
+delete_waveform
+~~~~~~~~~~~~~~~
 
     .. py:currentmodule:: nifgen.Session
 
-    .. py:method:: delete_script(script_name)
+    .. py:method:: delete_waveform(waveform_name_or_handle)
 
-            Deletes the specified script from onboard memory.
+            Removes a previously created arbitrary waveform from the signal generator memory and invalidates the waveform handle.
```
marcoskirsch: `and invalidates the waveform handle` only applies when a handle is used to call. I think you can just remove this part of the sentence.

docs/nifgen/class.rst
```diff
@@ -6634,11 +6508,10 @@ set_waveform_next_write_position
             a call to the nifgen_AllocateWaveform method or one of the following
             niFgen CreateWaveform methods:
 
-            -  nifgen_CreateWaveformF64
-            -  nifgen_CreateWaveformI16
-            -  nifgen_CreateWaveformFromFileI16
-            -  nifgen_CreateWaveformFromFileF64
-            -  nifgen_CreateWaveformFromFileHWS
+            -  :py:meth:`nifgen.Session.create_waveform`
```
marcoskirsch: If you have control over this, let's remove the duplication.

generated/nifgen/_library.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, ctypes_library):
         self.niFgen_DefineUserStandardWaveform_cfunc = None
         self.niFgen_DeleteNamedWaveform_cfunc = None
         self.niFgen_DeleteScript_cfunc = None
+        self.niFgen_DeleteWaveformDispatch_cfunc = None
```
marcoskirsch: These shouldn't be in the _library.py.

generated/nifgen/session.py
```diff
@@ -1969,6 +1969,31 @@ def delete_script(self, script_name):
...
+        Tip:
+        This method requires repeated capabilities (usually channels). If called directly on the
+        nifgen.Session object, then the method will use all repeated capabilities in the session.
+        You can specify a subset of repeated capabilities using the Python index notation on an
+        nifgen.Session instance, and calling this method on the result.:
+
+            session.channels['0,1'].delete_waveform(waveform_name_or_handle)
+
+        Args:
+            waveform_name_or_handle (str or int): The name (str) or handle (int) of an arbitrary waveform previously allocated with allocate_named_waveform or allocate_waveform.
```
marcoskirsch: there's also create_waveform...

generated/nifgen/session.py
```diff
@@ -1969,6 +1969,31 @@ def delete_script(self, script_name):
...
+
+        '''
+        if isinstance(waveform_name_or_handle, str):
```
marcoskirsch: Why not try using `try`/`except`? We've learned it's more Pythonic.
ctypes will complain, is my guess, based on the order in which you try.

generated/nifgen/_library.py
texasaggie97: Removed. Also removed WriteWaveformDispatcher since it shouldn't be there either.

generated/nifgen/session.py
texasaggie97: Added to all

generated/nifgen/session.py
texasaggie97: Do I change write_waveform too?

generated/nifgen/session.py
```diff
@@ -1969,6 +1969,31 @@ def delete_script(self, script_name):
...
+    @ivi_synchronized
+    def delete_waveform(self, waveform_name_or_handle):
+        '''delete_waveform
+
+        Removes a previously created arbitrary waveform from the signal generator memory.
+
+        Note: The signal generator must not be in the Generating state when you call this method.
+
+        Tip:
+        This method requires repeated capabilities (usually channels). If called directly on the
```
marcoskirsch: A lot of methods in NI-FGEN have channel inputs that are there for future proofing or are used +90% of the time with empty string.

Documentation bolerplate says "requires". This makes me uneasy.

This is food for thought outside of this PR.

PR #873, opened 2018-05-22
Fix documentation links in the case of incorrect case

### LINKED ISSUE #861: Bad function names in nifgen documentation
Some method names in the documentation (docstring, readthedocs) shows C function names instead of the Python names.

In reality, this happens due to the function names being misspelled, so our code-generator doesn't replace them properly.

We can hack something to "fix" up the names, or we can get NI to fix the source of the problem (NI Internal bug 699079).

Examples:

*   nifgen_AllocateWaveform
*   nifgen_CreateWaveformF64
*   nifgen_CreateWaveformI16
*   nifgen_CreateWaveformFromFileI16
*   nifgen_CreateWaveformFromFileF64
*   nifgen_CreateWaveformFromFileHWS

texasaggie97:
* Also do a lower case check for driver name when looking for functions to link to

* Fix #861 

### DIFF
--- a/generated/niscope/session.py
+++ b/generated/niscope/session.py
@@ -1114,7 +1114,7 @@ class _SessionBase(object):
     Timebase Rate <pSampleClockTimebaseRate.html>`__ and the actual
     sample rate, which can be higher. This property can be used in
     conjunction with the `Sample Clock Timebase Divisor
-    Property <pniscope_SampleClockTimebaseDivisor.html>`__.
+    Property <pSampleClockTimebaseDivisor.html>`__.
     Some devices use multiple ADCs to sample the same channel at an
     effective sample rate that is greater than the specified clock rate.
     When providing an external sample clock use this property to indicate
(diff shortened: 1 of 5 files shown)


generated/niscope/session.py
```diff
@@ -1114,7 +1114,7 @@ class _SessionBase(object):
     Timebase Rate <pSampleClockTimebaseRate.html>`__ and the actual
     sample rate, which can be higher. This property can be used in
     conjunction with the `Sample Clock Timebase Divisor
-    Property <pniscope_SampleClockTimebaseDivisor.html>`__.
```
texasaggie97: This is the only case I see where it incorrectly tries to match and create a link. Is this a showstopper for this PR?

marcoskirsch: No, it fixes more than it breaks so leaves us in better shape.
Why is this happening? Is it due to bad source data?

Issue #855, opened 2018-05-14
Does fetch_into method does not support the acqusition using 2 channels?


swapnil15689:
Hello,
I am using a PXI 5922 niscope with 8MB onboard memory per channel. I want to read the continuous samples from two channels for a longer time (so only one record but very long); so I use the fetch_into method.

My code works perfectly fine for single channel for any acquisition freq  but no matter what, it does not work for 2 channels and I end up with the error saying 'The requested data has been overwritten in memory so it is no longer available for fetching.'
The code follows:

``` python
scope = niscope.Session("Dev1")
scope.configure_horizontal_timing(50000,131072,0.0,1,True)
scope.channels['0,1'].configure_vertical(5.0, coupling=niscope.VerticalCoupling.DC,enabled=True)
scope.allow_more_records_than_memory=True
scope.initiate()
wfm=np.ndarray((4194304),dtype=np.float64)
wfm_info=scope.channels['0,1'].fetch_into(wfm,relative_to=FetchRelativeTo.READ_POINTER,offset=0,record_number=0, num_records=1,timeout=datetime.timedelta(seconds=-1.0))
scope.close()
```

The above code works perfectly fine for single channel. Is there something which I am missing..?

Thanks,
Swapnil

marcoskirsch:
* What version of the driver runtime are you using?
* What version of niscope Python module are you using?
* What kind of system are you running this on?

Thanks

swapnil15689:
I am using:
runtime 16.0
niscope python 0.8.0
Chassis: NI PXle-1071; Controller: NI PXle-8820; Windows 7 64 bits

Thanks

marcoskirsch:
Please try two things:

1. Swap the line in which you create the `np.ndarray` with the line in which you `initiate()`. This way the device is not acquiring while the memory is getting allocated, wasting precious on-board memory.
2. Try decreasing your sample rate to half or even less if needed. Your program may work with one channel, but maybe your system (2.2 GHz Intel Celeron 1020E) may truly not be able to keep up with twice the amount of data. If that's the case, then we have ruled out bugs and we can look into other ways to modify the program so that it can keep up.

swapnil15689:
1. I tried the first approach you suggested; but it did not work; I got the same error.

2. In fact, 50kHz is the minimum acquisition frequency for PXI 5922. For single channel it works fine for 50kHz and also even for 15MHz which is the maximum limit for acquisition frequency with PXI 5922.

Thanks

[...]

PR #880, opened 2018-05-30
Add niscope fetch forever example

texasaggie97:
* Adds example that shows how to effectively fetch forever in niscope 
* Useful for #855 

* Unit
* System


texasaggie97:
I cannot figure out what is wrong when testing the new example. It works from the command line.

marcoskirsch:
> I cannot figure out what is wrong when testing the new example. It works from the command line.

`NISCOPE_ERROR_NOT_A_SCOPE`
When you test it in the command line, are you testing on nimi-bot and using the same resource name?

texasaggie97:
It fails both on nimibot and on my test machine with this error:
```
src\niscope\examples\niscope_fetch_forever.py:69: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src\niscope\examples\niscope_fetch_forever.py:55: in _main
    example(args.resource_name, args.channels, args.option_string, args.time, args.voltage, args.sample_rate, args.samples_per_fetch)
src\niscope\examples\niscope_fetch_forever.py:23: in example
    with niscope.Session(resource_name) as session:
.tox\64\py36-system_tests\lib\site-packages\niscope\session.py:3818: in __init__
    self._vi = self._init_with_options(resource_name, id_query, reset_device, options)
.tox\64\py36-system_tests\lib\site-packages\niscope\session.py:4642: in _init_with_options
.tox\64\py36-system_tests\lib\site-packages\niscope\errors.py:92: in handle_error
    description = session._get_error_description(code)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = niscope.Session(repeated_capability_list=[], vi=None, library=None, encoding=None)
error_code = -1074118622

    def _get_error_description(self, error_code):
        '''_get_error_description
    
            Returns the error description.
            '''
        try:
            _, error_string = self._get_error()
            return error_string
        except errors.Error:
            pass
    
        try:
            '''
                It is expected for _get_error to raise when the session is invalid
                (IVI spec requires GetError to fail).
                Use _error_message instead. It doesn't require a session.
                '''
>           error_string = self._error_message(error_code)
E           AttributeError: 'Session' object has no attribute '_error_message'
```

Where are you getting `NISCOPE_ERROR_NOT_A_SCOPE`?

I am using the same option string to simulate the same device as the other scope examples.

texasaggie97:
Actually, it looks like maybe niscope really is missing `_error_message`?

texasaggie97:
`niscope` `_error_message` added in #885

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### NI-SCOPE
     * #### Added
+        * fetch forever example
```
marcoskirsch: Might as well use proper name `niscope_fetch_forever.py` so there's no ambiguity.

src/niscope/examples/niscope_fetch_forever.py
```diff
@@ -0,0 +1,74 @@
...
+import niscope
+import numpy as np
+import pprint
+import sys
+
+
+pp = pprint.PrettyPrinter(indent=4, width=80)
+
+
+# This example is based on the C FetchForever example
```
marcoskirsch: I don't think this is necessary. If NI-SCOPE modifies or renames or stops shipping that example, the comment will be wrong.

src/niscope/examples/niscope_fetch_forever.py
```diff
@@ -0,0 +1,74 @@
...
+def example(resource_name, channels, options, total_acquisition_time_in_seconds, voltage, sample_rate_in_hz, samples_per_fetch):
+    total_samples = int(total_acquisition_time_in_seconds * sample_rate_in_hz)
+    channel_list = channels.split(',')  # We need channels as a list
```
marcoskirsch: Can the user pass `channels` as empty string to signify "all"?
Can the user pass `channels` using ranges, such as "0-4"?
What if the user passes `channels` but adds whitespace such ass "0, 1, 2, 3"?

Seems like those will make this fail in unexpected ways. Why do the channels need to be passed in at all? Can we assume "all channels" and get those from the session instead?

src/niscope/examples/niscope_fetch_forever.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # 1. Creating numpy arrays
```
marcoskirsch: * I suggest using array.array so you don't need to depend on numpy.
* Explain why we do it: As an optimization, so we can use `fetch_into` which allows us to re-use the buffers and save a reallocation per function call.

src/niscope/examples/niscope_fetch_forever.py
```diff
@@ -0,0 +1,74 @@
...
+    waveforms = [np.ndarray(total_samples, dtype=np.float64) for c in channel_list]
+
+    # 2. Opening session
+    with niscope.Session(resource_name=resource_name, options=options) as session:
+        # 3. Configuring
+        session.configure_horizontal_timing(min_sample_rate=sample_rate_in_hz, min_num_pts=1, ref_position=0.0, num_records=1, enforce_realtime=True)
+        session.channels[channel_list].configure_vertical(voltage, coupling=niscope.VerticalCoupling.DC, enabled=True)
+        # We configure for a software trigger and then never send it
```
marcoskirsch: Can you explain why?

src/niscope/examples/niscope_fetch_forever.py
```diff
@@ -0,0 +1,74 @@
...
+        session.configure_trigger_software()
+        current_pos = 0
+        # 4. initating
+        with session.initiate():
+            while current_pos < total_samples:
+                # We fetch each channel at a time so we don't have to de-interleave afterwards
+                # We do not keep the wfm_info returned from fetch_into
+                for c, wfm in zip(channel_list, waveforms):
```
marcoskirsch: I'd expand `c` -> `channel`, `wfm` -> `waveform`.
Consistent with `channel_list` and `waveforms`.

src/niscope/examples/niscope_fetch_forever.py
```diff
@@ -0,0 +1,74 @@
...
+                    # 5. fetching
+                    session.channels[c].fetch_into(wfm[current_pos:current_pos + samples_per_fetch], relative_to=niscope.FetchRelativeTo.READ_POINTER,
+                                                   offset=0, record_number=0, num_records=1, timeout=datetime.timedelta(seconds=5.0))
+                current_pos += samples_per_fetch
+
+        # 6 fetch complete
+    # 7 session closed
```
marcoskirsch: I think it's ok to take out most of these comments. Several are just oddly placed and several mostly paraphrase the code.
I assume the C example was written this way?

CHANGELOG.md
texasaggie97: Changed

src/niscope/examples/niscope_fetch_forever.py
texasaggie97: Removed

src/niscope/examples/niscope_fetch_forever.py
texasaggie97: Other examples allow passing in channel list

Removed

src/niscope/examples/niscope_fetch_forever.py
texasaggie97: * Fetch into doesn't support `array.array`
* Added overview

src/niscope/examples/niscope_fetch_forever.py
texasaggie97: Because that is how they did it in the C example. Other than that, no I cannot explain why.

src/niscope/examples/niscope_fetch_forever.py
texasaggie97: They were originally `print()` so I could track what was going on.

src/niscope/examples/niscope_fetch_forever.py
texasaggie97: Changed

src/niscope/examples/niscope_fetch_forever.py
marcoskirsch: Does it still work if you stop doing it?

src/niscope/examples/niscope_fetch_forever.py
marcoskirsch: I think it's safe to delete. The code is really clean and self-explanatory.

src/niscope/examples/niscope_fetch_forever.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+# We use fetch_into which allows us to allocate a single buffer per channel and "fetch into" it a section at a time without having to
+# reconstruct the waveform once we are done
+def example(resource_name, options, total_acquisition_time_in_seconds, voltage, sample_rate_in_hz, samples_per_fetch):
+    total_samples = int(total_acquisition_time_in_seconds * sample_rate_in_hz)
+    # 1. Opening session
+    with niscope.Session(resource_name=resource_name, options=options) as session:
+        # We will acquire on all channels of the device
+        channel_list = [c for c in range(session.channel_count)]  # Need an actual list and not a range
```
marcoskirsch: neat!

src/niscope/examples/niscope_fetch_forever.py
texasaggie97: I changed the comment to match what was in the C example. It looks like it is required.

PR #883, opened 2018-05-30
Disallow repeated capability on session

### LINKED ISSUE #853: Directly indexing the session rather than the container works.
#718 introduced containers for repeated capabilities.

Before:
```
session[1].foo
session["MarkerEvent2"].bar
```

After:
```
session.channels[1].foo
session.marker_events[2].bar
```

It appears code written in the style of "Before" still works.

texasaggie97:
* Disallow repeated capability on session
* Explicitly disallow using a repeated capability on Session. `session[0].vertical_range = 1.0` will no longer work. Instead use `session.channels[0].vertical_range = 1.0`
* Fix #853 

* Unit
* System

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
     * #### Removed
+        * Explicitly disallow using a repeated capability on Session. `session['0'].vertical_range = 1.0` will no longer work. Instead use `session.channels['0'].vertical_range = 1.0` - [#853](https://github.com/ni/nimi-python/issues/853)
 * ### NI-DMM
     * #### Added
     * #### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
     * #### Removed
-        * Explicitly disallow using a repeated capability on Session. `session['0'].vertical_range = 1.0` will no longer work. Instead use `session.channels['0'].vertical_range = 1.0` - [#853](https://github.com/ni/nimi-python/issues/853)
+        * Explicitly disallow using a repeated capability on Session. `session[0].vertical_range = 1.0` will no longer work. Instead use `session.channels[0].vertical_range = 1.0` - [#853](https://github.com/ni/nimi-python/issues/853)
 * ### NI-DMM
     * #### Added
     * #### Changed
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
     * #### Removed
+        * Explicitly disallow using a repeated capability on Session. `session['0'].vertical_range = 1.0` will no longer work. Instead use `session.channels['0'].vertical_range = 1.0` - [#853](https://github.com/ni/nimi-python/issues/853)
```
marcoskirsch: I think in example snippets we should show the use of indices with numbers.
Meaning: `session[0].vertical_range = 1.0` as opposed to `session['0'].vertical_range = 1.0`

build/templates/session.py.mako
```diff
@@ -190,6 +190,9 @@ constructor_params = helper.filter_parameters(init_function, helper.ParameterUsa
             raise AttributeError("'{0}' object has no attribute '{1}'".format(type(self).__name__, key))
         object.__setattr__(self, key, value)
 
+    def __getitem__(self, key):
```
marcoskirsch: Is this needed? When you try to index a class that doesn't support it, the error is:
    
    TypeError: 'Foo' object does not support indexing

We wouldn't want this to be different. Let's let Python do its thing.

If you think the text in the exception is a good enough reason (I'm not convinced) then let's make it really useful:

* raise TypeError
* text should say `'Session' object does not support indexing. XXX
* XXX should be something that talks about the repeated capability containers for the class. The magic of codegen! (or we could at runtime figure it out)

Regardless, I'm not sure it's worth it.

generated/nifake/unit_tests/test_session.py
```diff
@@ -1211,6 +1211,14 @@ def test_matcher_prints(self):
         assert _matchers.CustomTypeMatcher(nifake.custom_struct, nifake.custom_struct(cs[0])).__repr__() == "CustomTypeMatcher(<class 'nifake.custom_struct.custom_struct'>, custom_struct(data=None, struct_int=42, struct_double=4.2))"
         assert _matchers.CustomTypeBufferMatcher(nifake.custom_struct, cs_ctype).__repr__() == "CustomTypeBufferMatcher(<class 'nifake.custom_struct.custom_struct'>, [custom_struct(data=None, struct_int=42, struct_double=4.2), custom_struct(data=None, struct_int=43, struct_double=4.3), custom_struct(data=None, struct_int=42, struct_double=4.3)])"
 
+    def test_channel_on_session(self):
+        with nifake.Session('dev1') as session:
+            try:
+                session['100'].read_write_double = 5.0
+                assert False
+            except AttributeError:
```
marcoskirsch: `TypeError`

PR #889, opened 2018-06-05
Clean up repeated capabilities tip

### LINKED ISSUE #854: Documentation shows outdated indexing of session in several places
In #718 things changed. We didn't update everything:

All drivers have this mistake:
```
docs/niscope/attributes.rst:            session['0,1'].accessory_gain = var
```
We should think carefully what we want to do.

niscope has:
```
docs/niscope/functions.rst:            wfm_info = session['0,1'].fetch_into(num_samples, waveform, timeout=5.0)
```

I think that's it.

texasaggie97:
* Add repeated capability container information to nifgen (only driver that currently has non 'channels' containers)
* Add default repeated capability container information of 'channels' where appropriate
* Use repeated capability container information in tips
* Fix #854

* Visual inspection of generated html and session files

### DIFF
--- a/build/helper/documentation_helper.py
+++ b/build/helper/documentation_helper.py
@@ -473,7 +473,7 @@ def get_function_rst(function, method_template, numpy, config, indent=0):
         session_declaration = ParameterUsageOptions.SESSION_NUMPY_INTO_METHOD_DECLARATION
 
     if function['has_repeated_capability'] is True:
-        function['documentation']['tip'] = rep_cap_method_desc_rst.format(config['module_name'], function['python_name'], get_params_snippet(function, session_method))
+        function['documentation']['tip'] = rep_cap_method_desc_rst.format(config['module_name'], function['repeated_capability_type'], function['python_name'], get_params_snippet(function, session_method))
 
     rst = '.. py:method:: ' + function['python_name'] + suffix + '('
     rst += get_params_snippet(function, session_method) + ')'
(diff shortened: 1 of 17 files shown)


docs/niswitch/class.rst
```diff
@@ -1183,15 +1183,15 @@ max_carry_dc_current
...
+            niswitch.Session object, then the set/get will use all repeated capabilities in the session.
             You can specify a subset of repeated capabilities using the Python index notation on an
-            max_carry_dc_current.Session instance, and calling set/get value on the result.:
+            niswitch.Session repeated capabilities container, and calling set/get value on the result.:
 
             .. code:: python
 
-                session['0,1'].max_carry_dc_current = var
-                var = session['0,1'].max_carry_dc_current
+                session.channels[0,1].max_carry_dc_current = var
```
marcoskirsch: On `niswitch`, it's pretty rare for channels to be 0,1... usually they are c0, r1, com2.
Most of these would error if called with two channels.

marcoskirsch: Furthermore, a lot of these are read-only properties.
I think we should merge this PR but open an issue about this additional problem.

texasaggie97: I do not know how to fix that, unless we add even more metadata somewhere.

texasaggie97: True. I was thinking of the `0,1` part. I can fix the r/w/rw pretty easy.

texasaggie97: Add r/w/rw specific tips

build/helper/documentation_helper.py
```diff
@@ -121,6 +127,22 @@ def get_rst_admonition_snippet(node, admonition, d, config, indent=0):
...
 
+def add_attribute_rep_cap_tip_rst(attr, config):
+    '''Add the appropriate (r/w/rw/none) rst formatted tip for an attribute'''
+    if 'repeated_capability_type' in attr:
+        if attr['access'] == 'read only':
+            tip = rep_cap_attr_desc_rst_r
+        elif attr['access'] == 'write only':
+            tip = rep_cap_attr_desc_rst_w
+        else:
+            tip = rep_cap_attr_desc_rst_rw
```
marcoskirsch: assert it's read/write

texasaggie97: Added

Issue #514, opened 2017-10-20
Need support for complex numbers in our NI-FGEN and NI-SCOPE APIs


Fladolcetta:
NI-FGEN and NI-SCOPE both support generating and acquiring complex waveforms. For example, NI-5541 and NI-5442 have Onboard Signal Processing that upconverts a complex waveform and generates it. The APIs use NIComplexI16 and NIComplexF64 to represent complex numbers.

nimi-python doesn't support those types.

marcoskirsch:
There are challenges and API design questions on how to best support those types (NIComplexF64 and NIComplexI16) in our Python APIs.

We estimate that only a very small number of clients use complex waveforms with NI-FGEN and NI-SCOPE – most of those customers use NI-RFSG and NI-RFSA which aren't in nimi-python and aren't planned for the [0.9 release](https://github.com/ni/nimi-python/milestone/2).

Given this, and the sheer amount of work remaining, there's a pretty good change that we won't get to this.

Fladolcetta:
Closing item until demand is requested for Complex numbers.

marcoskirsch:
When we add this to NI-FGEN, we should keep the same signature for the public method and dispatch to the proper function internally if possible, rather than adding new public methods.

MrDuff8:
We expect this is a low-priority item, and we haven't heard requests for it. It's very rare for this to be used on its own without an upconverter or downconverter, which we don't offer support for in python. Please comment if this is something your application needs.

Issue #189, opened 2017-08-28
Support complex waveform datatype


Fladolcetta:

Fladolcetta:
Closed as duplicate of #514

PR #891, opened 2018-06-05
Disable nifgen get_fir_filter_coefficients()

### LINKED ISSUE #535: Generated documentation for return parameters is incomplete
Example: NI-FGEN function:

```
    def get_fir_filter_coefficients(self):
        '''get_fir_filter_coefficients

        | Returns the FIR filter coefficients used by the onboard signal
          processing block. These coefficients are determined by NI-FGEN and
          based on the FIR filter type and corresponding attribute (Alpha,
          Passband, BT) unless you are using the custom filter. If you are using
          a custom filter, the coefficients returned are those set with the
          nifgen_ConfigureCustomFIRFilterCoefficients function coerced to the
          quantized values used by the device.
        | To use this function, first call an instance of the
          get_fir_filter_coefficients function with the
          **coefficientsArray** parameter set to VI_NULL. Calling the function
          in this state returns the current size of the **coefficientsArray** as
          the value of the **numberOfCoefficientsRead** parameter. Create an
          array of this size, and call the get_fir_filter_coefficients
          function a second time, passing the new array as the
          **coefficientsArray** parameter and the size as the **arraySize**
          parameter. This second function call populates the array with the FIR
          filter coefficients.
        | Refer to the FIR Filter topic for your device in the *NI Signal
          Generators Help* for more information about FIR filter coefficients.
          This function is supported only for the NI 5441.
        | **Default Value**: None

        Tip:
        This method requires repeated capabilities (usually channels). If called directly on the
        nifgen.Session object, then the method will use all repeated capabilities in the session.
        nifgen.Session instance, and calling this method on the result.:

            session['0,1'].get_fir_filter_coefficients()

        Args:
            array_size (int): Specifies the size of the coefficient array

        Returns:
            number_of_coefficients_read (int): Specifies the array of data containing the number of coefficients you
                want to read.
        '''
```

It should have an additional "coefficients_array" return value, before number_of_coefficients_read.

### LINKED ISSUE #596: Return values of nifgen_GetFIRFilterCoefficients seem to be incorrect
While writing the test given below, no matter what value was given to coeff_array, the value returned by session.get_fir_filter_coefficients() is always ([], 95). This is different from the behavior on LabVIEW and seems incorrect.

def test_fir_filter_coefficients():
    with nifgen.Session('', False, 'Simulate=1, DriverSetup=Model:5441;BoardType:PXI') as session:
        coeff_array = [1, 0, -1]
        session.configure_custom_fir_filter_coefficients(coeff_array)
        session.commit()
        array, size = session.get_fir_filter_coefficients()
        assert size == len(coeff_array)

texasaggie97:
* Remove `get_fir_filter_coefficients()` because it only applies for OSP and OSP is not included in the Python API yet in general
* Keep all created files and metadata in case we need to enable again in the future
  * Don't want to have to figure this out again.
* ~~Fixed nifgen `get_fir_filter_coefficients()`~~
  * ~~Create specific implementation instead of relying on code generator~~
    * ~~Not really ivi-dance because the size is returned in an out parameter instead of the return value (error)~~
    * ~~Not worth trying to make the code generator support this use case for this one function~~
  * ~~Don't use code generator for documentation~~
* Fix #535 
* Fix #596 
* We no longer care about #534 

* Unit
* System


docs/nifgen/class.rst
```diff
@@ -5656,66 +5656,57 @@ get_ext_cal_recommended_interval
...
-
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    .. py:currentmodule:: nifgen.Session
+
+    .. py:method:: get_fir_filter_coefficients()
+
+            | Returns the FIR filter coefficients used by the onboard signal
+              processing block. These coefficients are determined by NI-FGEN and
+              based on the FIR filter type and corresponding property (Alpha,
```
marcoskirsch: Since we have control over this text now, we _could_ type the actual property name so they get linked).

generated/nifgen/unit_tests/_mock_helper.py
```diff
@@ -105,8 +105,6 @@ def __init__(self):
         self._defaults['GetExtCalRecommendedInterval']['Months'] = None
         self._defaults['GetFIRFilterCoefficients'] = {}
         self._defaults['GetFIRFilterCoefficients']['return'] = 0
-        self._defaults['GetFIRFilterCoefficients']['numberOfCoefficientsRead'] = None
```
marcoskirsch: If this mocks the C function, why did it change at all?

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -388,16 +388,16 @@ def test_user_standard_waveform(session):
...
 
 
-# TODO(bhaswath): Doesn't work, issue #596
-'''
 def test_fir_filter_coefficients():
     with nifgen.Session('', '0', False, 'Simulate=1, DriverSetup=Model:5441;BoardType:PXI') as session:
-        coeff_array = [1, 0, -1]
+        coeff_array = [0 for i in range(95)]
+        coeff_array[0] = -1.0
+        coeff_array[2] = 1.0
```
marcoskirsch: This should error because the device expected symmetrical coefficients.
I think you need to also enable OSP so that the OSP configuration gets validated.

docs/nifgen/class.rst
```diff
@@ -1,10 +1,10 @@
 nifgen.Session
-==============
-
-.. py:module:: nifgen
-
-.. py:class:: Session(self, resource_name, channel_name=None, reset_device=False, options={})
-
+==============
```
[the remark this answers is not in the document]
texasaggie97: I am not sure. Looking into it but I don't see anything obvious

src/nifgen/system_tests/test_system_nifgen.py
texasaggie97: Test disabled since OSP disabled, can't check this.

marcoskirsch:
Do we have an issue about adding OSP to NI-FGEN?
If we don't we should open it and reference all those Issues that are getting closed now but would become a problem again when we start tackling OSP.

src/nifgen/templates/functions.rst/get_fir_filter_coefficients.rst.mako
```diff
@@ -0,0 +1,42 @@
+<%page args="function, config, method_template, indent"/>\
```
marcoskirsch: Are you keeping these on purpose because they may be useful in the future?
I'm not sure what the proper way of handling this sort of thing is.

texasaggie97: Yes, I am keeping this on purpose. I mentioned that in the PR description. I don't want to lose this knowledge since it isn't obvious how this needs to work.

texasaggie97:
We have an issue about complex datatypes (#189 and #514) and OSP only applies to complex data.

src/nifgen/templates/functions.rst/get_fir_filter_coefficients.rst.mako
marcoskirsch: It just seems like dead code so there's a risk of it being removed in the future too.

How about removing it but in the issue recording the last commit that contained it to show a solution?

marcoskirsch:
> Comments addressed?

I think we should leverage the Git repo, rather than have dead code in master.
This can be done by having a branch with the files (I don't like this idea) or simply documenting in the related issues, even in the test comment, the Git ID where the fix existed.

texasaggie97:
We have to keep the branch around if we want to be able to reference the commit hash. Once we delete the branch, that deleted all the commits as well.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

marcoskirsch:
What if we commit to master, then remove?
How is this typically done?

texasaggie97:
Add more comments about how these are not used until OSP is supported

PR #913, opened 2018-07-23
Add support for NI Switch Executive

Fladolcetta:

This adds python APIs for NI Switch Executive

Ran TOX, ran system tests, ran examples with and without parameters


CHANGELOG.md
```diff
@@ -43,6 +43,11 @@ All notable changes to this project will be documented in this file.
     * #### Added
     * #### Changed
     * #### Removed
+* ### NI Switch Executive
```
texasaggie97: Add this to the template at the bottom of the file.

generated/nise/session.py
```diff
@@ -0,0 +1,871 @@
...
+
+        '''
+        vi_ctype = _visatype.ViSession(self._vi)  # case S110
+        route_spec_ctype = (_visatype.ViChar * route_spec_size[0])()  # case C080
+        route_spec_size_ctype = get_ctypes_pointer_for_buffer(value=route_spec_size, library_type=_visatype.ViInt32)  # case B550
+        error_code = self._library.niSE_GetAllConnections(vi_ctype, route_spec_ctype, route_spec_size_ctype)
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
+        return route_spec_ctype.value.decode(self._encoding)
+
+    def get_ivi_device_session(self, ivi_logical_name):
```
texasaggie97: Do we need a way to instantiate a session object from an already open session? And then should this try to use that?

tools/build_release.py
```diff
@@ -96,7 +98,7 @@ def main():
 
     if args.upload:
         logging.info('Uploading to PyPI')
-        complete_twine_cmd = twine_cmd + ['upload', 'bin/nidcpower/dist/*', 'bin/nidmm/dist/*', 'bin/nimodinst/dist/*', 'bin/niswitch/dist/*', 'bin/nifgen/dist/*', 'bin/niscope/dist/*']
+        complete_twine_cmd = twine_cmd + ['upload', 'bin/nidcpower/dist/*', 'bin/nidmm/dist/*', 'bin/nimodinst/dist/*', 'bin/niswitch/dist/*', 'bin/nifgen/dist/*', 'bin/niscope/dist/*', 'bin/nise/dist/*']
```
texasaggie97: Did you check that `nise` was available on PyPI?

Fladolcetta: Yep! https://pypi.org/search/?q=nise

generated/nise/session.py
Fladolcetta: Since there is no IVI python API I will remove it

CHANGELOG.md
Fladolcetta: done

src/nise/metadata/config.py
```diff
@@ -0,0 +1,26 @@
+config = {
+    'metadata_version': '1.0',
+    'module_name': 'nise',
+    'module_version': '1.0.1.dev0',
```
texasaggie97: Do you really want to start with version 1.0.1? I think you might want to make the first release something like 0.9.0 or something just in case there are breaking changes required.

Fladolcetta: While I understand the reason that 0.9 signifies that we might break something, I don't like the idea of different versions for drivers and nimi-python. I like that nimi-python 1.0.1 supports nise 1.0.1.

 @marcoskirsch @texasaggie97  Thoughts?

marcoskirsch: While we may think of nimi-python as a whole, 99.99% of customers will download each module separately, think of them as distinct things, and will appreciate different versions.

Fladolcetta: Fixed!

docs/_static/nise_usage.inc
```diff
@@ -0,0 +1,13 @@
+Usage
+------
+
+The following is a basic example of using the **nise** module to open a session to a Switch Executive Virtual Device and connect a routegroup.
+
+.. code-block:: python
+
+    import nise
+    with nise.Session('SwitchExecutiveExample') as session:
```
marcoskirsch: Is "'SwitchExecutiveExample'" what we typically use in NISE examples? (LabVIEW, etc)

generated/nise/__init__.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
+# This file was generated
+
+from nise.enums import *          # noqa: F403,F401,H303
+from nise.errors import DriverWarning   # noqa: F401
+from nise.errors import Error     # noqa: F401
+from nise.session import Session  # noqa: F401
+
+
+def get_diagnostic_information():
```
marcoskirsch: Does this return sensical results for NISE?

generated/nise/session.py
```diff
@@ -0,0 +1,842 @@
...
+                syntax for **options** is a dictionary of properties with an assigned
+                value. For example:
+
+                { 'simulate': False }
+
+                You do not have to specify a value for all the properties. If you do not
+                specify a value for a property, the default value is used.
+
+                Advanced Example:
+                { 'simulate': True, 'driver_setup': { 'Model': '<model number>',  'BoardType': '<type>' } }
```
marcoskirsch: Do these apply to NISE?

generated/nise/session.py
```diff
@@ -0,0 +1,842 @@
...
+
+        self._is_frozen = True
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
+    def initiate(self):
```
marcoskirsch: Does this apply to NISE?

generated/nise/session.py
```diff
@@ -0,0 +1,842 @@
...
+                one of the following: NISE_VAL_EXPAND_TO_ROUTES (0) - expands the
+                route spec to routes. Converts route groups to their constituent routes.
+                NISE_VAL_EXPAND_TO_PATHS (1) - expands the route spec to paths.
+                Converts routes and route groups to their constituent square bracket
+                route spec strings. Example: [Dev1/c0->Dev1/r0->Dev1/c1]
+
+                Note:
+                One or more of the referenced values are not in the Python API for this driver. Enums that only define values, or represent True/False, have been removed.
+
+            expanded_route_spec_size (list of int): The routeSpecSize is an ViInt32 that is passed by reference into the
```
marcoskirsch: * This is not a list of int
* This should not be exposed in the Python API

src/nise/examples/nise_basic_example.py
```diff
@@ -0,0 +1,39 @@
...
+import nise
+import sys
+
+
+def example(virtual_device_name, connection):
+    print('Opening session to', virtual_device_name)
+    with nise.Session(virtual_device_name=virtual_device_name) as session:
+        print('Connecting', connection)
+        session.connect(connection)
+        print(connection, ' is now connected.')
```
marcoskirsch: I'd remove all prints but this "result" one.

src/nise/metadata/config.py
```diff
@@ -0,0 +1,26 @@
...
+    'metadata_version': '0.9',
+    'module_name': 'nise',
+    'module_version': '0.9.0.dev0',
+    'c_function_prefix': 'niSE_',
+    'driver_name': 'NI Switch Executive',
+    'session_class_description': 'An NI Switch Executive session',
+    'session_handle_parameter_name': 'vi',
+    'library_info':
+    {
+        'Windows': {
```
marcoskirsch: By not having Linux entries, users may get a misleading error. In other drivers we put something there even though the driver doesn't support Linux (i.e. nifgen). Please test it out at least.

src/nise/metadata/enums_addon.py
```diff
@@ -0,0 +1,139 @@
+# These dictionaries are applied to the generated enums dictionary at build time
```
marcoskirsch: What's up with this? Extraction didn't work so you had to type all this by hand?
Please:

* Make sure the names match the C NISE API
* Add comment to this file on why it's all written here and not in enums.py
* File internal bug report to NISE to fill out the proper source of metadata extraction.

Alternatively, we could source the enums as extracted in-house and export them just the same as we export the other ones. This one would just be hand-coded.

generated/nise/__init__.py
Fladolcetta: I didn't know this function existed.

It returns:

"{'os': {'name': 'Windows', 'version': '6.1.7601', 'bits': '64'}, 'python': {'version': '3.6.6 (v3.6.
6:4cf1f54eb7, Jun 27 2018, 03:37:03) [MSC v.1900 64 bit (AMD64)]', 'bits': '64', 'is_venv': False, 'packages': [{'name': 'wxpython', 'version': '4.0.3'}, {'name': 'virtualenv', 'version': '16.0.0'}, {'name': 'tox', 'version': '3.1.2'}, {'name': 'six', 'version': '1.11.0'}, {'name': 'setuptools', 'version': '39.0.1'}, {'name': 'pytz', 'version': '2018.5'}, {'name': 'python-dateutil', 'version': '2.7.3'}, {'name': 'pytest', 'version': '3.6.3'}, {'name': 'pypubsub', 'version': '4.0.0'}, {'name': 'pyparsing', 'version': '2.2.0'}, {'name': 'pyflakes', 'version': '1.6.0'}, {'name': 'pycodestyle', 'version': '2.3.1'}, {'name': 'py', 'version': '1.5.4'}, {'name': 'pluggy', 'version': '0.6.0'}, {'name': 'pip', 'version': '18.0'}, {'name': 'packaging', 'version': '17.1'}, {'name': 'numpy', 'version': '1.15.0'}, {'name': 'niswitch', 'version': '1.0.0'}, {'name': 'nise', 'version': '1.0.1.dev0'}, {'name': 'niscope', 'version': '1.0.1.dev0'}, {'name': 'nimodinst', 'version': '1.0.0'}, {'name': 'nidmm', 'version': '1.0.1.dev0'}, {'name': 'nidcpower', 'version': '1.0.1.dev0'}, {'name': 'more-itertools', 'version': '4.2.0'}, {'name': 'mccabe', 'version': '0.6.1'}, {'name': 'matplotlib', 'version': '2.2.3'}, {'name': 'kiwisolver', 'version': '1.0.1'}, {'name': 'flake8', 'version': '3.5.0'}, {'name': 'cycler', 'version': '0.10.0'}, {'name': 'colorama', 'version': '0.3.9'}, {'name': 'attrs', 'version': '18.1.0'}, {'name': 'atomicwrites', 'version': '1.1.5'}]}, 'driver': {'name': 'NI Switch Executive', 'version': 'Unknown'}, 'module': {'name': 'nise', 'version': '1.0.1.dev0'}}
"

README.rst
```diff
@@ -258,6 +259,45 @@ NI-SCOPE Python API Status
 
 
 
+NI Switch Executive Python API Status
+-------------------------------------
+
++-------------------------------+---------------------+
+| NI Switch Executive (nise)    |                     |
++===============================+=====================+
+| Driver Version Tested Against | 18.0.0              |
```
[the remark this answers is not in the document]
Fladolcetta: This is similar to ModInst and as discussed offline, we should keep it for consistency.

docs/_static/nise_usage.inc
Fladolcetta: That is the default virtual device automatically added to the system by default when you install Switch Executive.

generated/nise/__init__.py
Fladolcetta: Fixed

generated/nise/session.py
Fladolcetta: fixed

src/nise/examples/nise_basic_example.py
Fladolcetta: fixed

src/nise/metadata/config.py
Fladolcetta: fixed

generated/nise/session.py
Fladolcetta: In Switch Executive, for strings, you:
1. create a pointer 
2. Assign a value to that pointer with the size of the string buffer you have pre-allocated
3. Pass the pointer into the dll 
4. Read pointer containing the actual length of the string

This is similar to the IVI dance, except

1. The length of the string is not returned from the function. Switch Executive returns the error code always. If I used ivi-dance as a mechanism, then the string size is always 0, resulting in an empty string.
2. The direction is out instead of in

To resolve this for Switch Executive, I had to do the following.

- Change type from pointer to VIInt32 array of size 1 (that is essentially a pointer.)
- Set direction from out, to in (so both I and the dll could set the value, as not setting it would cause no string)
- Set default value of "Size" to the default used in examples,
- Use "Size" to allocated the size of the string buffer.

This enables the string to be read, and partially hidden from the customer (via defaults).

The other option is to create a completely new mechanism for filling buffers specific to Switch Executive.

Thoughts?

src/nise/metadata/enums_addon.py
Fladolcetta: Enums are extracted from attributes. 
There are no attributes in NI Switch Executive, therefore there are no extracted enums. 

There are no enum names to match to the C API, there were hand typed based on the LabVIEW API.

For other drivers, enums_addon.py was hand typed as well and was used for enums that are part of functions. The names for other drivers were also referencing the LabVIEW API. This is similar to other drivers.

The other drivers did not add info on why additional enums were needed. Should they have?

generated/nise/session.py
marcoskirsch: I think this is what we should do. But I don't think that should hold up this PR, rather it should be a separate discussion altogether. Please create an issue about it.

src/nise/metadata/enums_addon.py
marcoskirsch: > The other drivers did not add info on why additional enums were needed. Should they have?

I didn't remember this flaw in the extracted metadata... so I'd say that yes, it would probably be a helpful comment. Something along the lines of: "Extracted metadata is missing enums used only in functions. Those are added here manually."

PR #918, opened 2018-08-03
Add scripting example for nifgen

texasaggie97:
* Add example to show nifgen script

* Unit
* System


src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/python
```
marcoskirsch: do we need to add the utf8 thing to our examples?

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+import argparse
+import nifgen
+import numpy as np
```
marcoskirsch: I'm not a fan of having our examples depend on large third party dependencies.
If we're using bumpy and scipy for trivial things, I'd rather not use them.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+    return sinewave
+
+
+def generate_rampup():
+    ramp = np.arange(0, 0.5, 0.01)                # numpy.arange([start,]stop, [step,]dtype=None(
+    ramp_up = ramp.tolist()                       # List of Float
+    return ramp_up
+
+
+def generate_ramdown():
```
marcoskirsch: rampdown

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+        # 2-Configure Trigger:
+        # SOFTWARE TRIGGER: used in the script
+        session.script_triggers[0].script_trigger_type = nifgen.ScriptTriggerType.SOFTWARE_EDGE  # TRIG_NONE / DIGITAL_EDGE / DIGITAL_LEVEL / SOFTWARE_EDGE
+        session.script_triggers[0].digital_edge_script_trigger_edge = nifgen.ScriptTriggerDigitalEdgeEdge.RISING  # RISING / FAILING
+        # PFI TRIGGER: not used right now
+        # session.script_triggers[1].script_trigger_type = nifgen.script_triggers[0].ScriptTriggerType.DIGITAL_EDGE
+        # session.script_triggers[1].digital_edge_script_trigger_source = 'PFI0'
+        # session.script_triggers[1].digital_edge_script_trigger_edge = session.script_triggers[0].ScriptTriggerDigitalEdgeEdge.RISING
+
+        # 3-Generate Signal
```
marcoskirsch: Since we are in "arbitrary waveform generator" world, calling functions "generate something" is confusing, because one would naturally assume that the device will generate the signal which is not the case.

Perhaps use "calculate" as the term.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+        sine = generate_sinewave()
+        ramp_up = generate_rampup()
```
marcoskirsch: You should use underscores consistently, this applies elsewhere.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+        ramp_down = generate_ramdown()
+        square = generate_square()
+        triangle = generate_triangle()
+        noise = generate_gaussian_noise()
+
+        # 4-Writes data to the waveform in onboard memory.
```
marcoskirsch: This is awkwardly written.

"Writes waveform data to the device's onboard memory"

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+        session.channels[channel].write_waveform('sine', sine)        # (waveform_name, data)
```
marcoskirsch: Consider not creating the temporary variables that hold waveform data and just invoking the  functions from the parameter list.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+        session.channels[channel].write_waveform('rampup', ramp_up)
+        session.channels[channel].write_waveform('rampdown', ramp_down)
+        session.channels[channel].write_waveform('square', square)
+        session.channels[channel].write_waveform('triangle', triangle)
+        session.channels[channel].write_waveform('noise', noise)
+
+        # 5-Write Script
+        script_sine = 'script Script\nrepeat until scriptTrigger0\nGenerate sine \nend repeat\nend script'
```
marcoskirsch: If a user wanted to do what this example shows, they would use Arbitrary Waveform Mode.
I think it would be best to do something that shows the strengths of Script Mode.

Consider this:

A single script that generates a mix of all waveforms, repeating each one a different number of times.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+        script_ramp_up = 'script Script\nrepeat until scriptTrigger0\nGenerate rampup\nend repeat\nend script'
+        script_ramp_down = 'script Script\nrepeat until scriptTrigger0\nGenerate rampdown\nend repeat\nend script'
+        script_square = 'script Script\nrepeat until scriptTrigger0\nGenerate square\nend repeat\nend script'
+        script_triangle = 'script Script\nrepeat until scriptTrigger0\nGenerate triangle\nend repeat\nend script'
+        script_noise = 'script Script\nrepeat until scriptTrigger0\nGenerate noise \nend repeat\nend script'
```
marcoskirsch: For readability, can we use multi-line string literals?

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+            script = script_triangle
+        elif shape == 'NOISE':
+            script = script_noise
+        elif shape == 'MULTI':
+            script = script_multi
+        else:
+            print('Unknown shape: {}'.format(shape))
+            sys.exit(1)
+
+        session.channels[channel].write_script(script)
```
marcoskirsch: FYI: In NI-FGEN, if you assign different names to the scripts, you can just write them all.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,153 @@
...
+                for x in range(2):
+                    time.sleep(2)
+                    session.script_triggers[0].send_software_edge_trigger()
+            else:
+                session.script_triggers[0].send_software_edge_trigger()
+
+
+def _main(argsv):
+    parser = argparse.ArgumentParser(description='Continuously generates an arbitrary waveform.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-n', '--resource-name', default='AWG2', help='Resource name of a National Instruments Arbitrary Waveform Generator')
```
marcoskirsch: Didn't we make our default resource names something like PXI1Slot2 in all other examples?

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+        # 3 - Calculate and write different waveform data to the device's onboard memory
+        session.channels[channel].write_waveform('sine', calculate_sinewave())        # (waveform_name, data)
+        session.channels[channel].write_waveform('rampup', calculate_rampup())
+        session.channels[channel].write_waveform('rampdown', calculate_rampdown())
+        session.channels[channel].write_waveform('square', calculate_square())
+        session.channels[channel].write_waveform('triangle', calculate_triangle())
+        session.channels[channel].write_waveform('noise', calculate_gaussian_noise())
+
+        # 4 - Script to generate
+        ''' SINE / SQUARE / TRIANGLE / RAMPUP / RAMPDOWN / NOISE / MULTI '''
```
epage: Whats with this unused string?

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+    return triangle
+
+
+def calculate_gaussian_noise():
+    random_noise = np.random.normal(0, 0.1, 300)  # random.normal(center, scale, size)
+    noise = random_noise.tolist()
+    return noise
+
+
+script_all = '''
```
epage: Constants should be in SCREAMING_CASE

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+
+def example(resource_name, options, shape, channel):
+    with nifgen.Session(resource_name=resource_name, options=options, channel_name=channel) as session:
+        # CONFIGURATION
+        # 1 - Set the mode to Script
+        session.output_mode = nifgen.OutputMode.SCRIPT
+
+        # 2 - Configure Trigger:
+        # SOFTWARE TRIGGER: used in the script
+        session.script_triggers[0].script_trigger_type = nifgen.ScriptTriggerType.SOFTWARE_EDGE  # TRIG_NONE / DIGITAL_EDGE / DIGITAL_LEVEL / SOFTWARE_EDGE
```
epage: API Question: What does it mean to `[0]` on `script_triggers`?

Are there a predefined set of script triggers and this is just changing the values away from default?

marcoskirsch: There are 4 script triggers, 0 through 3.
In the script they are referred to as ScriptTrigger0 and so forth.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+
+def calculate_square():
+    time = np.arange(0, 10, 0.01)                 # np.arange(Start, Stop, Step)
+    square_build = signal.square(time, 0.5)       # signal.square(time,duty_cycle)
+    square = square_build.tolist()                # List of Float
+    return square
+
+
+def calculate_triangle():
+    time = np.linspace(0, 1, 200)                 # np.arange(Start, Stop, Step)
```
marcoskirsch: 3d parameter is not Step, it's num.
I think for all these, if you use named parameters then this sort of copy/paste error goes away and the need for the comment disappears.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+def calculate_gaussian_noise():
+    random_noise = np.random.normal(0, 0.1, 300)  # random.normal(center, scale, size)
+    noise = random_noise.tolist()
+    return noise
+
+
+script_all = '''
+script scriptmulti
+  repeat until scriptTrigger0
+    Generate rampup
```
marcoskirsch: language is case-insensitive, but if we are not capitalizing other keywords, then let's not capitalize `Generate`.

src/nifgen/examples/nifgen_script.py
texasaggie97: Changed to comment

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+    with nifgen.Session(resource_name=resource_name, options=options, channel_name=channel) as session:
+        # CONFIGURATION
+        # 1 - Set the mode to Script
+        session.output_mode = nifgen.OutputMode.SCRIPT
+
+        # 2 - Configure Trigger:
+        # SOFTWARE TRIGGER: used in the script
+        session.script_triggers[0].script_trigger_type = nifgen.ScriptTriggerType.SOFTWARE_EDGE  # TRIG_NONE / DIGITAL_EDGE / DIGITAL_LEVEL / SOFTWARE_EDGE
+        session.script_triggers[0].digital_edge_script_trigger_edge = nifgen.ScriptTriggerDigitalEdgeEdge.RISING  # RISING / FAILING
+        # PFI TRIGGER: not used right now
```
marcoskirsch: I think we can remove all this commented out code.

src/nifgen/examples/nifgen_script.py
texasaggie97: Changed

src/nifgen/examples/nifgen_script.py
texasaggie97: Yes, see the scripts defined in `SCRIPTS_ALL`

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # LAUNCH
+        with session.initiate():
+            for x in range(num_triggers):
+                time.sleep(10)
+                session.script_triggers[0].send_software_edge_trigger()
+
+
+def _main(argsv):
+    parser = argparse.ArgumentParser(description='Continuously generates an arbitrary waveform.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
marcoskirsch: outdated

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,185 @@
...
+    parser.add_argument('-n', '--resource-name', default='PXI1Slot2', help='Resource name of a National Instruments Arbitrary Waveform Generator')
+    parser.add_argument('-s', '--shape', default='SINE', help='Shape of the signal to generate')
```
marcoskirsch: In NI-FGEN parlance, these are usually known as waveforms, not shapes.

But I would argue for simplification, unless you disagree:

* Get rid of triggering.
* Get rid of multiple scripts. Just have a single script that generates forever a bunch of different waveforms. Once you do that then you no longer have to tell the driver what the active script is.

This is for simplicity's sake, if you disagree that's ok.

src/nifgen/examples/nifgen_script.py
texasaggie97: Switched to named parameters.

Fixed comments, some add information about defaults, so I kept them.

src/nifgen/examples/nifgen_script.py
texasaggie97: Changed

src/nifgen/examples/nifgen_script.py
texasaggie97: Removed

src/nifgen/examples/nifgen_script.py
texasaggie97: Changed

src/nifgen/examples/nifgen_script.py
texasaggie97: It seems useful to show how you can write multiple different scripts and then call by name. I would rather keep it.

src/nifgen/examples/nifgen_script.py
```diff
@@ -0,0 +1,181 @@
...
+import argparse
+import nifgen
+import numpy as np
+from scipy import signal
+import sys
+import time
+
+
+def calculate_sinewave():
+    time = np.arange(start=0, stop=10, step=0.01)    # np.arange([start,]stop, [step,]dtype=None)
```
marcoskirsch: Now you can remove the comment part.

texasaggie97: I like the comments because they show what defaults we are using. I can remove though if you don't think that is useful information.

PR #935, opened 2018-10-15
Cleanup waveform info - add explicit fields used

### LINKED ISSUE #934: Classes with dynamic members are user-hostile and maintenance-hostile
Example: WaveformInfo from https://github.com/ni/nimi-python/pull/933

User Problem
- Users can't rely on the code or documentation in front of them to know what they can access
- IDE's can't auto-complete
- mypy will complain

Maintenance problems
- The intent of the code is less clear, making it more likely someone will break it when changing it.

texasaggie97:
* Be more explicit with the fields we use in `niscope.WaveformInfo`

* Fix #934 

* System
* Unit
* Hand

### DIFF
--- a/generated/niscope/session.py
+++ b/generated/niscope/session.py
@@ -1822,10 +1822,10 @@ class _SessionBase(object):
 
         for i in range(len(wfm_info)):
             start = i * num_samples
-            end = start + wfm_info[i].actual_samples
+            end = start + wfm_info[i]._actual_samples
             # We use the actual number of samples returned from the device to determine the end of the waveform. We then remove it from the wfm_info
             # since the length of the wfm will tell us that information
-            wfm_info[i].actual_samples = None
+            wfm_info[i]._actual_samples = None
             if sys.version_info.major >= 3:
                 wfm_info[i].samples = mv[start:end]
             else:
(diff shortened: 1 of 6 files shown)


generated/niscope/session.py
```diff
@@ -1822,10 +1822,10 @@ def fetch(self, num_samples=None, relative_to=enums.FetchRelativeTo.PRETRIGGER,
 
         for i in range(len(wfm_info)):
             start = i * num_samples
+            end = start + wfm_info[i].actual_samples
```
epage: Would we be able to get rid of the `actual_samples` field on `WaveformInfo` if you did this work before the transformation in `_fetch`?

Referring to:
```python
return waveform_array, [waveform_info.WaveformInfo(wfm_info_ctype[i]) for i in range(self._actual_num_wfms())]
```

texasaggie97: `actual_samples` is in the struct that we get back from the driver.

All the different `_fetch` functions are completely code generated. In order to do something there
we would have to torque around the code generator for a special case in scope, or have even more hand maintained functions where we do this same manipulation in each one. I don't think either of those options are very good.

epage: Understandable.

Since this is purely an implementation detail, I'm just generally concerned about publicly exposing the,

What if we `_` prefix `actual_samples`?

texasaggie97: Updated to make `actual_samples` "private"

PR #949, opened 2018-10-29
Add support for string size mechanism of passed-in

### LINKED ISSUE #900: GetAStringWithSpecifiedMaximumSize does not exist in nifake's 'functions.py' but is called out in functions_addon.py
GetAStringWithSpecifiedMaximumSize does not exist in functions.py

https://github.com/ni/nimi-python/blob/0670ca9d755be3a4ab96d9f431914cd92829b2e1/src/nifake/metadata/functions_addon.py#L48

- The generator should have caught this.
- We are not testing this mechanism.

texasaggie97:
* Add support for string size mechanism of `passed-in`
* Add test in nifake to use new mechanism
* For some reason, after re-installing Windows, flake8 (of any version) would not look in `tox.ini` without telling it that that was the config file. I don't believe this will hurt and may keep this from happening to others.
* Fix #900 

* Unit

### DIFF
--- a/generated/nifake/session.py
+++ b/generated/nifake/session.py
@@ -956,6 +956,27 @@ class Session(_SessionBase):
         return a_string_ctype.value.decode(self._encoding)
 
     @ivi_synchronized
+    def get_a_string_with_specified_maximum_size(self, buffer_size):
+        r'''get_a_string_with_specified_maximum_size
+
+        Returns multiple types.
+
+        Args:
+            buffer_size (int): Number of bytes allocated for aString
+
+
+        Returns:
+            a_string (str): An string with passed in size.
+
+        '''
+        vi_ctype = _visatype.ViSession(self._vi)  # case S110
+        buffer_size_ctype = _visatype.ViInt32(buffer_size)  # case S190
+        a_string_ctype = (_visatype.ViChar * buffer_size)()  # case C090
+        error_code = self._library.niFake_GetAStringWithSpecifiedMaximumSize(vi_ctype, buffer_size_ctype, a_string_ctype)
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
+        return a_string_ctype.value.decode(self._encoding)
+
+    @ivi_synchronized
     def get_an_ivi_dance_string(self):
         r'''get_an_ivi_dance_string
 
--- a/generated/nifake/session.py
+++ b/generated/nifake/session.py
@@ -959,7 +959,7 @@ class Session(_SessionBase):
     def get_a_string_with_specified_maximum_size(self, buffer_size):
         r'''get_a_string_with_specified_maximum_size
 
-        Returns multiple types.
+        Method to test getting a string with the size passed in as a parameter.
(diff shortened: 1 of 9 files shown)


generated/nifake/session.py
```diff
@@ -955,6 +955,27 @@ def get_a_string_using_python_code(self, a_number):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return a_string_ctype.value.decode(self._encoding)
 
+    @ivi_synchronized
+    def get_a_string_with_specified_maximum_size(self, buffer_size):
+        r'''get_a_string_with_specified_maximum_size
+
+        Returns multiple types.
```
marcoskirsch: Is this right? Seems like a bug in the generator (if so, should be addressed in its own PR)

src/nifake/metadata/functions.py
```diff
@@ -923,6 +923,43 @@
...
+                    'description': 'Number of bytes allocated for aString',
+                },
+            },
+            {
+                'direction': 'out',
+                'enum': None,
+                'name': 'aString',
+                'type': 'ViChar[]',
+                'documentation': {
+                    'description': 'An string with passed in size.',
```
marcoskirsch: An -> A

src/nifake/metadata/functions.py
```diff
@@ -923,6 +923,43 @@
...
+                },
+            },
+        ],
+        'documentation': {
+            'description': 'Returns multiple types.',
```
marcoskirsch: Please fix. It through this was created by the generator ;)

generated/nifake/session.py
```diff
@@ -955,6 +955,27 @@ def get_a_string_using_python_code(self, a_number):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return a_string_ctype.value.decode(self._encoding)
 
+    @ivi_synchronized
+    def get_a_string_with_specified_maximum_size(self, buffer_size):
```
epage: Why is a python user expected to pass in a buffer size?  That feels like a C API bleeding over into Python.  Is there really no way we can work around this?

marcoskirsch:
Here's an additional problem: If we decide we absolutely need this in out public API:

```python
def get_a_string_with_specified_maximum_size(self, buffer_size):
 ```

We may or may not have our string-encoding story 100% correct yet and we get away with it becuase the APIs tend to return ASCII only. But that assumption will not hold forever. So the number passed in by the client (buffer_size) may not match the number of characters from the client's point of view when the API returns non-ASCII characters.

texasaggie97:
This is becoming too much for a mechanism we haven't needed so far.

I propose abandoning this PR and not adding this mechanism. Instead see #950 for how to close #900

marcoskirsch:
> This is becoming too much for a mechanism we haven't needed so far.

Furthermore, I don't think we'd ever want to expose it in this manner; at least not in the public API. I can see ourselves maybe using it in a private method?

I'm ok abandoning altogether.

Issue #906, opened 2018-07-11
Using the API when Debug Session is configured to "Using Breakpoints in C/C++/.NET Applications" causes access violation


TestStandTrent:
* Description of issue

    When using a simulated DCPower device in Max and not passing 'simulate' as an option when creating the session, the attached example generates an access violation.

* `python -c "import niscope; niscope.print_diagnostic_information()"` output (replace `niscope` with appropriate package name)

```
  OS:
    Name:      Windows
    Version:   10.0.17134
    Bits:      64
Driver:
    Name:      NI-DCPower
    Version:   18.1.0.49152
Module:
    Name:      nidcpower
    Version:   1.0.0
Python:
    Version:   3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 17:00:18) [MSC v.1900 64 bit (AMD64)]
    Bits:      64
    Is_Venv:   False
    Installed Packages:
        wheel==0.31.1
        six==1.11.0
        setuptools==39.0.1
        pip==10.0.1
        nidcpower==1.0.0
```

* Steps to reproduce issue

    1. Create a simulated 4140 in MAX
    2. run the script below and reference the simulated 4140 as you would a real one
    3. the script waits for a while and then errors, leaving the simulated device in an unrecoverable state.

```
import nidcpower
with nidcpower.Session(resource_name='PXI1Slot6', channels='0') as session:
    session.measure_record_length = 20
    session.measure_record_length_is_finite = True
    session.measure_when = nidcpower.MeasureWhen.AUTOMATICALLY_AFTER_SOURCE_COMPLETE
    session.voltage_level = 5.0

    session.commit()
    print('Effective measurement rate: {0} S/s'.format(session.measure_record_delta_time / 1))

    samples_acquired = 0
    print('  #    Voltage    Current    In Compliance')
    row_format = '{0:3d}:   {1:8.6f}   {2:8.6f}   {3}'
    with session.initiate():
        while samples_acquired < 20:
            measurements = session.fetch_multiple(count=session.fetch_backlog)
            samples_acquired += len(measurements)
            for i in range(len(measurements)):
                print(row_format.format(i, measurements[i].voltage, measurements[i].current, measurements[i].in_compliance))
```

Fladolcetta:
This works for me on Python 64bit version 3.6.2 using NI-DCPower 18.1 (driver) and nidcpower (python) 1.0.0.

Perhaps this is a 3.6.5 issue?

[...]

PR #951, opened 2018-11-01
Add global lock disable at codegen time

### LINKED ISSUE #896: No mechanism to globally not support locking
NI Swith Executive does not have a method to lock a session. The current session.py.mako requires a function called lock to work. We should allow a global config file disable for locking rather than disabling locking on every function manually in functions_addon.py

texasaggie97:
* Add global 'use_locking' configuration - this allows completely disabling session locking for a complete driver in addition to individual methods
    * nitclk will most likely need this as well
* Changes to remove specialized `nise` templates for `__init__.py` and `session.py`
    * Support name different than the driver name when looking in the registry for the driver version
        * Default is `config['driver_name']`
        * Can override by setting `config['driver_registry']`
    * `nise/__init__.py` now has `__version__` like it should have had
    * Specialized `session.py` template no longer needed with global lock disable
* Fix error report when trying to index a driver that has no repeated capabilities
* Fix a typo
* Fix #896 

* Unit
* System


build/templates/session.py.mako
```diff
@@ -87,6 +87,7 @@ class ${session_context_manager}(object):
 
 
 % endif
+% if config['use_locking']:
 # From https://stackoverflow.com/questions/5929107/decorators-with-parameters
 def ivi_synchronized(f):
     def aux(*xs, **kws):
```
epage: Unrelated to your change but I noticed you have a decorator not using [`wraps`](https://docs.python.org/3/library/functools.html#functools.wraps)

generated/nidmm/session.py
```diff
@@ -562,8 +537,12 @@ def __setattr__(self, key, value):
         object.__setattr__(self, key, value)
 
     def __getitem__(self, key):
-        rep_caps = []
-        raise TypeError("'Session' object does not support indexing. You should use the applicable repeated capabilities container(s): {}".format(', '.join(rep_caps)))
+        rep_cap_help_text = ''
+        raise TypeError("'Session' object does not support indexing." + rep_cap_help_text)
+
+    def _set_use_locking(self, use_locking):
+        '''Allow runtime disabling of session locking'''
```
epage: Is this meant to be used internally or by users?  The `_` suggests internal / not something that can be counted on.

If it is meant to be used by users, what about the problem of someone toggling this while the lock is held?

marcoskirsch: Since it's a workaround to an (obscure?) bug, I would not want this to not have a `_`.

But my actual preference for working around #906 would be to add a small example to the issue that teaches customers how to monkey patch their session with stub `[lock](https://nimi-python.readthedocs.io/en/1.1.0/nidmm/class.html#nidmm.Session.lock)` and `[unlock](https://nimi-python.readthedocs.io/en/1.1.0/nidmm/class.html#nidmm.Session.unlock)` implementations.

epage: I'd rather not encourage people to monkey patch the API.

texasaggie97: I put `_` because I don't plan on documenting it and I don't expect the vast majority of customers to need this. It is a workaround for a very specific issue in the driver.

I updated the docstring to be more specific when this should be called and a warning not to change it while a function call is in progress. Though if they are able to do that, then it means they don't need to set this anyway.

I also don't think we should be telling customers to monkey patch the API.

marcoskirsch: > I'd rather not encourage people to monkey patch the API.
 
> I also don't think we should be telling customers to monkey patch the API.

I agree, but I think that's better than adding code to the API itself to workaround a bug in the drivers. Monkey patching has a lot of code smell, as it should.

epage: > I put _ because I don't plan on documenting it and I don't expect the vast majority of customers to need this. It is a workaround for a very specific issue in the driver.

We should have a clear message / policy when it comes to `_` functions.  Examples
- `_` are "here be dragons" but we will not break someone using it without breaking semver
- `_` is an implementation detail and we do not guarantee its behavior from release to release. Use at your own risk (and pin your version)

epage: > I agree, but I think that's better than adding code to the API itself to workaround a bug in the drivers. Monkey patching has a lot of code smell, as it should.

I'm fine with the workaround being listed in an issue.  Listing it in the docs is what I'd be concerned about.  That adds a level of officialness to it in my mind.

marcoskirsch: > I'm fine with the workaround being listed in an issue.

Yep, that's what I meant with

> to add a small example to the issue

Should have been clearer.

marcoskirsch: > We should have a clear message / policy when it comes to _ functions. Examples

My understanding is that `_` means the second one:

> _ is an implementation detail and we do not guarantee its behavior from release to release. Use at your own risk (and pin your version)
 
I can make that clearer somewhere in the wiki if you think I should.

texasaggie97: I removed the runtime locking disabling.

build/templates/session.py.mako
texasaggie97: You can open a separate issue for this.

generated/nise/session.py
```diff
@@ -68,6 +72,15 @@ def __init__(self, vi, library, encoding, freeze_it=False):
...
         return '{0}.{1}({2})'.format('nise', self.__class__.__name__, self._param_list)
 
+    def __setattr__(self, key, value):
+        if self._is_frozen and key not in dir(self):
+            raise AttributeError("'{0}' object has no attribute '{1}'".format(type(self).__name__, key))
+        object.__setattr__(self, key, value)
+
+    def __getitem__(self, key):
+        rep_cap_help_text = ''
+        raise TypeError("'Session' object does not support indexing." + rep_cap_help_text)
```
epage: Whats the goal with these error checks?

build/templates/__init__.py.mako
```diff
@@ -4,6 +4,7 @@ ${template_parameters['encoding_tag']}
 enums = template_parameters['metadata'].enums
 config = template_parameters['metadata'].config
 module_name = config['module_name']
+registry_name = config['driver_registry'] if 'driver_registry' in config else config['driver_name']
```
epage: Please update the PR to mention support for this new metadata field.

generated/nise/__init__.py
```diff
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 # This file was generated
 
+
+__version__ = '0.2.1.dev0'
```
epage: I feel I'm missing what changed to cause this to show up?

build/templates/session.py.mako
texasaggie97: To a point.

I am not going to go create a separate PR to fix a typo or to fix an error report. At some point you add too much overhead and it becomes easier to just leave it be.

For these particular things, I knew what to do to fix the error report. For wraps I will need to read up on them and figure out how to use them, and then probably write a test. It is a bigger thing.

generated/nise/session.py
texasaggie97: So that they do not try `scope[0].range = 4.0`. If you do that on a `niscope.Session()` there will be a list of repeated capabilities that you might want to use.

`nise` does not have any repeated capabilities (similar to `nidmm`) so the error report would have said to try one of these repeated capability objects and then given an empty list.

build/templates/__init__.py.mako
texasaggie97: This is not new. It was used by nise. It is here so that nise does not need a specialized template.

I will update the decsription to mention this.

generated/nise/__init__.py
texasaggie97: `nise` no longer has specialized templates for `__init__.py` and `session.py`. This wasn't added to the `nise` template like it should have been. Basically I didn't realize that there was a specialized template.

generated/nidmm/session.py
```diff
@@ -562,8 +535,8 @@ def __setattr__(self, key, value):
         object.__setattr__(self, key, value)
 
     def __getitem__(self, key):
-        rep_caps = []
-        raise TypeError("'Session' object does not support indexing. You should use the applicable repeated capabilities container(s): {}".format(', '.join(rep_caps)))
+        rep_cap_help_text = ''
+        raise TypeError("'Session' object does not support indexing." + rep_cap_help_text)
```
marcoskirsch: "does not support indexing" is no longer descriptive. It was for a while back during development when you could do `session[0].foo()`.

texasaggie97: How is it no longer descriptive. Session does not in fact support indexing. `session[0].foo()` is illegal now. That is why we put this specific error in, so we would be more descriptive.

marcoskirsch: I see. I misunderstood the context in which the error shows up.

I went ahead to see what Python returns in this case in general. Perhaps we should use the same wording.

```
>>> a = 5
>>> a[4]
TypeError: 'int' object is not subscriptable
```

texasaggie97: I updated the wording to match. Is the wording added when there are repeated capabilities okay?

build/templates/session.py.mako
```diff
@@ -194,11 +201,17 @@ constructor_params = helper.filter_parameters(init_function, helper.ParameterUsa
...
         rep_caps = []
-% for rep_cap in config['repeated_capabilities']:
-        rep_caps.append("${rep_cap['python_name']}")
-% endfor
-        raise TypeError("'Session' object does not support indexing. You should use the applicable repeated capabilities container(s): {}".format(', '.join(rep_caps)))
+        for rep_cap in config['repeated_capabilities']:
+            rep_caps.append(rep_cap['python_name'])
+        r = ', '.join(rep_caps)
+%>\
+        raise TypeError("'Session' object is not subscriptable. Did you mean to use a repeated capabilities container: ${r}")
```
marcoskirsch: Should have  `?` at the end since it's a question.

marcoskirsch: I'd put the container in parenthesis...

> 'Session' object is not subscriptable. Did you mean to use a repeated capabilities container: (channels)?"

PR #956, opened 2018-11-07
Remove  __getitem__ from Session

### LINKED ISSUE #955: Should SessionBase override `__getitem__` and `__setattr__` to error?
This came up in https://github.com/ni/nimi-python/pull/951

Talking to @texasaggie97, it appears that `Session` originally supported these capabilities but they were moved off, so that these error were meant to help people migrate.

If we consider that migration story done, should we drop these?

Why?
- In Python you don't generally override every dunder function to say you don't support it but instead rely on the default behavior
- This might throw off some documentation, analysis tools, and "smart" libraries that check for non-default implementations (ran into this but with a different dunder when using enums)

Instead, the lack of a `getitem` and lack of documentation on `Session` should be sufficient.

texasaggie97:
* Remove `__getitem__()` from `Session`
    * Was originally implemented to help clarify that indexing for repeated capabilities belongs on the repeated capabilities containers
    * That was only during the prerelease stage and is now properly documented

* Fix #955 

* Unit
* System

### DIFF
--- a/generated/nidmm/session.py
+++ b/generated/nidmm/session.py
@@ -537,9 +537,6 @@ class _SessionBase(object):
             raise AttributeError("'{0}' object has no attribute '{1}'".format(type(self).__name__, key))
         object.__setattr__(self, key, value)
 
-    def __getitem__(self, key):
-        raise TypeError("'Session' object is not subscriptable")
-
     def _get_error_description(self, error_code):
         '''_get_error_description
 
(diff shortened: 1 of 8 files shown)


generated/nidmm/session.py
```diff
@@ -537,9 +537,6 @@ def __setattr__(self, key, value):
             raise AttributeError("'{0}' object has no attribute '{1}'".format(type(self).__name__, key))
         object.__setattr__(self, key, value)
```
epage: What about `setattr`?

texasaggie97: That has a different purpose that has not gone away. We do not want `session.rnge = 5.0` (notice the misspelling) to work.

PR #957, opened 2018-11-08
nidcpower advanced sequencing

### LINKED ISSUE #504: Parameter attribute_ids to nidcpower.Session.create_advanced_sequence() should be an enum
The way the C API works, the client is supposed to pass an array of attribute IDs that he/she wants included in the advanced sequence. In our Python API, right now, this is just an int :(

texasaggie97:
* Implements more Pythonic API for nidcpower advanced sequences
    * Option 1 from #504 
* Documentation and tests still remaining but looking for feedback before doing that work

* Fix #504


generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
...
+            for key in s:
+                attribute_ids_used.add(attribute_lookup[key]['id'])
+
+        # Create the step with the list of attr ids we have
+        self._create_advanced_sequence(sequence_name, list(attribute_ids_used), set_as_active_sequence)
+
+        for s in sequence:
+            self._create_advanced_sequence_step()
+            for key in s:
+                attribute_lookup[key]['property'] = s[key]
```
epage: So I'm assuming the intent is that `attribute_lookup` is suppose to have a reference to `@property`s and this assignment is invoking the function.

I don't think this is working that way though,  `self.output_enabled` will have called the property, getting the value.  We are then just modifying `attribute_lookup`.

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
...
+
+        Args:
+            sequence_name (str): Name of sequence.
+
+            sequence (list of int): Sequence.
+
+            set_as_active_sequence (bool): Specifies that this current sequence is active.
+
+        '''
+        attribute_lookup = {
```
epage: If it wasn't for the property lookups, I'd say this should be a constant outside of this functions (a `_ATTRIBUTE_LOOKUP`)

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
...
+        for s in sequence:
+            for key in s:
+                attribute_ids_used.add(attribute_lookup[key]['id'])
+
+        # Create the step with the list of attr ids we have
+        self._create_advanced_sequence(sequence_name, list(attribute_ids_used), set_as_active_sequence)
+
+        for s in sequence:
+            self._create_advanced_sequence_step()
+            for key in s:
```
epage: Why not `for key, value in s.items()` to avoid having to lookup the value when we could already have it?

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
...
+        '''create_advanced_sequence
+
+        Test.
+
+        Note: Test.
+
+        Args:
+            sequence_name (str): Name of sequence.
+
+            sequence (list of int): Sequence.
```
epage: So it looks like a `sequence` is a `Iterable[Dict[str, Any]]` with the `str` being the property name.

Why are we doing a stringly-typed API rather than using enums?  Would we want the `.value` to be the string or the id?

Or at worse, providing constants to the user (and having this function's source use the constants as well)

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
+    @ivi_synchronized
+    def create_advanced_sequence(self, sequence_name, sequence, set_as_active_sequence=True):
+        '''create_advanced_sequence
+
+        Test.
+
+        Note: Test.
```
epage: As you mention, documentation is still outstanding

generated/nidcpower/session.py
texasaggie97: Documentation is incomplete. There will need to be some changes to get the type right. I don't want to put in the effort to fix it in case we end up doing something different.

epage: Sorry if I wasn't clear.  This was a "good enough" place for me to open an issue regarding the data format.

generated/nidcpower/session.py
epage: > I don't want to put in the effort to fix it in case we end up doing something different.

Sounds like documentation isn't fixed.  I'm unresolving this to help us, the reviewers, track this being resolved.

generated/nidcpower/session.py
marcoskirsch: > Why are we doing a stringly-typed API rather than using enums?
> Would we want the .value to be the string or the id?
> Or at worse, providing constants to the user (and having this function's
> source use the constants as well)
 

We debated the API in #504 

If we had an enum, the client using the Advanced Sequences API would need to refer to attributes differently than how they normally do it:
* `my_session.output_function` when using the propoerty normally.
* `Properties.OUTPUT_FUNCTION` when referring to the same thing in the context of `create_advanced_sequence`.

We thought it would be nicer for the client to be able to simply use `"output_function"` in the context of `create_advanced_sequence`.

marcoskirsch: This is wrong

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
...
+
+            sequence (list of int): Sequence.
+
+            set_as_active_sequence (bool): Specifies that this current sequence is active.
+
+        '''
+        attribute_lookup = {
+            "range_check": {
+                'id': 1050002,
+                'property': self.range_check,
```
marcoskirsch: Is this not causing us to call "get" on the attributes?

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
...
+            },
+            "ovp_limit": {
+                'id': 1250003,
+                'property': self.ovp_limit,
+            },
+            "current_limit": {
+                'id': 1250005,
+                'property': self.current_limit,
+            },
+            "output_enabled": {
```
marcoskirsch: I think it's unfortunate that we are duplicating all this information we already have within the Session class just in dictionary form. I understand that since we use a code-generator, it doesn't add any maintenance burden. But it just seems _wrong_. We should be looking up the member of the class Session by the given name, then trying to get the attribute ID. If that fails, we can raise the appropriate exception.

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,599 @@ def export_attribute_configuration_file(self, file_path):
...
+            },
+        }
+
+        # First we need to get all possible properties we might be setting
+        attribute_ids_used = set()
+        for s in sequence:
+            for key in s:
+                attribute_ids_used.add(attribute_lookup[key]['id'])
+
+        # Create the step with the list of attr ids we have
```
marcoskirsch: This comment is incorrect. You are creating the sequence not the step.

In the comments, try to use "property" or "attribute" consistently. Here you have "attr ids" but above you have "all possible properties"

src/nidcpower/metadata/functions_addon.py
```diff
@@ -418,6 +417,52 @@
...
+                'documentation': {
+                    'description': 'Identifies a particular instrument session. **vi** is obtained from the niDCPower_InitExtCal or niDCPower_InitializeWithChannels function.',
+                },
+            },
+            {
+                'direction': 'in',
+                'name': 'SequenceName',
+                'type': 'ViString',
+                'documentation': {
+                    'description': 'Name of sequence.',
```
marcoskirsch: Does this match the LabVIEW VI string?

src/nidcpower/metadata/functions_addon.py
```diff
@@ -418,6 +417,52 @@
...
+                },
+            },
+            {
+                'direction': 'in',
+                'name': 'Sequence',
+                'type': 'ViInt32[]',  # Doesn't matter here
+                'documentation': {
+                    'description': 'Sequence.',
```
marcoskirsch: We should better explain things here. If you want to leave it to its own PR (not a bad idea) please TODO it.

generated/nidcpower/session.py
texasaggie97: Correct. So I changed how I do this.

generated/nidcpower/session.py
texasaggie97: Fixed.

generated/nidcpower/session.py
texasaggie97: Changed.

generated/nidcpower/session.py
texasaggie97: It does. I changed how I do this.

generated/nidcpower/session.py
texasaggie97: I am trying to keep everything isolated to this method. To do what you suggest would require modifying the attributes to allow getting the attribute ID without sneaking in and reading a 'private' variable.

generated/nidcpower/session.py
texasaggie97: Changed step to sequence.

Property is in Python, attr_id is a C thing. I used those deliberately since they are different.

src/nidcpower/metadata/functions_addon.py
texasaggie97: Please see PR description: 'Documentation and tests still remaining but looking for feedback before doing that work'

src/nidcpower/metadata/functions_addon.py
texasaggie97: Because we don't generate this method, this type doesn't matter.

Something will need to change for documentation though. I am not working on that yet, until we are sure of the implementation and how the API will look.

generated/nidcpower/session.py
texasaggie97: Please see PR description: 'Documentation and tests still remaining but looking for feedback before doing that work'

generated/nidcpower/session.py
marcoskirsch: >  To do what you suggest would require modifying the attributes to allow getting the attribute ID without sneaking in and reading a 'private' variable.

I think that's acceptable. It's all within the same module.

texasaggie97: Same module but different classes.

I will see what I can do so we can see what that would look like.

texasaggie97:
How do we handles channels?

Do we require the call happen on the channel rep_cap container, even though `create_advanced_sequence` and `create_advanced_sequence_step` themselves don't take a channel string?

Can the channel change per step? Or in a single step, can there be different configurations per channel? If so, option 1 doesn't work as is.

marcoskirsch:
After some experimentation.

How to get the attribute ID given the Python name of the attribute, DRY version:

```python
>>> nidmm.Session.__bases__[0].__dict__['range']._attribute_id
1250002
```

How to set an attribute given the Python name of the attribute, DRY version:

```python
>>> setattr(s, 'range', 100)
>>> s.range
100.0
```

I like this better than duplicating known information in the code, and adding complexity to the mako template.

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,37 @@ def export_attribute_configuration_file(self, file_path):
...
+
+        '''
+        # First we need to get all possible properties we might be setting
+        attribute_ids_used = set()
+        for s in sequence:
+            for key in s:
+                attribute_ids_used.add(Session.__base__.__dict__[key]._attribute_id)
+
+        # Create the sequence with the list of attr ids we have
+        self._create_advanced_sequence(sequence_name, list(attribute_ids_used), set_as_active_sequence)
```
marcoskirsch: I'm surprised you couldn't just pass the set and had to create a list. I'd expect duck typing to work since set is iterable.

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,37 @@ def export_attribute_configuration_file(self, file_path):
...
+            sequence (list of int): Sequence.
+
+            set_as_active_sequence (bool): Specifies that this current sequence is active.
+
+        '''
+        # First we need to get all possible properties we might be setting
+        attribute_ids_used = set()
+        for s in sequence:
+            for key in s:
+                attribute_ids_used.add(Session.__base__.__dict__[key]._attribute_id)
```
marcoskirsch: While this is nice and concise, one disadvantage is that if the client passes a string that's not the name of an attribute, the error will be pretty bad. If the string matches another non-attribute member it will be one thing. If the string doesn't match any members, it will be another thing. This would be a common thing with typos. Do you think the error raised is good enough?

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,37 @@ def export_attribute_configuration_file(self, file_path):
...
+
+        # Create the sequence with the list of attr ids we have
+        self._create_advanced_sequence(sequence_name, list(attribute_ids_used), set_as_active_sequence)
+
+        # Create and configure the steps
+        for s in sequence:
```
marcoskirsch: I'd just spell out `step`, the code becomes really readable then.

src/nidcpower/metadata/functions_addon.py
```diff
@@ -42,9 +42,8 @@
     'FetchMultiple':                              { 'codegen_method': 'private', 'method_name_for_documentation': 'fetch_multiple',                  },  # 'FancyFetchMultiple' Public wrapper
     'MeasureMultiple':                            { 'codegen_method': 'private', 'method_name_for_documentation': 'measure_multiple',                },  # 'FancyMeasureMultiple' Public wrapper
     'self_test':                                  { 'codegen_method': 'private', 'method_name_for_documentation': 'self_test',                       },  # 'fancy_self_test' Public wrapper that raises
-    'CreateAdvancedSequence':                     { 'codegen_method': 'private',  },  # Advanced sequence private until #504 has a fix
-    'CreateAdvancedSequenceStep':                 { 'codegen_method': 'private',  },  # Advanced sequence private until #504 has a fix
-    'DeleteAdvancedSequence':                     { 'codegen_method': 'private',  },  # Advanced sequence private until #504 has a fix
+    'CreateAdvancedSequence':                     { 'codegen_method': 'private',  },  # Advanced sequence private
```
marcoskirsch: Comments need to be updated "FancyCreateAdvancedSequence"

texasaggie97: No longer applicable

generated/nidcpower/session.py
texasaggie97: Done

generated/nidcpower/session.py
texasaggie97: We explicitly look for list in `get_ctypes_pointer_for_buffer()`

generated/nidcpower/session.py
```diff
@@ -4151,6 +4151,37 @@ def export_attribute_configuration_file(self, file_path):
...
+
+        Args:
+            sequence_name (str): Name of sequence.
+
+            sequence (list of int): Sequence.
+
+            set_as_active_sequence (bool): Specifies that this current sequence is active.
+
+        '''
+        # First we need to get all possible properties we might be setting
```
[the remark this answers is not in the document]
texasaggie97: Changed.

marcoskirsch:
I think we should close this PR until it's ready to be looked at.

generated/nidcpower/session.py
texasaggie97: Added error checking for both existence and type of the key

build/templates/tox-system_tests.ini.mako
```diff
@@ -80,7 +80,7 @@ deps =
 
 depends =
     ${module_name}-coverage: py{35,36,37,38}-${module_name}-system_tests
-% if config['supports_nitclk']:
+% if uses_other_wheel:
```
marcoskirsch: Seems unrelated to PR.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4694,6 +4694,43 @@ def export_attribute_configuration_file(self, file_path):
...
+
+        '''
+        # First we need to get all possible properties we might be setting. The way the NI-DCPower C API is designed,
+        # we need to know this upfront in order to call `niDCPower_CreateAdvancedSequence`. In order to find the attribute
+        # ID of each property, we look at the member Attribute objects of Session.
+        attribute_ids_used = set()
+        for step in sequence:
+            for key in step:
+                if key not in Session.__base__.__dict__:
+                    raise TypeError('{0} is not an property on the nidcpower Session class'.format(key))
```
marcoskirsch: Why `TypeError` and not `KeyError`?
`the nidcpower Session class` ->  `nidcpower.Session`

generated/nidcpower/nidcpower/session.py
```diff
@@ -4694,6 +4694,43 @@ def export_attribute_configuration_file(self, file_path):
...
+                if not isinstance(Session.__base__.__dict__[key], _attributes.Attribute) and not isinstance(Session.__base__.__dict__[key], _attributes.AttributeEnum):
+                    raise TypeError('{0} is not an attribute type: {1}'.format(key, type(Session.__base__.__dict__[key])))
+                attribute_ids_used.add(Session.__base__.__dict__[key]._attribute_id)
```
marcoskirsch: I think it's more Pythonic to simply try/except getting the _attribute_id rather than to use `isinstance`.

generated/nitclk/tox-system_tests.ini
```diff
@@ -50,6 +50,7 @@ deps =
 
 depends =
     nitclk-coverage: py{35,36,37,38}-nitclk-system_tests
+    nitclk-system_tests: py38-nitclk-wheel_dep,
```
marcoskirsch: unrelated

src/nidcpower/metadata/functions.py
```diff
@@ -883,6 +882,52 @@
...
+                'name': 'setAsActiveSequence',
+                'type': 'ViBoolean',
+                'default_value': True,
+                'documentation': { 'description': 'Specifies that this current sequence is active.', },
+            },
+        ],
+        'documentation':
+        {
+            'description': 'Test.',
+            'note': 'Test.',
```
marcoskirsch: Documentation is incomplete. If the intention is to merge this way and fix that in a separate PR, then add a TODO that we can point to an issue about it.

build/templates/tox-system_tests.ini.mako
texasaggie97: It is a fix I came across while working on this. Do you need a separate PR for this one change?

generated/nidcpower/nidcpower/session.py
texasaggie97: `KeyError` is an implementation detail, and I thought `TypeError` more appropriate. And to be consistent with the other error condition. Since I don't have a strong opinion though, do you want me to change it?

Changed to `nidcpower.Session`

generated/nidcpower/nidcpower/session.py
texasaggie97: But then we don't know the specific reason for the exception and cannot set the description appropriately.

generated/nitclk/tox-system_tests.ini
texasaggie97: This is the fix from above. There is a lot of overhead with separate PRs and waiting for reviews and approval and merging so this branch will work.

src/nidcpower/metadata/functions.py
texasaggie97: Yes, documentation is complete. As stated before elsewhere, I am waiting to make sure that the approach is good before spending the time on documentation that may change.

build/templates/tox-system_tests.ini.mako
marcoskirsch: Best practice is to make PRs small and single-purpose. It helps keep source control history clean and (we don't take that much advantage of it because this is a small project) makes cherry-picking changes viable.

generated/nidcpower/nidcpower/session.py
marcoskirsch: I think it's more correct to return KeyError when the name is bogus and TypeError when the name is for something that exists but isn't a property.

generated/nidcpower/nidcpower/session.py
marcoskirsch: ok

build/templates/tox-system_tests.ini.mako
texasaggie97: This change is now in master so this is no longer different.

marcoskirsch:
Closing this based on https://github.com/ni/nimi-python/issues/504#issuecomment-590586634
We're implementing #1289 instead. :(

PR #972, opened 2019-04-16
Internal/add fgen

texasaggie97:
* Update nifgen metadata


marcoskirsch:
nimi-bot is MIA

texasaggie97:
Caleb has it and will set it back up once y’all’s move is complete.

docs/nifgen/enums.rst
```diff
@@ -423,7 +423,7 @@ SampleClockSource
 
 
 
-    .. py:attribute:: SampleClockSource.PXI_TRIGGER_LINE_0RTSI_0
+    .. py:attribute:: SampleClockSource.PXI_TRIGGER_LINE_0_RTSI_0
```
marcoskirsch: Both old and new look wrong (old looks more wrong).
Constant in C is `NIFGEN_VAL_RTSI_0`

What should we be doing here?
Do we even want / need this in `nifgen`? It's used for the old enum-based routing API, we should be directing clients towards using string-based routing API.

generated/nifgen/enums.py
```diff
@@ -87,10 +87,10 @@ class DataMarkerEventLevelPolarity(Enum):
 
 class HardwareState(Enum):
     IDLE = 0
-    WAITING_FOR_START_TRIGGER = 1
-    RUNNING = 2
-    DONE = 3
-    HARDWARE_ERROR = 4
+    WAITING_FOR_START_TRIGGER = 100
```
marcoskirsch: CHANGELOG these

generated/nifgen/session.py
```diff
@@ -853,6 +843,16 @@ class _SessionBase(object):
 
     Returns a brief description of NI-FGEN.
     '''
+    major_version = _attributes.AttributeViInt32(1050503)
```
marcoskirsch: Do you know why this shown up in a different position? (not that it matter much)

generated/nifgen/session.py
```diff
@@ -3532,6 +3532,8 @@ def create_arb_sequence(self, waveform_handles_array, loop_counts_array):
         '''
         vi_ctype = _visatype.ViSession(self._vi)  # case S110
         sequence_length_ctype = _visatype.ViInt32(0 if waveform_handles_array is None else len(waveform_handles_array))  # case S160
+        if loop_counts_array is not None and len(loop_counts_array) != len(waveform_handles_array):  # case S160
```
marcoskirsch: Looks like a small bugfix. Worth noting?

generated/nifgen/session.py
```diff
@@ -3619,6 +3621,8 @@ def create_freq_list(self, waveform, frequency_array, duration_array):
         vi_ctype = _visatype.ViSession(self._vi)  # case S110
         waveform_ctype = _visatype.ViInt32(waveform.value)  # case S130
         frequency_list_length_ctype = _visatype.ViInt32(0 if frequency_array is None else len(frequency_array))  # case S160
+        if duration_array is not None and len(duration_array) != len(frequency_array):  # case S160
```
marcoskirsch: Looks like a small bugfix. Worth noting?

generated/nifgen/enums.py
texasaggie97: Done

generated/nifgen/session.py
texasaggie97: Because the actual attribute name is not `NIFGEN_ATTR_MAJOR_VERSION`, it is actually `NIFGEN_ATTR_SPECIFIC_DRIVER_MAJOR_VERSION`. But to not break backwards compatibility I had to keep the old name, but it is alphabetized by the actual name.

generated/nifgen/session.py
texasaggie97: I don't think we need to.

generated/nifgen/session.py
texasaggie97: I don't think we need to.

PR #983, opened 2019-05-17
Differentiate between tested with version and api version

texasaggie97:
* Differentiates between the version that the Python support is tested against and the version of the API Metadata that is used


README.rst
```diff
@@ -109,7 +109,9 @@ NI-DMM Python API Status
 +-------------------------------+----------------------+
 | NI-DMM (nidmm)                |                      |
 +===============================+======================+
-| Driver Version Tested Against | 19.1.0d11            |
+| Driver Version Tested Against | 18.1.0               |
++-------------------------------+----------------------+
+| Driver API Version            | 19.1.0d11            |
```
marcoskirsch: This is internal to NI... but ok

src/nimodinst/metadata/config_addon.py
```diff
@@ -1,4 +1,5 @@
 # We need to maintain the version here since it needs to be updated by the build process on GitHub
 config_additional_config = {
     'module_version': '1.1.1.dev0',
+    'last_tested_version': '18.0.0',
```
marcoskirsch: The name of this key is terrible. It sounds like the last version on Python module `nimodinst` that we tested is 18.0.0.

I think we should take this opportunity to rename this key to something that isn't confusing like `latest_runtime_version_tested_against`

README.rst
texasaggie97: Yes, but the same information shows up in DLL version information.

src/nimodinst/metadata/config_addon.py
texasaggie97: Changed key to `latest_runtime_version_tested_against`

PR #990, opened 2019-05-30
Add Switch Executive

texasaggie97:
* Add Switch Executive back
* Update Switch Executive metadata
* Unit


build/templates/status.inc.mako
```diff
@@ -11,7 +11,8 @@
 <%
 table_contents = []
 table_contents.append(['{0} ({1})'.format(driver_name, module_name), '', ])
-table_contents.append(['Driver Version Tested Against', config['last_tested_version']])
+table_contents.append(['Driver Version Tested Against', config['latest_runtime_version_tested_against']])
+table_contents.append(['Driver API Version', config['api_version']])
```
marcoskirsch: I wish changes like these, which are independent of the main prupose of the PR, had been pushed in their own PR.

docs/conf.py
```diff
@@ -55,15 +55,15 @@
 
 # General information about the project.
 project = 'NI Modular Instruments Python API'
-copyright = '2017-2018, National Instruments'
```
marcoskirsch: We should consider querying and using the current year for the second value.

docs/conf.py
```diff
@@ -55,15 +55,15 @@
...
+copyright = '2017-2019, National Instruments'
 author = 'National Instruments'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The full version, including alpha/beta/rc tags.
-release = '1.1.1.dev0'
+release = '0.2.1.dev0'
```
marcoskirsch: Why is this being downgraded?

docs/nidcpower/class.rst
```diff
@@ -3430,7 +3470,7 @@ pulse_current_level_range
             +----------------+------------+
             | Permissions    | read-write |
             +----------------+------------+
-            | Channel Based  | True       |
+            | Channel Based  | Yes        |
```
marcoskirsch: There are too many changes here that are independent from adding NISE. Can those occur in their own PR?

build/templates/status.inc.mako
texasaggie97: That would have required a bunch more work:
* Manually update all old metadata to new format
* Consume new format
* Update metadata with generated metadata that also uses new format

docs/conf.py
texasaggie97: We do. This is a generated file. From the mako template:
```
copyright = '2017-${current_year}, National Instruments'
```

docs/conf.py
texasaggie97: Because we only have one docs/conf.py and the version will be the last driver to build. nise hasn't had a 1.0 yet.

docs/nidcpower/class.rst
texasaggie97: See above about why.

PR #1005, opened 2019-06-17
[RFC] Add NI-TClk support

### LINKED ISSUE #293: Add NI-TClk Support
Add NI-TClk support.

How do we want to do this? NI-TClk is different from the other drivers supported by `nimi-python` in a fairly major way. 

The C API for nitclk is not session based like other drivers. Instead, the functions are stateless and you pass in an array of sessions you have gotten from drivers. The only functions that take a single session are the get/set functions. Some typical function prototypes:
``` c
ViStatus _VI_FUNC niTClk_ConfigureForHomogeneousTriggers(ViUInt32 sessionCount,
                                                         const ViSession sessions[]);
ViStatus _VI_FUNC niTClk_SetAttributeViReal64(ViSession session,
                                              ViConstString channelName,
                                              ViAttr attributeId,
                                              ViReal64 value);
```

~~From the examples I have seen we always pass in the same array of sessions to all functions, never one set to one function and a different set to a different function.~~ This is not always the case.

Given this, I propose that:
* If the driver supports NI-TClk and nitclk is installed, add the attributes to the driver session
    * This will be via a `nitclk.properties()` object that the driver's `_SessionBase()` will contain.
        ``` python
        class _SessionBase(object):  # the driver session class
            @property
            def tclk(self):
                import nitclk
                return nitclk.properties(self.get_session_reference(), self._repeated_capability)
        ```
    * This will allow us to use the already existing repeated capabilities mechanism
* Create a `nitclk.Session()` object that will be instantiated with the list of sessions and will own all non-attribute function calls.

An example of how this might look:
``` python
import niscope
import nifgen
import nitclk

scope1 = niscope.Session('PXI1Slot2')
scope2 = niscope.Session('PXI1Slot3')
fgen = nifgen.Session('PXI1Slot42')
tclk_all = nitclk.Session([scope1, scope2, fgen])
tclk_scope = nitclk.Session([scope1, scope2])
scope1.tclk.sync_pulse_source = 'PXI_Trig0'
fgen.script_triggers[0].tclk.script_trigger_master_session = scope1.get_session_reference()
tclk_all.configure_for_homogeneous_triggers()
tclk_all.synchronize()
tclk_all.initiate()

while not cows_are_home:
    tclk_scope.initiate()

scope1.close()
scope2.close()
fgen.close()
```

~~Even though the `nitclk.Session()` doesn't need to call a close or do any other cleanup, it would still be a context manager.~~ No reason to do this.

Some alternatives would be:
* Make everything a separate, stateless function, similar to C and LabVIEW
* Have an attribute "session" but leave the other function as individual stateless functions
* Something else?

texasaggie97:
* Add nitclk to nimi-python - see below for details
* Fix #293 

* Flake8

Attached code does/~~does not~~ reflect this description
#### Definitions
* driver session - Class that encapsulates the session in Python - for example `niscope.Session`
* driver session num - this is the actual number that represents the session - for IVI drivers, this is returned when creating the IVI session
* tclk session num - this is the actual number that represents the session to tclk - this may or may not be the same actual number as driver session num
#### SessionReference
New class that encapsulates the tclk session num. This class does not open or close any sessions. It is instantiated with the tclk session num.

`nimi-python` based driver sessions will include a tclk member that is the SessionReference associated with that session.
``` python
import niscope
scope = niscope.Session('dev1')
print(type(scope.tclk))
```
Will produce something like `nitclk.SessionReference`

Any place that needs a tclk session num in the C API can accept:
* nimi-python Session (assuming NI-TClk support enabled)
* `nitclk.SessionReference`
* tclk session num

Any place that returns a tclk session num from the C API will return a `nitclk.SessionReference`

#### Function
NI-TClk is a stateless API, I.e. there is no session that ties all the different function calls together like other IVI based drivers. Because of this there is no `Session` object in the public API. Instead, the API consists of functions that take the list of sessions to act on. For example:
``` python
def synchronize(sessions, min_time):
    ...
```
``` python
import nifgen
import niscope
import nitclk

scope = niscope.Session('dev1')
fgen = nifgen.Session('dev2')
# scope1 configure
fgen.write_waveform(calculated_sine)
nitclk.configure_for_homogeneous_triggers([scope, fgen])
nitclk.synchronize([scope, fgen], 0.00042)
nitclk.initiate([fgen, scope])
wfm = scope.fetch()
# Do something with data
scope.close()
fgen.close()
```

`sessions` is an iterable where each item is one of the list above.

The reason for not using a session based API is that different calls to nitclk functions may need to act on a different subset of the sessions.

#### Properties
Because properties are set per session, we need to handle them differently.

Properties will be added to the `SessionReference` class. They will then be set just like properties on driver sessions are.

Then any nimi-python driver that claims support for nitclk will add a `tclk` attribute which will be a `nitclk.SessionReference` for that driver session. 
```python
import niscope

scope1 = niscope.Session('dev1')
scope2 = niscope.Session('dev2')
scope1.nitclk.sample_clock_delay = .0000042
scope2.nitclk].sample_clock_delay = .0000043
```

At this point we will not support repeated capabilities on NI-TClk properties. We will open an issue regarding the one property that does use repeated capabilities and use that to track priority and need for this feature.

#### Packaging
Any driver that declares in metadata that it supports nitclk will have a dependency added for the nitclk package. This way when you `pip install niscope` you will also get nitclk support. We know that if the driver is installed, then NI-TClk is installed as well, and we know that other ADEs also install NI-TClk support when the driver support is installed.

We could make nitclk being installed optional, but the cost of installing it is very small and it would complicate the generated code quite a bit in order to handle it not being installed. Unlike `numpy` which is optional, nitclk from PyPI doesn't install anything else whereas `numpy` installs DLLs in addition to the Python code.

#### Functions
Internally, there is a class that contains the actual implementation of the functions. The main purpose of this is to allow reusing the default function template as is. Not doing this would require either a separate template whose only difference is indentation, or adding variable indentation to the existing template which would make it very messy. @marcoskirsch didn't like variable indenting when originally attempted for something in the pre 1.0 time frame.

This class will be instantiated as a singleton and each call to the function will get this singleton and then call the actual implementation.
``` python
def _get_session_class():
    '''Internal function to return session singleton'''
    global _session_instance
    global _session_instance_lock

    with _session_instance_lock:
        if _session_instance is None:
            _session_instance = _Session()

        return _session_instance


def wait_until_done(sessions, timeout):
    '''wait_until_done

    <doc string snipped for brevity>
    '''
    return _get_session_class().wait_until_done(sessions, timeout)
```


epage:
> We could make nitclk being installed optional, but the cost of installing it is very small and it would complicate the generated code quite a bit in order to handle it not being installed. Unlike numpy which is optional, nitclk from PyPI doesn't install anything else whereas numpy installs DLLs in addition to the Python code.

Do the drivers require NI-TCLK be installed?  Is that an intrinsic requirement or an implementation detail?

From there, does having `.nitclk` property require NI-TCLK be installed or only if you actually call NI-TCLK functionality?

marcoskirsch:
>sessions is an iterable where each item is the actual session number, or an object with a get_session() method that returns the actual session number (get_session() support will be added in a future commit)

I'm saddened by the need to expose session handles in Python. What if we didn't publicly expose it in the other APIs (meaning, no `get_session()` method) and simply made it so that `nitclk` knows about this implementation detail? I think that's probably OK, since they are all maintained by NI.

marcoskirsch:
> Any driver that declares in metadata that it supports nitclk will have a dependency added for the nitclk package.

I think this is fine.

marcoskirsch:
> > This class will be instantiated as a singleton and each call to the function will get this singleton and then call the actual implementation.
> 
> * Delay initialization of the internal session causes extra synchronization points between all the APIs.  That seems like a bad thing
> * Is there a downside to every user operation sharing the same internal TClk session?

Sounds like the reason why the internal implementation uses a Session singleton is to make it easy to re-use the existing templates in nimi-python.

Perhaps we should ask the questions:

* Is NI-TClk as an API different enough that it should be its own thing separate from nimi-python?
* Is NI-TClk small enough and a one-off in which codegeneration is not the best course of action?

I think the answer to both is "no", there is enough benefit in leveraging the nimi-python infrastructure and in single sourcing of API metadata. But want to make sure we think about it.

texasaggie97:
> > We could make nitclk being installed optional, but the cost of installing it is very small and it would complicate the generated code quite a bit in order to handle it not being installed. Unlike numpy which is optional, nitclk from PyPI doesn't install anything else whereas numpy installs DLLs in addition to the Python code.

> Do the drivers require NI-TCLK be installed? Is that an intrinsic requirement or an implementation detail?

The drivers do require that NI-TClk be installed and have a runtime dependency
 
> From there, does having `.nitclk` property require NI-TCLK be installed or only if you actually call NI-TCLK functionality?

The main API DLL (nitclk_64.dll) will not be loaded until setting a property (or calling a function) but other underlying DLLs will already be loaded by the driver(s).

> > This class will be instantiated as a singleton and each call to the function will get this singleton and then call the actual implementation.

> * Delay initialization of the internal session causes extra synchronization points between all the APIs.  That seems like a bad thing

There is not an NI-TClk session. Each entry point takes the sessions from the driver sessions. Even getting or setting an attribute takes a driver session.

texasaggie97:
> > sessions is an iterable where each item is the actual session number, or an object with a get_session() method that returns the actual session number (get_session() support will be added in a future commit)

> I'm saddened by the need to expose session handles in Python. What if we didn't publicly expose it in the other APIs (meaning, no `get_session()` method) and simply made it so that `nitclk` knows about this implementation detail? I think that's probably OK, since they are all maintained by NI.

I know there has been at least one request to be able to use this from other APIs not supported by nimi-python. This is why I didn't want to rely on any internal implementation details of how we did things.

epage:
> I know there has been at least one request to be able to use this from other APIs not supported by nimi-python. This is why I didn't want to rely on any internal implementation details of how we did things.

if this is the case, I'd recommend defining an ABC for sessions that work with TClk.

texasaggie97:
> > I know there has been at least one request to be able to use this from other APIs not supported by nimi-python. This is why I didn't want to rely on any internal implementation details of how we did things.

> if this is the case, I'd recommend defining an ABC for sessions that work with TClk.

What would that look like? NI-TClk needs the internal IVI session, or whatever the driver uses to identify sessions. I was going to just rely on duck typing to see if there is a `get_session()`. If there isn't, assume it is just the actual session and let NI-TClk validate it in the driver.

epage:
>What would that look like? NI-TClk needs the internal IVI session, or whatever the driver uses to identify sessions. I was going to just rely on duck typing to see if there is a get_session(). If there isn't, assume it is just the actual session and let NI-TClk validate it in the driver.

```python
import abc

class TClkDriverSession(abc.ABC):

    @abc.abstractmethod
    def get_session(self):
        ...
```

In TClk, you can
- Set the type annotation to be a `typing.Union[TClkDriverSession, <sessiontype>]`
- `if isinstance(TClkDriverSession)`

Random notes:
- I just made up a name for the ABC.  I'm sure there is a better one
- I'd personally recommend a more descriptive name for the method.

So why do this over duck typing?
- We can better document the requirement
- We can clarify the requirement via type checking
- We can verify the user's intent rather than getting something unrelated.

homechicken80:
> > > We could make nitclk being installed optional, but the cost of installing it is very small and it would complicate the generated code quite a bit in order to handle it not being installed. Unlike numpy which is optional, nitclk from PyPI doesn't install anything else whereas numpy installs DLLs in addition to the Python code.
> 
> > Do the drivers require NI-TCLK be installed? Is that an intrinsic requirement or an implementation detail?
> 
> The drivers do require that NI-TClk be installed and have a runtime dependency
> 
Technically the drivers don't depend on NI-TClk and NI-TClk doesn't depend on the drivers.  Both depend on a thin third component (NI-JEL pronounced Nigel).  However, NI-JEL doesn't currently ship with NI-TClk (although it doesn't have to).

marcoskirsch:
> I was going to just rely on duck typing to see if there is a get_session(). If there isn't, assume it is just the actual session and let NI-TClk validate it in the driver.

To be clear, _if_ we go this route, the method would be called `get_session_handle()` not `get_session()`. The distinction is important.

generated/nitclk/session.py
```diff
@@ -0,0 +1,1140 @@
...
+            chassis/PC is 200 ns. If the specified value is less than 200 ns,
+            NI-TClk automatically coerces minTime to 200 ns. For multichassis
+            synchronization, adjust this value to account for propagation delays
+            through the various devices and cables.
+
+    '''
+    return _get_session_class().setup_for_sync_pulse_sender_synchronize(sessions, min_time)
+
+
+def synchronize(sessions, min_time):
```
auchter: Consider adding a default for `min_time` like the LabVIEW VIs do. The majority of users won't need to adjust this to anything greater than 200 ns.

auchter: Also, `min_time` is a bit of an odd name for the parameter... `min_period` (or `min_tclk_period` if you're not into the whole brevity thing) would probably be more descriptive.

texasaggie97: Changed the name to `min_tclk_period`
Set default of 0
Allow datetime.timedelta() type

generated/nitclk/session.py
```diff
@@ -0,0 +1,1140 @@
...
+            chassis/PC is 200 ns. If the specified value is less than 200 ns,
+            NI-TClk automatically coerces minTime to 200 ns. For multichassis
+            synchronization, adjust this value to account for propagation delays
+            through the various devices and cables.
+
+    '''
+    return _get_session_class().synchronize(sessions, min_time)
+
+
+def syncronize_to_sync_pulse_sender(sessions, min_time):
```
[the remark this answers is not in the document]
texasaggie97: Fixed

texasaggie97:
I believe all comments have been addressed and the code has been updated.

If there are no other concerns, is NI-TClk's Python API approved?

marcoskirsch:
> If there are no other concerns, is NI-TClk's Python API approved?

I think it's ready for a more formal review.

texasaggie97:
Closed by other PRs culminating with #1054

PR #1011, opened 2019-06-21
Add documentation for close()

### LINKED ISSUE #1010: Public documentation is missing for close()
* Description of issue

Documentation on readthedocs is missing an entry for `close()`

texasaggie97:
* Add documentation for close()
* Updated .codecov.yml for current state
* Fix #1010 


build/helper/documentation_snippets.py
```diff
@@ -96,5 +96,97 @@
 { 'simulate': True, 'driver_setup': { 'Model': '<model number>',  'BoardType': '<type>' } }
 '''
 
+default_close_function_doc = '''
+Closes the driver session and cleans up.
+'''
+
+close_function_note = '''
+This function is not needed when using the session context manager
```
marcoskirsch: Do you think clients will understand what "the session context manager" is?
Honest question.

texasaggie97: By default we will use the documentation from the driver's close function if it exists. 

The note will be added regardless of whether the default description is used or the description from the driver. I am open on what the note should say to make clear that normally you don't need to use this function. I agree this isn't very good, but couldn't come up with a better way to say this.

marcoskirsch: Looking at docs:
https://docs.python.org/3/whatsnew/2.6.html#pep-343-the-with-statement

They talk of the `with` statement, rather than of context managers.
I suggest following this terminology here as well as in initate/abort/initalize.

Let's refine that in its own PR so we don't hold this one.

build/helper/documentation_snippets.py
```diff
@@ -96,5 +96,97 @@
...
+'''
+
+def close_function_def_for_doc(functions):
+    # This is very specific to IVI. We look for a 'close' function and if we find it,
```
marcoskirsch: NISE and niModInst have close. How is this specific to IVI?

docs/nidigital/class.rst
```diff
@@ -3685,6 +3687,21 @@ clock_generator_initiate
...
+
+    .. py:currentmodule:: nidigital.Session
+
+    .. py:method:: close()
+
+            Closes the driver session and cleans up.
+
+            
+
+            .. note:: This method is not needed when using the session context manager
```
marcoskirsch: We shall refine the wording.

generated/nidcpower/session.py
```diff
@@ -3724,6 +3724,26 @@ def initiate(self):
...
+        that NI-DCPower reserves. If power output is enabled when you call this
+        method, the output channels remain in their existing state and
+        continue providing power. Use the ConfigureOutputEnabled
+        method to disable power output on a per channel basis. Use the
+        reset method to disable power output on all channel(s).
+
+        **Related Topics:**
+
+        `Programming
+        States <REPLACE_DRIVER_SPECIFIC_URL_1(programmingstates)>`__
```
marcoskirsch: ?

build/helper/documentation_snippets.py
marcoskirsch: https://github.com/ni/nimi-python/issues/1020

build/helper/documentation_snippets.py
texasaggie97: Changed to `# This is very specific to session based APIs ...`

generated/nidcpower/session.py
texasaggie97: This is to make fixing up links in help. The original link is internal to the help file and does not work here, so I replace it with something that can be easily searched for.

This is not new, pre 1.0.

build/helper/documentation_snippets.py
```diff
@@ -96,6 +96,14 @@
 { 'simulate': True, 'driver_setup': { 'Model': '<model number>',  'BoardType': '<type>' } }
 '''
 
+default_close_function_doc = '''
+Closes the driver session and cleans up.
```
marcoskirsch: Should we add: "After calling this, the Session object is no longer valid and cannot be used."

build/helper/documentation_snippets.py
```diff
@@ -96,6 +96,14 @@
...
+'''
+
+close_function_note = '''
+This function is not needed when using the session context manager
+'''
+
 default_initiate_function_doc = '''
 Calls initiate
```
marcoskirsch: > "`Calls initiate`"
Is this the documentation we'd get for `initiate()`?
It's useless, no?

build/helper/documentation_snippets.py
```diff
@@ -105,6 +113,45 @@
...
+            function_def['documentation'] = {}
+        if 'description' not in function_def['documentation']:
+            function_def['documentation']['description'] = default_close_function_doc
+        if 'note' not in function_def['documentation']:
+            function_def['documentation']['note'] = []
+        if type(function_def['documentation']['note']) is not list:
+            function_def['documentation']['note'] = [function_def['documentation']['note']]
+        function_def['documentation']['note'].append(close_function_note)
+        function_def['python_name'] = 'close'
+    else:
```
marcoskirsch: Maybe we should consider breaking the build if we don't have documentation for `close` (or `initiate`).
I assume having the `else` case is nice right now because `nidigital` doesn't have documentation. If that's the reason for having this, we should consider adding documentation via add-on just for these two functions.

build/helper/documentation_snippets.py
texasaggie97: Yes, that seems like a good thing to add.

build/helper/documentation_snippets.py
texasaggie97: This is only used if initiate doesn't have its own documentation. In theory, you should never see this. Given that, I am open to a better, generic description.

marcoskirsch: > This is only used if initiate doesn't have its own documentation.

I'd rather have this be a build error.

build/helper/documentation_snippets.py
texasaggie97: Done. Assert when close or initiate function doesn't exist

Also found that `'close_function'` in config metadata was incorrect. Fixed here and internally.

.codecov.yml
```diff
@@ -6,9 +6,10 @@ coverage:
   status:
     project:
       default:
-        threshhold: 5%
+        target: 90%
```
marcoskirsch: This change appears to be unrelated to the PR, as it is not mentioned in the description.
Is this a merge problem?

build/helper/documentation_snippets.py
```diff
@@ -105,6 +114,29 @@
 '''
 
 
+def close_function_def_for_doc(functions, config):
+    # This is very specific to session based APIs. We look for a 'close' function and if we find it,
```
marcoskirsch: I think this is missing context and is now a bit outdated. Suggestion:


> This is very specific to session based APIs. We look for a 'close' function in the metadata which is to become a code-generated "private" method of Session for which we provide a public wrapper.
Copy its documentation so we apply it to the said public wrapper.

build/helper/documentation_snippets.py
```diff
@@ -96,6 +96,15 @@
 { 'simulate': True, 'driver_setup': { 'Model': '<model number>',  'BoardType': '<type>' } }
 '''
 
+default_close_function_doc = '''
+Closes the driver session and cleans up. After calling this the Session object
+is no lonber valid and cannot be used.
```
marcoskirsch: longer (typo)

build/helper/documentation_snippets.py
```diff
@@ -105,6 +114,29 @@
...
+        function_def['python_name'] = 'close'
+    else:
+        assert False, "No '{}' function defined".format(close_name)
+
+    return function_def
+
+
 def initiate_function_def_for_doc(functions, config):
     # This is very specific to the IVI state model which not all drivers in nimi-python follow.
     # We look for a 'close' function and if we find it, We will copy that and modify it to be
```
marcoskirsch: this is also outdated... there is no more "if we find it".

.codecov.yml
texasaggie97: It is the same change as is in #1023 to correctly control the limits.

I will add something to the description to mention it.

build/helper/documentation_snippets.py
texasaggie97: You want to require there be supplied documentation? Is that reasonable?

build/helper/documentation_snippets.py
texasaggie97: Changed

build/helper/documentation_snippets.py
texasaggie97: Fixed

build/helper/documentation_snippets.py
texasaggie97: Changed to something similar as for close.

PR #1018, opened 2019-08-29
Add initial nitclk metadata

texasaggie97:

* Initial nitclk metadata
* Step 1 towards #293 

* None


src/nitclk/metadata/config.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# This file is generated from NI-TClk API metadata version 255.0.0d0
+config = {
+    'api_version': '255.0.0d0',
+    'c_function_prefix': 'niTClk_',
+    'close_function': None,
+    'context_manager_name': {
+        'abort_function': None,
+        'initiate_function': 'Initiate',
+        'task': 'acquisition'
```
marcoskirsch: NI-TClk can be used for generation too (like in NI-FGEN).
Do you think it matters? Where does this get reflected?

src/nitclk/metadata/config.py
```diff
@@ -0,0 +1,40 @@
...
+                'name': 'nitclk_64.dll',
+                'type': 'cdll'
+            }
+        }
+    },
+    'metadata_version': '2.0',
+    'module_name': 'nitclk',
+    'repeated_capabilities': [
+    ],
+    'session_class_description': 'An NI-TClk session.',
```
marcoskirsch: Since there isn't an NI-TClk sessions, I'd leave this blank or even remove it if possible.

texasaggie97: This is not really true though. The TClk functions take a list of TClk sessions. True, they don't come from instantiating a tclk session object, but they do exist.

src/nitclk/metadata/config.py
texasaggie97: Don't know yet. Maybe nowhere? It is called initiate in the tclk API though.

marcoskirsch: I was referring to:
```
        'task': 'acquisition'
 ```

File: src/nitclk/metadata/config.py (hunk 2 above)
Comment (marcoskirsch) [reply src/nitclk/metadata/config.py]: > The TClk functions take a list of TClk sessions.

I thought they take a list of `nitclk.SessionReference` objects.
Is that what `'session_class_description'` is going to document? If so, it needs to be adjusted. It's a reference to an instrument / driver session, not an NI-TClk session.

File: src/nitclk/metadata/config.py (hunk 1 above)
Comment (texasaggie97) [reply src/nitclk/metadata/config.py]: Actually, I don't think nitclk can have a context manager, so removed.

File: src/nitclk/metadata/config.py (hunk 2 above)
Comment (texasaggie97) [reply src/nitclk/metadata/config.py]: You're right. I was confusing where this was used.

Removed.

File: src/nitclk/metadata/config.py (hunk 1 above)
Comment (marcoskirsch) [reply src/nitclk/metadata/config.py]: Is that because there is no "abort" function?

I think there is an expectation we've created for nimi-python users that you can use a context manager, and it would be bad not to support it in nitclk.

```
with nitclk.initiate(session_reference_list)
   # do stuff
# more code after aborting
```

The difference is that we'd need a context manager that on `__exit__` iterates over the `nitclk.SessionReference` objects and calls `abort()` on them. Which means `nitclk.SessionReference` objects need to internally hold on to `nifoo.Session`. But with duck-typing it should all "just work" I believe.

texasaggie97: We do not do this in other languages. Is that really the right way to program NI-TClk? It doesn't look like it.

marcoskirsch: I have created a detailed issue to track this separately since it's larger than just adding the metadata.
https://github.com/ni/nimi-python/issues/1024

PR #1019, opened 2019-08-29
Clean tox and fix build break (pypy3 again)

texasaggie97:

* Fixes pypy3 build break
    * Workaround for pytest issue pytest-dev/pytest#5807
    * Use older version of pytest for `pypy` and `pypy3`
* Removes some no longer needed version requirements

* Unit
* System

### DIFF
--- a/tools/build_release.py
+++ b/tools/build_release.py
@@ -8,7 +8,7 @@ import sys
 
 pp = pprint.PrettyPrinter(indent=4, width=100)
 
-default_python_cmd = ['c://Python27//python.exe']
+default_python_cmd = ['c:/Python27/python.exe']
 
 
 def configure_logging(lvl=logging.WARNING, logfile=None):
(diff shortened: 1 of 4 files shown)


tools/build_release.py
```diff
@@ -37,13 +37,13 @@ def main():
         * Update the changelog to show the version of the release
         * Change unreleased in TOC to new version
         * Commit to master
-    * `c:\Python36\python.exe tools\build_release.py --build --update --upload --release`
+    * `c:\\Python36\\python.exe tools\\build_release.py --build --update --upload --release`
```
marcoskirsch: Are you sure you need to escape the backslashes if they are inside the backticks?
At least here in GitHub markdown it doesn't appear to be the case.

`c:\\Python36\\python.exe tools\\build_release.py --build --update --upload --release`
c:\\Python36\\python.exe tools\\build_release.py --build --update --upload --release

texasaggie97: Either that or turn off that check in flake8 altogether. The new version started complaining that `\P` was an undefined escape sequence.

marcoskirsch: Makes sense, I thought it was in a comment bit it's actually inside a multi-line string literal. I think you got it right.

PR #1022, opened 2019-09-03
Add tclk/simple codegen

texasaggie97:

* Next step for adding nitclk support
* Build most of the files using unchanged templates
    * nitclk specific changes will be in a separate PR
    * Additional nitclk specific template(s) will be added in a separate PR
* Nothing for nitclk yet other than flake8 on generated files


docs/nitclk/class.rst
```diff
@@ -0,0 +1,960 @@
+nitclk.Session
+==============
+
+.. py:module:: nitclk
+
+.. py:class:: Session(self)
```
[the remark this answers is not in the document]
texasaggie97: As mentioned elsewhere, documentation for nitclk needs much cleanup.

Possibly completely overriding some of the files. For now, tclk documentation is not correct, even when enabled in a separate PR. All that one does is add it to the appropriate TOCs, it does not try to fix it.

docs/nitclk/installation.inc
```diff
@@ -0,0 +1,18 @@
+
+.. _nitclk_installation-section:
+
+Installation
+------------
+
+As a prerequisite to using the nitclk module, you must install the NI-TClk runtime on your system. Visit `ni.com/downloads <http://www.ni.com/downloads/>`_ to download the driver runtime for your devices.
```
marcoskirsch: I don't think this is accurate. I think NI-TClk is installed for you when you install the runtime for a driver that supports it.

marcoskirsch: I imagine that drivers that have NI-TClk property `nitclk.SessionReference` will need to install the `nitclk` module too. We can revisit all this then.

PR #1026, opened 2019-09-04
Add nitclk template changes

texasaggie97:

* Updates existing templates with changes required for nitclk
* Lowers targets for code coverage
    * Can be raised once tests for nitclk are added
    * Changed codecov patch limit to 35% and overall limit to 85%
    * Change coveralls delta limit from 0.5% to 1.0%
* Unit
* System


build/templates/_converters.py.mako
```diff
@@ -222,6 +222,36 @@ def convert_init_with_options_dictionary(values, encoding):
     return init_with_options_string
 
 
+% if config['module_name'] == 'nitclk':
+# nitclk specific converters
+def convert_to_nitclk_session_num(item):
+    '''Convert from supported objects to NI-TClk Session Num
```
marcoskirsch: Let's remove functionality so that this converter only works for Session classes (in other words "object with a `tclk` property of type `nitclk.SessionReference`
(one way to do things)

We can add more if/when justified.

build/templates/_converters.py.mako
```diff
@@ -222,6 +222,36 @@ def convert_init_with_options_dictionary(values, encoding):
     return init_with_options_string
 
 
+% if config['module_name'] == 'nitclk':
+# nitclk specific converters
+def convert_to_nitclk_session_num(item):
```
marcoskirsch: The term "NI-TClk Session Num" doesn't exist anywhere, right?

In our driver APIs in which NI-TClk is supported we have a VI called "Get Session Reference".
In C you just pass the ViSession.

I think `convert_session_reference_to_vi_session` would be pretty self explanatory.

build/templates/_converters.py.mako
```diff
@@ -222,6 +222,36 @@ def convert_init_with_options_dictionary(values, encoding):
     return init_with_options_string
 
 
+% if config['module_name'] == 'nitclk':
```
marcoskirsch: Similar to Attribute classes, not all drivers use all converters yet the code is there as dead code. I'd leave it alone here as well. Less conditional logic all over the place.

build/templates/_converters.py.mako
```diff
@@ -222,6 +222,36 @@ def convert_init_with_options_dictionary(values, encoding):
...
+    try:
+        return item.get_session_number()
+    except KeyError:
+        pass
+
+    # If we haven't gotten a SessionReference, we assume the item is the actual nitclk session num and return it
+    return item
+
+
+def convert_to_nitclk_session_num_list(item_list):
```
marcoskirsch: `convert_session_reference_list_to_vi_session_list` ?

docs/nifgen/class.rst
```diff
@@ -5338,11 +5338,11 @@ create_waveform_from_file_f64
                 |
                 | ****Default Value**:** :py:data:`~nifgen.ByteOrder.LITTLE`
 
-                +-------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------+
+                +-------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------+
```
marcoskirsch: needs to be merged from master... (encoding screwing up width problem)

generated/nitclk/__init__.py
```diff
@@ -6,7 +6,18 @@
 
 from nitclk.errors import DriverWarning   # noqa: F401
 from nitclk.errors import Error     # noqa: F401
```
marcoskirsch: minor: why are there different number of spaces before the `#`?

generated/nitclk/__init__.py
```diff
@@ -6,7 +6,18 @@
...
-from nitclk.session import Session  # noqa: F401
+from nitclk.session import SessionReference  # noqa: F401
+
+# Function imports
+from nitclk.session import configure_for_homogeneous_triggers  # noqa: F401
```
marcoskirsch: > F401 | module imported but unused

Is this needed?

build/templates/__init__.py.mako
```diff
@@ -14,7 +17,22 @@ from ${module_name}.enums import *          # noqa: F403,F401,H303
...
+# nitclk, we import the Session class like before. If it is nitclk then we will
+# import each function and the SessionReference class
+%>\
+% if config['module_name'] != 'nitclk':
 from ${module_name}.session import Session  # noqa: F401
+% else:
+from ${module_name}.session import SessionReference  # noqa: F401
+
+# Function imports
+%   for func_name in sorted([functions[k]['python_name'] for k in functions if not functions[k]['render_in_session_base']]):
```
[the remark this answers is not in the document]
texasaggie97: Added this comment:
``` python
# There are two types of functions in nitclk: functions that take a single sessions (get/set attribute) and 
# functions that take a list of sessions. The functions that take a list of sessions are considered the main
# API for this and these are the ones we need to explicitly import. We can tell which type of function each
# one is based on 'render_in_session_base'. When this is True, it means it is one of the get/set attribute
# funtions, so we skip it.
```

build/templates/_converters.py.mako
texasaggie97: The type is `ViSession`, but more specifically it is the nitclk session number. For some APIs this is different than the ViSession that is the main session number. This is why I named it that. If that is a special enough case, I can change it, but it may be misleading in some instances.

generated/nitclk/__init__.py
texasaggie97: Yes. It is imported and not used in this file. `flake8` doesn't care that it is `__init__.py` and is supposed to do things like this.

generated/nitclk/__init__.py
texasaggie97: At one point they all lined up. Since they no longer do, and some lines have variables in them so we don't know how many spaces are required, made them all `<space><space>#<space>`

build/templates/_converters.py.mako
marcoskirsch: ok... so "session num" is to mean "ViSession used by NI-TClk to refer to an instrument session.

Well, at least spell out `number`.

build/templates/__init__.py.mako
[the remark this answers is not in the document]
marcoskirsch: Ok, sounds like you are reusing `render_in_session_base` to mean "methods in SessionReference" and the rest of the functions (the ones not `render_in_session_base`) are to become functions under `nitclk`.

Allow me to suggest clearer wording based on my understanding, since this is "clever".

```
There two types of functions in `nitclk`:

1. Functions that take a single SessionReference (get/set attribute)
2. Functions that take in a list of SessionReference

The second type are the public functions that clients will call, so we need to import them explicitly into the `nitclk` namespace. We are using the `render_in_session_base` metadata in order to distinguish them
```

marcoskirsch:
codecov dropped a bunch. If the drop is expected, then we should change the config in the same PR so the new baseline is established. We should work on getting coverage back up in subsequent PRs (we may want to open an issue in order to track).

build/templates/__init__.py.mako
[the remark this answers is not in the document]
texasaggie97: Changed

build/templates/_attributes.py.mako
```diff
@@ -112,4 +113,18 @@ class AttributeEnum(object):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+% if config['module_name'] == 'nitclk':
```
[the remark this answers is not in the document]
texasaggie97: See #1039 for a possible solution, but for this PR, always add for all drivers.

build/templates/_converters.py.mako
texasaggie97: See #1039 for a possible solution, but for this PR, always add for all drivers.

build/templates/_converters.py.mako
texasaggie97: Changed to `number`

docs/nifgen/class.rst
texasaggie97: Changed

build/templates/_converters.py.mako
marcoskirsch: Opened #1042 to follow up

PR #1027, opened 2019-09-05
[#1026] Add tclk session template

texasaggie97:
* Add nitclk session template (based on #1005)
* Build session.py


generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+        return (library_type * len(value))(*value)
+    else:
+        if library_type is not None and size is not None:
+            return (library_type * size)()
+        else:
+            return None
+
+
+class SessionReference(object):
+    '''Properties container for NI-TClk attributes.'''
```
marcoskirsch: That's a strange docstring, doesn't say much. Suggestion:

>Reference to an instrument's session. NI-TClk specific properties can be configured through this.

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+    Empty string. Empty string is a valid value, indicating that the signal is  not exported.
+    PXI Devices -  'PXI_Trig0' through  'PXI_Trig7' and device-specific settings
+    PCI Devices -  'RTSI_0' through  'RTSI_7' and device-specific settings
+    Examples of Device-Specific Settings
+    - NI PXI-5122 supports  'PFI0' and  'PFI1'
+    - NI PXI-5421 supports  'PFI0',  'PFI1',  'PFI4', and  'PFI5'
+    - NI PXI-6551/6552 supports  'PFI0',  'PFI1',  'PFI2', and  'PFI3'
+    Default Value is empty string
+    '''
+    pause_trigger_master_session = _attributes.AttributeViInt32SessionReference(6)
```
marcoskirsch: I think this class should be `AttributeSessionReference`, since that's the Python-facing type.

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+    PXI Devices -  'PXI_Trig0' through  'PXI_Trig7' and device-specific settings
+    PCI Devices -  'RTSI_0' through  'RTSI_7' and device-specific settings
+    Examples of Device-Specific Settings
+    - NI PXI-5122 supports  'PFI0' and  'PFI1'
+    - NI PXI-5421 supports  'PFI0',  'PFI1',  'PFI2', and  'PFI3'
+    - NI PXI-6551/6552 supports  'PFI0',  'PFI1',  'PFI2', and  'PFI3'
+    Default Value - Empty string. This default value directs  synchronize to set this property when all the synchronized devices  are in one PXI chassis. To synchronize a multichassis system, you must set  this property before calling synchronize.
+    '''
+    tclk_actual_period = _attributes.AttributeViReal64(8)
+    '''Type: float
```
marcoskirsch: Do we want to support time data type?
That may be for its own PR. We probably will need to review the whole API once we're done with codegen.

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+
+    Indicates the computed TClk period that will be used during the acquisition.
+    '''
+
+    def __init__(self, session, repeated_capability_list='', encoding='windows-1251'):
```
marcoskirsch: I thought we were not going to support `repeated_capability_list`?
Should `session` be `session_handle` or `vi`? (is it the number?)

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+            It is expected for _get_error to raise when the session is invalid
+            (IVI spec requires GetError to fail).
+            Use _error_message instead. It doesn't require a session.
+            '''
+            error_string = self._get_extended_error_info()
+            return error_string
+        except errors.Error:
+            return "Failed to retrieve error description."
+
+    def get_session_number(self):
```
marcoskirsch: why does this need to be public?

(member probably should be _session_number)

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+        channel_name_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case C010
+        attribute_id_ctype = _visatype.ViAttr(attribute_id)  # case S150
+        value_ctype = ctypes.create_string_buffer(value.encode(self._encoding))  # case C020
+        error_code = self._library.niTClk_SetAttributeViString(session_ctype, channel_name_ctype, attribute_id_ctype, value_ctype)
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
+        return
+
+
+class _Session(object):
+    '''Hidden class for template reuse'''
```
marcoskirsch: I don't think docstrings in generated code should reference code-generation aspects. We should document what the class does in the context of the code it is documenting.

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+
+    def __init__(self):
+        '''Hidden class for template reuse'''
```
marcoskirsch: ?

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+        session_count_ctype = _visatype.ViUInt32(0 if sessions is None else len(sessions))  # case S160
+        sessions_ctype = get_ctypes_pointer_for_buffer(value=_converters.convert_to_nitclk_session_num_list(sessions), library_type=_visatype.ViSession)  # case B520
+        error_code = self._library.niTClk_ConfigureForHomogeneousTriggers(session_count_ctype, sessions_ctype)
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
+        return
+
+    def finish_sync_pulse_sender_synchronize(self, sessions, min_time):
+        r'''finish_sync_pulse_sender_synchronize
+
+        TBD
```
marcoskirsch: :)

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+
+        Args:
+            sessions (list of int): sessions is an array of sessions that are being synchronized.
```
marcoskirsch: needs converted (can be addressed in separate PR)

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+        error_string_size_ctype = _visatype.ViUInt32()  # case S170
+        error_code = self._library.niTClk_GetExtendedErrorInfo(error_string_ctype, error_string_size_ctype)
+        errors.handle_error(self, error_code, ignore_warnings=True, is_error_handling=True)
+        error_string_size_ctype = _visatype.ViUInt32(error_code)  # case S180
+        error_string_ctype = (_visatype.ViChar * error_string_size_ctype.value)()  # case C060
+        error_code = self._library.niTClk_GetExtendedErrorInfo(error_string_ctype, error_string_size_ctype)
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=True)
+        return error_string_ctype.value.decode(self._encoding)
+
+    def init_for_documentation(self):
```
marcoskirsch: ?

generated/nitclk/session.py
```diff
@@ -0,0 +1,1156 @@
...
+            chassis/PC is 200 ns. If the specified value is less than 200 ns,
+            NI-TClk automatically coerces minTime to 200 ns. For multichassis
+            synchronization, adjust this value to account for propagation delays
+            through the various devices and cables.
+
+    '''
+    return _get_session_class().finish_sync_pulse_sender_synchronize(sessions, min_time)
+
+
+def init_for_documentation(self):
```
marcoskirsch: I'm confused why the methods show up twice in this file.

src/nitclk/templates/session.py.mako
```diff
@@ -0,0 +1,217 @@
...
+    module_name       = config['module_name']
+    c_function_prefix = config['c_function_prefix']
+
+    functions = helper.filter_codegen_functions(functions)
+%>\
+
+import array
+import ctypes
+import datetime
+import threading
```
marcoskirsch: are we using this? I suppose flake8 complains if not?

src/nitclk/templates/session.py.mako
```diff
@@ -0,0 +1,217 @@
...
+
+% if attributes:
```
marcoskirsch: needed?

src/nitclk/templates/session.py.mako
```diff
@@ -0,0 +1,217 @@
...
+
+# Used for __repr__ and __str__
+import pprint
+pp = pprint.PrettyPrinter(indent=4)
+
+_session_instance = None
+_session_instance_lock = threading.Lock()
+
+
+# Helper functions for creating ctypes needed for calling into the driver DLL
```
marcoskirsch: We should move to a separate file for DRY. This can be own PR.

generated/nitclk/session.py
texasaggie97: Documentation will be updated in a different PR. To keep them small.

generated/nitclk/session.py
texasaggie97: Yes, this should probably be one of the `datetime` based types. We will definitely need to review the API as a whole once things are more complete.

generated/nitclk/session.py
texasaggie97: `session` came from the initial metadata (CVI) but we can make it what ever we want.

`repeated_capabilities_list` removed. Gives you more files to review.

generated/nitclk/session.py
texasaggie97: Member name is based on metadata. Changed

Not sure that it does need to be. I can't think of a use case where someone will need to get it. Made private

generated/nitclk/session.py
texasaggie97: Removed in a later PR - not needed any more.

generated/nitclk/session.py
texasaggie97: It is an implementation detail: From #1005 

> Internally, there is a class that contains the actual implementation of the functions. The main purpose of this is to allow reusing the default function template as is. Not doing this would require either a separate template whose only difference is indentation, or adding variable indentation to the existing template which would make it very messy. @marcoskirsch didn't like variable indenting when originally attempted for something in the pre 1.0 time frame.

src/nitclk/templates/session.py.mako
texasaggie97: No, removed.

src/nitclk/templates/session.py.mako
texasaggie97: Yes, it should be separate since it will touch many different places.

src/nitclk/templates/session.py.mako
texasaggie97: We are, for now. Removed in later PR

generated/nitclk/session.py
texasaggie97: Changed

generated/nitclk/session.py
texasaggie97: Made longer to explain what is going on. The purpose of this class _is_ code generation related though so the comment is still about it, just more wordy.

generated/nitclk/session.py
texasaggie97: Unlike other drivers, there isn't different documentation for the class vs `__init__()`, so removing from `__init__()`

generated/nitclk/session.py
texasaggie97: TBD here means that there was no documentation from the source.

generated/nitclk/session.py
texasaggie97: What do you mean by "needs converted"?

Documentation will be cleaned up in a later PR, but only what is generated, not in the content itself. So "list of int" will change but "sessions is an array of sessions that are being synchronized." will not

marcoskirsch: I meant "needs a converter".

generated/nitclk/session.py
marcoskirsch: So the production code is made more complex and more unlike what a human would directly write in order to simplify the templates?

docs/conf.py
```diff
@@ -63,7 +63,7 @@
 # built documents.
 #
 # The full version, including alpha/beta/rc tags.
-release = '1.1.3.dev0'
+release = '0.1.0.dev0'
```
marcoskirsch: Is this right? It appears to be a global version.

texasaggie97: It is global, but I wasn't going to worry about it until we are ready to release.

We also need to figure out a better way to handle it, but not this PR.

generated/nitclk/session.py
texasaggie97: Correct. There is one extra level of indirection in order to reuse templates.

PR #1037, opened 2019-09-12
Use python3 for everything but tests

texasaggie97:

* Use python3 for all steps other than unit and system tests. There's no need to do codegen with an old version of Python and we may want to take advantage of Python 3 features in the future. 
* In Travis-CI explicitly use python3 for those steps regardless of version being used for tests.
* Update instructions.

* Unit
* System


CONTRIBUTING.md
```diff
@@ -100,21 +100,24 @@ In order to run **[nimi-python](https://github.com/ni/nimi-python)** System Test
...
    * Run [flake8](http://flake8.pycqa.org/)
    * Generate [HTML documentation](http://www.sphinx-doc.org/)
 
+   The second line will iterate over all python versions and run the NI-FAKE unit tests
+   for each version that is installed and registered using `tox --scan`
+
 1. To clean everything and start fresh, type:
 
-         tox -e clean
+         <path to python3>/python3 -m tox -e clean
```
marcoskirsch: Do you think we should, instead of writing `<path to python3>` everytwhere, teach how to add it to the path?

texasaggie97: Some places we explicitly need python 3.x and other places we explicitly need python 2.7.

At least for now.

tox.ini
```diff
@@ -3,7 +3,7 @@
 # test suite on all supported python versions. To use it, "pip install tox"
 # and then run "tox" from this directory.
 [tox]
-envlist = build_test, codegen, py{27,34,35,36,37,38,py,py3}-test, flake8, docs, pkg
+envlist = py{27,34,35,36,37,38,py,py3}-test
```
sbethur: What was the reason for this change? The old default of do-everything-except-run-system-tests was convenient.

texasaggie97: The way tox work is that it will run each environment in the envlist, even if previous environments fail.

It also has the ability to expand expressions so something like `py{27,36}-test` becomes `py27-test` and `py36-test`.

I find it much easier to have command that I use to build that explicitly calls each environment so that it can stop if their is a failure. Doing this though can't take advantage of the expansion.

tox.ini
```diff
@@ -70,34 +70,34 @@ commands =
...
+    clean: python3 --version
+    clean: python3 -c "import platform; print(platform.architecture())"
     clean: make clean {posargs}
-    codegen: python --version
-    codegen: python -c "import platform; print(platform.architecture())"
-    codegen: python -m pip install pip --upgrade pip
+    codegen: python3 --version
+    codegen: python3 -c "import platform; print(platform.architecture())"
+    codegen: python3 -m pip install pip --upgrade pip
     codegen: make {posargs} PYTHON_CMD=python
```
sbethur: Shouldn't this be changed to python3 as well?

texasaggie97: It shouldn't matter since in the venv that is created both python and python3 should be the same thing.

However, I changed it anyway to be consistent.

texasaggie97:
test this please

texasaggie97:
test this please

texasaggie97:
test this please

texasaggie97:
test this please

PR #1048, opened 2019-09-19
Correct dev status for PyPI

### LINKED ISSUE #1046: "Development status" in PyPI is inaccurate for nimi-python modules
# Description of issue

All our modules are marked as "4 - Beta".
Most should be " 5 - Production/Stable".
nidigital should be Pre-Alpha.

texasaggie97:

* We were not correctly setting the Development Status level in `setup.py`
* Based on module version
    * version >= 1.0
        * .devN or .aN - Alpha
        * .bN, .cN or .rcN - Beta
        * \<nothing\> or .postN  - Stable
    * version < 1.0 and version >= 0.5 - Beta
    * version < 0.5 - Alpha
* Fix #1046 
* Visual inspection of generated `setup.py`


generated/niswitch/setup.py
```diff
@@ -51,7 +51,7 @@ def read_contents(file_to_read):
     tests_require=['pytest'],
     test_suite='tests',
     classifiers=[
-        "Development Status :: 4 - Beta",
+        "Development Status :: 3 - Alpha",
```
marcoskirsch: I'm confused... This should not say Alpha.

build/templates/setup.py.mako
```diff
@@ -2,7 +2,34 @@
 # -*- coding: utf-8 -*-
 # This file was generated
 <%
-config        = template_parameters['metadata'].config
+config         = template_parameters['metadata'].config
+module_version = config['module_version']
+
+from packaging.version import Version
```
marcoskirsch: you should factor out a function that determines the dev_status, so we can unit test.

IMO:

1 - Planning: Starting at 0.0
2 - Pre-Alpha: Starting at 0.3
3 - Alpha: Starting at 0.5
4 - Beta: Starting at 0.9
5 - Production/Stable Starting at 1.0
6 - Mature N/A
7 - Inactive N/A

generated/niswitch/setup.py
texasaggie97: Yes, it should because the current version is X.Y.Z.dev1. Once we go to release to PyPI, .dev1 will go away and this will be updated to Stable.

I updated the comments and CHANGELOG.md to address this but I hadn't updated the PR with this level of granularity. Added now.

build/templates/setup.py.mako
texasaggie97: You think we need this level of granularity?

What about one the driver has gone to 1.0? Do we still need `Planning` and `Pre-Alpha`? See PR description for current algorithm for drivers > 1.0.

texasaggie97: Moved into separate function and tested. Did not add all the granularity. Waiting on comment.

marcoskirsch: I think how we have it is fine.

PR #1057, opened 2019-09-25
Add repeated capabilities to Session() documentation

texasaggie97:

* Add some documentation about repeated capabilities and which ones are supported by the particular driver

* Visual inspection of html pages


build/templates/class.rst.mako
```diff
@@ -62,6 +62,61 @@ ${helper.get_rst_header_snippet(item, '-')}
...
+    :language: c
+
+.. role:: python(code)
+    :language: python
+
+${helper.get_rst_header_snippet('Repeated Capabilities', '=')}
+
+    Repeated capabilities attributes are used to set the `channel_string` parameter to the
+    underlying driver function call. This can be the actual function based on the :py:class:`Session`
+    method being called, or it can be the appropriate Get/Set Attribute function, such as :c:`${config['c_function_prefix']}SetAttributeViInt32()`.
```
marcoskirsch: * I don't think we should talk about how this works. The fact that there is an "underlying driver function call" is an implementation detail to the Python customer. We should document this purely from the perspective of the Python customer.
* We should call them "properties" in Python, since they are presented as Python properties.

marcoskirsch: I'm going to take a stab, feel free to improve or push back:

>This class supports repeated capabilites. These are multiple instances of the same type of functionality within the session.
>Repeated capability containers are properties of `Session` and can be indexed using a string, list, tuple, or slice (range).
> The repeated capability containers supported by this driver are:
```

The prefix is an implementation detail in my opinion. As a fact, I think we should downplay the fact that you can pass a string at all, instead use the more pythonic aproaches whenever we have examples.

What do you think?

File: build/templates/class.rst.mako [hunk 2]
```diff
@@ -62,6 +62,61 @@ ${helper.get_rst_header_snippet(item, '-')}
...
+
+        .. code:: python
+
+            session.${name}['0-2'].channel_enabled = True
+
+        passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
+
+        If an invalid repeated capability is passed to the driver, the driver will return an error.
+
+        You can also explicitly use the prefix as part of the parameter, but it must be the correct prefix
```
marcoskirsch: Too much information?

docs/nidcpower/class.rst
```diff
@@ -1569,6 +1569,39 @@ wait_for_event
...
+    Some repeated capabilities use a prefix before the number and this is optional
+
+channels
+--------
+
+    .. py:attribute:: nidcpower.Session.channels[]
+
+        .. code:: python
+
+            session.channels['0-2'].channel_enabled = True
```
marcoskirsch: Let's use range or a list literal instead which is more pythonic than our convoluted syntax.

build/templates/class.rst.mako
texasaggie97: Removed references to and examples of prefix.
Updated overall description and per rep cap description. The per rep cap description seems redundant though.

image

build/templates/class.rst.mako
texasaggie97: See above

docs/nidcpower/class.rst
texasaggie97: See above

Issue #489, opened 2017-10-18
Confusing errors when building session.py and functions.rst


Fladolcetta:
When I run tox I get the following errors:

> FGEN:
   session.py
      Warning: "init" not found in function metadata. Typo? Generated code will be funky!
   functions.rst
      Warning: "init" not found in function metadata. Typo? Generated code will be funky!
DCPower
   session.py
      Warning: "ClearError" not found in function metadata. Typo? Generated code will be funky!
      Warning: "MeasureMultiple" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureOutputFunction" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureCurrentLimit" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureOutputFunction" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureVoltageLimit" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureOutputEnabled" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureOutputEnabled" not found in function metadata. Typo? Generated code will be funky!
   functions.rst
      Warning: "MeasureMultiple" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureOutputFunction" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureCurrentLimit" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureOutputFunction" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureVoltageLimit" not found in function metadata. Typo? Generated code will be funky!
DMM
   session.py
      Warning: "ConfigureThermistorType" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureThermistorType" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureThermistorType" not found in function metadata. Typo? Generated code will be funky!
   functions.rst
      Warning: "ConfigureThermistorType" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureThermistorType" not found in function metadata. Typo? Generated code will be funky!
      Warning: "ConfigureThermistorType" not found in function metadata. Typo? Generated code will be funky!
NI-SWITCH
   session.py
      Warning: "ClearError" not found in function metadata. Typo? Generated code will be funky!
[...]

PR #1086, opened 2019-11-04
Add 'codegen-method':  'library-only'

texasaggie97:
* Adds another supported 'codegen-method': 'library-only'
    * Needed for nidigital
    * Useful for creating ctypes wrapper function but not function in session.py
    * 'ivi-dance[-with-a-twist]' does not support multiplying two out parameters together
    * Usage will be in a later PR
* Remove all 'codegen-method': 'no' functions from nifgen - causing build issues
    * All 'codegen-method': 'no' functions will be removed when all metadata updated to latest version
* Remove one 'codegen-method': 'no' function from nidigital - will be added back in a later PR

* Visual inspection of generated code and as well as diffs
* System
* Unit


generated/nidcpower/session.py
```diff
@@ -3073,24 +3070,21 @@ def query_in_compliance(self):
         The compliance limit is the current limit when the output method is
         set to OutputFunction.DC_VOLTAGE. If the output is operating at the
         compliance limit, the output reaches the current limit before the
-        desired voltage level. Refer to the ConfigureOutputFunction
-        method and the ConfigureCurrentLimit method for more
+        desired voltage level. Refer to the configure_output_function
```
marcoskirsch: This is an odd side effect of the change... before we'd use the C function name minus the prefix. Now we use the Python name of a non-existent method.

Was another side effect of this making the "funky" warnings from  #489 go away? I don't think that's desirable - the warnings are effectively masked but the problem they originally pointed out is still there. If that's the case I think we need a bit of further discussion.

generated/niswitch/session.py
```diff
@@ -181,11 +181,8 @@ class _SessionBase(object):
...
-    Some cases exist where the end-user must specify instrument driver  options at initialization time.  An example of this is specifying  a particular instrument model from among a family of instruments  that the driver supports.  This is useful when using simulation.   The end-user can specify driver-specific options through  the DriverSetup keyword in the optionsString parameter to the  InitWithOptions method, or through the IVI Configuration Utility.
+    Some cases exist where the end-user must specify instrument driver  options at initialization time.  An example of this is specifying  a particular instrument model from among a family of instruments  that the driver supports.  This is useful when using simulation.   The end-user can specify driver-specific options through  the DriverSetup keyword in the optionsString parameter to the  init_with_options method, or through the IVI Configuration Utility.
     If the user does not specify a Driver Setup string, this property returns an empty string.
-
-    Note:
-    One or more of the referenced methods are not in the Python API for this driver.
```
marcoskirsch: This note was useful and now it's gone. That's a problem.

generated/nidcpower/session.py
texasaggie97: A very long explanation why this changed.

* Metadata on GitHub currently includes functions with `'codegen_method': 'no'`
* When I update the metadata soon (I am currently working on this) then those functions will no longer be added to the metadata on GitHub and will just go away.
* Before this change, we only expanded functions that were in the API (I.e. part of `Session`).
* After the change, we need to expand functions that are not part of the API, but are also _not_ `'codegen_method': 'no'`. I.e. `'library-only'`

I could have modified the code generator even more to avoid these changes, and I would have if the changes would have become permanent, but they aren't.

Once the metadata is updated without the unused functions, the documentation (and the note) will go back to the way they were.

It seemed foolish to torque the code generator around just to avoid temporary changed to the documentation.

generated/niswitch/session.py
texasaggie97: See comment above. The note will come back when the metadata is updated without `'codegen-method': 'no'` functions.

generated/nidcpower/session.py
marcoskirsch: I don't fully follow why we need `libary-only` now but in the future the metadata on GitHub won't contain those methods at all.

#1103 created to track this temporary problem.

Issue #1072, opened 2019-10-30
Improve usability of nidigital API by making method args optional when applicable


sbethur:
* Description of issue

Improve usability of nidigital API by making method args optional when applicable.
These are the methods that can be improved:
- create_pin_map(self, dut_pin_list, system_pin_list)
- apply_levels_and_timing(self, site_list, levels_sheet, timing_sheet, initial_state_high_pins, - initial_state_low_pins, initial_state_tristate_pins)
- tdr(self, apply_offsets)
- burst_pattern(self, site_list, start_label, select_digital_function, wait_until_done, timeout)
- burst_pattern_synchronized
- HRAM configuration (tracked [here](https://github.com/ni/nimi-python/issues/1071))
- unload_all_patterns(self, unload_keep_alive_pattern)
- export_signal(self, signal, signal_identifier, output_terminal)
- fetch_capture_waveform_u32
- write_sequencer_flag(self, flag, value)
- write_sequencer_flag_synchronized
- clock_generator_generate_clock(self, frequency, select_digital_function)
- wait_until_done(self, timeout)
- send_software_edge_trigger(self, trigger, trigger_identifier)
```

Issue #1069, opened 2019-10-30
write_source_waveform_site_unique_u32 missing in nidigital API


sbethur:
# Description of issue

write_source_waveform_site_unique_u32 missing in nidigital API.
Corresponding method in C header is `niDigital_WriteSourceWaveformSiteUniqueU32`

> Is this issue only to enable this method or to make a "fancy" version of it?

We need a fancy version.

`write_source_waveform_site_unique_u32` method is closely related to already implemented `write_source_waveform_broadcast_u32`.

`write_source_waveform_broadcast_u32` needs to be updated to take in only `waveform_name` and `waveform_data`. `waveform_data` can be of type list or array.array or numpy.ndarray. This work is tracked [here](https://github.com/ni/nimi-python/issues/1070).

Parameters for `write_source_waveform_site_unique_u32` should be:

1. `site_list` of type str
2. `waveform_name` of type str
3. `waveform_data` of type "2D array" of ViUInt32. I think we should accept:
    - numpy.ndarray
    - list of lists or array.arrays

We should also validate that size of all arrays inside `waveform_data` is same.

Method should not return anything.

marcoskirsch:
@sbethur to split this up into two issues.

texasaggie97:
@sbethur, Is this issue only to enable this method or to make a "fancy" version of it?

If fancy, what should the parameter list look like, and what should the output be and in what format?

sbethur:
> If fancy, what should the parameter list look like, and what should the output be and in what format?

Updated issue description.

sbethur:
Rethinking the types for this method.. we should consider `site_list` and `waveform_data` into a dict or namedtuple. It should consistent with what's used in `fetch_capture_waveform`.

PR #1096, opened 2019-11-06
Add fetch_capture_waveform() fancy wrapper

### LINKED ISSUE #1076: fetch_capture_waveform_u32 missing in nidigital API
# Description of issue

fetch_capture_waveform_u32 missing in nidigital API.
Corresponding method in C header is `niDigital_FetchCaptureWaveformU32`

> @sbethur, Is this issue only to enable this method or to make a "fancy" version of it?

There should be a fancy version.

Parameters should be:

1. `site_list` of type str
3. `samples_to_read` of type int
4. `timeout` of type float or datetime.timedelta

Return value should be "2D array" of VIUInt32. To be consistent with similar fetch method in niscope API, we can make the actual type be a list of memoryview of array.array.

texasaggie97:

* Create a "fancy" wrapper around `fetch_capture_waveform_u32()`
    * Returns array of named tuples Measurement(data, site)
    * Gets site list and uses that when creating named tuple

* Fixes #1076 

* Visual inspection


src/nifgen/metadata/functions.py
```diff
@@ -18,33 +18,6 @@
         'python_name': 'abort',
         'returns': 'ViStatus'
     },
-    'AdjustSampleClockRelativeDelay': {
```
sbethur: Are changes in this file supposed to be part of this PR?

texasaggie97: It required #1086 first. I have changed the title to reflect that.

So until that one is accepted, merged into master, and then merged into this branch, the changes from there will show up here as well.

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,76 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
         return
 
+    @ivi_synchronized
+    def _fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):
+        # This is slightly modified codegen from the function
+        vi_ctype = _visatype.ViSession(self._vi)  # case S110
+        site_list_ctype = ctypes.create_string_buffer(site_list.encode(self._encoding))  # case C020
+        waveform_name_ctype = ctypes.create_string_buffer(waveform_name.encode(self._encoding))  # case C020
+        samples_to_read_ctype = _visatype.ViInt32(samples_to_read)  # case S150
+        timeout_ctype = _visatype.ViReal64(timeout)  # case S150
```
sbethur: Will this work when type of timeout is datetime.timedelta?

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,76 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+
+        data, actual_num_waveforms, actual_samples_per_waveform = self._fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout)
+
+        # Get the site list
+        site_list = self.get_site_results_site_numbers(site_list, enums.SiteType.CAPTURE_WAVEFORM)
+        assert len(site_list) == actual_num_waveforms
+
+        Measurement = collections.namedtuple('Measurement', ['data', 'site'])
+
+        measurements = []
```
sbethur: Rename `Measurement` to `Waveform`, `'Measurement'` to `'Waveform'`, and `measurements` to `waveforms`

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,76 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+        import collections
+        import sys
+
+        data, actual_num_waveforms, actual_samples_per_waveform = self._fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout)
+
+        # Get the site list
+        site_list = self.get_site_results_site_numbers(site_list, enums.SiteType.CAPTURE_WAVEFORM)
+        assert len(site_list) == actual_num_waveforms
+
+        Measurement = collections.namedtuple('Measurement', ['data', 'site'])
```
sbethur: Is there any convention for member order in a namedtuple? <Identifier, data> seems more intuitive to me than the other way around.

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,76 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+
+        for i in range(actual_num_waveforms):
+            start = i * actual_samples_per_waveform
+            end = start + actual_samples_per_waveform
+            if sys.version_info.major >= 3:
+                measurements.append(Measurement(data=mv[start:end], site=site_list[i]))
+            else:
+                # memoryview in Python 2 doesn't support numeric types, so we copy into an array.array to put in the wfm. :( You should be using Python 3!
+                # Or use the _into version. memoryview in Python 2 only supports string and bytearray, not array.array or numpy.ndarray of arbitrary types.
+                measurements.append(Measurement(data=array.array('d', data[start:end]), site=site_list[i]))
```
sbethur: Shouldn't the type of array.array be 'L'?

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,76 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+
+            waveform_name (str):
+
+            samples_to_read (int):
+
+            timeout (float or datetime.timedelta):
+
+
+        Returns:
+            data (list of int):
```
sbethur: Type is incorrect.

src/nidigital/metadata/functions_addon.py
```diff
@@ -2,6 +2,61 @@
...
+        ],
+        'documentation': {
+            'description': 'TBD'
+        },
+        'parameters': [
+            {
+                'direction': 'in',
+                'documentation': {
+                    'description': 'The instrument handle you obtain from niScope_init that identifies a particular instrument session.'
+                },
```
sbethur: Not valid, remove it.

docs/nidigital/class.rst
```diff
@@ -1557,6 +1557,56 @@ export_signal
...
+            :type samples_to_read: int
+            :param timeout:
+
+
+                
+
+
+            :type timeout: float or datetime.timedelta
+
+            :rtype: list of int
```
sbethur: Incorrect

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### NI-Digital Pattern Driver
     * #### Added
+        * `fetch_capture_waveform()` - returns named tuple `Measurement(data, site)`
```
sbethur: Change to `Wavefrom()`. Also namedtuple is one word.

docs/nidcpower/class.rst
```diff
@@ -135,7 +135,7 @@ abort
             when you call the :py:meth:`nidcpower.Session.abort` method, the output channels remain
             in their current state and continue providing power.
 
-            Use the :py:meth:`nidcpower.Session.ConfigureOutputEnabled` method to disable power
+            Use the :py:meth:`nidcpower.Session.configure_output_enabled` method to disable power
```
marcoskirsch: These changes shouldn't be shown in this PR.

docs/nidigital/class.rst
```diff
@@ -1557,6 +1557,66 @@ export_signal
 
             :type output_terminal: str
 
+fetch_capture_waveform
+----------------------
+
+    .. py:currentmodule:: nidigital.Session
+
+    .. py:method:: fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout)
```
marcoskirsch: * I thought sites were a repeated capability in which case I expect the signature to be different.
* In most nimi-python drivers we have default timeouts.

If either of these is meant to be fixed separately please reply with a reference to the corresponding issue.

marcoskirsch: I believe sites are like channels in the sense that from Python we should support `"0"`, `"0-1"`, `0`, `[0, 1]`, etc.

docs/nidigital/class.rst
```diff
@@ -1557,6 +1557,66 @@ export_signal
...
+
+            Returns a list of named tuples (Waveform) that <FILL IN THE BLANK HERE>
+
+            Fields in Waveform:
+
+            - **site** (int)
+            - **data** (array.array of int)
```
marcoskirsch: I wonder if named tuple of int/array of int is the best way to return this.
What are all the places in the API in which this type could be used? Can it ever be an input?

docs/nidigital/class.rst
```diff
@@ -1557,6 +1557,66 @@ export_signal
...
+                
+
+
+            :type timeout: float or datetime.timedelta
+
+            :rtype: list of Waveform
+            :return:
+
+
+                    List of named tuples with fields:
```
marcoskirsch: Is the order important?
Maybe it would be nicer for clients to return a dictionary of site -> data?

docs/nidigital/class.rst
```diff
@@ -2114,7 +2174,7 @@ get_site_results_site_numbers
                 
 
 
-            :type site_result_type: int
+            :type site_result_type: :py:data:`nidigital.SiteResult`
```
marcoskirsch: This does not appear to be related to this PR.

docs/nidigital/enums.rst
```diff
@@ -83,6 +83,19 @@ Signal
 
 
 
+SiteResult
```
marcoskirsch: This does not appear to be related to this PR.

docs/nidmm/class.rst
```diff
@@ -1543,16 +1535,14 @@ perform_short_cable_comp
             Performs the short cable compensation measurements for the current
             capacitance/inductance range, and returns short cable compensation
             **Resistance** and **Reactance** values. You can use the return values
-            of this method as inputs to :py:meth:`nidmm.Session.ConfigureShortCableCompValues`.
+            of this method as inputs to :py:meth:`nidmm.Session.configure_short_cable_comp_values`.
```
marcoskirsch: These does not appear to be related to this PR.

docs/niswitch/class.rst
```diff
@@ -371,12 +371,10 @@ close
             deallocates any memory resources the driver uses. Notes: (1) You must
             unlock the session before calling :py:meth:`niswitch.Session._close`. (2) After calling
             :py:meth:`niswitch.Session._close`, you cannot use the instrument driver again until you
-            call :py:meth:`niswitch.Session.init` or :py:meth:`niswitch.Session.InitWithOptions`.
+            call :py:meth:`niswitch.Session.init` or :py:meth:`niswitch.Session.init_with_options`.
```
marcoskirsch: This does not appear to be related to this PR.

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,85 @@ def export_signal(self, signal, signal_identifier, output_terminal):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
+    @ivi_synchronized
+    def _fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):
+        # This is slightly modified codegen from the function
```
marcoskirsch: Why wasn't this code-generated?

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,85 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+        data_array = array.array("L", [0] * data_size)  # case B620
+        data_ctype = get_ctypes_pointer_for_buffer(value=data_array, library_type=_visatype.ViUInt32)  # case B620
+        error_code = self._library.niDigital_FetchCaptureWaveformU32(vi_ctype, site_list_ctype, waveform_name_ctype, samples_to_read_ctype, timeout_ctype, data_buffer_size_ctype, data_ctype, None if actual_num_waveforms_ctype is None else (ctypes.pointer(actual_num_waveforms_ctype)), None if actual_samples_per_waveform_ctype is None else (ctypes.pointer(actual_samples_per_waveform_ctype)))
+        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
+        return data_array, actual_num_waveforms_ctype.value, actual_samples_per_waveform_ctype.value
+
+    def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):
+        '''fetch_capture_waveform
+
+        Returns a list of named tuples (Waveform) that <FILL IN THE BLANK HERE>
```
marcoskirsch: How are we tracking that this still needs documentation work?

generated/nidigital/session.py
```diff
@@ -2770,6 +2770,85 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+
+        # Get the site list
+        site_list = self.get_site_results_site_numbers(site_list, enums.SiteType.CAPTURE_WAVEFORM)
+        assert len(site_list) == actual_num_waveforms
+
+        Waveform = collections.namedtuple('Waveform', ['site', 'data'])
+
+        waveforms = []
+
+        if sys.version_info.major >= 3:
```
marcoskirsch: @MrDuff8 when can we drop Python 2.7?

docs/nidigital/class.rst
texasaggie97: We don't know what sites will become. They may become rep_caps, but they may not.

docs/nidigital/class.rst
texasaggie97: @sbethur?

docs/nidigital/class.rst
marcoskirsch: #1111 opened

generated/nidigital/session.py
texasaggie97: All of digital needs documentation work.

generated/nidigital/session.py
texasaggie97: Because the code generator doesn't support an 'ivi-dance' where there are two values returned and then multiplied together.

docs/nidigital/class.rst
texasaggie97: @sbethur?

docs/nidcpower/class.rst
texasaggie97: I have redone this as a new PR to get rid of the extraneous changes. See #1112

texasaggie97:
I have moved this to a new PR to no longer include extraneous changes.

See #1112

docs/nidigital/class.rst
sbethur: > In most nimi-python drivers we have default timeouts.

There is a [issue](https://github.com/ni/nimi-python/issues/1072) open to have defaults where it makes sense and make the parameters optional.

docs/nidigital/class.rst
sbethur: It's used in two places:

1) `niDigital_SortSiteResultsViUInt32Waveform` - This will NOT be added in Python, since its use is superseded by multi-instrument sessions.
2) `niDigital_WriteSourceWaveformSiteUniqueU32` - This should use the same type as what we use here for `fetch_capture_waveform`.  Tracked [here](https://github.com/ni/nimi-python/issues/1069).

I looked at one of the use-cases for `fetch_capture_waveform` - streaming capture waveform data from h/w test program. It would be ideal for `fetch_capture_waveform` to accept a buffer and populate that with the waveform data (like [fetch_into](https://nimi-python.readthedocs.io/en/master/niscope/class.html#fetch-into) in niscope). 

But the streaming use-case is not common and even then the size is about tens of MBs. If we learn about a use-case where performance is not sufficient, we can add `fetch_into` later. Usability of streaming with `fetch_capture_waveform` that returns a dict is not bad:
```
# Pattern burst is configured to fetch 1024 samples
 num_samples_expected = 16
 samples_per_fetch = 4
 waveforms = collections.defaultdict(list)
 for i in range(num_samples_expected // samples_per_fetch):
     fetched_waveform = multi_instrument_session.fetch_capture_waveform(site_list='', waveform_name='capt_wfm', samples_to_read=samples_per_fetch, timeout=10.0)
     for key in fetched_waveform:
         waveforms[key] += fetched_waveform[key]
```

So, now the question is whether to use dict or namedtuple..

One drawback for dict is iteration order is not deterministic (at least in Python 3.5). It looks like in 3.6 and later, iteration is ordered. Either way, I think it's better to not guarantee any order, because trying to do that is little tricky:
 
If method is called with `site=''` then the order will 0,1,2... with disabled sites skipped.
If method is called with `site='3,0,1'` (crazy but allowed), then the order will be 3,0,1 with disabled sites skipped.

Given that order depends on input and also can contain gaps, users should not really rely on order. Even if they want to iterate in a particular order, it's not difficult to do that in Python (Everything's easy in Python )

In summary, the recommendation is dict(int, array of unsigned int)

docs/nidigital/class.rst
sbethur: See answer above.

TL;DR - dict is better.

PR #1098, opened 2019-11-06
[#1112] Make get_pin_results_pin_information() fancy

### LINKED ISSUE #1089: Make get_pin_results_pin_information in nidigital API fancy
### Description of issue

`get_pin_results_pin_information` method returns 3 lists of int of same size (pin_indexes, site_numbers, and channel_indexes). It should be made fancy i.e. Have it return a single list of named tuples.

texasaggie97:

* Add fancy wrapper around `get_pin_results_pin_information()`
    * Returns named tuple `PinInfo(pin_indexes, site_numbers, channel_indexes)`

* Fix #1089 

* Visual inspection of generated code


src/nidigital/metadata/functions_addon.py
```diff
@@ -2,6 +2,106 @@
...
+                'documentation': {
+                    'description': '\nList of named tuples with fields:\n\n- **pin_indexes** (int)\n- **site_numbers** (int)\n- **channel_indexes** (int)\n'
+                },
+                'name': 'pin_info',
+                'python_type': 'PinInfo',
+                'size': {
+                    'mechanism': 'python-code',
+                    'value': None
+                },
+                'type': 'ViReal64[]'
```
sbethur: Type is incorrect

texasaggie97: Type doesn't matter here since there isn't an underlying DLL function call, and 'python_type' will be used in documentation.

Changed to `ViInt32[]` since that is a little closer though.

docs/nidigital/class.rst
```diff
@@ -1564,7 +1564,12 @@ fetch_capture_waveform
 
     .. py:method:: fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout)
 
-            TBD
+            Returns a list of named tuples (Waveform) that <FILL IN THE BLANK HERE>
+
+            Fields in Waveform:
+
+            - **site** (int)
+            - **data** (array.array of int)
```
marcoskirsch: * Is `array.array of int` the right type for the data of digital waveforms?
* When I think of `Waveform` I think of an analog signal by default. Should we call this DigitalWaveform or is it enough that it is an `nidigital.Waveform`?

generated/nidigital/session.py
```diff
@@ -2830,7 +2830,12 @@ def _fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, tim
     def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):
         '''fetch_capture_waveform
 
-        TBD
+        Returns a list of named tuples (Waveform) that <FILL IN THE BLANK HERE>
```
marcoskirsch: We should add an issue about `<FILL IN THE BLANK HERE>` unless we have another way to track it in place.

generated/nidigital/session.py
```diff
@@ -2843,21 +2848,25 @@ def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, time
 
 
         Returns:
-            data (list of int):
+            waveform (list of Waveform): List of named tuples with fields:
```
marcoskirsch: Is the order important?
If not, it might make sense for this to be a dictionary of `site`->`waveform data`
And perhaps we should consider replacing the name tuple with the same... one less type.

generated/nidigital/session.py
```diff
@@ -2867,13 +2876,13 @@ def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, time
             start = i * actual_samples_per_waveform
             end = start + actual_samples_per_waveform
             if sys.version_info.major >= 3:
-                measurements.append(Measurement(data=mv[start:end], site=site_list[i]))
+                waveforms.append(Waveform(site=site_list[i], data=mv[start:end]))
             else:
                 # memoryview in Python 2 doesn't support numeric types, so we copy into an array.array to put in the wfm. :( You should be using Python 3!
```
marcoskirsch: @MrDuff8 when can we drop Python 2.7?

src/nidigital/templates/session.py/fancy_fetch_capture_waveform.py.mako
```diff
@@ -34,15 +34,16 @@
         import collections
         import sys
 
-        data, actual_num_waveforms, actual_samples_per_waveform = self._fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout)
+        timeout_secs = _converters.convert_timedelta_to_seconds(timeout, _visatype.ViReal64)
+        data, actual_num_waveforms, actual_samples_per_waveform = self._fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout_secs)
```
marcoskirsch: Why is `_fetch_capture_waveform` not code-generated?
The comment doesn't say so.

src/nidigital/templates/session.py/fancy_fetch_capture_waveform.py.mako
```diff
@@ -34,15 +34,16 @@
         import collections
         import sys
```
marcoskirsch: Can't leave a comment on line 3
`    '''Dispatches to the appropriate "fetch waveform into" method based on the waveform type.'''`

That seems wrong.

src/nidigital/templates/session.py/fancy_fetch_capture_waveform.py.mako
```diff
@@ -52,11 +53,11 @@
...
-                measurements.append(Measurement(data=mv[start:end], site=site_list[i]))
+                waveforms.append(Waveform(site=site_list[i], data=mv[start:end]))
             else:
                 # memoryview in Python 2 doesn't support numeric types, so we copy into an array.array to put in the wfm. :( You should be using Python 3!
                 # Or use the _into version. memoryview in Python 2 only supports string and bytearray, not array.array or numpy.ndarray of arbitrary types.
-                measurements.append(Measurement(data=array.array('d', data[start:end]), site=site_list[i]))
+                waveforms.append(Waveform(site=site_list[i], data=array.array('L', data[start:end])))
 
-        return measurements
+        return waveforms
```
marcoskirsch: We are missing test coverage for this hand-coded function. Seems like we could write a unit test by mocking  the private function (if we can get it to be code-generated) else we could mock the library.

Or can we get adequate coverage through sytem tests?

texasaggie97: None of the main drivers currently have unit tests and none do any mocking.

This would be a large change to how we are doing things.

generated/nidigital/session.py
texasaggie97: The earliest will be when the Python org drops it, which is 1/1/20.

src/nidigital/templates/session.py/fancy_fetch_capture_waveform.py.mako
marcoskirsch: Or can we get adequate coverage through sytem tests?

texasaggie97:
Replaced by #1120 to get rid of other files

Issue #1092, opened 2019-11-05
Bounds checking not done for int values


sbethur:
### Description of issue

Bounds checking is not done for int type method arguments and property values. It appears that an int value larger than the underlying ctypes type gets silently truncated and passed down to the C driver DLL. It would be nice if bounds checking is done in the Python layer.

### System report

`python -c "import niscope; niscope.print_diagnostic_information()"` output

```
OS:
    Name:      Windows
    Version:   10.0.16299
    Bits:      64
Driver:
    Name:      NI-SCOPE
    Version:   19.0.0.49152
Module:
    Name:      niscope
    Version:   1.1.3
Python:
    Version:   3.6.8 (tags/v3.6.8:3c6b436a57, Dec 24 2018, 00:16:47) [MSC v.1916 64 bit (AMD64)]
    Bits:      64
    Is_Venv:   True
    Installed Packages:
        zipp==0.6.0
        wcwidth==0.1.7
        six==1.12.0
        pytest==5.2.1
        pyparsing==2.4.2
        py==1.8.0
        pluggy==0.13.0
        packaging==19.2
        numpy==1.17.3
        nitclk==0.1.0.dev0
        niscope==1.1.3
        nidigital==0.1.1.dev0
        nidcpower==1.1.2
        more-itertools==7.2.0
        importlib-metadata==0.23
        future==0.18.1
        enum34==1.1.6
        colorama==0.4.1
        attrs==19.3.0
        atomicwrites==1.3.0
        setuptools==40.8.0
        pip==19.0.3
```

### Steps to reproduce issue

```
with niscope.Session(resource_name='my_device', options='Simulate=1, DriverSetup=Model:5164; BoardType:PXIe') as session:
    # Type of tv_trigger_line_number is ViInt32. val is 1 followed by 32 0s.
    val = 4294967296
    session.tv_trigger_line_number = val
    assert session.tv_trigger_line_number == 0
```

Run the above snippet and observe the incorrect value in `assert`

texasaggie97:
Should this error checking be done in all attributes and in all functions every time? I think: if we want to ever do the check, we need to always do the check

Are we concerned about any effect on performance these extra checks will add? I think: In general, if you care about performance, maybe Python isn't for you...

What error/exception should we return? I think: OverflowError

sbethur:
I agree that it should be always or never. We could provide a way to turn off bounds checking for users who want to squeeze out some performance.

[...]

PR #1112, opened 2019-11-12
Add fancy fetch capture waveform

### LINKED ISSUE #1074: nidigital fetch_capture_waveform_u32 returns empty list
* Description of issue

In nidigital driver, calling fetch_capture_waveform_u32 method after executing a pattern with digital capture returns an empty list.

I managed to fix this issue by modifying session.py line 3058 as follow:
<- data_buffer_size_ctype = _visatype.ViInt32(error_code)  # case S180
-> data_buffer_size_ctype = _visatype.ViInt32(actual_num_waveforms_ctype.value * actual_samples_per_waveform_ctype.value)

* `python -c "import niscope; niscope.print_diagnostic_information()"` output (replace `niscope` with appropriate package name)

nidigital version: 0.1.0

* Steps to reproduce issue

    1. Step 1: Initialise nidigital session
    2. Step 2: load pinmap, spec, levels, timings, pattern, digital capture
    3. Step3: Execute pattern with digital capture (session.burst_pattern)
    4. Step4: Get digital capture (fetch_capture_waveform_u32)

### LINKED ISSUE #1076: fetch_capture_waveform_u32 missing in nidigital API

texasaggie97:

* Replaces #1096
* Create a "fancy" wrapper around `fetch_capture_waveform_u32()`
    * Returns array of named tuples Measurement(data, site)
    * Gets site list and uses that when creating named tuple

* Fixes #1076 
* Fixes #1074 

* Visual inspection


generated/nidigital/session.py
```diff
@@ -2676,6 +2676,85 @@ def export_signal(self, signal, signal_identifier, output_terminal):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
+    @ivi_synchronized
+    def _fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):
+        # This is slightly modified codegen from the function
```
marcoskirsch: The reason we have this hand-coded implementation is that "the codegenerator does not support ivi-dance with two values that are multiplied together".

The reason should be documented here.
It's a non-trival amount of code to hand-maintain. It worries me because the fix #1092 will likely affect this sort of code and we are at risk of not updating this hand maintained copy.

generated/nidigital/session.py
```diff
@@ -2676,6 +2676,85 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+        vi_ctype = _visatype.ViSession(self._vi)  # case S110
+        site_list_ctype = ctypes.create_string_buffer(site_list.encode(self._encoding))  # case C020
+        waveform_name_ctype = ctypes.create_string_buffer(waveform_name.encode(self._encoding))  # case C020
+        samples_to_read_ctype = _visatype.ViInt32(samples_to_read)  # case S150
+        timeout_ctype = _visatype.ViReal64(timeout)  # case S150
```
marcoskirsch: If this was codegenerated, the timeout conversion would go here. I think we should make this as close to what the codegen would spit out as possible.

We should possibly open an issue on the generator's limitation and reference it here. We may not fix it but at least we have the opportunity to discuss it so the reasoning is clear.

generated/nidigital/session.py
texasaggie97: Already added.

generated/nidigital/session.py
texasaggie97: Issue #1114 created

Template updated to move the conversion to the "code generated" part.

generated/nidigital/session.py
```diff
@@ -2676,6 +2676,88 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+                - **data** (array.array of int)
+
+        '''
+        import collections
+        import sys
+
+        data, actual_num_waveforms, actual_samples_per_waveform = self._fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout)
+
+        # Get the site list
+        site_list = self.get_site_results_site_numbers(site_list, enums.SiteType.CAPTURE_WAVEFORM)
```
sbethur: This requires `get_site_results_site_numbers` method to be updated to accept enum param.

generated/nidigital/session.py
```diff
@@ -2676,6 +2676,88 @@ def export_signal(self, signal, signal_identifier, output_terminal):
...
+        waveform_name_ctype = ctypes.create_string_buffer(waveform_name.encode(self._encoding))  # case C020
+        samples_to_read_ctype = _visatype.ViInt32(samples_to_read)  # case S150
+        timeout_ctype = _converters.convert_timedelta_to_seconds(timeout, _visatype.ViReal64)  # case S140
+        data_buffer_size_ctype = _visatype.ViInt32(0)  # case S190
+        data_ctype = None  # case B610
+        actual_num_waveforms_ctype = _visatype.ViInt32()  # case S220
+        actual_samples_per_waveform_ctype = _visatype.ViInt32()  # case S220
+        error_code = self._library.niDigital_FetchCaptureWaveformU32(vi_ctype, site_list_ctype, waveform_name_ctype, samples_to_read_ctype, timeout_ctype, data_buffer_size_ctype, data_ctype, None if actual_num_waveforms_ctype is None else (ctypes.pointer(actual_num_waveforms_ctype)), None if actual_samples_per_waveform_ctype is None else (ctypes.pointer(actual_samples_per_waveform_ctype)))
+        errors.handle_error(self, error_code, ignore_warnings=True, is_error_handling=False)
+        data_buffer_size_ctype = _visatype.ViInt32(actual_num_waveforms_ctype.value)  # case S200
```
sbethur: This should be `actual_num_waveforms_ctype.value * actual_samples_per_waveform_ctype.value`

generated/nidigital/session.py
texasaggie97: Updated `get_site_results_site_numbers()` to use enum

generated/nidigital/session.py
texasaggie97: Fixed

src/nidigital/metadata/functions_addon.py
```diff
@@ -4,4 +4,61 @@
 functions_override_metadata = {
 }
```
sbethur: Is this still needed?

texasaggie97: Right now, no. But I like to leave it there empty for when we do need to add something.

src/nidigital/templates/session.py/fancy_fetch_capture_waveform.py.mako
```diff
@@ -0,0 +1,52 @@
...
+
+    def ${f['python_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.SESSION_NUMPY_INTO_METHOD_DECLARATION)}):
+        '''${f['python_name']}
+
+        ${helper.get_function_docstring(f, False, config, indent=8)}
+        '''
+        data, actual_num_waveforms, actual_samples_per_waveform = self._fetch_capture_waveform(site_list, waveform_name, samples_to_read, timeout)
+
+        # Get the site list
+        site_list = self.get_site_results_site_numbers(site_list, enums.SiteType.CAPTURE_WAVEFORM)
```
sbethur: Should be `enums.SiteResult`

texasaggie97: Fixed

generated/nidigital/session.py
```diff
@@ -2677,6 +2677,68 @@ def export_signal(self, signal, signal_identifier, output_terminal):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
+    @ivi_synchronized
+    def _fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):
+        # This is slightly modified codegen from the function
+        # We cannot use codegen without major modifications to the code generator
+        # This function uses two 'ivi-dance' parameters and then multiplies them together - see
+        # the (modified) line below
+        # Also, we want to return the two sized that normally wouldn't be returned
```
marcoskirsch: I don't understand this sentence...

PR #1113, opened 2019-11-12
Update nidcpower metadata

texasaggie97:

* Updates nidcpower metadata to 19.5.0

### ~~What testing has been done?~~
* Unit
* System


docs/nidcpower/class.rst
```diff
@@ -129,7 +129,7 @@ abort
     .. py:method:: abort()
 
             Transitions the NI-DCPower session from the Running state to the
-            Committed state. If a sequence is running, it is stopped. Any
+            Uncommitted state. If a sequence is running, it is stopped. Any
```
marcoskirsch: Weird, I didn't know we had an "uncommitted" state. Where did this get pulled from?

docs/nidcpower/class.rst
```diff
@@ -6017,7 +6585,7 @@ source_mode
             +----------------+------------------+
             | Permissions    | read-write       |
             +----------------+------------------+
-            | Channel Based  | No               |
+            | Channel Based  | Yes              |
```
marcoskirsch: There's a lot of changes here of properties going from "Channel Based  | No" to "Channel Based  | Yes". I believe these are bug fixes.

I suppose since the Python generated code is no different per our current design, this is just a documentation fix. But it may be worth noting in the CHANGELOG.

docs/nidcpower/class.rst
texasaggie97: From their internal metadata

docs/nidcpower/class.rst
texasaggie97: Added

texasaggie97:
This will be done again with more trace-ability

Issue #1111, opened 2019-11-12
`nidigital` should expose sites as repeated capabilities


marcoskirsch:
This is up for discussion.
I believe sites in `nidigital` should be repeated capabilities similar to how we do channels.

`my_session.sites[[3,4]]. fetch_capture_waveform("my_waveform", 1024)`

sbethur:
Depending on whether we make `sites` a repeated capability or not, the type for sites in `write_source_waveform_site_unique` and `fetch_capture_waveform` may need to change.

siteList in C API is a const string. Its contents are _Site numbers listed as a comma-delimited list of strings of form siteN, where N is the site  number._

Currently, type of sites in `write_source_waveform_site_unique` and `fetch_capture_waveform`, at public Python API layer, is int. We internally convert it to _"siteA,siteB"_, as required.

When working on this issue, we should ensure the type is kept consistent at the Python API layer.

sbethur:
If sites is not made a repeated capability, we should make `site_list` parameter in public methods optional by giving it a default value of `''` i.e. empty string, which means all sites.

sbethur:
There are a bunch of methods that accept a site list as "site0,site1,site2". These could easily be made more usable by making sites a rep-cap (with "site" as prefix). Users would be able to write code like:
 ```
session.sites[0,1].burst_pattern("pat", true)
```
 
Then there are methods that accept a single site as "site0". Should we update these methods to treat site as a rep-cap? ..so users can write something like:
```
session.sites[0].fetch_history_ram_cycle_information(0, 0)
```
and let the driver return an error if they specify more than one site?
 
 
Then there are methods that expect a `pinList`. The pin list can include "site-restricted" pins like site0/PinA. To make this easier to use we can allow users to specify both sites and pins and pins as repeated capabilities. It would be something like:
 ```
session.sites[2].pins['VCC,CLK'].ppmu_voltage_level = 4
```
We need to figure out the usability and implementation feasibility of something like above.

PR #1119, opened 2019-11-13
Add fancy write_source_waveform_unique

### LINKED ISSUE #1069: write_source_waveform_site_unique_u32 missing in nidigital API

texasaggie97:
* Add fancy wrapper for `write_source_waveform_unique()`

* Fix #1069 

* Visual inspection


generated/nidigital/session.py
```diff
@@ -2687,6 +2749,48 @@ def self_test(self):
...
+
+            site_list.append(str(site))
+
+            start = i * actual_samples_per_waveform
+            end = start + actual_samples_per_waveform
+            mv[start:end] = waveform_data[site]
+
+            i += 1
+
+        site_list_str = ','.join(site_list)
```
sbethur: We need to create a string of the form `"site0,site1"` before passing it down to C API.

Before 1.0, we need to figure out the right type to use for sites across the API. Tracked in #1111.

generated/nidigital/session.py
```diff
@@ -2687,6 +2749,48 @@ def self_test(self):
...
+        actual_samples_per_waveform = len(waveform_data[next(iter(waveform_data))])
+        data = array.array('L', [0] * (len(waveform_data) * actual_samples_per_waveform))
+        mv = memoryview(data)
+
+        i = 0
+        for site in waveform_data:
+            if len(waveform_data[site]) != actual_samples_per_waveform:
+                raise ValueError('Mismatched length of waveforms. All must be the same length.')
+            if waveform_data[site].typecode != 'L':
+                raise ValueError('Wrong array element type. Must be unsigned 32 bit int "l", was {}'.format(waveform_data[site].typecode))
```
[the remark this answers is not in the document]
texasaggie97: Fixed

generated/nidigital/session.py
texasaggie97: Fixed

generated/nidigital/session.py
```diff
@@ -2750,6 +2750,48 @@ def self_test(self):
...
+
+        TBD
+
+        Args:
+            waveform_name (str):
+
+            waveform_data ({ site: data, site: data, ... }): Dictionary where each key is the site number and the value is array.array of unsigned int
+
+        '''
+        if len(waveform_data) == 0:
```
marcoskirsch: Is this what the C library does? (as opposed ot an error)
I would have expected an error. Please confirm.

generated/nidigital/session.py
```diff
@@ -2750,6 +2750,48 @@ def self_test(self):
...
+        # Get an entry from the dictionary from https://stackoverflow.com/questions/30362391/how-do-you-find-the-first-key-in-a-dictionary
+        actual_samples_per_waveform = len(waveform_data[next(iter(waveform_data))])
+        data = array.array('L', [0] * (len(waveform_data) * actual_samples_per_waveform))
+        mv = memoryview(data)
+
+        i = 0
+        for site in waveform_data:
+            if len(waveform_data[site]) != actual_samples_per_waveform:
+                raise ValueError('Mismatched length of waveforms. All must be the same length.')
+            if waveform_data[site].typecode != 'L':
```
marcoskirsch: This goes against ducktyping pretty strongly right?
It forces clients to use array.array even though they may have numpy.array or even a plain old Python list. Seems like something we shouldn't enforce so strictly.

marcoskirsch: We should have system tests for these combinations.

texasaggie97: Because we wrap the array directly without copying, we would need to explicitly handle the other types. When it comes to ctypes, duck typing doesn't really work.

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ All notable changes to this project will be documented in this file.
 * ### NI-Digital Pattern Driver
     * #### Added
         * `fetch_capture_waveform()` - returns dictionary { site: data, site: data, ... }
+        * `write_source_waveform_site_unique()` - takes waveform_name and dictionary { site: data, site: data, ... }
```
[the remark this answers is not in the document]
sbethur: > It seems like most of the time the user would want to use the same waveform data for all sites. This function lets you specify different waveform data for each site, right?

Yes, this method lets users specify different waveform for each site.

> Is this common?

It is common enough that we should have it as part of the API. Any part specific data (like serial number or some unique ID) that needs to be programmed during production test would need this method.

> How does one write the same wavefrom data to all sites at once?

[`write_source_waveform_broadcast`](https://github.com/ni/nimi-python/blob/master/generated/nidigital/session.py#L3548)

generated/nidigital/session.py
sbethur: C entry point includes 2 size-related parameters - `numWaveforms` and `samplesPerWaveform`.

If `numWaveforms` is 0: `NIDIGITAL_ERROR_INCONSISTENT_NUM_WAVEFORMS_AND_SITES` (The number of waveforms in the 2D array provided to niDigital Write Source Waveform (Site Unique U32) does not match the number of sites specified.) error is thrown.
If `samplesPerWaveform` is 0: `ivierrors::ERR_INVALID_VALUE` is thrown.

generated/nidigital/session.py
sbethur: > We should have system tests for these combinations.

I'm working on them. You mean combinations of different values for number of sites and actual_samples_per_waveform?

generated/nidigital/session.py
texasaggie97: I cannot use `next(iter(waveform_data))` on an empty list/array.

I can put is special code to just return early, or put in special code to pass in 0 to the driver and let it do what ever it is going to do, which looks like throw an exception.

generated/nidigital/session.py
texasaggie97: Do we want to explicitly support other types?

marcoskirsch: > When it comes to ctypes, duck typing doesn't really work.

Sure. But that's our problem. We've done work so that this isn't the client's problem before. NI-FGEN is an example of this.

> I'm working on them. You mean combinations of different values for number of sites and actual_samples_per_waveform?

No, I mean client passing different things for the data, such as array.array, list of int, numpy.array.

> Do we want to explicitly support other types?

Is it reasonable to force clients to use array.array? That doesn't seem pythonic to me. So I _think_ we do.


How about this:

* @texasaggie97 Open an issue that these functions should support data as things other than array.array
* We can discuss there and not hold this PR
* @sbethur write the test but have it disabled and point to the issue.

generated/nidigital/session.py
marcoskirsch: I think `put in special code to pass in 0 to the driver and let it do what ever it is going to do` will be the most consistent with the other ADEs and also return the richest error.

texasaggie97: Done

generated/nidigital/session.py
texasaggie97: #1132 created

src/nidigital/templates/session.py/fancy_write_source_waveform_site_unique.py.mako
```diff
@@ -0,0 +1,39 @@
...
+
+            start = i * actual_samples_per_waveform
+            end = start + actual_samples_per_waveform
+            mv[start:end] = waveform_data[site]
+
+            i += 1
+
+        site_list_str = ','.join(site_list)
+
+        self._write_source_waveform_site_unique(site_list_str, waveform_name, len(waveform_data), actual_samples_per_waveform, data)
```
sbethur: Should be `_write_source_waveform_site_unique_u32` (_u32 suffix is missing)

texasaggie97: Fixed

src/nidigital/metadata/functions_addon.py
```diff
@@ -4,6 +4,44 @@
...
+                'name': 'waveformName',
+                'type': 'ViConstString'
+            },
+            {
+                'direction': 'in',
+                'documentation': {
+                    'description': '\nDictionary where each key is the site number and the value is array.array of unsigned int\n'
+                },
+                'name': 'waveform_data',
+                'type': 'ViUInt32',
```
marcoskirsch: If this is ignored because "python-only" please add a comment.

PR #1122, opened 2019-11-14
Add pins rep cap

### LINKED ISSUE #1093: Make nidigital methods that accept pinList argument easier to use
### Description of issue

Many methods in nidigital API accept `channel_list` parameter. Here's the description of that parameter in C API reference:

> List of channel names or list of pins. Do not pass a mix of channel names and pin names. An empty string denotes all digital pattern instrument channels. Pin names and pin groups apply to all enabled sites, unless the pin name explicitly specifies the site. You can specify a pin in a specific site  using the form siteN/pinName,  where N is the site number.

Valid values of `channel_list` parameter are:
- Channel number(s). e.g. '0', '0-7'
- Comma-delimited pins/pin groups that apply to all sites. e.g. 'VCC, CLK'
- Comma-delimited pins/pin groups that applies to specific site. e.g. 'site2/VCC, site2/CLK'

Ideally, the above cases should be coded in Python as:
```
session.channels['0-7'].ppmu_voltage_level = 4
session.pins['VCC,CLK'].ppmu_voltage_level = 4
session.sites[2].pins['VCC,CLK'].ppmu_voltage_level = 4
```

The 1st way is already available in nidigital API. The 2nd way would require adding `pins` as a repeated capability.
The 3rd way would require the session object to hold 2 active repeated capability objects, which would require changes to code generator.

texasaggie97:

* `pins` is now a repeated capability
    ``` python
    session.pins[['VCC','CLK']].ppmu_voltage_level = 4
    session.pins['VCC,CLK'].ppmu_voltage_level = 4
    ```

* Fix #1093 

### DIFF
--- a/docs/nidigital/class.rst
+++ b/docs/nidigital/class.rst
@@ -3384,6 +3384,18 @@ channels
         passes a string of :python:`'0, 1, 2'` to the set attribute function.
 
 
+pins
+----
+
+    .. py:attribute:: nidigital.Session.pins[]
+
+        .. code:: python
+
+            session.pins['0-2'].channel_enabled = True
+
+        passes a string of :python:`'0, 1, 2'` to the set attribute function.
+
+
 
 Properties
 ==========
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### NI-Digital Pattern Driver
     * #### Added
+        * `pins` is now a valid repeated capability
     * #### Changed
         * Fix get/set properties - [#1062](https://github.com/ni/nimi-python/issues/1062)
         * Removed array-size parameter from apply_tdr_offsets() and write_source_waveform_broadcast_u32() methods - [#1070](https://github.com/ni/nimi-python/issues/1070)
--- a/generated/nidigital/session.py
+++ b/generated/nidigital/session.py
@@ -1833,6 +1833,7 @@ class Session(_SessionBase):
 
         # Instantiate any repeated capability objects
         self.channels = _RepeatedCapabilities(self, '')
+        self.pins = _RepeatedCapabilities(self, '')
 
         self.tclk = nitclk.SessionReference(self._vi)
 
--- a/src/nidigital/metadata/config.py
+++ b/src/nidigital/metadata/config.py
@@ -39,7 +39,11 @@ config = {
         {
             'prefix': '',
             'python_name': 'channels'
-        }
+        },
+        {
+            'prefix': '',
+            'python_name': 'pins',
+        },
     ],
     'session_class_description': 'An NI-Digital session',
     'session_handle_parameter_name': 'vi',


marcoskirsch:
Please put some usage examples in the description.
Would this make adding a "sites" repeated capability redundant?

texasaggie97:
Usage (from issue) added.

From my understanding pins and sites are two different things so this is needed regardless of what we decide about sites. `sites` may need a new solution as it doesn't look like it is a normal repeated capability.

sbethur:
> From my understanding pins and sites are two different things so this is needed regardless of what we decide about sites.

That's correct. We need `pins` to be a rep cap irrespective of what we do for sites. This is mainly so users can do
```
session.pins['VCC,CLK'].ppmu_voltage_level = 4
```
instead of
```
session.channels['VCC,CLK'].ppmu_voltage_level = 4
```

If any operation needs to be done on specific sites, with this change, you can do
```
session.pins['site0/VCC, site0/CLK, site1/VCC, site1/CLK'].
```
This is obviously not good in terms of usability. It's slightly better if users setup pin groups in their pin maps. Then they can do
```
session.pins['site0/PinGroupA, site1/PinGroupA'].
``` 
But this is still not ideal. It would nice to do
```
session.sites[0,1].pins['PinGroupA'].
```
We could still need to allow using _siteN_ in `pins` rep cap, to allow users to do
```
session.pins['site0/VCC, site1/CLK'].   # This use-case is uncommon
```

docs/nidigital/class.rst
```diff
@@ -3384,6 +3384,18 @@ channels
...
 
 
+pins
+----
+
+    .. py:attribute:: nidigital.Session.pins[]
+
+        .. code:: python
+
+            session.pins['0-2'].channel_enabled = True
```
sbethur: In Digital driver, pin names cannot begin with a number. Also, range doesn't make sense for pins.

The example should show something like `'PinA,PinB'`

This example does bring up one thing.. Since many methods that take in `channel_list` can accept both channel numbers and pins, doing something like `pins['0-2']` will work. Similarly, doing `channels['PinA']` will also work. I wonder if we should put some checks in Python to prevent such confusing misuse.

texasaggie97: We do not do any of these checks in other drivers either. Now that our internal metadata is getting more complete, this might be possible, but would require changes to the metadata and to how we generate the code. So something like this works as well:
``` python
with nifgen.Session('dev1') as session:
    session.channel['marker1'].some_trigger = 'dev2/clk_out'
```

sbethur: First, I want to point out that in nidigital C API, there are functions that accept "channelList" (e.g. niDigital_TDR) param - These functions actually accept both channel numbers and pin lists.
 
Then there are functions that accept only pinList (e.g. niDigital_ConfigureTimeSetDriveFormat) param - These functions accepts only pin lists.
 
There are no functions that accept only channel numbers.

> We do not do any of these checks in other drivers either. Now that our internal metadata is getting more complete, this might be possible, but would require changes to the metadata and to how we generate the code. So something like this works as well:
>```
>with nifgen.Session('dev1') as session:
>    session.channel['marker1'].some_trigger = 'dev2/clk_out'
>```

But the difference is, in nifgen, driver will return an error for `channel['marker1']`. But in nidigital, `pins['0-2']` and `channels['PinA']` will both work for methods that accept channel_list parameter.
 
Also, I think we need changes to codegen and/or metadata to make pins rep_cap work for methods that accept `channel_list` i.e. both channel numbers and pin lists.

marcoskirsch:
> But this is still not ideal. It would nice to do
> `session.sites[0,1].pins['PinGroupA'].`

This will require more generator work. #1093 is already open for this discussion.

> `sites` may need a new solution as it doesn't look like it is a normal repeated capability.

Agreed. Let's have that discussion in #1111

docs/nidigital/class.rst
marcoskirsch: This is not a good example. How difficult is it to do something `nidigital` specific?
#1125 created

docs/nidigital/class.rst
```diff
@@ -3384,6 +3384,18 @@ channels
...
+
+        passes a string of :python:`'0, 1, 2'` to the set attribute function.
```
marcoskirsch: We need to stop documenting implementation details. The customer doesn't have to know or care how this ends up calling into the C API.

#1126 created

docs/nidigital/class.rst
marcoskirsch: > The example should show something like 'PinA,PinB'

Or the more pythonic ['PinA', 'PinB].

PR #1143, opened 2019-11-21
Fix Linux support

### LINKED ISSUE #1140: Erroneously passing lib and .so decorations to find_library on Linux
Question about library singleton.py. It looks like we recently changed it to pass the DLL name through find_library ([here](https://github.com/ni/nimi-python/commit/54ab723c78415ca90ca3986b7c2a91f588e268d3)), but the DLL names contain "lib" and ".so" -fixes ([here](https://github.com/ni/nimi-python/blob/master/generated/nidmm/_library_singleton.py#L15)), but find_library does not like those prefixes ([here](https://docs.python.org/3/library/ctypes.html#finding-shared-libraries)). This appears to be causing an issue on @Fladolcetta's machine with DMM on OpenSUSE, should we put together a PR for that?

lkolbly:

(unknown)

Does not pass libraries with lib/.so trimmings to ctypes find_library.

* Fix #1140 

Tested that modifying the generated file on a linux machine fixes the crash outlined in the issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,12 +45,14 @@ About
 The **nimi-python** repository generates Python bindings (Application Programming Interface) for interacting with the Modular Instrument drivers. Currently, the following drivers are supported:
 
 * NI-DCPower (Python module: nidcpower)
+* NI-Digital Pattern Driver (Python module: nidigital)
 * NI-DMM (Python module: nidmm)
 * NI-FGEN (Python module: nifgen)
-* NI-SCOPE (Python module: niscope)
-* NI-SWITCH (Python module: niswitch)
 * NI-ModInst (Python module: nimodinst)
+* NI-SCOPE (Python module: niscope)
 * NI Switch Executive (Python module: nise)
+* NI-SWITCH (Python module: niswitch)
+* NI-TClk (Python module: nitclk)
 
 It is implemented as a set of `Mako templates <http://makotemplates.org>`_ and per-driver metafiles that produce a Python module for each driver. The driver is called through its public C API using the
 `ctypes <https://docs.python.org/2/library/ctypes.html>`_ Python library.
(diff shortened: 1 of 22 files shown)


README.rst
```diff
@@ -440,10 +438,10 @@ Driver specific installation instructions can be found on Read The Docs:
 * `nidcpower <http://nimi-python.readthedocs.io/en/master/nidcpower.html#installation>`_
 * `nidmm <http://nimi-python.readthedocs.io/en/master/nidmm.html#installation>`_
 * `nifgen <http://nimi-python.readthedocs.io/en/master/nifgen.html#installation>`_
-* `nimodinst <http://nimi-python.readthedocs.io/en/master/nimodinst.html#installation>`_
 * `niscope <http://nimi-python.readthedocs.io/en/master/niscope.html#installation>`_
-* `nise <http://nimi-python.readthedocs.io/en/master/nise.html#installation>`_
 * `niswitch <http://nimi-python.readthedocs.io/en/master/niswitch.html#installation>`_
+* `nise <http://nimi-python.readthedocs.io/en/master/nise.html#installation>`_
+* `nimodinst <http://nimi-python.readthedocs.io/en/master/nimodinst.html#installation>`_
```
marcoskirsch: These are all unrelated changes from previous PRs. Please merge latest master.

src/nifgen/metadata/config.py
```diff
@@ -16,7 +16,7 @@
     'library_info': {
         'Linux': {
             '64bit': {
-                'name': 'libfgen.so',
+                'name': 'fgen',
```
marcoskirsch: Please make this `nifgen`. Even though that driver has no Linux support, I am fairly sure the library would not be "fgen".

README.rst
```diff
@@ -45,14 +45,12 @@ About
 The **nimi-python** repository generates Python bindings (Application Programming Interface) for interacting with the Modular Instrument drivers. Currently, the following drivers are supported:
 
 * NI-DCPower (Python module: nidcpower)
-* NI-Digital Pattern Driver (Python module: nidigital)
```
sbethur: @marcoskirsch, changes in this file is because I updated README without realizing that it's codegen'd. I'm looking into how to update this file the correct way.

README.rst
lkolbly: Based on @sbethur 's comment below, I think it is correct to include this change?

sbethur: @lkolbly, I'm creating a PR to fix this. Please exclude this file from your PR.

sbethur:
ok to test

README.rst
lkolbly: @sbethur Pushed a commit that should revert it.

texasaggie97:
ok to test

PR #1155, opened 2019-12-03
Add back tests that had intermittent failures

### LINKED ISSUE #650: Add intermittent system tests back to NI-SCOPE system tests
test_configure_trigger_video and test_filter_coefficients caused intermittent test failures. Internal issue report 565504 has been filed to resolve these failures.

We need to re-add this test to complete coverage.

marcoskirsch:

* Add commented out tests back, they were failing due to driver runtime bugs that have been since fixed. We have already updated nimi-bot.
* Add files created during local system test runs to `.gitignore`.

* Fix #650

Will let nimi-bot do its magic.


src/niscope/system_tests/test_system_niscope.py
```diff
@@ -248,17 +248,22 @@ def test_configure_chan_characteristics(session):
...
     with niscope.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:5142; BoardType:PXI') as session:  # filter coefficients methods are available on devices with OSP
-        assert [1.0, 0.0, 0.0] == session.get_equalization_filter_coefficients(3)
+        assert [1.0] + [0.0] * 34 == session.get_equalization_filter_coefficients() # coefficients list should have 35 items
         try:
             filter_coefficients = [1.0, 0.0, 0.0]
             session.configure_equalization_filter_coefficients(filter_coefficients)
         except niscope.Error as e:
-            assert e.code == -1074135024  # coefficients list should have 35 items
-'''
+            assert "Incorrect number of filter coefficients." in e.description
```
texasaggie97: Are you sure you want to look for specific words in the description? There is no guarantee that it won't change from driver release to driver release.

marcoskirsch: I'm not sure, but this is something we've done before in system tests:

```
(base) kirsch@us-aus-wireless-10-0-146-136 nimi-python % grep -r e\.description src | grep system_tests
src/niscope/system_tests/test_system_niscope.py:        assert e.description.find('Simulation does not support the selected model and board type.') != -1
src/niscope/system_tests/test_system_niscope.py:        assert e.description.find('Attribute is read-only.') != -1
src/nidcpower/system_tests/test_system_nidcpower.py:        assert e.description.find('The option string parameter contains an entry with an unknown option value.') != -1
src/nidcpower/system_tests/test_system_nidcpower.py:        assert e.description.find('Attribute is read-only.') != -1
src/nidcpower/system_tests/test_system_nidcpower.py:        assert e.description.find('The requested function is not available when multiple channels are present in the same session.') != -1
src/nidmm/system_tests/test_system_nidmm.py:        assert e.description.find('The option string parameter contains an entry with an unknown option value.') != -1
src/niswitch/system_tests/test_system_niswitch.py:        assert e.description.find('Invalid resource name.') != -1
src/nise/system_tests/test_system_nise.py:        assert e.description.find('The route specification string contains invalid characters or could not be understood.') != -1
src/nifgen/system_tests/test_system_nifgen.py:        assert e.description.find('Insufficient location information or resource not present in the system.') != -1
src/nifgen/system_tests/test_system_nifgen.py:        assert e.description.find('Attribute is read-only.') != -1
```

I am also considering having the system test verify that the bug we're working around is there. This way whenever we install a new driver runtime on nimi-bot that includes a fix, system tests would fail and remind us to remove the workaround. What do you think?

texasaggie97: Hmmm. Since it's already there, then this isn't any worse. We maybe shouldn't have done it at all, but since we already do, there is precedence.

texasaggie97: I think it depends on how hacky the work around is. And this one doesn't seem that bad since it is reasonable that someone would explicitly put the channel even on a single channel device. 

Given that though, I do not have a strong feeling one way or the other for this one.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -248,17 +248,27 @@ def test_configure_chan_characteristics(session):
...
+            assert e.code == -1074135024
+        filter_coefficients = [0.01] * 35
+        try:
+            # TODO(marcoskirsch): The following should work. It doesn't due to internal NI-SCOPE driver bug 959625.
+            #                     The workaround is to explicitly pass channel "0" rather than the default "" to the driver runtime
+            #                     even though the two should be equivalent on a PXI-5142 (a 1 channel device).
```
ni-jfitzger: #                     even though the two should be equivalent on a PXI-5142 (a 1 channel device).
 This is incorrect.  The PXI-5142 has 2 channels.

marcoskirsch: Fixed in #1248

PR #1176, opened 2020-01-03
Fix LF in repo and locally

texasaggie97:

* Be explicit that we want text files to use LF (unix style) line endings both locally and in the repo
* This will not fix things immediately, but going forward, as files are updated, I think it will help

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,6 +3,15 @@
 *.py text=set eol=lf
 *.rst text=set eol=lf
 *.inc text=set eol=lf
+*.md text=set eol=lf
+*.ini text=set eol=lf
+*.txt text=set eol=lf
+*.yml text=set eol=lf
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,5 +1,8 @@
 # Auto detect text files and perform LF normalization
-* text=auto
+# We explicitly set eol to LF on checkin and checkout
+*.py text=set eol=lf
+*.rst text=set eol=lf
+*.inc text=set eol=lf
```
marcoskirsch: Why specific extensions? What would be the drawback of wild-carding this?
For example, we're missing markdown.

texasaggie97: Because I don't want to force the conversion on any binary files (like the zip files that will be added in a different PR)

Added many more text types.

Issue #1174, opened 2020-01-03
Simplify testing by keeping installable copy of generated files in git


texasaggie97:
# Background
Right now we keep a copy of the generated files in git in `$(ROOT_DIR)/generated/<driver>`. But, this copy is not installable since the files are not all in the correct relative positions.

We keep these files here for ease of validating any changes: We can see what, if anything, changed in the generated files for a given PR.

This also makes the build (and makefiles) more complicated in order to copy the generated files to the generated directory in addition to the location where they are actually built into.

# Proposal
Do not add the files in generated to git and instead, add a subset of the files generated into `bin`. This subset would be:
* `bin/<driver>/<driver>/**`
* `bin/<driver>/setup.py`
* `bin/<driver>/README.rst`

We would not add any of the generated wheels or sdist files.

# Benefit
There are a couple of benefits by doing this.
1. We know exactly what was tested since we are testing what is in git
1. The tests will run faster since we would not need to run the codegen step on every platform/python version we want to test
    1. We would build the wheel from the files in git for each platform/python version since we do want to test the wheel itself
1. Simplify the make files to no longer need the extra copy to generated.

marcoskirsch:
So instead of copying generated files to `generated/` after code-generation, we'll just keep them in their original codegeneration location and add those to git.

I can see how this is confusing to someone looking at our repo:

* It is not clear that things inside `bin/` are source files
* It is not clear that on git only a subset of the files that are produced during a build are included.

I wonder if we can better show intent by playing around with the folder names/structure inside `bin/`.

PR #1179, opened 2020-01-03
Move generated files

texasaggie97:
* No longer use bin and instead generate into generated files, but in the correct locations to install
    * `git mv` generated files to actual location
    * The hierarchy allows for installing from the generated directory whereas before, the hierarchy did not

* First step for #1174

* Visual inspection of generated files added to git


generated/nidcpower/README.rst
```diff
@@ -0,0 +1,201 @@
+Overall Status
```
sbethur: Was this change and other README.rst changes supposed to show up as moves from somewhere?

.gitignore
```diff
@@ -1,7 +1,17 @@
 *.pyc
 .cache
 .com.apple.timemachine.supported
-bin/
+
+generated/docs/
```
marcoskirsch: Add comment

```
# The "generated" folder is created during a build.
# Normally you wouldn't want those files to be in source control.
# But we actually do keep the latest copy of some of the generated code, specifically the Python modules, for two reasons:
#   1. So that we can make changes to the generator and easily review what effect those changes have in the generated code.
#   2. So that our system tests can create a wheel out of the files and quickly install them, so that they don't need to go through the full code-generation process.
```

(please improve the wording, review for accuracy, and format as needed)

tools/install_local_wheel.py
```diff
@@ -27,7 +27,7 @@ def configure_logging(lvl=logging.WARNING, logfile=None):
 def main():
     # Setup the required arguments for this script
     usage = """
-Install the wheel found in bin/<driver>/dist
+Install the wheel found in generated/<driver>/dist
```
sbethur: I grepped for "bin" and found two similar places where updates might be required:

- https://github.com/ni/nimi-python/blob/969ceaa214392e68850d841d1b2599cb50ffde74/tools/build_release.py#L137
- https://github.com/ni/nimi-python/blob/969ceaa214392e68850d841d1b2599cb50ffde74/tools/simple_mako.py#L78

Makefile
```diff
@@ -11,8 +11,8 @@ export ROOT_DIR
 BUILD_HELPER_DIR := $(ROOT_DIR)/build
 export BUILD_HELPER_DIR
 
-BIN_DIR := $(ROOT_DIR)/bin
-export BIN_DIR
+GENERATED_DIR := $(ROOT_DIR)/generated
```
marcoskirsch: FWIW I think this name is more self explanatory, I like it.

build/Makefile
```diff
@@ -62,10 +62,10 @@ $(foreach t,$(DEFAULT_TARGETS),$(eval $(call per_target,$(t))))
 
 SUPPRESS_ERROR_OUTPUT ?= 2> /dev/null
 
-# We need to ensure bin exists for Travis CI builds
+# We need to ensure generated exists for Travis CI builds
```
marcoskirsch: add / after generated to make it clear it's a directory

build/helper/metadata_add_all.py
```diff
@@ -609,7 +609,7 @@ def add_all_metadata(functions, attributes, enums, config, persist_output=True):
     fixup_enum_names(config)
 
     pp_persist = pprint.PrettyPrinter(indent=4, width=200)
-    metadata_dir = os.path.join('bin', 'processed_metadata')
+    metadata_dir = os.path.join('generated', 'processed_metadata')
```
marcoskirsch: observation: it's too bad this is not single-sourced
Should not be fixed in this PR

generated/nidigital/README.rst
```diff
@@ -0,0 +1,185 @@
+Overall Status
```
marcoskirsch: Other files are shown in diff as: "File renamed without changes."
Except this one. What gives?

tox.ini
```diff
@@ -36,13 +36,13 @@ changedir =
...
     flake8: .
     pkg: .
-    # pkg-nidcpower: bin/nidcpower
-    # pkg-nidmm: bin/nidmm
-    # pkg-nifgen: bin/nifgen
-    # pkg-niscope: bin/niscope
-    # pkg-niswitch: bin/niswitch
-    # pkg-nimodinst: bin/nimodinst
-    # pkg-nise: bin/nise
+    # pkg-nidcpower: generated/nidcpower
```
marcoskirsch: don't know what this does but notice (orthogonal issue) `nidigital` is missing.

generated/nidcpower/README.rst
texasaggie97: Yes, generated already was where we were putting the generated files, but just not in a format that we could re-use. By `git mv`ing them we can keep the history of the files.

.gitignore
texasaggie97: Added this comment:
```
# The "generated" folder is created during a build.
# Normally you wouldn't want those files to be in source control.
# But we actually do keep the latest copy of some of the generated code, specifically the Python modules, for two reasons:
#   1. So that we can make changes to the generator and easily review what effect those changes have in the generated code.
#   2. So that our system tests can create a wheel out of the files and quickly install them, instead of going through the full code-generation process.
# These are the generated files we do NOT want it git
```

build/Makefile
texasaggie97: Added

generated/nidigital/README.rst
texasaggie97: `README.rst`'s are new to generated so they really are an add.

We need them in git though so that building the wheel will work without having to generate any files.

tools/install_local_wheel.py
texasaggie97: First one is changed to generated
Second the description was fixed - it was wrong before.

tox.ini
texasaggie97: They didn't do anything since they were commented out

All removed.

.gitignore
marcoskirsch: those files -> build artifacts
(small improvement)

generated/nidigital/README.rst
marcoskirsch: ah ok

generated/nifake/README.rst
```diff
@@ -0,0 +1,21 @@
+This is the MIT license: http://www.opensource.org/licenses/mit-license.php
+
+Copyright (c) 2017, National Instruments Corp.
+NI-DMM, NI-DCPower, NI-FGEN, NI-SCOPE, NI-SWITCH, NI Switch Executive are trademarks of National
+Instruments.
```
marcoskirsch: year is old, list of drivers is old

Is this true elsewhere? May need to open an issue.

texasaggie97: See #1183

PR #1180, opened 2020-01-06
Improve test efficiency

### LINKED ISSUE #1174: Simplify testing by keeping installable copy of generated files in git

texasaggie97:
Improve test efficiency by not running codegen on all python versions
* Add "installers" tox environment
    * This calls the `installers' target in make which only builds the wheel and sdist per driver
    * Moves readme, setup.py and documentation files to the default target `all` instead of being dependencies of the wheel and sdist
    * This allows the only thing to be built during installers to be the wheel and sdist
    * Wheel and sdist build use whatever is in `generated/` at that time
    * 'installers' environment runs on all supported python versions so that we have wheels to test with
* `clean`, `codegen`, `flake8`, `docs`, & `pkg` have been limited to running on Python 3.8 only
    * This allows us to use tox as it is intended
    * On Travis CI, Python 3.8 will run all environents for complete coverage
    * All other versions will only run installers and unit tests in order to speed up testing
* Orders versions of Python to test on Travis CI based on length of time it takes to complete - longer first
    * For some reason PyPy3 takes twice as long as the next one
    * All Python 3.8 since all tests happen here (codegen, flake8, docs, pkg)
* Added tests for dev builds of all supported Python versions
    * Allowed to fail

* Fix #1174 

* Unit
* System
* Visual inspection of build/test logs to verify what builds when


texasaggie97:
test this please

tools/install_local_wheel.py
```diff
@@ -27,7 +27,7 @@ def configure_logging(lvl=logging.WARNING, logfile=None):
 def main():
```
sbethur: I remember talking last week about installing from wheel for nifake, and from source files for other drivers. Is that not what we're going with?

.travis.yml
```diff
@@ -2,38 +2,45 @@ group: travis_latest
...
       # so multiple others can be running while these continue to run.
-      - python: pypy
       - python: pypy3
-      - python: 2.7
-      - python: 3.4
+      # We put Python 3.8 based towards since this is where additional tests are ran so we give it more time to run
+      - python: 3.8
+        dist: xenial
+        sudo: true
+      - python: 3.8-dev
```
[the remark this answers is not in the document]
texasaggie97: Added `# Allow tests using under development builds of Python to fail without failing the PR/change` to the allow_failures section.

.travis.yml
```diff
@@ -2,38 +2,45 @@ group: travis_latest
...
-      # We put 3.6 first since this is where flake8 and docs are tested
-      - python: 3.6
-      # We put pypy towards the beginning because for some reason the build takes much longer, this starts it earlier 
+      # We put pypy at the beginning because for some reason the build takes much longer, this starts it earlier 
       # so multiple others can be running while these continue to run.
-      - python: pypy
       - python: pypy3
-      - python: 2.7
-      - python: 3.4
+      # We put Python 3.8 based towards since this is where additional tests are ran so we give it more time to run
```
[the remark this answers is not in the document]
texasaggie97: Reworded to `# We put Python 3.8 based tests towards the start. This is where additional tests are ran so we give it more time to run`

tools/install_local_wheel.py
texasaggie97: We never install the wheel for nifake since we do not run system tests on it.

We do install the locally built nitclk wheel since that needs to be there to satisfy the dependencies when running nifake unit tests.

In the system tests, we build and install the wheel for each driver (and nitclk if the driver depends on it).

sbethur: Got it. Thanks.

.gitignore
```diff
@@ -2,6 +2,12 @@
 .cache
 .com.apple.timemachine.supported
 
+# The "generated" folder is created during a build.
+# Normally you wouldn't want build artifacts to be in source control.
+# But we actually do keep the latest copy of some of the generated code, specifically the Python modules, for two reasons:
+#   1. So that we can make changes to the generator and easily review what effect those changes have in the generated code.
+#   2. So that our system tests can create a wheel out of the files and quickly install them, instead of going through the full code-generation process.
+# These are the generated files we do NOT want it git
```
marcoskirsch: it -> in

.travis.yml
```diff
@@ -5,7 +5,7 @@ matrix:
       # We put pypy at the beginning because for some reason the build takes much longer, this starts it earlier 
       # so multiple others can be running while these continue to run.
       - python: pypy3
-      # We put Python 3.8 based towards since this is where additional tests are ran so we give it more time to run
+      # We put Python 3.8 based tests towards the start. This is where additional tests are ran so we give it more time to run
```
marcoskirsch: put -> rum

marcoskirsch: are ran -> are run
more time to run -> more time.

.travis.yml
```diff
@@ -29,6 +29,7 @@ matrix:
         dist: xenial
         sudo: true
     allow_failures:
+      # Allow tests using under development builds of Python to fail without failing the PR/change
```
marcoskirsch: The comment is helpful, but why do we even run tests under non-released version of Python?

texasaggie97: To hopefully see if they break something. Should I remove them?

marcoskirsch: But if they do, do we even notice it since we allow failures?
We are making every single PR slower than it needs to be, right?

texasaggie97: Removed

PR #1190, opened 2020-01-10
Restructure testing

### LINKED ISSUE #1173: System Tests do not report back code coverage results
### Description of issue

Our system tests run code coverage checks, but they are never reported back to GitHub nor are they used as pass/fail criteria.

We should look at the coverage results and actually fail when they go down below a threshold.

### LINKED ISSUE #1188: Eliminate the need to run GNU make while executing system tests
nimi-bot builds the module installers (wheels) and installs them before it runs system tests. In order to do so,  GNU make is involved which means that nimi-bot needs to have mingw installed. We are no longer recommending / supporting [MinGW](http://www.mingw.org). Customers should use WSL.

So one alternative is WSL. But WSL image with 32-bit Python is not easy to find. Also WSL solution would involve 2 different python interpreters (one of Windows and one for WSL).

We think it's a better solution to completely eliminate the need for running GNU make. This will simplify the setup of nimi-bot altogether.

We could have a way to build the installers outside of GNU make, or we could simply install from `generated/` using setup.py with a small loss of test coverage (we would not be running the wheels anymore). Either way it's better than requiring MinGW.

texasaggie97:
* Restructure how tests are managed
    * `tox.ini` split into several different files
        * `tox.ini` is the default and is what is used when no command line parameter set - envlist is defined to work for developers running tox manually
        * `tox-travis.ini` - Very similar to `tox.ini` with slightly different envlist

            ``` makefile
            # tox.ini and tox-travis.ini are copies of the same content with different envlist commented out
            # tox-travis.ini will have pyXX-clean and all pyXX-installers in the default envlist, while the developer tox.ini 
            # does not have clean and only has one pyXX-installers
            ```

        * `generated/<driver>/tox-driver.ini` - generated from make template, per driver.
            * Knows how to generate and install nitclk wheel for drivers needing it
            * Reports coverage to codecov
            * Each driver runs separately in different venvs
            * This is in anticipation of generating jobs per driver per bitness: This allows them to run in parallel as well as moving some of the Jenkins configuration into source
* System tests now run via a batch file `tools/system_tests.bat`
    * This moves how tests should be ran into source instead of in Jenkins configuration
    * Currently the jobs on the new Jenkins server are configured to use the old method if the file doesn't exist, and to use the file if it does.
* Tag unittests and systemtests on codecov
* Name where the code comes from as well as a link to the build on codecov

* Fix #1173 
* Fix #1188 

* System tests
* Unit tests
* Inspection of codecov results


texasaggie97:
test this please

tools/system_tests.bat
```diff
@@ -0,0 +1,48 @@
+@echo off
+
+echo Testing nidcpower
+pushd generated\nidcpower
+c:\python38\python.exe -m tox -c tox-driver.ini
```
sbethur: Can we not hardcode the Python version and path?

tox-travis.ini
```diff
@@ -0,0 +1,220 @@
+# Tox (http://tox.testrun.org/) is a tool for running tests
```
sbethur: Should we rename this file to be not tied to "travis"? maybe tox-ci.ini?

tox.ini
```diff
@@ -100,36 +104,6 @@ commands =
...
-    system_tests: coverage run --rcfile=tools/coverage_system_tests.rc --append --source nimodinst -m py.test src/nimodinst/examples --junitxml=generated/junit/junit-modinst-{envname}-{env:BITNESS:64}.xml {posargs}
-    system_tests: coverage run --rcfile=tools/coverage_system_tests.rc --append --source nimodinst -m py.test src/nimodinst/system_tests --junitxml=generated/junit/junit-modinst-{envname}-{env:BITNESS:64}.xml {posargs}
-    system_tests: python tools/install_local_wheel.py --driver nise
-    system_tests: coverage run --rcfile=tools/coverage_system_tests.rc --source nise -m py.test src/nise/examples --junitxml=generated/junit/junit-nise-{envname}-{env:BITNESS:64}.xml {posargs}
-    system_tests: coverage run --rcfile=tools/coverage_system_tests.rc --source nise -m py.test src/nise/system_tests --junitxml=generated/junit/junit-nise-{envname}-{env:BITNESS:64}.xml {posargs}
-    system_tests: coverage report --rcfile=tools/coverage_system_tests.rc
-    system_tests: coverage html --rcfile=tools/coverage_system_tests.rc --directory=generated/htmlcov/system_tests
     nidcpower_system_tests: python --version
```
sbethur: Why are these lines still required here?

sbethur: I mean `<driver>_system_tests` rules

tools/system_tests.bat
texasaggie97: To do that would require making sure the version we want is the first version on the system path on the test machine. I would rather be explicit about it here than take a chance that the wrong version is used.

tox-travis.ini
texasaggie97: I am not sure that a different CI system would be able to use this file as is. So at this point, I am only confident it will work with Travis CI.

tox.ini
texasaggie97: Because the internal test suites for the drivers assume tox.ini. I think it is beyond the scope of this PR to fix all those up, though it may be something we should do if we do go with this change.

They could be removed from `tox-travis.ini` but that would mean the two files are no longer the same other than the env list.

.travis.yml
```diff
@@ -31,9 +31,9 @@ install:
   - travis_retry pip install --upgrade tox tox-travis coveralls codecov
 
 script:
-  - tox
+  - tox -c tox-travis.ini
 
 after_success:
   - travis_retry coveralls --rcfile=tools/coverage_unit_tests.rc
-  - travis_retry codecov
+  - travis_retry codecov --flags unittests
```
marcoskirsch: Where is this flag defined? Everywhere else we refer to unit tests as `unit_tests`.

build/templates/tox-driver.ini.mako
```diff
@@ -0,0 +1,71 @@
...
+    ${module_name}-system_tests: python ../../tools/install_local_wheel.py --driver nitclk --start-path ../..
+% endif
+    ${module_name}-system_tests: coverage run --rcfile=../../tools/coverage_system_tests.rc --source ${module_name} -m py.test ../../src/${module_name}/examples --junitxml=../../generated/junit/junit-${module_name}-{envname}-{env:BITNESS:64}.xml {posargs}
+    ${module_name}-system_tests: coverage run --rcfile=../../tools/coverage_system_tests.rc --source ${module_name} -m py.test ../../src/${module_name}/system_tests --junitxml=../../generated/junit/junit-${module_name}-{envname}-{env:BITNESS:64}.xml {posargs}
+    # Create the report to upload
+    ${module_name}-system_tests: coverage xml -i --rcfile=../../tools/coverage_system_tests.rc
+    # Display the coverage results
+    ${module_name}-system_tests: coverage report --rcfile=../../tools/coverage_system_tests.rc
+    # token is from codecov
+    ${module_name}-system_tests: codecov -X gcov --token=4c58f03d-b74c-489a-889a-ab0a77b7809f --no-color --flags systemtests --name ${module_name} --root ../.. --file ../../generated/${module_name}/coverage.xml
```
marcoskirsch: flag should be `system_tests` (even though it doesn't matter... for consistency)

build/templates/tox-driver.ini.mako
```diff
@@ -0,0 +1,71 @@
...
+    module_name = config['module_name']
+    if config['supports_nitclk']:
+        nitclk_env = 'py38-{}-nitclk_wheel,'.format(module_name)
+    else:
+        nitclk_env = ''
+%>\
+# Tox (http://tox.testrun.org/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox" from this directory.
```
marcoskirsch: > and then run "tox" from this directory

is this still accurate?

build/templates/tox-driver.ini.mako
```diff
@@ -0,0 +1,71 @@
...
+description =
+% if config['supports_nitclk']:
+    ${module_name}-nitclk_wheel: Build the nitclk wheel
+% endif
+    ${module_name}-system_tests: Run ${module_name} system tests (requires driver runtime to be installed)
+
+changedir =
+% if config['supports_nitclk']:
+    ${module_name}-nitclk_wheel: ../../generated/nitclk
+% endif
```
marcoskirsch: this is pretty cool

build/templates/tox-driver.ini.mako
```diff
@@ -0,0 +1,71 @@
...
+    ${module_name}-system_tests: .
+
+commands =
+% if config['supports_nitclk']:
+    ${module_name}-nitclk_wheel: python.exe setup.py bdist_wheel --universal
+% endif
+    ${module_name}-system_tests: python --version
+    ${module_name}-system_tests: python -c "import platform; print(platform.architecture())"
+    ${module_name}-system_tests: python -c "import ${module_name}; ${module_name}.print_diagnostic_information()"
```
marcoskirsch: This is good, I noticed we were only printing this for NI-DCPower.

tools/system_tests.bat
```diff
@@ -0,0 +1,48 @@
+@echo off
```
marcoskirsch: This whole file is Windows-specific.
Given that Linux supports a subset of the runtimes, do you expect to have a `.sh` file that is invoked in the Linux case?

Ideally we could have a single file (could be Python) that has a list of supported runtimes by OS and programmatically executes the appropriate list.

Have you considered this? I'm ok if that's handled in a separate PR but may be easier to do the groundwork now. It would also be less repetitive. We could even consider codegen.

tox-travis.ini
```diff
@@ -0,0 +1,220 @@
...
+    docs: Generate documentation
+    pkg: Verify the package
+    nidcpower_system_tests: Run system tests for nidcpower
+    nidigital_system_tests: Run system tests for nidigital
+    nidmm_system_tests: Run system tests for nidmm
+    nifgen_system_tests: Run system tests for nifgen
+    niscope_system_tests: Run system tests for niscope
+    niswitch_system_tests: Run system tests for niswitch
+    nimodinst_system_tests: Run system tests for nimodinst
+    nise_system_tests: Run system tests for nise
```
marcoskirsch: Do you think it would make sense to codegen this file?
Maybe then we wouldn't have two copies of so much repetitive stuff.

(could be done in a separate PR if you agree, we just need to open an issue)

build/templates/tox-driver.ini.mako
texasaggie97: No, changed.

.travis.yml
texasaggie97: This is used as a tag on codecov (though flags are temporarily disabled there while refactoring). They do not allow using a `_` in a flag for some reason.

build/templates/tox-driver.ini.mako
texasaggie97: Yes, but that was when they were all ran at the same time. I didn't see a need to print the same information over and over per driver. Now that each driver/platform will be a separate job, it does make sense to get this information into the log file.

build/templates/tox-driver.ini.mako
texasaggie97: Same as above. codecov doesn't allow `_` in flags.

tools/system_tests.bat
texasaggie97: Somewhere there will be platform specific code. My thinking was there would also be a `system_tests.sh` file once we add Linux system tests, and the jobs would be generated to call the appropriate file. There is a comment to that in the .groovy file in the next PR in the chain.

tox-travis.ini
texasaggie97: I thought about that, and we could. #1202

.codecov.yml
```diff
@@ -6,7 +6,7 @@ coverage:
   status:
     project:
       default:
-        target: 85%
+        target: 70%
```
marcoskirsch: Please add TODO and link to the issue explaining why it went down and when it needs to be restored.

build/templates/tox-driver.ini.mako
```diff
@@ -0,0 +1,71 @@
...
+    module_name = config['module_name']
+    if config['supports_nitclk']:
+        nitclk_env = 'py38-{}-nitclk_wheel,'.format(module_name)
+    else:
+        nitclk_env = ''
+%>\
+# Tox (http://tox.testrun.org/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox -c tox-driver.ini" from the driver directory.
```
marcoskirsch: Shouldn't this be "tox-${module_name}.ini" ?
Can we replace "the driver directory" with something specific?

The reason is that "driver" is very ambiguous. We refer to things interchangeably as "driver" (ambiguous), Python module, Python API bindings.

We sometimes refer to the "driver runtime" as "driver".

I want to minimize ambiguity and use consistent terminology.

build/templates/tox-driver.ini.mako
```diff
@@ -0,0 +1,71 @@
...
+ignore_basepython_conflict=True
+# We put the .tox directory outside of the workspace so that it isn't wiped with the rest of the repo
+toxworkdir = ../../../.tox
+
+[testenv]
+description =
+% if config['supports_nitclk']:
+    ${module_name}-nitclk_wheel: Build the nitclk wheel
+% endif
+    ${module_name}-system_tests: Run ${module_name} system tests (requires driver runtime to be installed)
```
marcoskirsch: We have the name of the driver runtime in metadata, can you make it so it says:

>(requires the NI-FGEN driver runtime to be installed)

?

.codecov.yml
texasaggie97: Added

build/templates/tox-driver.ini.mako
marcoskirsch: After discussion with @texasaggie97 , a more descriptive name `tox-system_tests.ini` would be better.

texasaggie97: Changed the name

build/templates/tox-driver.ini.mako
texasaggie97: Added driver name

generated/nifake/tox-system_tests.ini
```diff
@@ -0,0 +1,51 @@
+# Tox (http://tox.testrun.org/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox -c tox-driver.ini" from the driver directory.
```
marcoskirsch: This is wrong

texasaggie97: Fixed

PR #1210, opened 2020-01-17
Make nidigital.Session.send_software_edge_trigger() fancy

### LINKED ISSUE #1193: Improve usability of send_software_edge_trigger in nidigital
### Description of issue

[send_software_edge_trigger](https://github.com/ni/nimi-python/blob/afa1df2493212124809c6cd7dbbaa973eefa3fe7/generated/nidigital/nidigital/session.py#L3527) in nidigital should be updated to resemble its counterpart in [nifgen](https://github.com/ni/nimi-python/blob/afa1df2493212124809c6cd7dbbaa973eefa3fe7/generated/nifgen/nifgen/session.py#L2174); so users can write code like:

```
session.send_software_edge_trigger()   # Trigger "start" via software
session.conditional_jump_triggers[0].send_software_edge_trigger()   # Trigger conditional jump trigger 0 via software
```

texasaggie97:

* Makes nidigital `send_software_edge_trigger()` "fancy"
    * To send a SW start trigger, use on the session - `session.send_software_edge_trigger()`
    * To send a SW conditional jump trigger, use on the repeated capability - `session.conditional_jump_trigger.send_software_edge_trigger()`
* Change VI -> method in nifgen version of this "fancy" function

* Fix #1193 

* Unit
* System


CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ All notable changes to this project will be documented in this file.
         * `conditional_jump_triggers` and `pattern_opcode_events` repeated capabilities - [#1191](https://github.com/ni/nimi-python/issues/1191), [#1192](https://github.com/ni/nimi-python/issues/1192)
     * #### Changed
         * `write_source_waveform_site_unique()` now supports `numpy.array` and `list` as site waveform types
+        * `send_software_edge_trigger()` no longer takes any parameters. To send a SW start trigger, use the session `session.send_software_edge_trigger()`. To send a Conditional Jump Trigger, use the repeated capability `session.conditional_jump_triggers[1].send_software_edge_trigger()` - [#1193](https://github.com/ni/nimi-python/issues/1193)
```
sbethur: How will this design scale when we add a new single-instance trigger type?

texasaggie97: The code would need to look at what the repeated capability string begins with and select the trigger type that way.

sbethur: Hmm, so code would look like:
```
session.shiny_new_trigger[0].send_software_edge_trigger()
```
..even if the driver supports only one instance of shiny_new_trigger?

texasaggie97: Yes, and `shiny_new_trigger` would need to have a unique string in the repeated capabilities, like `"conditionalJumpTrigger"`. Unless you know of another way to differentiate the different trigger types.

marcoskirsch: #1220 for discussing this

docs/nidigital/class.rst
```diff
@@ -2471,28 +2471,25 @@ send_software_edge_trigger
...
     .. py:currentmodule:: nidigital.Session
 
-    .. py:method:: send_software_edge_trigger(trigger, trigger_identifier)
-
-            TBD
-
-            
+    .. py:method:: send_software_edge_trigger()
 
+        Sends a command to trigger the signal generator. This VI can act as an
```
marcoskirsch: this device is not a signal generator

instrument?

src/nidigital/templates/session.py/send_software_edge_trigger.py.mako
```diff
@@ -0,0 +1,34 @@
+<%page args="f, config"/>\
```
marcoskirsch: An alternative aproach so we duplicate less lower level code would be to let the generator generate `_send_software_edge_trigger` from metadata but as private. We could even have the enum for selecting the trigger which _should_ be in metadata.

Then the fancy wrapper calls `_send_software_edge_trigger`.

docs/nidigital/class.rst
texasaggie97: Used "Digital Pattern Instrument" in the docstring so using that here.

src/nidigital/templates/session.py/send_software_edge_trigger.py.mako
sbethur: If we go with the solution of using enum values to differentiate between different kinds of trigger, we would probably need a new python-only enum in nidigital because the [existing enum](https://github.com/ni/nimi-python/blob/f047438c6f40b93f9e2377d5aee806d5d3068b99/src/nidigital/metadata/enums.py#L350) is named `Signal` (which is weird in the context of `send_software_edge_trigger`) and it has values that cannot be passed to `send_software_edge_trigger` - NIDIGITAL_VAL_PATTERN_OPCODE_EVENT and NIDIGITAL_VAL_REF_CLOCK.

The reason for this mess is the use of same constants between `send_software_edge_trigger` and `export_signal` in C API. In python API, we got rid of `export_signal`. So we could probably give the enum a more appropriate name and get rid of the values that are not needed.

I think this is not a deal-breaker for the "enums" solution, just something extra to do.

sbethur:
We [decided ](https://github.com/ni/nimi-python/issues/1220#issuecomment-592168536)to not go the route of making `send_software_edge_trigger` fancy.

PR #1235, opened 2020-01-29
nitclk documentation cleanup

### LINKED ISSUE #1225: `nitclk` documentation problems
* says "list of list of"... Should be "list of" and the int needs to go away.
* TBDs need to be filled out
* SessionReference constructor add snippet "Clients should have no need to construct objects of this type directly" or something like that.
* Properties of type SessionReference need Datatype documentation updated

texasaggie97:

* Clean up nitclk documentation in preparation for 1.0
    * Update parameter type for documentation
        * Remove reference to the session number since it is no longer supported
        * Fix how it is displayed since it is a list
    * Update SessionReference type properties type in documentation to remove session number
    * Replace TBD documentation with string from C help

* Fix #1225 


src/nitclk/templates/class.rst.mako
```diff
@@ -63,10 +63,7 @@ ${helper.get_rst_header_snippet('SessionReference', '=')}
     Helper class that contains all NI-TClk properties. This class is what is returned by
     any nimi-python Session class tclk attribute when the driver supports NI-TClk
 
-    .. code:: python
-
-        with niscope.Session('dev1') as session:
-            session.tclk.sample_clock_delay = .42
```
sbethur: Why remove the code snippet?

generated/nitclk/nitclk/session.py
```diff
@@ -73,13 +76,13 @@ class SessionReference(object):
     Default Value is empty string
     '''
     pause_trigger_master_session = _attributes.AttributeSessionReference(6)
-    '''Type: nimi-python Session class, nitclk.SessionReference, NI-TClk Session Number
+    '''Type: nimi-python Session class or nitclk.SessionReference
```
marcoskirsch: * I wonder if "nimi-python Session class" makes sense to users.
* Why have "class" there but not after `nitclk.SessionReference`?

Not sure best way to reword. Maybe just "Driver Session or nitclk.SessionReference" ?

texasaggie97: I don't have a better way either so I will just use "Driver Session or nitclk.SessionReference"

src/nitclk/templates/class.rst.mako
texasaggie97: Added back

PR #1237, opened 2020-01-29
Download link points to release on GitHub, either one associated with this build, or last one

### LINKED ISSUE #1226: When I build nimi-python, the example zip files show as changed on git
If I run
```
python3 -m tox -e clean
python3 -m tox -e codegen
```
...everything gets code-generated fine, but I see the example zip files shown as changed, even though the examples being zipped contain no changes.

This is going to cause source control bloat and noise.

```
(base) kirsch@Kirsch0 nimi-python % git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   generated/examples/nidigital_examples.zip
	modified:   generated/examples/nidmm_examples.zip
	modified:   generated/examples/nifgen_examples.zip
	modified:   generated/examples/nimodinst_examples.zip
	modified:   generated/examples/niscope_examples.zip
	modified:   generated/examples/nise_examples.zip
	modified:   generated/examples/niswitch_examples.zip
	modified:   generated/examples/nitclk_examples.zip
```

texasaggie97:

* Use Python script for zipping examples
* Copy to temp location
* Change timestamp to known time (1/1/2020)
* Zip from there, using store only (no compression)
* Do not store example zips in git
* Link on examples documentation page will point to:
    * Latest release (if this is building non-release build)
    * Release that is being built, when this is a release build

* Fix #1226 


tools/zip_examples.py
```diff
@@ -0,0 +1,86 @@
...
+pp = pprint.PrettyPrinter(indent=4, width=100)
+
+
+def configure_logging(lvl=logging.WARNING, logfile=None):
+    root = logging.getLogger()
+    root.setLevel(lvl)
+
+    formatter = logging.Formatter("[%(asctime)s] [%(levelname)8s] --- %(message)s (%(filename)s:%(funcName)s:%(lineno)s)", "%Y-%m-%d %H:%M:%S")
+    if logfile is None:
+        hndlr = logging.StreamHandler(sys.stdout)
```
sbethur: `handler` seems better

texasaggie97: See #1238

marcoskirsch: Can you use newly merged shared implementation?

tools/zip_examples.py
```diff
@@ -0,0 +1,86 @@
...
+import argparse
+import logging
+import os
+import pprint
+import shutil
+import sys
+import tempfile
+import time
+import zipfile
+
```
marcoskirsch: I think it makes sense to have top level documentation of what this script does, because to the regular viewer this seems overly convoluted.

It should be displayed when the script is invoked with `--help`.

Suggested:

```
Script for compressing a directory of Python examples as a ZIP file in a way that prevents the resulting file from being binary different if the source files didn't change.
* Reset the file modification timestamps on each file because these are preserved during zipping.
* Ignore .pyc files that may be encountered.
```

marcoskirsch:
I have regenerated on macOS and pushed my changes.
@texasaggie97 please pull latest and codegen to see if the ZIP files appear different.

Also: Removed `examples/nidigital_examples.zip` which I believe was added to git by accident https://github.com/ni/nimi-python/commit/5439bbfbecfd5e201d010815977fe973d41ba157

tools/zip_examples.py
texasaggie97: Changed

build/rules.mak
```diff
@@ -75,7 +75,7 @@ $(DRIVER_DOCS_DIR)/%.inc: %.inc.mako $(BUILD_HELPER_SCRIPTS) $(METADATA_FILES)
 
 $(DRIVER_EXAMPLES_ZIP_FILE): $(EXAMPLE_FILES)
 	$(call trace_to_console, "Zipping",$@)
-	$(_hide_cmds)$(call log_command,cd src/$(DRIVER)/examples && zip -u -r -9 $@ * || ([ $$? -eq 12 ] && exit 0) || exit)
+	$(_hide_cmds)$(call log_command,python tools/zip_examples.py --zip-file $@ --src-path src/$(DRIVER)/examples)
```
marcoskirsch: We're no longer trying to create bit-by-bit identical ZIP files. So I think we should go back to the much simpler direct use of `zip` rather than using our own Python script, date blanking, etc.

build/templates/examples.rst.mako
```diff
@@ -23,7 +23,9 @@
     if v.dev is None and v.pre is None:
         url = 'https://github.com/ni/nimi-python/releases/download/{0}/{1}_examples.zip'.format(global_version, module_name)
     else:
-        url = 'https://github.com/ni/nimi-python/blob/master/generated/examples/{0}_examples.zip'.format(module_name)
+        with open('./LAST_RELEASE') as vf:
```
marcoskirsch: I think a clearer term would be "latest release". Last release implies finality.

texasaggie97:
test this please

LAST_RELEASE
```diff
@@ -0,0 +1 @@
+1.1.5
```
marcoskirsch: Is this now a hand-maintained file?
How do we minimize the risk of this being stale?

tools/build_release.py
```diff
@@ -50,6 +50,8 @@ def main():
         * Upload to PyPI - you will need to type in your credentials
     * Merge the pull request to origin/master
     * Create a release on GitHub using the portion from the changelog for this release for the description
+        * Add example zip files as artifacts
+        * Update LAST_RELEASE with the version of this release (1.1.5, 1.1.6, 1.2.0)
```
marcoskirsch: To make this clearer to newcomers:

* Add the ZIP files under `generated/examples` for each module as a release artifact.
* Update contents of LATEST_RELEASE with the version of the release being created.

This said, I see we have a VERSION file in the root as well. Do you think it would make sense to have both "version" and "latest released version" exist as metadata inside the `src/` folder?

LAST_RELEASE
texasaggie97: Added to release instructions. Eventually, once we automate creating a release on GitHub, it can be part of that script.

tools/build_release.py
texasaggie97: VERSION is generated so it should not go in src. Eventually, LATEST_RELEASE should also be generated, so it also should not go in src.

build/templates/examples.rst.mako
texasaggie97: Changed.

build/rules.mak
texasaggie97: Forgot to make a new PR. This one will be closed and a new one created:

texasaggie97:
#1257 supersedes this PR

tools/build_release.py
marcoskirsch: Then shouldn't VERSION go inside generated/ ?
How do we make LATEST_RELEASE be generated?

These are things worth describing in an issue.

texasaggie97: It is standard for repos to have a VERSION file in the root. Ours just happens to be generated based on the VERSION files of each of the drivers, which are also generated based on the version in their metadata.

When "we" create a script to create a GitHub release, then one of the things that script can (should) do is update LATEST_RELEASE.

PR #1263, opened 2020-02-11
Update code generator to support passing in and returning lists of timedelta instances

### LINKED ISSUE #1255: Add codegen support for using list of timedate.timedelta objects as method parameters or return values
### Description of issue

In nidigital API, there's a [method](https://github.com/ni/nimi-python/blob/31e2064908aa141610220720259a3a274008f1c8/generated/nidigital/nidigital/session.py#L642) that accepts a list of `floats` (representing time) as a parameter. We need to improve its usability by allowing users to pass either list of `floats` or list of `timedate.timedelta` objects.

Code generator needs to updated to add support for the above, specifically we need to add a new converter to convert list of passed-in `timedate.timedelta` objects to list of `floats` that we can then pass down to C DLL.

Similarly, we also need to add codegen support for methods that want to return list of time-based values, to allow them to return list of `timedate.timedelta` objects.

sbethur:

- Add code generator support for methods that accept list of values representing time
  Methods that accept list of `datetime.timedelta` instances
  - Users can also pass list of numbers representing time
  - Users can also pass array of numbers representing time
 - Add code generator support for methods that return list of `datetime.timedelta` instances

* Fix #1255 

Added unit tests.


generated/nifake/nifake/_library.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, ctypes_library):
         self._library = ctypes_library
         # We cache the cfunc object from the ctypes.CDLL object
         self.niFake_Abort_cfunc = None
+        self.niFake_AcceptListOfTimeValues_cfunc = None
```
marcoskirsch: Are these in internal NI metadata? Please confirm.

generated/nifake/nifake/session.py
```diff
@@ -782,6 +782,28 @@ def abort(self):
...
         return
 
+    @ivi_synchronized
+    def accept_list_of_time_values(self, timestamps, delays):
+        r'''accept_list_of_time_values
+
+        Accepts lists of values representing time.
+
+        Args:
+            timestamps (list of float in seconds or datetime.timedelta): A collection of timestamp values.
```
marcoskirsch: I don't know if timestamps should be timedeltas rather than aboslute times?
Do you really need both parameters? Seems like unecessary complexity.

generated/nifake/nifake/session.py
```diff
@@ -1295,7 +1317,7 @@ def import_attribute_configuration_buffer(self, configuration):
         '''
         vi_ctype = _visatype.ViSession(self._vi)  # case S110
         size_in_bytes_ctype = _visatype.ViInt32(0 if configuration is None else len(configuration))  # case S160
-        configuration_ctype = get_ctypes_pointer_for_buffer(value=_converters.convert_import_buffer_to_array(configuration), library_type=_visatype.ViInt8)  # case B520
+        configuration_ctype = get_ctypes_pointer_for_buffer(value=_converters.convert_import_buffer_to_array(configuration, _visatype.ViInt8), library_type=_visatype.ViInt8)  # case B520
```
marcoskirsch: Getting rid of `library_type` seems like a step back. Is this because of that `unused` parameter?
Maybe we should rethink this a bit.

generated/nifake/nifake/unit_tests/test_session.py
```diff
@@ -1405,6 +1405,49 @@ def test_nitclk_integration(self):
         with nifake.Session('dev1') as session:
             assert str(type(session.tclk)) == "<class 'nitclk.session.SessionReference'>"
 
+    def test_accept_time_values(self):
+        self.patched_library.niFake_AcceptListOfTimeValues.side_effect = self.side_effects_helper.niFake_AcceptListOfTimeValues
+        # Pass one argument as list and one as array.array
+        timestamps = [-1.5, 2.0]
```
marcoskirsch: these aren't time values, they are floats.
Is this testing the conversion?

generated/nifake/nifake/_library.py
sbethur: They are in a changelist for now. Once I get this merged, I'll send a review for that.

generated/nifake/nifake/session.py
sbethur: Yes, bad naming. I'll use a different variable name.

generated/nifake/nifake/session.py
sbethur: Yes, the `unused` parameter results in `library_type` to be passed into the converter.
The other options I considered:
1. Not passing in `library_type` to converters. This would require us to implement N versions (one for each supported number type) of `convert_timedeltas_to_seconds`, `convert_timedeltas_to_milliseconds`, and `convert_timedeltas_to_microseconds` each.
2. Add a new case in code generator for something like `python_api_two_parameter_converter_name`, which goes against the desire to keep code generator as simple as possible.

Another thing to keep in mind is the case of using the converter for a single `timedelta`, where the converter takes two parameters - `time` and `library_type`. It would be nice to be consistent with it.

marcoskirsch: Can we keep the `library_type` parameter, and instead of `unused` assert that it's what we expect? Or does that complicate things too much?

generated/nifake/nifake/unit_tests/test_session.py
sbethur: By "time values" I mean number types representing time, as opposed to `timedelta` instance representing time.
I can rename the test method as `test_accept_time_as_floats`. 

This is testing the case where the converter does not have to convert anything, but just pass through the list of number types it gets.

generated/nifake/nifake/session.py
sbethur: Added an assert.

generated/nifake/nifake/unit_tests/test_session.py
sbethur: Updated the test names

src/nifake/unit_tests/test_session.py
```diff
@@ -1405,6 +1405,56 @@ def test_nitclk_integration(self):
...
+
+    def test_return_timedeltas(self):
+        self.patched_library.niFake_ReturnListOfTimedeltas.side_effect = self.side_effects_helper.niFake_ReturnListOfTimedeltas
+        time_values = [-1.5, 2.0]
+        time_values_ctype = (nifake._visatype.ViReal64 * len(time_values))(*time_values)
+        expected_timedeltas = [datetime.timedelta(seconds=i) for i in time_values]
+        self.side_effects_helper['ReturnListOfTimedeltas']['timedeltas'] = time_values_ctype
+        with nifake.Session('dev1') as session:
+            returned_timedeltas = session.return_list_of_timedeltas(len(expected_timedeltas))
+            assert len(returned_timedeltas) == len(expected_timedeltas)
```
texasaggie97: You don't check that the values returned are correct, only that the length of the array is correct.

sbethur: done

build/templates/_converters.py.mako
```diff
@@ -237,7 +275,8 @@ def convert_to_nitclk_session_number(item):
     raise TypeError('Unsupported type for nitclk session: {}'.format(type(item)))
 
 
-def convert_to_nitclk_session_number_list(item_list):
+# The unused parameter is required for case B520 in codegen, which expects a two-parameter converter.
```
marcoskirsch: we've renumbered the cases serveral times, I feel like this will accidentally get stale at some point. I'm not sure if we're better off removing it though because it certainly is useful.

Do we have converters that only take one parameter? Maybe we should standardize the interface and always take two parameters. If you agree then we should do that prep-work in its own PR.

build/templates/_converters.py.mako
```diff
@@ -237,7 +275,8 @@ def convert_to_nitclk_session_number(item):
...
+def convert_to_nitclk_session_number_list(item_list, unused):
```
marcoskirsch: I'd rather you call `unused` -> `library_type` so caller can state the name explicitly.
Optionally you could assert the library_type if/when it makes sense to.

build/templates/_converters.py.mako
sbethur: Discussed with Marcos offline. We decided that we can standardize on all converters to have one parameter - created a PR - #1302.

sbethur: Don't have `unused` parameter anymore :)

build/templates/_converters.py.mako
sbethur: Don't have `unused` parameter anymore :)

PR #1265, opened 2020-02-12
Explain why niscope test code uses persistent simulation

marcoskirsch:
It was not obvious to me, it will be much less obvious to someone who hasn't been involved in the project from day one.

### DIFF
--- a/src/niscope/system_tests/test_system_niscope.py
+++ b/src/niscope/system_tests/test_system_niscope.py
@@ -8,7 +8,15 @@ import sys
 import tempfile
 
 
-# We need a lock file so multiple tests aren't hitting the simulated HW at the same time
+# There are system tests below that need either a PXI-5124 or a PXI-5142 instead of the PXIe-5164 we use everywhere else
+# because of specific capabilities on those models. Due to internal NI bug 969274, opening a simulated session to those models
+# sometimes fails. As a workaround, the nimi-bot VMs are configured with one simulated instrument of each kind respectively
+# named "5124" and "5142". If you want to run these tests on your own system, you will need to create these two simulated
+# instruments.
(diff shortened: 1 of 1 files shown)


src/niscope/system_tests/test_system_niscope.py
```diff
@@ -8,7 +8,15 @@
 import tempfile
 
 
-# We need a lock file so multiple tests aren't hitting the simulated HW at the same time
+# There are system tests below that need either a PXI-5124 or a PXI-5142 instead of the PXIe-5164 we use everywhere else
+# because of specific capabilities on those models. Due to internal NI bug 969274, opening a simulated session to those models
+# sometimes fails. As a workaround, the nimi-bot VMs are configured with one simulated instrument of each kind respectively
```
sbethur: I think we should qualify usages of "simulated" with the kind of simulation i.e. session-based vs. persistent.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -8,7 +8,15 @@
...
+# named "5124" and "5142". If you want to run these tests on your own system, you will need to create these two simulated
+# instruments.
+# In addition, we need a global lock in order to keep us from opening more than one session to the same simulated instrument
```
sbethur: Here again we should explicitly call out persistent simulation.

src/niscope/system_tests/test_system_niscope.py
marcoskirsch: Fixed in #1286

src/niscope/system_tests/test_system_niscope.py
marcoskirsch: Fixed in #1286

PR #1269, opened 2020-02-12
Use per driver tox for testsuites

### LINKED ISSUE #1266: There is more than one tox INI supporting running of system tests and one is incorrect
We use `tox` to run system tests.
If a user wants to run system tests, they can either:

(1)
```
python3 -m tox -e niswitch_system_tests
```
 which uses the environments from the top level `tox.ini` file, or

(2)
```
pushd generated\niswitch & python3 -m tox -c tox-system_tests.ini & popd 
```
which uses the code-generated `tox-system_tests.ini` file.

nimi-bot uses (2) when running system tests.
The reason why support wasn't removed from the top level `tox.ini` file is that Automated Test Suites internal to NI rely on the top-level INI file.

Problems with this:
* Top level `tox.ini` is incorrect. For example, it is missing the `fastening` dependency on the modules that need it.
* We aren't exercising those environments in nimi-python CI, which allows for bugs like the above one to go unnoticed.
* Two ways of doing the same thing is confusing to humans.
   * Why two?
   * How are they different?
   * Which one should I be using?

I think we need to get rid of method (1). This will simplify that `tox.ini` and eliminate the redundancy. But this will require internal NI Automated Test Suites to make a change.

texasaggie97:
* Removes per driver system tests from `tox.ini` and `tox-travis.ini`
* `run_python_system_tests.py` modified to run test from the per driver tox.ini file

* Fix #1266 

* System


build/templates/tox-system_tests.ini.mako
```diff
@@ -57,19 +61,36 @@ commands =
     # token is from codecov
     ${module_name}-coverage: codecov -X gcov --token=4c58f03d-b74c-489a-889a-ab0a77b7809f --no-color --flags ${module_name}systemtests --name ${module_name} --root ../.. --file ../../generated/${module_name}/coverage.xml
 
+    ${module_name}-test_suite: python --version
```
marcoskirsch: I think `print_diagnostic_information` includes version. It may also include platform.architecture.
If it doesn't, then we should consider making it so it does so se can simplify these files and also have more control over formatting.

build/templates/tox-system_tests.ini.mako
```diff
@@ -57,19 +61,36 @@ commands =
...
+    ${module_name}-test_suite: python -c "import platform; print(platform.architecture())"
+    ${module_name}-test_suite: python -c "import ${module_name}; ${module_name}.print_diagnostic_information()"
+    ${module_name}-test_suite: python -m pytest ../../src/${module_name}/system_tests --json=generated/kibana/${module_name}_system_test_result.json {posargs}
```
marcoskirsch: In #1268 I added `--durations`.
This may need integration whichever goes in second.

build/templates/tox-system_tests.ini.mako
texasaggie97: Removed printing the version and architecture separately.

build/templates/tox-system_tests.ini.mako
texasaggie97: Noted

generated/nidcpower/tox-system_tests.ini
```diff
@@ -41,6 +40,7 @@ deps =
     nidcpower-system_tests: numpy
     nidcpower-system_tests: scipy
     nidcpower-system_tests: fasteners
+
```
marcoskirsch: is this blank like intentional?

texasaggie97:
The test suites do not pull from master. They pull the zip artifact from the last release and unzip that, and use the system_tests that were zipped up as part of that.

Still, changed to use the code from that artifact for testing.

generated/nidcpower/tox-system_tests.ini
texasaggie97: Yes, separating the dependencies for the different environments.

tools/run_python_system_tests.py
```diff
@@ -55,7 +54,7 @@
 
 
 print('****Running system tests in tox.****')
-tox_dir = os.path.join(zip_folder, os.listdir(zip_folder)[0])
+tox_dir = os.path.join(zip_folder, os.listdir(zip_folder)[0], 'generated', args.driver)
 os.chdir(tox_dir)
-result = os.system('tox ' + args.python_bitness + ' -e ' + args.python_version + '-' + args.driver + '_system_tests')
+result = os.system('tox {0} -e {1}-{2}-system_tests -c tox-system_tests.ini'.format(args.python_bitness, args.python_version, args.driver))
```
marcoskirsch: if we use `python -m pip` above, I think we should be  consistent and use `python -m tox` here. Unless I'm missing something.

texasaggie97: Changed

tox.ini
```diff
@@ -108,30 +92,6 @@ commands =
...
-    niswitch_system_tests: python -m pytest src/niswitch/system_tests --json=generated/kibana/niswitch_system_test_result.json --durations=5
-    niscope_system_tests: python --version
-    niscope_system_tests: python -c "import platform; print(platform.architecture())"
-    niscope_system_tests: python -m pytest src/niscope/system_tests --json=generated/kibana/niscope_system_test_result.json --durations=5
-    nimodinst_system_tests: python --version
-    nimodinst_system_tests: python -c "import platform; print(platform.architecture())"
-    nimodinst_system_tests: python -m pytest src/nimodinst/system_tests --json=generated/kibana/nimodinst_system_test_result.json --durations=5
-    nise_system_tests: python --version
-    nise_system_tests: python -c "import platform; print(platform.architecture())"
-    nise_system_tests: python -m pytest src/nise/system_tests --json=generated/kibana/nise_system_test_result.json --durations=5
```
adityabhat7: Is running system tests completely removed or is it handled in any other file?

I tried running the test by downloading these files and I didn't find any json report.

texasaggie97: If you use `run_python_system_tests.py` it knows how to run the system tests in their new location. If you are running them using something else, you will need to make changes similar to what was changed in `run_python_system_tests.py`.

marcoskirsch: If`run_python_system_tests.py` isn't being used by the internal NI test suites, then it's dead code and we should consider deleting it.

Need input from @adityabhat7 and @ni-jfitzger

ni-jfitzger: tox is not a tool I am very familiar with and I do not really understand the affects of these changes.
However, using the zip from last night's niscope test run, I ran the code that executes our system tests and then ran them, changed the name of or deleted the results file, deleted run_python_system_tests.py and executed again, then compared the results.
I did this with both python 2.7 and python 3.6.
In all instances, a test results file was produced.  In python 3, it was very clear that the only differences in results was duration of each test.  In python 2, results are much harder to compare but the results are there.

**I think it is safe to say that run_python_system_tests.py is dead code.**

adityabhat7: It is very tricky to use 'run_python_system_tests.py'. Since this file is available at github and we need the same file to download the code. We need to write a script for downloading only run_python_system_tests.py from master and running it.

I didn't find any json report being created. Are we removing it?

adityabhat7:
Default value for python version is set to py27 at '' (https://github.com/ni/nimi-python/blob/f692f244f7de13ff2af7d652861e3232589577e0/tools/run_python_system_tests.py#L13).
I think it should be change to py35. Since we are removing python 27 support.

tox.ini
texasaggie97: I added json output.

There are a few ways to handle this:
1. You can get the latest version of the file by downloading [run_python_system_tests.py](https://raw.githubusercontent.com/ni/nimi-python/master/tools/run_python_system_tests.py). There are command line utilities wget and curl, or you can write a very simple script in the language of your choice. Curl is part of Windows. At least Windows 10. This is very simple to do. `curl -o test.py https://raw.githubusercontent.com/ni/nimi-python/master/tools/run_python_system_tests.py`
1. You can keep the logic to download and unzip the latest release artifact in your test suite, and then invoke a script in the tools directory to actually run the system tests. This is similar to what we do for running system tests in the CI. `tools/system_tests.bat` is used by the Jenkins jobs to actually invoke the test. This way these jobs don't need to be updated every time the system tests change.
1. A combination of the first two. Move the logic that downloads and unzips the release artifact into one script. This won't change unless GitHub's rest API changes. Have a second script in tools to invoke the system tests that you run from the unzipped artifact. 

I think the last one is the best from a maintenance standpoint. Also, the script you invoke to run the system tests will match the system tests you are running. I.e. the ones in the release artifact.

texasaggie97:
@adityabhat7, Changed default python version to py38. If you add a comment on the line of code, I can reply in line there.

tox.ini
texasaggie97: I went ahead and implemented option 3. There is now a `prepare_python_system_tests.py` that will download the latest release artifact, unzip it, then execute `run_python_system_tests.py` from the downloaded artifact.

Here is a simple batch file that would use it:
``` batch
mkdir %temp%\nimi-python
chdir %temp%\nimi-python
curl -o prepare_python_system_tests.py https://raw.githubusercontent.com/ni/nimi-python/master/tools/prepare_python_system_tests.py
python prepare_python_system_tests.py --driver niscope --python-version py36
```

adityabhat7:
I ran the tests by downloading the repo and using prepare_python_system_tests.py. It worked.
We will modify the test to use prepare_python_system_tests.py.

tools/prepare_python_system_tests.py
```diff
@@ -0,0 +1,72 @@
+import argparse
+import json
+import os
+import subprocess
+import tempfile
+import time
+import urllib.request
+import zipfile
+
+parser = argparse.ArgumentParser(description='Downloads the latest release nimi-python and runs system tests on the specified driver.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
marcoskirsch: "release of" or "released".

tools/prepare_python_system_tests.py
```diff
@@ -0,0 +1,72 @@
...
+parser.add_argument('-d', '--driver', required=True, type=str, help='Driver Name.')
```
marcoskirsch: I feel weird calling this driver name. It's really the Python module "nidmm" not the driver name "NI-DMM".

The description above also calls it "driver". Consider: 
* "on the specified Python module".
* --driver -> --module with help "Python module name e.g. `nidmm`, `niscope`, etc.".

tools/prepare_python_system_tests.py
```diff
@@ -0,0 +1,72 @@
...
+parser.add_argument('-pv', '--python-version', required=False, type=str, help='Python version to be run.', default=None)
+parser.add_argument('-pb', '--python-bitness', required=False, type=str, help='Python bitness to be run.', default=None)
```
marcoskirsch: Have examples so syntax is obvious. Is bitness "32" or "32-bit"?  Is version "3.8" or "38" ?

tools/prepare_python_system_tests.py
```diff
@@ -0,0 +1,72 @@
...
+
+print('****Parsing GitHub releases to obtain the latest zip file URL.****')
+release_url = 'https://api.github.com/repos/ni/nimi-python/releases'
+with urllib.request.urlopen(release_url) as response:
+    html = response.read()
+url_data = json.loads(html.decode('utf-8'))
+zip_url = url_data[0]['zipball_url']
+print(zip_url)
+
+print('****Downloading latest zip file from github.****')
```
marcoskirsch: GitHub

tools/prepare_python_system_tests.py
texasaggie97: Changed to `'Downloads the latest release atrifacts from nimi-python and runs system tests on the specified python package.'`

tools/prepare_python_system_tests.py
texasaggie97: Changed the help string to `'Python package name.'`

tools/prepare_python_system_tests.py
texasaggie97: Changed the output to look like this:
```
usage: prepare_python_system_tests.py [-h] -d {nidcpower,nidigital,nidmm,nifake,nifgen,nimodinst,niscope,nise,niswitch,nisync,nitclk}
                                      [-pv {py35,py36,py37,py38}]
                                      [-pb {None,--32}]

Downloads the latest release artifacts from nimi-python and runs system tests
on the specified python package.

optional arguments:
  -h, --help            show this help message and exit
  -d {nidcpower,nidigital,nidmm,nifake,nifgen,nimodinst,niscope,nise,niswitch,nisync,nitclk}, --driver {nidcpower,nidigital,nidmm,nifake,nifgen,nimodinst,niscope,nise,niswitch,nisync,nitclk}
                        Python package name. (default: None)
  -pv {py35,py36,py37,py38}, --python-version {py35,py36,py37,py38}
                        Python version to be run. This is used to invoke the
                        appropriate tox environment. (default: py38)
  -pb {None,--32}, --python-bitness {None,--32}
                        Python bitness to be run. This is passed to tox.
                        (default: None)
```

tools/prepare_python_system_tests.py
texasaggie97: Fixed

tools/run_python_system_tests.py
```diff
@@ -1,61 +1,25 @@
...
-import os
 import subprocess
-import sys
-import tempfile
-import time
-import urllib.request
-import zipfile
 
 parser = argparse.ArgumentParser(description='Downloads the latest release nimi-python and runs system tests on the specified driver.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 parser.add_argument('-d', '--driver', required=True, type=str, help='Driver Name.')
```
texasaggie97: Changed here to match other script.

tools/prepare_python_system_tests.py
```diff
@@ -0,0 +1,78 @@
...
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument('-d', '--driver', required=True, type=str,
+                    help='Python package name.',
+                    choices=os.listdir('src'))
+parser.add_argument('-pv', '--python-version', required=False, type=str,
+                    help='Python version to be run. This is used to invoke the appropriate tox environment.',
+                    choices=['py35', 'py36', 'py37', 'py38', ], default='py38')
+parser.add_argument('-pb', '--python-bitness', required=False, type=str,
+                    help='Python bitness to be run. This is passed to tox.',
+                    choices=[None, '--32'], default=None)
```
marcoskirsch: Seems weird that you cannot choose 64 bits. And it isn't clear what `None` does.
Consider making choices ["32", "64", "any"]. It will require the choices to do a lookup so you can pull out the string to pass `tox`.

texasaggie97: Updated.

PR #1270, opened 2020-02-12
Skip numpy tests when running under PyPy

texasaggie97:
* Do not install `numpy` or `scipy` when running PyPy
    * There is not a precompiled wheel for PyPy3
    * Trying to compile numpy every time (or even just the times it would need to be installed or upgraded) would add a significant amount of time to the system tests
    * Vistual C++ is not installed on the test VMs
* Skip any test that uses `numpy` when running under PyPy
* Remove any `numpy` array from test data when running under PyPy
* System tests


marcoskirsch:
This is no longer needed because of #1271

Issue #1273, opened 2020-02-13
In nidigital, History RAM system tests cannot be run on simulated session


sbethur:
### Description of issue

nidigital python API contains fancy HRAM method. Digital driver does not simulate HRAM functions in the C API "well enough", to allow testing of the fancy HRAM method. During development of fancy HRAM method, the system tests were run on real h/w. But these tests are disabled on the nimi-bot runs.

We need to re-enable the tests by doing one of the following:

1. Mock the C entry points
2. Improve the simulation in C API
3. Use real h/w to run tests in CI
4. Something else

marcoskirsch:
The fact that simulation is deficient should be considered a driver bug. Please report that in the internal bug tracking system and post the number here.

I wouldn't mock the C API (in other words the _Library class). Through `nifake` we validate that the logic from the generator is correct. If you want to test the "fancy" handcoded function then mocking the private `nidigital.Session` methods it calls into is the right thing to do.

sbethur:
Created a bug in internal bug tracking system, number is 992370.

marcoskirsch:
This bug is about running the tests in simulation. It was closed by covering this code via unit testing. I think that's the workaround but this bug should remain open until driver bug 992370 is fixed and then these unit tests replaced with regular system tests.

Issue #1066, opened 2019-10-30
nidigital API methods and properties should accept enums when possible


sbethur:
* Description of issue

API methods and properties should accept enum parameters when possible, instead of int parameters.

e.g. `configure_time_set_drive_format` should be updated to accept DriveEdgeSetFormat enum.

PR #1274, opened 2020-02-13
Add fancy History RAM API method to allow fetching multiple History RAM samples in a single API call

### LINKED ISSUE #1071: Improve usability of nidigital History RAM API
* Description of issue
Currently, History RAM (HRAM) configuration can be done via a set of properties or using a set of `configure_history_ram_xyz` methods. Fetching of HRAM samples is done by calling `get_history_ram_sample_count` to get the number of samples and then calling `fetch_history_ram_cycle_xyz` in a loop.

Usability can be improved by getting rid of `configure_history_ram_xyz` methods and providing a multi-sample fetch API.

sbethur:

Provide multi-sample History RAM fetch API

TODO: List of issues.

* Fix #1071 

Added new system tests and ran them on real h/w (See #1273 for more info)


src/nidigital/custom_types/history_ram_cycle_information.py
```diff
@@ -0,0 +1,45 @@
...
+            'vector_number={}'.format(self.vector_number),
+            'cycle_number={}'.format(self.cycle_number),
+            'scan_cycle_number={}'.format(self.scan_cycle_number),
+            'expected_pin_states={}'.format(self.expected_pin_states),
+            'actual_pin_states={}'.format(self.actual_pin_states),
+            'per_pin_pass_fail={}'.format(self.per_pin_pass_fail),
+        ]
+
+        return '{0}({1})'.format(self.__class__.__name__, ', '.join(parameter_list))
+        return self.__class__.__name__
```
texasaggie97: The second return is redundant, I think.

src/nidigital/metadata/functions_addon.py
```diff
@@ -142,3 +142,71 @@
...
+                'documentation': {
+                    'description': 'Site on which to retrieve History RAM data. Specify site as a string in the form of siteN,\nwhere N is the site number. The VI returns an error if more than one site is specified.'
+                },
+                'name': 'site',
+                'type': 'ViConstString'
+            },
+            {
+                'direction': 'in',
+                'documentation': {
+                    'description': 'Pins for which to retrieve History RAM data. If empty, the pin list from the pattern\ncontaining the start label is used. Call get_pattern_pin_list or get_pattern_pin_names with the start\nlabel to retrieve the pins associated with the pattern burst.'
```
texasaggie97: Reference methods as `niDigital_BlahBlah` so they will be turned into a link in help.

src/nidigital/metadata/functions_addon.py
```diff
@@ -142,3 +142,71 @@
...
+        'codegen_method': 'python-only',
+        'method_templates': [
+            {
+                'documentation_filename': 'default_method',
+                'method_python_name_suffix': '',
+                'session_filename': 'fancy_fetch_history_ram_cycle_information',
+            }
+        ],
+        'documentation': {
+            'description': '\nReturns the pattern information acquired for the specified cycles.\n\nIf the pattern is using the edge multiplier feature, cycle numbers represent tester cycles, each of which may\nconsist of multiple DUT cycles. When using pins with mixed edge multipliers, pins may return\nNIDIGITAL_VAL_PIN_STATE_NOT_ACQUIRED for DUT cycles where those pins do not have edges defined.\n'
```
texasaggie97: Does `PIN_STATE_NOT_ACQUIRED` exist as an enum value somewhere? It looks like this isn't correctly being turned into a link, and that would be one reason it wouldn't be.

src/nidigital/metadata/functions_addon.py
```diff
@@ -142,3 +142,71 @@
...
... set for the acquired cycle.\n-  **vector_number** (int) Vector number within the pattern for the acquired cycle. Vector numbers start\nat 0 from the beginning of the pattern.\n-  **cycle_number** (int) Cycle number acquired by this History RAM sample. Cycle numbers start at 0\nfrom the beginning of the pattern burst.\n-  **scan_cycle_number** (int) Scan cycle number acquired by this History RAM sample. Scan cycle numbers\nstart at 0 from the first cycle of the scan vector. Scan cycle numbers are -1 for cycles that do not\nhave a scan opcode.\n-  **expected_pin_states** (list) Pin state as expected by the loaded pattern in the order specified in\nthe pin list. Pins without defined edges in the specified DUT cycle will have a value of\nNIDIGITAL_VAL_PIN_STATE_NOT_ACQUIRED.\n-  **actual_pin_states** (list) Pin state acquired by History RAM in the order specified in the pin\nlist. Pins without defined edges in the specified DUT cycle will have a value of\nNIDIGITAL_VAL_PIN_STATE_NOT_ACQUIRED.\n-  **per_pin_pass_fail** (list) pass fail information for pins in the order specified in the pin list.\nPins without defined edges in the specified DUT cycle will have a value of pass (True).\n'
```
texasaggie97: For very long descriptions, it may be easier to format it using the triple quote:
```python
'description': """
This is a very long description
that has multiple
lines breaks as well as
* a list
* because
* why not?
"""
```

src/nidigital/custom_types/history_ram_cycle_information.py
sbethur: Yeah, that was a placeholder until I implemented the method body. Forgot to delete it later.

src/nidigital/metadata/functions_addon.py
sbethur: Done

src/nidigital/metadata/functions_addon.py
sbethur: I'm working on another PR where I update to using enums everywhere. I initially thought of including this in that PR, but I don't really have to wait until that PR. I fixed it here.

src/nidigital/metadata/functions_addon.py
sbethur: This is much better. Thanks for the tip!

generated/nidigital/nidigital/history_ram_cycle_information.py
```diff
@@ -0,0 +1,46 @@
...
+        row_format_d = '{:<20}: {:,}\n'
+        row_format_s = '{:<20}: {:}\n'
+
+        string_representation = ''
+        string_representation += row_format_s.format('Pattern Name', self.pattern_name)
+        string_representation += row_format_s.format('Time Set Name', self.time_set_name)
+        string_representation += row_format_d.format('Vector Number', self.vector_number)
+        string_representation += row_format_d.format('Cycle Number', self.cycle_number)
+        string_representation += row_format_d.format('Scan Cycle Number', self.scan_cycle_number)
+        string_representation += row_format_s.format('Expected Pin States', self.expected_pin_states)
```
texasaggie97: You format pin states differently in `__repr__`, should you here as well?

generated/nidigital/nidigital/session.py
```diff
@@ -2265,6 +2267,121 @@ def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, time
...
+
+        if samples_to_read < -1:
+            raise ValueError('samples_to_read should be greater than or equal to -1.')
+
+        samples_available = self.get_history_ram_sample_count(site)
+        if position >= samples_available:
+            raise ValueError('position: Specified value = {0}, Maximum value = {1}.'.format(position, samples_available - 1))
+
+        if samples_to_read == -1:
+            if not self.history_ram_number_of_samples_is_finite:
```
texasaggie97: I may be (probably?) reading this wrong, but shouldn't the exception be when number of samples _is_ infinite? Not when it isn't?

generated/nidigital/nidigital/session.py
```diff
@@ -2265,6 +2267,121 @@ def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, time
...
+                raise RuntimeError(
+                    'Specifying -1 to fetch all History RAM samples is not supported when the digital pattern instrument is '
+                    'configured for continuous History RAM acquisition. You must specify an exact number of samples to fetch.')
+            samples_to_read = samples_available - position
+
+        if position + samples_to_read > samples_available:
+            raise ValueError(
+                'position: Specified value = {0}, samples_to_read: Specified value = {1}; Samples available = {2}.'
```
texasaggie97: You don't actually say why this is an exception. Maybe add something about requesting beyond the samples available? Not sure what the wording should be though.

src/nidigital/metadata/enums_addon.py
```diff
@@ -2,5 +2,7 @@
 # Any changes to the API should be made here. enums.py is code generated
 
 enums_override_metadata = {
+    'DigitalState': {
+        'codegen_method': 'public',
```
texasaggie97: Add a comment why you need to set this.

generated/nidigital/nidigital/history_ram_cycle_information.py
sbethur: I was debating between verbose info for pin states in `__str__` i.e. `<DigitalState.D: 8>` vs. just `DigitalState.D`. Thinking more about it, I don't see any value in the verbose way. I changed it to match `__repr__`.

generated/nidigital/nidigital/session.py
sbethur: The name of `history_ram_number_of_samples_is_finite` is confusing. If it is set to `false`, HRAM streaming mode is enabled. In this mode, users are not allowed to pass -1 to `samples_to_read` (which is supposed to mean "all available samples").
The code is correct.

generated/nidigital/nidigital/session.py
sbethur: The error description includes the number of samples user requested and number of samples available. This plus the fact that type of exception is `ValueError` should be enough information to let the user know what's wrong. I'm trying not to deviate too much from other nidigital API when it comes to error description.

Looking closely at this error message revealed a bug in the value calculated within the error message :)

src/nidigital/metadata/enums_addon.py
sbethur: done

generated/nidigital/nidigital/session.py
texasaggie97: But you have to already know the relationship between the values to know that `position + samples_to_read > samples_available` is the error condition. If you get a ValueError with the description "position: Specified value = 5, samples_to_read: Specified value = 100; Samples available = 40." there is nothing to say _why_ that is bad.

src/nidigital/templates/session.py/fancy_fetch_history_ram_cycle_information.py.mako
```diff
@@ -0,0 +1,73 @@
...
+            if pattern_index not in pattern_names:
+                pattern_names[pattern_index] = self.get_pattern_name(pattern_index)
+            pattern_name = pattern_names[pattern_index]
+
+            if time_set_index not in time_set_names:
+                time_set_names[time_set_index] = self.get_time_set_name(time_set_index)
+            time_set_name = time_set_names[time_set_index]
+
+            vector_number = vector_number
+            cycle_number = cycle_number
```
texasaggie97: Why are you setting these equal to themselves?

sbethur: No reason :). Forgot to delete this after renaming and restructuring some code. Cleaned them up.

docs/nidigital/class.rst
```diff
@@ -1188,72 +1188,13 @@ fetch_history_ram_cycle_information
 
     .. py:currentmodule:: nidigital.Session
 
-    .. py:method:: fetch_history_ram_cycle_information(site, sample_index)
+    .. py:method:: fetch_history_ram_cycle_information(site, pin_list, position, samples_to_read)
```
marcoskirsch: with sites becoming a repeated capability ( #1290 ) we will need to integrate this carefully.

docs/nidigital/class.rst
```diff
@@ -1262,87 +1203,74 @@ fetch_history_ram_cycle_pin_data
...
-
-
-                per_pin_pass_fail (list of bool): 
-
-
-                    
-
-
-
-fetch_history_ram_scan_cycle_number
```
marcoskirsch: a bunch of methods went away, consider mentioning that in CHANGELOG.md. It can still be a single "changed" item.

docs/nidigital/class.rst
```diff
@@ -1262,87 +1203,74 @@ fetch_history_ram_cycle_pin_data
...
+                Number of samples to fetch. A value of -1 specifies to fetch all available samples.
 
                 
 
 
-            :type sample_index: int
+            :type samples_to_read: int
 
-            :rtype: int
+            :rtype: list of HistoryRAMCycleInformation
```
marcoskirsch: Is it worth creating a type for this?
Why not return a list of dictionaries?

(consistency with `niscope`?)

generated/nidigital/nidigital/_library.py
```diff
@@ -6,6 +6,8 @@
 
 from nidigital._visatype import *  # noqa: F403,H303
 
+import nidigital.history_ram_cycle_information as history_ram_cycle_information  # noqa: F401
+
```
marcoskirsch: why is this needed by `_library.py`?

generated/nidigital/nidigital/session.py
```diff
@@ -2267,6 +2269,119 @@ def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, time
...
+                   Length of the inner list will be equal to the number of pins requested.
+
+        '''
+        if position < 0:
+            raise ValueError('position should be greater than or equal to 0.')
+
+        if samples_to_read < -1:
+            raise ValueError('samples_to_read should be greater than or equal to -1.')
+
+        samples_available = self.get_history_ram_sample_count(site)
```
marcoskirsch: does `self.get_history_ram_sample_count` need to be public?

generated/nidigital/nidigital/session.py
```diff
@@ -2267,6 +2269,119 @@ def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, time
...
+        if position >= samples_available:
+            raise ValueError('position: Specified value = {0}, Maximum value = {1}.'.format(position, samples_available - 1))
```
marcoskirsch: There's a bunch of parameter value checks here. What's the harm in letting the runtime handle it instead?

src/nidigital/metadata/enums_addon.py
```diff
@@ -2,5 +2,10 @@
 # Any changes to the API should be made here. enums.py is code generated
 
 enums_override_metadata = {
+    # Codegen deduces the enum types to generate based on usage of enum types in public metadata.
```
marcoskirsch: Do we intend for these things to remain in `_addon.py` files in the long term?

src/nidigital/nidigital.mak
```diff
@@ -8,5 +8,9 @@ MODULE_FILES_TO_COPY := $(DEFAULT_PY_FILES_TO_COPY)
 
 RST_FILES_TO_GENERATE := $(DEFAULT_RST_FILES_TO_GENERATE)
 
+# Add custom types to copy
```
marcoskirsch: Comment is not very useful. More useful would be to say why this needs to be copied and/or where to.
I assume you copied this from `niscope` so whatever you do do it in both locations - remove / leave alone / reword

docs/nidigital/class.rst
marcoskirsch: Ok, there's the `__repr__` stuff too. Resolving.

docs/nidigital/class.rst
sbethur: Yes, whichever PR goes last will merge the other's changes.

docs/nidigital/class.rst
sbethur: Had already update the changelog; added another item to better convey the changes.

generated/nidigital/nidigital/session.py
sbethur: Yes, there are multiple use-cases where users may want to know the total count of HRAM samples. e.g. In HRAM streaming, users just want to know the number of failures in the pattern burst, but not necessarily fetch the HRAM data.

`get_history_ram_sample_count` is also in LV and .NET APIs.

generated/nidigital/nidigital/session.py
sbethur: Runtime doesn't do these checks since it doesn't support multi-sample fetch; it doesn't have any notion of `position` or `samples_to_read`.

These same checks are present in other APIs that support multi-sample fetch (LV and .NET), which is not good :(

src/nidigital/metadata/enums_addon.py
sbethur: No, it will go away when #1066 is fixed.

src/nidigital/nidigital.mak
sbethur: I removed it. None of the other rules in the file have any comments.

generated/nidigital/nidigital/_library.py
sbethur: In this case it's not needed. But in the case of other custom types (like `waveform_info` in niscope) the module also contains a type that's passed to the C function. Unconditionally adding the import keeps the codegen and/or metadata simpler.

PR #1281, opened 2020-02-17
Limit one concurrent job per node

texasaggie97:

* We want to allow concurrent builds, but we need to limit it to one per node
    * This prevents issues with two jobs trying to modify the venv at the same time

* System

### DIFF
--- a/tools/system_test_jobs.groovy
+++ b/tools/system_test_jobs.groovy
@@ -21,6 +21,8 @@ def genJob(driver, platform) {
         label(platform)
 
         // We want to allow multiple builds running at the same time, just not on one node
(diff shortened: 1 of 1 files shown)


texasaggie97:
test this please

tools/system_test_jobs.groovy
```diff
@@ -19,7 +19,12 @@ def genJob(driver, platform) {
         description "Run system tests for ${driver} on ${platform}"
 
         label(platform)
+
+        // We want to allow multiple builds running at the same time, just not on one node
```
marcoskirsch: we should say why

texasaggie97: Done

PR #1303, opened 2020-03-02
Make nifgen send sw trig consistent with current proposal

### LINKED ISSUE #1300: Update send_software_edge_trigger in nifgen to make the behavior explicit
### Description of issue

Currently, `send_software_edge_trigger` in nifgen behaves as:
```
session.send_software_edge_trigger()   # Trigger "start" via software
session.script_trigger[0].send_software_edge_trigger()   # Trigger script trigger 0 via software
```

There are problems with this design. See [here](https://github.com/ni/nimi-python/issues/1220#issue-553136854) for details.

We need to update `send_software_edge_trigger` to make it [behave ](https://github.com/ni/nimi-python/issues/1193#issuecomment-592196375) like what we're planning to do in nidigital.

In addition, we should do two things:
1. Preserve backwards compatibility by using appropriate parameter defaults 
2. Deprecate the old way by "raising a warning", if used that way

texasaggie97:

* Makes `send_software_edge_trigger()` consistent with how we decided nidigital's should look
* Backwards compatible with previous method
    * Sets DeprecationWarning when used this way
* Updates old tests to look for warning
* Add new tests to not

* Fix #1300 

* System


generated/nifgen/nifgen/session.py
```diff
@@ -2173,35 +2173,55 @@ def _lock_session(self):
...
+                +----------------+
+                | Trigger.START  |
+                +----------------+
+                | Trigger.SCRIPT |
+                +----------------+
+
+                Note:
+                One or more of the referenced values are not in the Python API for this driver. Enums that only define values, or represent True/False, have been removed.
+
+            trigger_id (str):
```
sbethur: documentation missing?

generated/nifgen/nifgen/session.py
```diff
@@ -2173,35 +2173,55 @@ def _lock_session(self):
...
 
-            session.script_triggers[1].send_software_edge_trigger()
+                +----------------+
+                | Trigger.START  |
+                +----------------+
+                | Trigger.SCRIPT |
+                +----------------+
+
+                Note:
+                One or more of the referenced values are not in the Python API for this driver. Enums that only define values, or represent True/False, have been removed.
```
sbethur: Why this note?

generated/nifgen/nifgen/session.py
```diff
@@ -2173,35 +2173,55 @@ def _lock_session(self):
...
-        This method does not override external digital edge triggers of the
-        NI 5401/5411/5431.
         '''
-        # We look at whether we are called directly on the session or a repeated capability container to determine how to behave
-        if len(self._repeated_capability) > 0:
-            trigger_id = self._repeated_capability
-            trigger = 103  # enums.Trigger.SCRIPT
+        if trigger is None or trigger_id is None:
+            import warnings
+            warnings.warn('trigger and trigger_id should now always be passed in to the function', category=DeprecationWarning)
```
sbethur: Use `method` instead of `function`

generated/nifgen/nifgen/session.py
```diff
@@ -2173,35 +2173,55 @@ def _lock_session(self):
...
+                trigger_id = "None"
+                trigger = enums.Trigger.START
+
+        elif trigger is not None and trigger_id is not None:
+            pass  # This is how the function should be called
+
         else:
-            trigger_id = "None"
-            trigger = 1004  # enums.Trigger.START
+            raise ValueError('Both trigger ({0}) and trigger_id ({1}) should be passed in to the function'.format(str(trigger), str(trigger_id)))
```
sbethur: I don't see when `else` case will be hit.

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -433,11 +434,38 @@ def test_fir_filter_coefficients():
 '''
 
 
+def test_send_software_edge_trigger_start_old(session):
```
sbethur: I feel the name - `test_send_software_edge_trigger_start_deprecated` would be more descriptive.

generated/nifgen/nifgen/session.py
texasaggie97: Added

generated/nifgen/nifgen/session.py
texasaggie97: Changed

generated/nifgen/nifgen/session.py
texasaggie97: It will be hit if only one parameter is set and the other is left as the default. `session.send_software_edge_trigger(trigger_id='None')`

src/nifgen/system_tests/test_system_nifgen.py
texasaggie97: Changed

generated/nifgen/nifgen/session.py
sbethur: Then the `if` statement:
```
if trigger is None or trigger_id is None:
```
would catch that, right?

generated/nifgen/nifgen/session.py
sbethur: Should we open an issue for it?

src/nifgen/metadata/functions.py
```diff
@@ -2435,23 +2435,27 @@
...
+                    ],
                 },
+                'enum': 'Trigger',
                 'name': 'trigger',
-                'type': 'ViInt32'
+                'type': 'ViInt32',
             },
             {
+                'documentation': {
+                    'description': 'Trigger ID specifies the Script Trigger to use for triggering.',
```
sbethur: I think we should provide more info on what the acceptable values are, like in [C reference](http://zone.ni.com/reference/en-XX/help/370524V-01/fgencref/cvinifgen_sendsoftwareedgetrigger/).

marcoskirsch:
I suppose we should know that when version is bumped to 2.0 we need to go remove all deprecated warnings and codepaths.

src/nifgen/metadata/functions.py
sbethur: Discussed offline and decided to leave it as is to keep it consistent with LV documentation.

PR #1304, opened 2020-03-02
Change converters to have single parameter

### LINKED ISSUE #1302: Standardize the function signature of all converters
### Description of issue

Some converters like:
```
convert_timedelta_to_seconds
convert_timedelta_to_milliseconds
convert_timedelta_to_microseconds
convert_init_with_options_dictionary
convert_repeated_capabilities_from_init
```
have two parameters.

Some converters like:
```
convert_month_to_timedelta
convert_to_nitclk_session_number
convert_to_bytes
```
have one parameter.

Usage of these converters are code generated. If we standardize the number of parameters, the code generators become consistent and adding new converters will be easy.

Based on the currently present converters and the new ones that will be added for #1255, single parameter converters look like the right choice. Standardizing on single-parameter may lead to higher number of converters (e.g. `convert_timedelta_to_seconds` will become `convert_timedelta_to_seconds_real64`, `convert_timedelta_to_seconds_int32`, etc.), but that's better since it reduces the inter-dependency between code generator and metadata. Besides, given the maturity of nimi-python, it's unlikely to have a huge increase in number of coverters.

sbethur:

 - Changed converters to have single parameter
 - Got rid of converters that are not being used

* Fix #1302 

Ran unit tests


generated/nidcpower/nidcpower/_converters.py
```diff
@@ -170,7 +165,7 @@ def convert_month_to_timedelta(months):
 
 # This converter is not called from the normal codegen path for function. Instead it is
 # call from init and is a special case. Also, it just returns a string rather than a ctype object
```
marcoskirsch: Is this comment accurate? Converters don't return a ctype object. I think you can fix it.
Also I'm surprised the `encoding` parameter was unused all along!

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -493,7 +428,7 @@ def test_repeated_capabilies_slice_prefix():
 
 
 def test_repeated_capabilies_from_init():
-    test_result = convert_repeated_capabilities_from_init((slice(0, 1), '2', [4, '5-6'], '7-9', '11:14', '16, 17'), '')
+    test_result = convert_repeated_capabilities_from_init((slice(0, 1), '2', [4, '5-6'], '7-9', '11:14', '16, 17'))
```
marcoskirsch: You had to change this test for a reason. Testing `''` made sense, I think.
Are we ok with this behavior change?

generated/nidcpower/nidcpower/_converters.py
sbethur: It is still accurate, although I don't see the value of it. I'll remove the part about return type.

generated/nidcpower/nidcpower/_converters.py
sbethur: I didn't get your question.

marcoskirsch: I misread the code.

generated/nidcpower/nidcpower/_converters.py
sbethur: done

PR #1308, opened 2020-03-04
Remove misleading rep cap tip

texasaggie97:
- [X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).
- [X] ~~I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~
- [X] ~~I've added tests applicable for this pull request~~
### What does this Pull Request accomplish?
* Remove code from repeated capabilities tip since there is no way to make the code snippet work for all cases
* If we have more metadata (I.e. what repeated capability it applies to) we might be able to add this back in a more correct way
### List issues fixed by this Pull Request below, if any.
* Fix #1125 

### What testing has been done?
* Visual inspection of generated files

PR #1317, opened 2020-03-09
Move rep cap docs

texasaggie97:

* Moves repeated capabilities documentation from the middle of `class.rst` to its own file, `rep_caps.rst`
* This raises the visibility of repeated capabilities and allows us to put more content there

### ~~List issues fixed by this Pull Request below, if any.~~
* Based on comment in PR #1308 


src/nidmm/nidmm.mak
```diff
@@ -6,7 +6,7 @@ MODULE_FILES_TO_GENERATE := $(DEFAULT_PY_FILES_TO_GENERATE)
 
 MODULE_FILES_TO_COPY := $(DEFAULT_PY_FILES_TO_COPY)
 
-RST_FILES_TO_GENERATE := $(DEFAULT_RST_FILES_TO_GENERATE)
+RST_FILES_TO_GENERATE := $(filter-out rep_caps.rst,$(DEFAULT_RST_FILES_TO_GENERATE))
```
marcoskirsch: It's a shame that this knowledge (the fact that this driver doesn't have repeated capabilities) is duplicated - since it is also present in config.py.

That said, I think it's better to use additive rather than substractive logic:

* DEFAULT_RST_FILES_TO_GENERATE could be renamed to RST_FILES_FOR_ALL_MODULES
* On those that *do* have repeated capability, add rep_caps.rst

This may mean modifying more .mak since most drivers have repeated capability. I think the clarity is worth it. This has a WTH factor to it.

texasaggie97: We have been filtering out from the beginning so this is not new. If we make the change only for rep_caps.rst then it doesn't match the others, and if we change all of them, there will be unrelated changes in this PR. If you want to make this change, it should be a separate PR after this one is merged (or just closed).

Current filter out stats:
`rep_caps.rst`: nidmm, nimodinst, nise, nitclk
`enums.py`: nimodinst, nitclk
`_attributes.py`: nimodinst, nise
`enums.rst`: nimodinst, nitclk

Changing to additive rather than subtractive would mean adding these files to all the others not listed for that particular file.

PR #1328, opened 2020-03-13
Change the type of method parameters and properties in nidigital to enums

### LINKED ISSUE #1066: nidigital API methods and properties should accept enums when possible

### LINKED ISSUE #1244: Do not use the abbreviation for History RAM - HRAM in public API
### Description of issue

We should not be using the abbreviation for History RAM - HRAM in public API.

We need to update:

- `enums.HramCyclesToAcquire`
- `enums.HramTriggerType`

sbethur:

- Change the type of method parameters and properties in nidigital to enums, where possible.
- Do not use the abbreviation for History RAM - HRAM in public API.

* Fix #1066 
* Fix #1244 

Visual inspection of generated files and running existing system tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
         * `get_pattern_pin_names` - [#1292](https://github.com/ni/nimi-python/issues/1292)
     * #### Changed
-        * Change the type of method parameters and properties to enums - [#1066](https://github.com/ni/nimi-python/issues/1066)
+        * Change the type of applicable method parameters and properties to enums - [#1066](https://github.com/ni/nimi-python/issues/1066)
     * #### Removed
         * `get_pattern_pin_list`, `get_pattern_pin_indexes` and `get_pin_name` - [#1292](https://github.com/ni/nimi-python/issues/1292)
 * ### NI-TClk
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
         * `get_pattern_pin_names` - [#1292](https://github.com/ni/nimi-python/issues/1292)
     * #### Changed
+        * Change the type of method parameters and properties to enums - [#1066](https://github.com/ni/nimi-python/issues/1066)
     * #### Removed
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ All notable changes to this project will be documented in this file.
     * #### Added
         * `get_pattern_pin_names` - [#1292](https://github.com/ni/nimi-python/issues/1292)
     * #### Changed
+        * Change the type of method parameters and properties to enums - [#1066](https://github.com/ni/nimi-python/issues/1066)
```
marcoskirsch: ...`applicable method parameters and`...

docs/nidigital/enums.rst
```diff
@@ -77,6 +90,79 @@ DigitalState
...
+    .. py:attribute:: HistoryRAMTriggerType.PATTERN_LABEL
+
+
+
+PPMUCurrentLimitBehavior
+------------------------
+
+.. py:class:: PPMUCurrentLimitBehavior
+
+    .. py:attribute:: PPMUCurrentLimitBehavior.REGULATE
```
marcoskirsch: Is this correct, an enum of 1 value?
If so, we should consider (whenever we go over defaults) to make this value the default if it makes sense to and other ADEs do so (in nimi-python as a general rule we use default values less often than LabVIEW APIs)

src/nidigital/metadata/enums.py
```diff
@@ -127,7 +127,7 @@
             }
         ]
     },
-    'HramCyclesToAcquire': {
+    'HistoryRAMCyclesToAcquire': {
```
marcoskirsch: Where did all these (not only the changed ones) enum names come from?
Do they match the .ctl in the LabVIEW API?

For internal sanity we should strive to make them match.

sbethur: I created an [issue ](https://github.com/ni/nimi-python/issues/1330) to address this.

CHANGELOG.md
sbethur: done

docs/nidigital/enums.rst
sbethur: Yes, there is only one value for this enum. It is not used in any methods, it's only used in one property. We could make this property private since it's useless but it would be easy to forget to make it public when another value is added; better leave it public.

Issue #1216, opened 2020-01-20
Use datetime.timedelta for all time related properties and method arguments in nidigital


sbethur:
### Description of issue

Use datetime.timedelta for all time related properties and method arguments in nidigital. Users would be able to pass either floats (to represent time in seconds) or datetime.timedelta objects. This will make nidigital consistent with other API in nimi-python.

PR #1332, opened 2020-03-18
Add default values to applicable method parameters

### LINKED ISSUE #1072: Improve usability of nidigital API by making method args optional when applicable

sbethur:

Add default values to applicable method parameters

* Fix #1072 

Ran system tests


docs/nidigital/class.rst
```diff
@@ -188,7 +188,7 @@ burst_pattern
 
     .. py:currentmodule:: nidigital.Session
 
-    .. py:method:: burst_pattern(site_list, start_label, select_digital_function, wait_until_done, timeout)
+    .. py:method:: burst_pattern(site_list, start_label, select_digital_function=True, wait_until_done=True, timeout=10.0)
```
marcoskirsch: timeout is the wrong type... have you addressed those converters or is there still a pending task about it?

docs/nidigital/class.rst
```diff
@@ -2194,7 +2194,7 @@ unload_all_patterns
 
     .. py:currentmodule:: nidigital.Session
 
-    .. py:method:: unload_all_patterns(unload_keep_alive_pattern)
+    .. py:method:: unload_all_patterns(unload_keep_alive_pattern=False)
```
marcoskirsch: The default contradicts the name of the function "unload all" but implicitly "not one of them".

docs/nidigital/class.rst
```diff
@@ -2249,7 +2249,7 @@ wait_until_done
 
     .. py:currentmodule:: nidigital.Session
 
-    .. py:method:: wait_until_done(timeout)
+    .. py:method:: wait_until_done(timeout=10.0)
```
marcoskirsch: type

src/nidigital/metadata/functions_addon.py
```diff
@@ -120,6 +120,7 @@
                 'type': 'ViInt32'
             },
             {
+                'default_value': 'datetime.timedelta(seconds=10.0)',
```
marcoskirsch: why is this default in `_addon.py`?
why is this the only default that uses `datetime.timedelta` ?

src/nidigital/system_tests/test_system_nidigital.py
```diff
@@ -210,12 +193,7 @@ def test_fetch_capture_waveform(multi_instrument_session):
 
     multi_instrument_session.pins['HighPins'].create_capture_waveform_parallel(waveform_name='capt_wfm')
 
-    multi_instrument_session.burst_pattern(
-        site_list='',
-        start_label='new_pattern',
-        select_digital_function=True,
-        wait_until_done=True,
-        timeout=5)
+    multi_instrument_session.burst_pattern(site_list='', start_label='new_pattern')
```
marcoskirsch: This is gone, but you may want to review the rest of the tests to verify that you prefer passing appropriate time data types whenever applicable.

docs/nidigital/class.rst
sbethur: Agreed, this is tricky function to design. Users typically (90%) don't use `keep_alive` pattern, so it makes sense to not require them to pass any parameter by using a default value. For those that do use `keep_alive` pattern, they typically don't want to unload `keep_alive` pattern; hence the default of `False`.

We could go for a different design like renaming the method or changing its parameters, but that would make it inconsistent with LV and C API. Even though it's semantically wrong, it doesn't lead the users to do the wrong thing. I'd say we keep it.

src/nidigital/metadata/functions_addon.py
sbethur: >why is this default in _addon.py?

Because it's part of the fancy function that's public.

> why is this the only default that uses datetime.timedelta ?

When this fancy function was developed, `timedelta` was used. Many other methods will also get the `timedelta` default when #1216 is fixed.

sbethur:
> I'm ok merging it, but I want to remind you that we need to do an API review meeting before release in which we carefully walk down the API with NI-Digital Pattern Driver experts.

Agreed, I'll setup a review meeting when we're close to done. I've been in constant discussion with PBD experts while designing the API and they have reviewed the changes to internal metadata

docs/nidigital/class.rst
marcoskirsch: Ok, the first part makes sense.

> but that would make it inconsistent with LV and C API

Are you using defaults in your C API?
Per our [API design guidelines](https://github.com/ni/nimi-python/wiki/API-design-guidelines#default-values-for-methods):

>For the most part, defaults are a subset of the defaults in the corresponding LabVIEW VIs
Less defaults than in LabVIEW APIs, where we may have overdone it
Defaults can always be added later
But never removed

So the fact that it is defaulted in LabVIEW doesn't suggest strongly that we have to.

In any case, I'm ok with the default given your explanation.

PR #1369, opened 2020-04-03
Use 54x3 in get_hardware_state test

marcoskirsch:
Use this device which is a lot faster to simulate and doesn't require synchronization.

Uses 54x3 instead of 5421 in `test_get_hardware_state`. Support was added in NI-FGEN 17.8.
Using 54x3 is preferrable because simulated sessions open much faster, and because we don't need to synchronize opening/closing those sessions in order to workaround driver bugs.

None

Let nimi-bot run the system test.


sbethur:
test this please

PR #1384, opened 2020-04-07
Improve formatting of niswitch can_connect documentation

mgunyho:

Improve formatting of docs of the `niswitch.Session.can_connect` function. I also fixed `NISWITCH_VAL_RSRC_IN_USE` that was incorrect in the metadata. Compare the generated docs:

(before)
image

(after)
image

Note that the warning is gone.

I understand that the metadata is generated from the API docs, so maybe they should be fixed there instead, feel free to close this PR if it's not worth it to merge here. I can also commit the changes to the generated files if it makes sense.


src/niswitch/metadata/enums.py
```diff
@@ -314,7 +314,7 @@
                 'documentation': {
                     'description': 'Resource in use'
                 },
-                'name': 'NISWITCH_VAL_RESOURCE_IN_USE',
+                'name': 'NISWITCH_VAL_RSRC_IN_USE',
```
marcoskirsch, line 317: This is a good catch! We need to fix our internal metadata.

marcoskirsch:
> I can also commit the changes to the generated files if it makes sense.

We do that with every PR. I think @sbethur will take care of it. Thanks for the PR!

src/niswitch/metadata/enums.py
sbethur: Changing the enum value would break source compatibility; I'll update the documentation to match the enum value.

sbethur:
test this please

PR #1385, opened 2020-04-07
Link nidigital docs and other documentation cleanup

texasaggie97:
* Links to nidigital documentation, what there is of it
* Fixes two sphinx warnings
    * One fixes formatting of code block

* Inspect documentation in web browser

### DIFF
--- a/build/templates/rep_caps.rst.mako
+++ b/build/templates/rep_caps.rst.mako
@@ -7,6 +7,7 @@
 
 %>\
 .. py:module:: ${module_name}
+    :noindex:
 
 .. py:currentmodule:: ${module_name}.Session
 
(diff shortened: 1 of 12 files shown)


build/templates/rep_caps.rst.mako
```diff
@@ -7,6 +7,7 @@
 
 %>\
 .. py:module:: ${module_name}
+    :noindex:
```
sbethur, line 10: What does this do?

texasaggie97: `.. py:module:: <module_name>` is also at the top of class.rst. Only one instance can be in the index, and I figured the class instance was more important.

This gets rid of a warning from Sphinx when generating the html files.

texasaggie97:
>That's a lot of places to manually edit. Should some of this move to codegeneration?

It is but it only has to happen when we add another driver and that doesn't happen very often.

Issue #1393, opened 2020-04-09
In nidigital, remove unnecessary prefix for file_path parameter in methods that load files


sbethur:
### Description of issue

```
    def load_levels(self, levels_file_path):
    def load_pin_map(self, pin_map_file_path):
    def load_specifications(self, specifications_file_path):
    def load_timing(self, timing_file_path):
    def unload_specifications(self, specifications_file_path):
```

In all the above methods, the parameter name can be `file_path`.

PR #1394, opened 2020-04-09
Improve usability of loading specifications, levels, and timing files and unloading specifications

### LINKED ISSUE #1392: Improve usability of loading specifications, levels, and timing files and unloading specifications in nidigital
### Description of issue

Users typically load multiple levels files, multiple timing files, and at least one specifications file. Currently this requires `load_levels()`, `load_timing()`, and `load_specifications()` to be called in separate loops.

Usability can be improved by exposing a single method that allows multiple files of all three kinds to be loaded, similar to [Load Specifications Levels and Timing](https://www.ni.com/documentation/en/ni-digital/latest/ni-digital-node-ref/load-spec-levels-timing-multiple/) in LV.

Similarly API should provide a method to [unload multiple specifications files](https://www.ni.com/documentation/en/ni-digital/latest/ni-digital-node-ref/nidigital-unload-specs-single/) in a single method call.

sbethur:

Improve usability of loading specifications, levels, and timing files and unloading specifications:
- Added `load_specifications_levels_and_timing` that allows loading of multiple specs, levels, and/or timing files in a single call
- Removed  `load_specifications`, `load_levels`, and `load_timing`
- Modified `unload_specifications` to allow unloading of one or more specs files in a single call

* Fix #1392

Added and executed new system tests

### DIFF
--- a/generated/nidigital/nidigital/session.py
+++ b/generated/nidigital/nidigital/session.py
@@ -2775,11 +2775,11 @@ class Session(_SessionBase):
         return
 
     @ivi_synchronized
-    def _load_files(self, loader, files):
+    def _file_helper(self, action, files):
         if isinstance(files, str):
             files = [files]
         for f in files:
-            loader(f)
+            action(f)
 
     def load_specifications_levels_and_timing(self, specifications_file_paths=None, levels_file_paths=None, timing_file_paths=None):
         '''load_specifications_levels_and_timing
@@ -2800,11 +2800,11 @@ class Session(_SessionBase):
 
         '''
         if specifications_file_paths is not None:
-            self._load_files(self._load_specifications, specifications_file_paths)
+            self._file_helper(self._load_specifications, specifications_file_paths)
         if levels_file_paths is not None:
-            self._load_files(self._load_levels, levels_file_paths)
+            self._file_helper(self._load_levels, levels_file_paths)
         if timing_file_paths is not None:
-            self._load_files(self._load_timing, timing_file_paths)
+            self._file_helper(self._load_timing, timing_file_paths)
 
     @ivi_synchronized
     def self_test(self):
@@ -2816,6 +2816,26 @@ class Session(_SessionBase):
         if code:
             raise errors.SelfTestError(code, msg)
         return None
+
+    @ivi_synchronized
+    def unload_specifications(self, file_paths):
+        '''unload_specifications
+
+        Unloads the given specifications sheets present in the previously loaded
+        specifications files that you select.
+
+        You must call LoadSpecificationsLevelsAndTiming to reload the files with updated
+        specifications values. You must then call apply_levels_and_timing in order to apply
+        the levels and timing values that reference the updated specifications values.
+
+        Note:
+        One or more of the referenced methods are not in the Python API for this driver.
+
+        Args:
+            file_paths (str or iterable of str): Absolute file path of one or more loaded specifications files.
+
+        '''
+        self._file_helper(self._unload_specifications, file_paths)
 
     @ivi_synchronized
     def write_source_waveform_site_unique(self, waveform_name, waveform_data):
--- a/generated/nidigital/nidigital/session.py
+++ b/generated/nidigital/nidigital/session.py
@@ -2825,11 +2825,13 @@ class Session(_SessionBase):
         return
 
     @ivi_synchronized
-    def _file_helper(self, action, files):
+    def _call_method_with_iterable(self, method, files):
+        if files is None:
+            return
         if isinstance(files, str):
             files = [files]
         for f in files:
-            action(f)
+            method(f)
 
     def load_specifications_levels_and_timing(self, specifications_file_paths=None, levels_file_paths=None, timing_file_paths=None):
         '''load_specifications_levels_and_timing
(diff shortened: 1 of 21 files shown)


marcoskirsch:
conflicts

generated/nidigital/nidigital/session.py
```diff
@@ -2774,6 +2774,38 @@ def delete_all_time_sets(self):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
+    @ivi_synchronized
+    def _file_helper(self, action, files):
```
marcoskirsch, line 2778: `    @staticmethod`

generated/nidigital/nidigital/session.py
```diff
@@ -2774,6 +2774,38 @@ def delete_all_time_sets(self):
...
+        if isinstance(files, str):
+            files = [files]
+        for f in files:
+            action(f)
+
+    def load_specifications_levels_and_timing(self, specifications_file_paths=None, levels_file_paths=None, timing_file_paths=None):
```
marcoskirsch, line 2784: specifications, levels are plural. Timing is singular. Shouldn't this be `timing_sheets_file_paths`?
If the function was called `load` would that be an improvement?
Interested in the corresponding names for VI and inputs.

generated/nidigital/nidigital/session.py
```diff
@@ -2774,6 +2774,38 @@ def delete_all_time_sets(self):
...
+
+            timing_file_paths (str or iterable of str): Absolute file path of one or more timing sheet files.
+
+        '''
+        if specifications_file_paths is not None:
+            self._file_helper(self._load_specifications, specifications_file_paths)
+        if levels_file_paths is not None:
+            self._file_helper(self._load_levels, levels_file_paths)
+        if timing_file_paths is not None:
+            self._file_helper(self._load_timing, timing_file_paths)
```
marcoskirsch, line 2807: I don't think this is very Pythonic, although it works.
An improvement would be to move the None check into the function.

generated/nidigital/nidigital/session.py
marcoskirsch, line 2778: Can you give this a more descriptive name? like `_call_method_with_iterable`
rename `action` -> `method`

generated/nidigital/nidigital/session.py
```diff
@@ -2774,6 +2774,38 @@ def delete_all_time_sets(self):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return
 
+    @ivi_synchronized
+    def _file_helper(self, action, files):
+        if isinstance(files, str):
```
marcoskirsch, line 2779: "Ask for forgiveness, not for permission"

So you can try `[] + files`which raises `TypeError` if invoked with string; in which case you can use `append`

generated/nidigital/nidigital/session.py
```diff
@@ -2970,8 +3019,8 @@ def is_done(self):
         return bool(done_ctype.value)
 
     @ivi_synchronized
-    def load_levels(self, levels_file_path):
-        r'''load_levels
+    def _load_levels(self, levels_file_path):
```
marcoskirsch, line 3022: The parameter names of the public function are different:
specifications_file_paths, levels_file_paths, timing_file_paths

Please be consistent.

src/nidigital/metadata/functions.py
```diff
@@ -2085,6 +2086,7 @@
         'returns': 'ViStatus'
     },
     'LoadSpecifications': {
+        'codegen_method': 'private',
         'documentation': {
             'description': 'TBD'
         },
```
marcoskirsch, line 2092: I assume we can remove documentation from private methods?

generated/nidigital/nidigital/session.py
sbethur: I followed the VI and input names used in [LabVIEW](https://www.ni.com/documentation/en/ni-digital/latest/ni-digital-node-ref/load-spec-levels-timing-multiple/); which unfortunately is not consistent in many ways.

We need to decide whether to be consistent with LabVIEW or break from it for the sake of better names.

I think `load` is too generic and isn't consistent with `load_patterns` and `load_pinmap`. In our documentation, `timing` is used with `levels` quite pervasively, we even have the [apply_levels_and_timing](https://nimi-python.readthedocs.io/en/master/nidigital/class.html#apply-levels-and-timing) public method.

We could update parameter names to `levels_sheets_file_paths` and `timing_sheets_file_path`; this would be a good improvement without being inconsistent. I like the extra context it provides about sheets and files that a new user would likely find confusing.

generated/nidigital/nidigital/session.py
sbethur: There's an issue - #1393 to address the parameter names of these methods. The thought was to make the parameter names be just `file_path` since the method name implies the kind of file and since only one file can be loaded at a time.

src/nidigital/metadata/functions.py
sbethur: I think so. Created #1405 to address it.

generated/nidigital/nidigital/session.py
sbethur: Done

generated/nidigital/nidigital/session.py
sbethur: Done

generated/nidigital/nidigital/session.py
marcoskirsch: ok

Issue #1391, opened 2020-04-08
Update apply_levels_and_timing() in nidigital to accept initial state parameters as a list


sbethur:
### Description of issue

`apply_levels_and_timing()` requires the 3 initial states parameters to be passed in as comma-delimited string representing pins:
https://github.com/ni/nimi-python/blob/8540e86f00dc4b7aa9dbf5d905275db7255d9c7b/generated/nidigital/nidigital/session.py#L572

We should allow the parameters to be `list` of 'str` too, as that is more Pythonic.

Also, check if there are other methods where we can do something similar.

Issue #1386, opened 2020-04-07
Improve usability of get_channel_name in nidigital API


sbethur:
### Description of issue

Currently, there are two very similar methods in the public API:

1. `get_channel_name()`
    - Accepts a single channel number as `int`.
    - Range is [1 : ChannelCount * InstrumentCount]
    - One-based channel input
    - On multi-instruemnt session, returns a string that represents a single fully-qualified channel name e.g. 'PXI1Slot2/0'
    - On single-instruemnt session, returns a string that represents a single channel number e.g. '0'
2. `get_channel_name_from_string()`
    - Accepts a `str` which can be one of the following formats:
        - A comma-separated list - for example, "0,2,3,1"
        - A range using a hyphen - for example, "0-3"
        - A range using a colon - for example, "0:3"
    - Range is [0 : (ChannelCount * InstrumentCount) - 1]
    - Zero-based channel input
    - On multi-instruemnt session, returns a comma-delimited string of fully-qualified channel names e.g. 'PXI1Slot2/0,PXI1Slot5/31'
    - On single-instruemnt session, returns a comma-delimited string of channel numbers e.g. '0,31'

Proposal:
1. Get rid of `get_channel_name()`
2. Rename `get_channel_name_from_string()` to `get_channel_name()`
3. Make the new `get_channel_name()` use `channels` repeated capability - This will make this method consistent with rest of the API w.r.t. using `channels` repeated capability and it will allow the implementation to leverage support for ranges,lists,slices etc. that are available for repeated capabilities.

marcoskirsch:
This does not work.
The `get_channel_name_from_string` method takes in 0-based _channel indices_. The repeated capability is for _channel names_, which in multi-instrument session **must be** fully qualified channel names like "Dev1/4". 

We cannot mix the two concepts.

We still need to figure out what to do about these methods though.

sbethur:
After discussing offline, the proposed solution is:

Any module in nimi-python that needs to provide an API for translating channel indices to names will have only one method - `get_channel_names(channel_indices)`.
  - `channel_indices` input represents channels numbers. Supported types:
    - str - List (comma delimited)
    - str - Range (using '-' or ':')
    - str - single item
    - int
    - list of str
    - tuple of str
    - range of str
    - slice of str
- `channel_indices` will be 0-based **(Needs to be clearly documented)**
- Accepted range of `channel_indices`: [0 : (num_channels_per_instrument * num_instruments) - 1]
[...]

PR #1404, opened 2020-04-13
Add codegen support for using converters to convert parameter of type string sequence to comma-delimited string

### LINKED ISSUE #1401: Add codegen support for using converters to convert string sequence to comma-delimited string
### Description of issue

#1391 requires logic in Python bindings to convert input of basic sequence types of `str` elements to comma-delimited string before calling into C library functions. The concept of converters as used in nimi-python is well suited for this job. But codegen does not support using converters in such cases.

sbethur:

- Rename convert_repeated_capabilities_from_init() to convert_repeated_capabilities_without_prefix(). Other than its current use for option_string in init(), it will be used for indices in get_channel_names() (#1386) and for initial state parameters in apply_levels_and_timing() (#1391)
- Update _get_ctype_variable_definition_snippet_for_string() to add case for string input with converter and special case for NOT using converter in init function

* Fix #1401 

Added new unit tests. Ran all unit tests. CI will run system tests.

### DIFF
--- a/build/helper/codegen_helper.py
+++ b/build/helper/codegen_helper.py
@@ -172,7 +172,7 @@ class IviDanceStep(Enum):
     'Step 2: Allocate the buffer, call back into the driver to get the actual data.'
 
 
-def get_ctype_variable_declaration_snippet(parameter, parameters, ivi_dance_step, config, use_numpy_array=False):
+def get_ctype_variable_declaration_snippet(parameter, parameters, ivi_dance_step, method_name, config, use_numpy_array=False):
     '''Returns array of python snippets that declares and initializes a ctypes variable for the parameter that can be passed to the Library.
 
     Logic for creating the appropriate snippet is split up in two helper functions. One for scalars and one for buffers.
@@ -191,7 +191,7 @@ def get_ctype_variable_declaration_snippet(parameter, parameters, ivi_dance_step
         module_name = '_visatype'
 
     if parameter['is_string'] is True:
-        definitions = _get_ctype_variable_definition_snippet_for_string(parameter, parameters, ivi_dance_step, module_name)
+        definitions = _get_ctype_variable_definition_snippet_for_string(parameter, parameters, ivi_dance_step, module_name, method_name, config)
     elif parameter['is_buffer'] is True:
         definitions = _get_ctype_variable_definition_snippet_for_buffers(parameter, parameters, ivi_dance_step, use_numpy_array, custom_type, module_name)
     else:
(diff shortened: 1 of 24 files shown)


build/helper/codegen_helper.py
```diff
@@ -200,12 +200,13 @@ def get_ctype_variable_declaration_snippet(parameter, parameters, ivi_dance_step
...
 
 
-def _get_ctype_variable_definition_snippet_for_string(parameter, parameters, ivi_dance_step, module_name):
+def _get_ctype_variable_definition_snippet_for_string(parameter, parameters, ivi_dance_step, module_name, method_name, config):
     '''These are the different cases for initializing the ctype variables for strings
 
     C010. Input repeated capability:                                           ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))
     C020. Input string:                                                        ctypes.create_string_buffer(parameter_name.encode(self._encoding))
     C030. Input string enum:                                                   ctypes.create_string_buffer(parameter_name.value.encode(self._encoding))
+    C040. Input uses converter:                                                ctypes.create_string_buffer(_converter.convert(parameter_name).encode(self._encoding))
```
marcoskirsch, line 209: this example is not good.
`_converters.` for starters.
Instead of `convert` which is a real-looking name, use something like `convert_foo` so it's clearly a placeholder.

build/helper/codegen_helper.py
```diff
@@ -221,6 +222,10 @@ def _get_ctype_variable_definition_snippet_for_string(parameter, parameters, ivi
             definition = 'ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case C010'
         elif parameter['enum'] is not None:
             definition = 'ctypes.create_string_buffer({0}.value.encode(self._encoding))  # case C030'.format(parameter['python_name'])
+        # _init_function is a special case - Converters should not be called within it since they are
```
marcoskirsch, line 225: These special cases (`method_name is not config['functions']['_init_function']['python_name']`) have a really high code-smell. It also seems to be the only reason why you had to pipe down `method_name` all the way here.

We should try to come up with a cleaner alternative.

@texasaggie97 what do you think?

texasaggie97: Instead of making a special case here (and having to plumb the method name all the way down) why not just remove the converter from `functions['_init_function']`? It is a copy of whichever init is listed in config: https://github.com/ni/nimi-python/blob/bcf6d2204bb1821ef398c1f91685a0e0294bb808/build/helper/metadata_add_all.py#L317

sbethur: > why not just remove the converter from functions['_init_function']?

I can't do that because it is used to generate the code that calls the converters if required, in `Session.__init__`:
https://github.com/ni/nimi-python/blob/bcf6d2204bb1821ef398c1f91685a0e0294bb808/build/templates/session.py.mako#L263

Converters need to be called here to get string representations of parameters, for use in `__repr__`.

build/helper/codegen_helper.py
sbethur: Fixed

PR #1409, opened 2020-04-15
Adding nidigital static i/o and sequencer function system tests

ni-jfitzger:

* Adds the nidigital system test `test_read_static`
* Adds the nidigital system test `test_write_static`
* Adds the nidigital system test `test_read_sequencer_flag`
* Adds the nidigital system test `test_write_sequencer_flag`
* Adds the nidigital system test `test_read_sequencer_register`
* Adds the nidigital system test `test_write_sequencer_register`

Ran the new system tests on a system with nidigital 19.0.1 installed.

### DIFF
--- a/src/nidigital/system_tests/test_system_nidigital.py
+++ b/src/nidigital/system_tests/test_system_nidigital.py
@@ -20,6 +20,37 @@ def multi_instrument_session():
         yield simulated_session
 
 
+def test_reset(multi_instrument_session):
+    multi_instrument_session.selected_function = nidigital.SelectedFunction.PPMU
+    assert multi_instrument_session.selected_function == nidigital.SelectedFunction.PPMU
+    multi_instrument_session.reset()
+    assert multi_instrument_session.selected_function == nidigital.SelectedFunction.DISCONNECT
+
+
+def test_reset_device(multi_instrument_session):
+    multi_instrument_session.selected_function = nidigital.SelectedFunction.PPMU
+    assert multi_instrument_session.selected_function == nidigital.SelectedFunction.PPMU
+    multi_instrument_session.reset_device()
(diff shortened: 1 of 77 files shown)


sbethur:
add to whitelist

src/nidigital/system_tests/test_system_nidigital.py
```diff
@@ -558,6 +558,46 @@ def test_ppmu_source(multi_instrument_session):
...
+def test_read_static(multi_instrument_session):
+    test_name = 'simple_pattern'
+    configure_session(multi_instrument_session, test_name)
+
+    pin_states = multi_instrument_session.pins['site0/LO0', 'site1/HI0'].read_static()
+
+    assert len(pin_states) == 2
+
+
+def test_write_static(multi_instrument_session):
```
sbethur, line 570: I think you can combine read() and write() into a single test - Write a value (preferably non-default), read it back and verify. It makes the test more robust.

The same applies to seq. flag and seq. register.

ni-jfitzger: I almost did it that way to begin with, but decided it was best to keep the tests separate and simple unless you asked me to do otherwise, since the goal seemed to be primarily to just call the api.

sbethur: We should not go out of the way to test more than the API. But if we can do more while testing the Python bindings, I don't see any reason why we shouldn't.

ni-jfitzger: We discussed this offline and decided not to combine the tests because there aren't really any examples of it being done elsewhere in the system tests (possibly due to lack of opportunity) and because writing values doesn't affect the read values in simulation for these functions.

However, we did agree that the assertions could be made more specific.

PR #1410, opened 2020-04-15
Improve usability of getting channel names from channel indices in nidigital

### LINKED ISSUE #1386: Improve usability of get_channel_name in nidigital API

sbethur:

- Add get_channel_names()
- Remove get_channel_name() and get_channel_name_from_string()

* Fix #1386

Added new system test and ran all unit tests. System tests will be run by CI.

### DIFF
--- a/docs/nidigital/class.rst
+++ b/docs/nidigital/class.rst
@@ -1265,9 +1265,9 @@ get_channel_names
                 Specifies indices for the channels in the session.
                 Valid values are from zero to the total number of channels in the session minus one.
                 The following types and formats are supported:
-                  - int - for example, 0
-                  - Basic sequence types (list, tuple, range, slice) - for example, [0, range(2, 4)]
-                  - str - for example, "0, 2, 3, 1", "0-3", "0:3"
+                  - int - example: 0
+                  - Basic sequence - example: [0, range(2, 4)]
+                  - str - example: "0, 2, 3, 1", "0-3", "0:3"
 
                 The input can contain any combination of above types. Both out-of-order and repeated indices are
                 supported ([2,3,0], [1,2,2,3]). White space characters, including spaces, tabs, feeds, and
--- a/docs/nidigital/class.rst
+++ b/docs/nidigital/class.rst
@@ -1252,6 +1252,9 @@ get_channel_names
 
             Returns a list of channel names for given channel indices.
 
+            This is useful in multi-instrument sessions, where channels are expected to be
+            referenced by their fully-qualified names, for example, PXI1Slot3/0.
+
             
 
 
@@ -1262,19 +1265,12 @@ get_channel_names
                 Specifies indices for the channels in the session.
                 Valid values are from zero to the total number of channels in the session minus one.
                 The following types and formats are supported:
-                  - Basic Sequence types
-                    - list - for example, [0, 2, 3, 1]
-                    - tuple - for example, (0, 2, 3, 1)
-                    - range - for example, range(0, 15, 2)
-                  - slice - for example, slice(0, 15, 2)
-                  - str
-                    - A comma-separated list - for example, "0,2,3,1"
-                    - A range using a hyphen - for example, "0-3"
-                    - A range using a colon - for example, "0:3"
                   - int - for example, 0
+                  - Basic sequence types (list, tuple, range, slice) - for example, [0, range(2, 4)]
+                  - str - for example, "0, 2, 3, 1", "0-3", "0:3"
 
                 The input can contain any combination of above types. Both out-of-order and repeated indices are
-                supported ("2,3,0", "1,2,2,3"). White space characters, including spaces, tabs, feeds, and
+                supported ([2,3,0], [1,2,2,3]). White space characters, including spaces, tabs, feeds, and
                 carriage returns, are allowed within strings. Ranges can be incrementing or decrementing.
 
                 
--- a/docs/nidigital/class.rst
+++ b/docs/nidigital/class.rst
@@ -1243,59 +1243,50 @@ frequency_counter_measure_frequency
 
 
 
-get_channel_name
-----------------
+get_channel_names
+-----------------
 
     .. py:currentmodule:: nidigital.Session
 
-    .. py:method:: get_channel_name(index)
+    .. py:method:: get_channel_names(indices)
 
-            TBD
+            Returns a list of channel names for given channel indices.
 
             
 
 
 
-            :param index:
+            :param indices:
 
 
-                
+                Specifies indices for the channels in the session.
+                Valid values are from zero to the total number of channels in the session minus one.
+                The following types and formats are supported:
+                  - Basic Sequence types
+                    - list - for example, [0, 2, 3, 1]
+                    - tuple - for example, (0, 2, 3, 1)
+                    - range - for example, range(0, 15, 2)
+                  - slice - for example, slice(0, 15, 2)
+                  - str
+                    - A comma-separated list - for example, "0,2,3,1"
+                    - A range using a hyphen - for example, "0-3"
+                    - A range using a colon - for example, "0:3"
+                  - int - for example, 0
 
-
-            :type index: int
-
-            :rtype: str
-            :return:
-
-
-                    
-
-
-
-get_channel_name_from_string
-----------------------------
-
-    .. py:currentmodule:: nidigital.Session
-
-    .. py:method:: get_channel_name_from_string(index)
-
-            TBD
-
-            
-
-
-
-            :param index:
-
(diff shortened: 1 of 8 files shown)


docs/nidigital/class.rst
```diff
@@ -1243,60 +1243,51 @@ frequency_counter_measure_frequency
...
+                carriage returns, are allowed within strings. Ranges can be incrementing or decrementing.
 
                 
 
 
-            :type index: str
+            :type indices: basic sequence types or str or int
 
-            :rtype: str
+            :rtype: list of str
```
marcoskirsch, line 1285: I think most clients are going to call this with a single index: `my_session.get_channel_names(0)`. It's going to be annoying that it returns a list.

I'm not sure if that's ok or not.

docs/nidigital/class.rst
```diff
@@ -1243,60 +1243,51 @@ frequency_counter_measure_frequency
...
+                The following types and formats are supported:
+                  - Basic Sequence types
+                    - list - for example, [0, 2, 3, 1]
+                    - tuple - for example, (0, 2, 3, 1)
+                    - range - for example, range(0, 15, 2)
+                  - slice - for example, slice(0, 15, 2)
+                  - str
+                    - A comma-separated list - for example, "0,2,3,1"
+                    - A range using a hyphen - for example, "0-3"
+                    - A range using a colon - for example, "0:3"
```
marcoskirsch, line 1273: These are not very Pythonic. I'd either remove from documentation or condense into a single line to de-emphasize importance.

```
- str - for example "0", "0,2,3", "0-3"...

File: docs/nidigital/class.rst [hunk 3]
```diff
@@ -1243,60 +1243,51 @@ frequency_counter_measure_frequency
...
+                  - int - for example, 0
```
marcoskirsch, line 1274: This is the most basic one so should go first

docs/nidigital/class.rst
```diff
@@ -1243,60 +1243,51 @@ frequency_counter_measure_frequency
...
-
-            :param index:
+                Specifies indices for the channels in the session.
+                Valid values are from zero to the total number of channels in the session minus one.
+                The following types and formats are supported:
+                  - Basic Sequence types
+                    - list - for example, [0, 2, 3, 1]
+                    - tuple - for example, (0, 2, 3, 1)
+                    - range - for example, range(0, 15, 2)
+                  - slice - for example, slice(0, 15, 2)
```
marcoskirsch, line 1269: can these be condensed to "iterable" or something like that?

generated/nidigital/nidigital/_library.py
```diff
@@ -434,13 +434,13 @@ def niDigital_GetChannelName(self, vi, index, name_buffer_size, name):  # noqa:
                 self.niDigital_GetChannelName_cfunc.restype = ViStatus  # noqa: F405
         return self.niDigital_GetChannelName_cfunc(vi, index, name_buffer_size, name)
 
-    def niDigital_GetChannelNameFromString(self, vi, index, name_buffer_size, name):  # noqa: N802
+    def niDigital_GetChannelNameFromString(self, vi, indices, name_buffer_size, names):  # noqa: N802
```
marcoskirsch, line 437: is this `index` to `indices` to be consistent with other ADEs (like LabVIEW) and other APIs (like NI-SCOPE)?

generated/nidigital/nidigital/session.py
```diff
@@ -2940,30 +2940,46 @@ def write_source_waveform_site_unique(self, waveform_name, waveform_data):
         self.sites[site_list]._write_source_waveform_site_unique_u32(waveform_name, len(waveform_data), actual_samples_per_waveform, data)
 
     @ivi_synchronized
-    def get_channel_name_from_string(self, index):
-        r'''get_channel_name_from_string
+    def get_channel_names(self, indices):
+        r'''get_channel_names
 
-        TBD
+        Returns a list of channel names for given channel indices.
```
marcoskirsch, line 2946: Where did this documentation come from? It won't be obvious to NI-Digital users why this is useful. The answer is that it's useful in multi-instrument sessions. The same is true for NI-SCOPE.

For NI-DCPower users there are more scenarios where this will be useful.

I think we should mention it.

src/nidigital/templates/session.py/fancy_get_pin_results_pin_information.py.mako
```diff
@@ -18,7 +18,7 @@
         pin_infos = []
         for i in range(len(pin_indexes)):
             pin_name = "" if pin_indexes[i] == -1 else self._get_pin_name(pin_indexes[i])
-            channel_name = self.get_channel_name(channel_indexes[i])
+            channel_name = self._get_channel_name(channel_indexes[i])
```
marcoskirsch, line 21: Consider removing `_get_channel_name` and implementing this in term of `get_channel_names`. I think that's less code overall.

generated/nidigital/nidigital/_library.py
sbethur: No, it's to be consistent with the use of "names" in method name - `get_channel_names`. I expect it to be consistent across all bindings in nimi-python but not with LabVIEW and C where `index` is used, to go with function/VI name - `GetChannelNameFromString`.

I'm ok with using `index` in Python too, for the sake of consistency.

docs/nidigital/class.rst
marcoskirsch: > Maybe we can qualify it along the lines of "use this as a last resort"?

I wouldn't use that wording :)
If you want to keep it, we'll just make it the last option.

generated/nidigital/nidigital/_library.py
marcoskirsch: I'm ok either way

sbethur: I'll keep it as `indices`

docs/nidigital/class.rst
sbethur: Neither is perfect, but we decided offline to use `list`.

docs/nidigital/class.rst
```diff
@@ -1243,60 +1243,47 @@ frequency_counter_measure_frequency
...
-            
 
+            :param indices:
 
 
-            :param index:
+                Specifies indices for the channels in the session.
+                Valid values are from zero to the total number of channels in the session minus one.
+                The following types and formats are supported:
+                  - int - for example, 0
```
marcoskirsch, line 1268: I think you should use colon rather than comma, and the word "for" isn't needed.

```
- int - example: 0
- Basic sequence - example: [0, 1, 2], [0, range(2, 4)]
- str - example: "0, 2, 3, 1", "0-3", "0:3"
```

so that all lines have mu

docs/nidigital/class.rst
marcoskirsch: Let's leave it as list which is the most semantically accurate thing. I hope we don't regret it ever :)
I think in most cases the results are passed to the `channels` repeated capability container as-is so nobody will care to even inspect them.

docs/nidigital/class.rst
sbethur: I'll keep it. It's already the last option.

docs/nidigital/class.rst
sbethur: Fixed

PR #1420, opened 2020-04-20
Add codegen support for basic sequence of strings type properties

### LINKED ISSUE #1419: Add codegen support for basic sequence type properties
### Description of issue

Currently, there are properties that are returned as comma-delimited string:
- driver_setup
- group_capabilities
- supported_instrument_models

Codegen support should be added so that these properties can be returned as `list` of `str`, which is semantically more correct.

This needs to fixed for #1417 and #1418.

sbethur:

~Add codegen support for basic sequence of strings type properties~
Add the ability to have attributes that return a list of strings when the runtime returns a comma-separated string and vice-versa.

* Fix #1419

Added new nifake unit tests. All tests will be run by CI.

### DIFF
--- a/build/templates/_attributes.py.mako
+++ b/build/templates/_attributes.py.mako
@@ -70,6 +70,15 @@ class AttributeViString(Attribute):
         session._set_attribute_vi_string(self._attribute_id, value)
 
 
+class AttributeViStringSequence(Attribute):
+
+    def __get__(self, session, session_type):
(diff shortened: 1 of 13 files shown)


sbethur:
> I am not convinced we should push this in until there is a need for it. We could close and leave the branch alone and it will be referenced from the issue.
> 
> The big drawback is that the branch will diverge from master which will make future merging harder.

Agreed. The change is simple enough, we don't even need to keep the branch around. PR diff can be used as reference when needed.

Issue #1426, opened 2020-04-22
nidigital.HistoryRAMCycleInformation().repr() returns an ambiguous string


ni-jfitzger:
### Description of issue

We have a system test for nidigital, `test_history_ram_cycle_information_representation`, which tests eval(repr()) of a `HistoryRAMCycleInformation` object.  However, the behavior it tests only works if HistoryRAMCycleInformation and PinState have been imported directly into the global namespace.  This seems to be a contradiction of the intended behavior of `__repr__` which is to be unambiguous.  See [this page](https://stackoverflow.com/questions/1436703/difference-between-str-and-repr) for more information about `__repr__` vs `__str__`.

### Steps to reproduce issue

Run the code below.

    ``` python
    import nidigital
    cycle_info = nidigital.HistoryRAMCycleInformation(
        pattern_name='pat',
        time_set_name='t0',
        vector_number=42,
        cycle_number=999,
        scan_cycle_number=13,
        expected_pin_states=[[nidigital.PinState.D, nidigital.PinState.D], [nidigital.PinState.V, nidigital.PinState.V]],
        actual_pin_states=[[nidigital.PinState.PIN_STATE_NOT_ACQUIRED, nidigital.PinState.PIN_STATE_NOT_ACQUIRED], [nidigital.PinState.NOT_A_PIN_STATE, nidigital.PinState.NOT_A_PIN_STATE]],
        per_pin_pass_fail=[[True, True], [False, False]])
    recreated_cycle_info = eval(repr(cycle_info))
    ```

PR #1430, opened 2020-04-23
Return custom objects that implements Sequence protocol, from fetch_capture_waveform()

### LINKED ISSUE #1389: Provide string representation for data returned by fetch_capture_waveform() in nidigital
### Description of issue

`fetch_capture_waveform` returns a dictionary:
https://github.com/ni/nimi-python/blob/8540e86f00dc4b7aa9dbf5d905275db7255d9c7b/generated/nidigital/nidigital/session.py#L1252

The documentation claims the dict values are of type `array.array` but the method returns `memoryview`. More importantly, the default string representation of `memoryview` does not return the data, which makes it hard for users to print the returned data or save it to file.

`memoryview` is used in the implementation because underlying C entry-point returns a flat buffer containing data for all the sites. In Python, `memoryview` objects are created for slices within the buffer and returned.

sbethur:

- Add a custom class - `CaptureWaveform` to implements `collections.abc.Sequence` interface
- Update `fetch_capture_waveform()` to return `CaptureWaveform` objects in the dictionary

**Note**: The PR is not complete. I'm creating a draft version to get some initial feedback before spending more time on it. The following items have not been done yet:
- Add necessary tests
- Update documentation
- Update CHANGELOG 

* Fix #1389

New tests added. CI will run all tests.

### DIFF
--- a/generated/nidigital/nidigital/__init__.py
+++ b/generated/nidigital/nidigital/__init__.py
@@ -11,6 +11,8 @@ from nidigital.session import Session  # noqa: F401
 
 from nidigital.history_ram_cycle_information import HistoryRAMCycleInformation  # noqa: F401
 
+from nidigital.capture_waveform import CaptureWaveform  # noqa: F401
+
 
 def get_diagnostic_information():
     '''Get diagnostic information about the system state that is suitable for printing or logging
--- a/generated/nidigital/nidigital/_library.py
+++ b/generated/nidigital/nidigital/_library.py
@@ -8,6 +8,8 @@ from nidigital._visatype import *  # noqa: F403,H303
 
 import nidigital.history_ram_cycle_information as history_ram_cycle_information  # noqa: F401
 
+import nidigital.capture_waveform as capture_waveform  # noqa: F401
+
 
 class Library(object):
     '''Library
--- a/generated/nidigital/nidigital/capture_waveform.py
+++ b/generated/nidigital/nidigital/capture_waveform.py
@@ -1,0 +1,21 @@
+import collections.abc
+
+
+class CaptureWaveform(collections.abc.Sequence):
+    '''Used by fetch_capture_waveform() to return the capture waveform data'''
+    def __init__(self, capture_waveform_data):
+        self._capture_waveform_data = capture_waveform_data
+
+    def __repr__(self):
+        return str(self)
+
+    def __str__(self):
+        return str(self._capture_waveform_data.tolist())
+
+    def __len__(self):
+        return len(self._capture_waveform_data)
+
+    def __getitem__(self, key):
+        if isinstance(key, slice):
+            return CaptureWaveform(self._capture_waveform_data[key])
+        return self._capture_waveform_data[key]
--- a/generated/nidigital/nidigital/session.py
+++ b/generated/nidigital/nidigital/session.py
@@ -15,6 +15,8 @@ import nidigital.errors as errors
 
 import nidigital.history_ram_cycle_information as history_ram_cycle_information  # noqa: F401
 
+import nidigital.capture_waveform as capture_waveform  # noqa: F401
+
 import nitclk
 
 # Used for __repr__
@@ -1265,7 +1267,7 @@ class _SessionBase(object):
         for i in range(actual_num_waveforms):
             start = i * actual_samples_per_waveform
             end = start + actual_samples_per_waveform
-            waveforms[site_list[i]] = mv[start:end]
+            waveforms[site_list[i]] = capture_waveform.CaptureWaveform(mv[start:end])
 
         return waveforms
 
--- a/src/nidigital/custom_types/capture_waveform.py
+++ b/src/nidigital/custom_types/capture_waveform.py
@@ -1,0 +1,21 @@
+import collections.abc
+
+
+class CaptureWaveform(collections.abc.Sequence):
+    '''Used by fetch_capture_waveform() to return the capture waveform data'''
+    def __init__(self, capture_waveform_data):
+        self._capture_waveform_data = capture_waveform_data
+
+    def __repr__(self):
+        return str(self)
+
+    def __str__(self):
+        return str(self._capture_waveform_data.tolist())
+
+    def __len__(self):
+        return len(self._capture_waveform_data)
+
+    def __getitem__(self, key):
+        if isinstance(key, slice):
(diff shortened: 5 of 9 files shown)


generated/nidigital/nidigital/__init__.py
```diff
@@ -11,6 +11,8 @@
 
 from nidigital.history_ram_cycle_information import HistoryRAMCycleInformation  # noqa: F401
 
+from nidigital.capture_waveform import CaptureWaveform  # noqa: F401
```
thejcannon, line 14: Something I've noticed is that the `__module__` and `__qualname__` of the re-exported types is exactly what we expect logically. But might not be what the client expects.

```py
>>> from nigitial import CaptureWaveform
>>> repr(CaptureWaveform(...))
nidigital.capture_waveform.CaptureWaveform(...)
```

I've been playing around with setting the class's `__module__` when re-exporting in `__init__.py`.
This has the added bonus of "using" the class so no more suppressing `F401`.

I haven't found anything online about others doing this or pros/cons, so just some food for thought 👍

generated/nidigital/nidigital/_library.py
```diff
@@ -8,6 +8,8 @@
 
 import nidigital.history_ram_cycle_information as history_ram_cycle_information  # noqa: F401
 
+import nidigital.capture_waveform as capture_waveform  # noqa: F401
```
thejcannon, line 11: For my own curiosity, why do we do import this is we don't use it?

generated/nidigital/nidigital/capture_waveform.py
```diff
@@ -0,0 +1,21 @@
+import collections.abc
+
+
+class CaptureWaveform(collections.abc.Sequence):
+    '''Used by fetch_capture_waveform() to return the capture waveform data'''
+    def __init__(self, capture_waveform_data):
+        self._capture_waveform_data = capture_waveform_data
+
+    def __repr__(self):
+        return str(self)
```
thejcannon, line 10: From https://docs.python.org/3.4/reference/datamodel.html#object.__repr__
> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment).

Usually this means You'll want the `__module__` and `__qualname__` of the class in there along with the args.

E.g.
```py
def __repr__(self):
   return "{}.{}({})".format(self.__class__.__module__, self.__class__.__qualname__, ...)
```

generated/nidigital/nidigital/capture_waveform.py
```diff
@@ -0,0 +1,21 @@
+import collections.abc
+
+
+class CaptureWaveform(collections.abc.Sequence):
+    '''Used by fetch_capture_waveform() to return the capture waveform data'''
+    def __init__(self, capture_waveform_data):
+        self._capture_waveform_data = capture_waveform_data
```
thejcannon, line 7: The class is already named "CaptureWaveform", so having the argument and attribute also have "capture waveform" in it is superfluous.

generated/nidigital/nidigital/__init__.py
sbethur: This comment is very timely - Jay created an [issue](https://github.com/ni/nimi-python/issues/1426) just yesterday about using `__module__` in `__repr__`.

Regarding overriding class' `__module__`, I found this SO post warning against doing it: https://stackoverflow.com/a/10194338

We should look into it more.

generated/nidigital/nidigital/_library.py
sbethur: Unlike the custom type in this PR, some types like:
https://github.com/ni/nimi-python/blob/397d60511f77852dba4494420dee8e343492fc5d/generated/niscope/niscope/waveform_info.py#L9
are passed in/out of the C library. We don't have a way to differentiate between the two kinds in metadata, so we unconditionally add them to `_library`. We could add something to metadata but we try to keep metadata keys as minimal as possible.

generated/nidigital/nidigital/capture_waveform.py
sbethur: > Usually this means You'll want the __module__ and __qualname__ of the class in there along with the args.

This is good to know. I couldn't find any best practice tips for `__repr__` online. I checked a couple of popular external libraries and didn't find them including `__module__`:
```
>>> import numpy as np
>>> n = np.arange(2)
>>> repr(n)
'array([0, 1])'

>>> import pandas as pd
>>> t = pd.CategoricalDtype(categories=['b', 'a'], ordered=True)
>>> repr(t)
"CategoricalDtype(categories=['b', 'a'], ordered=True)"
```

generated/nidigital/nidigital/__init__.py
thejcannon: If you look at this through the lens of "what am I defining my API to be" it makes a lot of sense to re-define it.
1. We tell clients to `from nigitial import CaptureWaveform`, guaranteeing that this will work (up until we make an API breaking change). It logically follows the `__module__` for that type would be `nidigital`.
2. We would avoid exposing an implementation detail. If we moved where the class was defined are we sure we aren't going to break someone? Someone JSON-ifying the data, or pickling it? Setting `__module__` to `nidgitial` means we can move the class as much as we want without worrying about exposing implementation details.

It's also worth pointing out that the SO post was about dynamically generated functions/classes.

Lastly, the new way leads to shorter strings, and who doesn't love that?

Anyways, just food for thought.

generated/nidigital/nidigital/capture_waveform.py
thejcannon: Look at internal libraries, as they're probably the best citizens:

```py
>>>import datetime
>>> datetime.datetime(1, 2, 3)
datetime.datetime(1, 2, 3, 0, 0)
>>> import operator
>>> operator.attrgetter("foo")
operator.attrgetter('foo')
```

Although that's not consistent either:

```py
>>> import collections
>>> collections.OrderedDict()
OrderedDict()
```

🤷‍♂️ Since you're defining this same type in several libraries, it might be good to tell the user _which_ library it came from.

generated/nidigital/nidigital/__init__.py
sbethur: Created an issue for this - #1433

sbethur:
PR rejected. See [here](https://github.com/ni/nimi-python/issues/1389#issuecomment-619259270) for details.

Issue #744, opened 2018-02-17
datetime.timedelta not precise enough for all nimi-python use cases


marcoskirsch:
#659 is merged which is great because it's slick. However, it introduced one problem: Python's built-in datetime.timedelta is not precise enough for all of nimi-python's use cases.

For example:
The NI-SCOPE attribute NISCOPE_ATTR_END_OF_RECORD_TO_ADVANCE_TRIGGER_HOLDOFF is specified as a ViReal64 (which is a double) in the C API and can have values with resolution greater than a microsecond.
But in the Python API this corresponds to [`niscope.Session.end_of_record_to_advance_trigger_holdoff`](http://nimi-python.readthedocs.io/en/master/niscope/attributes.html#niscope.Session.end_of_record_to_advance_trigger_holdoff) which is a [datetime.timedelta](https://docs.python.org/2/library/datetime.html#timedelta-objects) and can only be specified in resolution of microseconds but not smaller.

marcoskirsch:
So how to solve this?

Using float in the Python APIs like before #659 would be an un-pythonic step back.

What would be really neat, is a drop-in, duck-type replacement for datetime.datetime that supports resolutions higher than than the built-in class. Then nimi-python could switch to using that wherever appropriate.

Fladolcetta:
The workaround is to use a float in seconds via #796

ni-jfitzger:
Additional examples of this problem for niscope:
* Time Per Record: https://nimi-python.readthedocs.io/en/master/niscope/class.html#horz-time-per-record
* Runt Time Low Limit: https://nimi-python.readthedocs.io/en/master/niscope/class.html#runt-time-low-limit
* Runt Time High Limit: https://nimi-python.readthedocs.io/en/master/niscope/class.html#runt-time-high-limit

There may be others; I haven't gone over the API and devices with a fine-toothed comb.
Of these, Time Per Record is the most important and most common, by a long shot.  It is supported by all devices and is commonly used for horizontal configuration, though NI examples tend to steer users toward alternative settings.  It's not unreasonable that a user would configure a scope in such a way that Time Per Record would be less than 1 us.

Issue #1382, opened 2020-04-06
nitclk.SessionReference.sample_clock_delay loses precision when read


marcoskirsch:
### Description of issue

Property `nitclk.SessionReference.sample_clock_delay` returns a `timedelta`. Values from driver will usually be in the nanosecond range. But `timedelta` supports 1 microsecond precision.

We should use [hightime](https://github.com/ni/hightime). We need to review the whole API as other things may be affected.

PR #1445, opened 2020-04-29
Change the type of applicable properties and method parameters from `datetime.timedelta` to `hightime.timedelta` in nidigital

### LINKED ISSUE #1368: datetime.timedelta does not provide the necessary resolution for many methods and properties in nidigital
### Description of issue

`datetime.timedelta` has a microsecond resolution, which is not sufficient for many method parameters, return values, and properties. e.g. edge placement, vector period, TDR offset etc.

### LINKED ISSUE #1397: Return type of get_time_set_edge() and get_time_set_period() should be change hightime.TimeDelta
### Description of issue

All time-based method parameters and properties were updated to use timedelta type. But looks like `get_time_set_edge` and `get_time_set_period` got missed. When #1368 is fixed, we should also change the return type of aforementioned methods.

sbethur:

Change the type of applicable properties and method parameters from `datetime.timedelta` to `hightime.timedelta` in nidigital. `hightime.timedelta` is a drop-in replacement for `datetime.timedelta` but with sub-microsecond resolution.

- Update code generator to generate code using `datetime.timedelta` or `hightime.timedelta`, depending on `uses_hightime` key in `config.py`
- Update applicable nidigital attribute and function metadata to use `hightime.timedelta`
- Add nidigital unit tests to verify usage of `hightime.timedelta`

* Fix #1368
* Fix #1397 

Add nidigital unit tests to verify usage of `hightime.timedelta` in applicable property and method. CI runs all the tests.

Note that the tests were added to nidigital instead of nifake. This is because each module can specify whether it uses `datetime` or `hightime` via config metadata. I chose to have nifake continue to use `datetime` since other nimi-python modules use `datetime` and rely on nifake unit tests. This allows having test coverage for both `datetime` and `hightime`.

### DIFF
--- a/build/templates/_attributes.py.mako
+++ b/build/templates/_attributes.py.mako
@@ -3,10 +3,15 @@ ${template_parameters['encoding_tag']}
 <%
     module_name = template_parameters['metadata'].config['module_name']
     config = template_parameters['metadata'].config
+    timedelta_type = config['timedelta_type']
 %>\
 import ${module_name}._converters as _converters
 
+% if config['uses_hightime']:
+import hightime
+% else:
 import datetime
+% endif
 
 
 class Attribute(object):
@@ -28,7 +33,7 @@ class AttributeViInt32(Attribute):
 class AttributeViInt32TimeDeltaMilliseconds(Attribute):
 
     def __get__(self, session, session_type):
-        return datetime.timedelta(milliseconds=session._get_attribute_vi_int32(self._attribute_id))
+        return ${timedelta_type}(milliseconds=session._get_attribute_vi_int32(self._attribute_id))
 
     def __set__(self, session, value):
         session._set_attribute_vi_int32(self._attribute_id, _converters.convert_timedelta_to_milliseconds_int32(value).value)
@@ -55,7 +60,7 @@ class AttributeViReal64(Attribute):
 class AttributeViReal64TimeDeltaSeconds(Attribute):
 
     def __get__(self, session, session_type):
-        return datetime.timedelta(seconds=session._get_attribute_vi_real64(self._attribute_id))
+        return ${timedelta_type}(seconds=session._get_attribute_vi_real64(self._attribute_id))
 
     def __set__(self, session, value):
         session._set_attribute_vi_real64(self._attribute_id, _converters.convert_timedelta_to_seconds_real64(value).value)
--- a/build/templates/_attributes.py.mako
+++ b/build/templates/_attributes.py.mako
@@ -3,7 +3,7 @@ ${template_parameters['encoding_tag']}
 <%
     module_name = template_parameters['metadata'].config['module_name']
     config = template_parameters['metadata'].config
-    timedelta_type = config['timedelta_type']
+    datetime_library = config['datetime_library']
 %>\
 import ${module_name}._converters as _converters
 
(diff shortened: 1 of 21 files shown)


sbethur:
Converted to draft PR since this cannot be merged until https://github.com/ni/hightime/pull/36 is merged and published on PyPI.

sbethur:
test this please

build/templates/_attributes.py.mako
```diff
@@ -3,10 +3,15 @@ ${template_parameters['encoding_tag']}
...
     config = template_parameters['metadata'].config
+    datetime_library = config['datetime_library']
 %>\
 import ${module_name}._converters as _converters
 
+% if config['uses_hightime']:
+import hightime
+% else:
 import datetime
+% endif
```
marcoskirsch, line 14: So we either use `hightime` all the time or `datetime` all the time?
I think it would be better to use `datetime` all the time __unless__ hightime is needed. This means we always import `datetime` and optionally import `hightime`.

sbethur: Yes, it's one or the other. Given `hightime` is designed to be type-compatible with `datetime`, I thought it would be better for usability to have all `timedelta` objects to be the same type.

As an added bonus, development and maintenance becomes easier and less error-prone. If we use both types, attribute classes:
- `AttributeViInt32TimeDeltaMilliseconds`
- `AttributeViReal64TimeDeltaSeconds`
and converters:
- `convert_seconds_real64_to_timedelta`
- `convert_seconds_real64_to_timedeltas`
- `convert_timedelta_to_seconds_real64`
- `convert_timedelta_to_milliseconds_int32`
- `convert_timedeltas_to_seconds_real64`
will need to be duplicated. Metadata will have to carefully maintained to use the correct attribute class or converter.

Can you please elaborate how targeted use of `hightime` would be better?

marcoskirsch: > Can you please elaborate how targeted use of hightime would be better?
 
1. Similar methods across different nimi-python modules would return precisely the same type.
2. When we fix #744 adn #1382 we won't change the type returned by some existing functions that are precise enough with `datetime.timedelta`, duck typing notwithstanding.
3. When other driver runtimes add functions that require more precision, we won't be impacting existing functions like in no. 2.

I welcome the opinion of @texasaggie97 and @thejcannon

marcoskirsch: > As an added bonus, development and maintenance becomes easier and less error-prone.

I do see this benefit.

sbethur: Hmm, I'm sold on points 2 and 3. I'll start looking into making the required changes. Should be relatively straight-forward.

thejcannon: I don't think I really have an opinion here. 

`hightime` sometimes works for me:
If we do `hightime` sometimes, we can always switch to `hightime` always if we want. It's a lot harder if we do `hightime` always and want to switch it to do `hightime` sometimes.

texasaggie97: I think consistency within the driver > consistency across drivers
Also, since the normal `datetime` and `timedelta` are base classes for hightime, something like `dt 
= hightime.datetime(...); isinstance(dt, datetime.datetime)` will return true.

Overall, it seems like a lot of overhead for something that won't make a material difference.

sbethur: TL;DR: After offline discussion, we decided to replace `datetime.timedelta` with `hightime.timedelta` across the board in all usages of `timedelta` in all modules in nimi-python.

Pros:
- Code generator is simpler, no `if-else` cases
- No need to carefully curate metadata, with respect to which `timedelta` to use
- Consistency within and across nimi-python modules
- No churn if underlying driver runtime or a new device family starts supporting higher resolution for some property or method parameter

Cons:
- Currently, certain modules in nimi-python do not have a need for higher resolution provided by `hightime`. Users of those modules will be forced to install a third-party library - `hightime` as a dependency and use `hightime.timedelta` type.

Given that `datetime.datetime` and `datetime.timedelta` are base classes for the respective `hightime` variants, with duck-typing, users should not have to do anything different to use `hightime` types, other than importing from a different package.

Pros clearly outweigh the cons, thus the decision to move to `hightime.timedelta` wholesale.

sbethur:
The changes are moved  to new PR - https://github.com/ni/nimi-python/pull/1449

Issue #1388, opened 2020-04-07
For properties and method parameters that accept multiple types, documentation should promote the recommended type


sbethur:
### Description of issue

The main idea being documentation should guide the users to use the most-pythonic type for any property or method parameter. e.g.:
https://github.com/ni/nimi-python/blob/8540e86f00dc4b7aa9dbf5d905275db7255d9c7b/generated/nidigital/nidigital/session.py#L607

documentation says:
```
offsets (list of float in seconds or datetime.timedelta):
```

In this case, documentation should mention that `datetime.timedelta` is the recommended type, as that the pythonic way of representing time intervals.

Another instance would be the documentation for repeated capabilities:

> Repeated capabilities attributes use the indexing operator [] to indicate the repeated capabilities. The parameter can be a string, list, tuple, or slice (range). Each element of those can be a string or an integer. If it is a string, you can indicate a range using the same format as the driver: ‘0-2’ or ‘0:2’

I think it's still valuable to document all the different types that _can_ be passed in, for cases where the recommended type cannot be used; but there should be a clear distinction between recommended type and the others.

sbethur:
Along with documentation, examples should also be updated to use recommended types.

PR #1449, opened 2020-05-14
Replace usage of `datetime` library types with `hightime` types

### LINKED ISSUE #744: datetime.timedelta not precise enough for all nimi-python use cases

### LINKED ISSUE #1368: datetime.timedelta does not provide the necessary resolution for many methods and properties in nidigital

### LINKED ISSUE #1382: nitclk.SessionReference.sample_clock_delay loses precision when read

### LINKED ISSUE #1397: Return type of get_time_set_edge() and get_time_set_period() should be change hightime.TimeDelta

sbethur:

In all nimi-python modules, change the type of applicable properties and method parameters from `datetime.timedelta` to `hightime.timedelta` and from `datetime.datetime` to `hightime.datetime`.

`hightime` types are drop-in replacements for `datetime` counterparts, with sub-microsecond resolution. See https://github.com/ni/nimi-python/pull/1445#discussion_r424031427 for the rationale behind this change.

- Update code generator to generate code using `hightime` types
- Update applicable attribute and function metadata to use `hightime` types
- Update nifake unit tests to verify usage of `hightime` types

* Fix #744
* Fix #1368 
* Fix #1382 
* Fix #1397 

Update unit tests for converter and nifake to verify usage of `hightime` types. CI runs all the tests.

### DIFF
--- a/docs/nidcpower/class.rst
+++ b/docs/nidcpower/class.rst
@@ -739,7 +739,7 @@ fetch_multiple
                 .. note:: When setting the timeout interval, ensure you take into account any triggers so that the timeout interval is long enough for your application.
 
 
-            :type timeout: float in seconds or hightime.timedelta
+            :type timeout: hightime.timedelta, datetime.timedelta, or float in seconds
 
             :rtype: list of Measurement
             :return:
@@ -1785,7 +1785,7 @@ wait_for_event
                     application.
 
 
-            :type timeout: float in seconds or hightime.timedelta
+            :type timeout: hightime.timedelta, datetime.timedelta, or float in seconds
 
 
 Properties
@@ -3370,17 +3370,17 @@ measure_complete_event_delay
 
         The following table lists the characteristics of this property.
 
-            +----------------+----------------------------------------+
-            | Characteristic | Value                                  |
-            +================+========================================+
-            | Datatype       | float in seconds or hightime.timedelta |
-            +----------------+----------------------------------------+
-            | Permissions    | read-write                             |
-            +----------------+----------------------------------------+
-            | Channel Based  | Yes                                    |
-            +----------------+----------------------------------------+
-            | Resettable     | No                                     |
-            +----------------+----------------------------------------+
+            +----------------+-------------------------------------------------------------+
+            | Characteristic | Value                                                       |
+            +================+=============================================================+
+            | Datatype       | hightime.timedelta, datetime.timedelta, or float in seconds |
+            +----------------+-------------------------------------------------------------+
+            | Permissions    | read-write                                                  |
+            +----------------+-------------------------------------------------------------+
+            | Channel Based  | Yes                                                         |
+            +----------------+-------------------------------------------------------------+
+            | Resettable     | No                                                          |
+            +----------------+-------------------------------------------------------------+
 
         .. tip::
             This property corresponds to the following LabVIEW Property or C Attribute:
@@ -3529,17 +3529,17 @@ measure_record_delta_time
 
         The following table lists the characteristics of this property.
 
-            +----------------+----------------------------------------+
-            | Characteristic | Value                                  |
-            +================+========================================+
-            | Datatype       | float in seconds or hightime.timedelta |
-            +----------------+----------------------------------------+
-            | Permissions    | read only                              |
-            +----------------+----------------------------------------+
-            | Channel Based  | Yes                                    |
-            +----------------+----------------------------------------+
-            | Resettable     | No                                     |
-            +----------------+----------------------------------------+
+            +----------------+-------------------------------------------------------------+
+            | Characteristic | Value                                                       |
+            +================+=============================================================+
+            | Datatype       | hightime.timedelta, datetime.timedelta, or float in seconds |
+            +----------------+-------------------------------------------------------------+
+            | Permissions    | read only                                                   |
+            +----------------+-------------------------------------------------------------+
+            | Channel Based  | Yes                                                         |
+            +----------------+-------------------------------------------------------------+
+            | Resettable     | No                                                          |
+            +----------------+-------------------------------------------------------------+
 
         .. tip::
             This property corresponds to the following LabVIEW Property or C Attribute:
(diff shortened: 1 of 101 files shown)


docs/nidcpower/class.rst
```diff
@@ -739,7 +739,7 @@ fetch_multiple
                 .. note:: When setting the timeout interval, ensure you take into account any triggers so that the timeout interval is long enough for your application.
 
 
-            :type timeout: float in seconds or datetime.timedelta
+            :type timeout: float in seconds or hightime.timedelta
```
marcoskirsch, line 742: Since we document this can take a float, why wouldn't we document this can take a `datetime.timedelta`?
Applies to all input parameters.

generated/nimodinst/setup.py
```diff
@@ -45,6 +45,7 @@ def read_contents(file_to_read):
     install_requires=[
         'enum34;python_version<"3.4"',
         'singledispatch;python_version<"3.4"',
+        'hightime',
```
marcoskirsch, line 48: This is a bit of a shame - nimodinst truly doesn't use any timedeltas yet we're installing it.

src/nidcpower/disabled_examples/nidcpower_advanced_sequence.py
```diff
@@ -26,7 +26,7 @@ def example(resource_name, channels, options, steps, voltage_start, voltage_fina
     with nidcpower.Session(resource_name=resource_name, channels=channels, options=options) as session:
 
         session.source_mode = nidcpower.SourceMode.SEQUENCE
-        session.source_delay = datetime.timedelta(seconds=0.1)
+        session.source_delay = hightime.timedelta(seconds=0.1)
```
marcoskirsch, line 29: I see you changed client code. I think that's fine.
But do we have system tests in which we exercise and verify the higher precision of hightime?

src/nidcpower/disabled_examples/nidcpower_advanced_sequence.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/python
```
marcoskirsch, line 1: Why is this example disabled? We shipped #957.
Did we just forget?

marcoskirsch: Or merge error?

src/nidcpower/metadata/attributes.py
```diff
@@ -666,7 +666,7 @@
         'name': 'MEASURE_COMPLETE_EVENT_DELAY',
         'resettable': False,
         'type': 'ViReal64',
-        'type_in_documentation': 'float in seconds or datetime.timedelta'
+        'type_in_documentation': 'float in seconds or hightime.timedelta'
```
marcoskirsch, line 669: It is regrettable that we have this documentation copied over and over and over.
You'd hope we could derive it from the `atttribute_class`.

docs/nidcpower/class.rst
sbethur: Yup, we should. I will update. I'll also reorder it as:
```
hightime.timedelta or datetime.timedelta or float in seconds
```

generated/nimodinst/setup.py
sbethur: Typically user applications will include other nimi-python modules along with nimodinst. In the end, maybe it doesn't really matter that we have an unnecessary dependency?

docs/nidcpower/class.rst
marcoskirsch: Ok, remember there are several that take integer in milliseconds, since you'll be copying and pasting.

I've said this before - I'd rather not even mention the fact that you can pass a number and I sort of regret it.

sbethur: > remember there are several that take integer in milliseconds

Good point, thanks for reminding me. I made the appropriate changes.

> I've said this before - I'd rather not even mention the fact that you can pass a number and I sort of regret it.

Agreed. There is an issue for it - https://github.com/ni/nimi-python/issues/1388

src/nidcpower/disabled_examples/nidcpower_advanced_sequence.py
sbethur: Given that system tests run on simulated sessions, it's difficult to verify the higher precision of `hightime` in system tests. I've updated nifake unit tests to verify the precision, in both properties and method parameters.

src/nidcpower/disabled_examples/nidcpower_advanced_sequence.py
sbethur: I think we just forgot. Created an issue for it - https://github.com/ni/nimi-python/issues/1454.

src/nidcpower/metadata/attributes.py
sbethur: Agreed. Created an issue for it - https://github.com/ni/nimi-python/issues/1455.

PR #1452, opened 2020-05-15
Update NI-SCOPE metadata to use 20.0 released API

omejia10:
- [x ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

- [x ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md)

Adds attributes released with the 20.0 NI-SCOPE API

- built the component
- CI


sbethur:
ok to test

CHANGELOG.md
```diff
@@ -52,6 +52,14 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### NI-SCOPE
     * #### Added
+        * `end of acquisition event terminal name` attribute
```
marcoskirsch, line 55: those are not the names of the attributes, please write them exactly as they appear on the API

marcoskirsch: You could also add them all in one liner and say something like:

* API parity with NI-SCOPE 20.0
   * `foo`
   * `bar`
   * `baz`

src/niscope/metadata/attributes.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-SCOPE API metadata version 20.0.0d13
+# This file is generated from NI-SCOPE API metadata version 20.1.0d1
```
marcoskirsch, line 2: 20.0.0f0 won't do, huh? Please confirm. If that's the case well that's too bad and alos why we don't put this version in the README.

omejia10: yeah, sadly 20.0 won't cut it.  The attributes can only made available in this version.

CHANGELOG.md
```diff
@@ -52,6 +52,14 @@ All notable changes to this project will be documented in this file.
...
 * ### NI-SCOPE
     * #### Added
+        * API parity with NI-SCOPE 20.0
+            * `END_OF_RECORD_EVENT_TERMINAL_NAME`
+            * `ADVANCE_TRIGGER_TERMINAL_NAME`
+            * `REF_TRIGGER_TERMINAL_NAME`
+            * `START_TRIGGER_TERMINAL_NAME`
+            * `READY_FOR_ADVANCE_EVENT_TERMINAL_NAME`
+            * `READY_FOR_REF_EVENT_TERMINAL_NAME`
+            * `READY_FOR_START_EVENT_TERMINAL_NAME`
```
marcoskirsch, line 62: These should be the Python propery names. As added, they are metadata names, which clients would not recognize unless for some obscure reason they looked at the repo.

For less ambiguity, consider typing `Session.end_of_record_event_terminal_name`. If it was a method, you'd add `()` at the end.

CHANGELOG.md
```diff
@@ -52,6 +52,14 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### NI-SCOPE
     * #### Added
+        * API parity with NI-SCOPE 20.0 adding properties:
```
sbethur, line 55: `end_of_acquisition_event_terminal_name` is missing in this list.

sbethur, line 55: Better wording suggestion:
```
API parity with NI-SCOPE 20.0 by adding the following properties:
```

PR #1457, opened 2020-05-19
Fix tox env ordering issue

sbethur:

Fix tox env ordering issue:
- flake8 and docs depend on codegen
- test depends on installers (for `install_local_wheel.py --driver nitclk`), not codegen
- pkg depends on installers

None

Ran `tox` in serial and parallel modes and verified the ordering is correct.

### DIFF
--- a/tox-travis.ini
+++ b/tox-travis.ini
@@ -140,8 +140,11 @@ deps =
 
 depends =
     codegen: py38-clean
-    test: py38-codegen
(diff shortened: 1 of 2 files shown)


sbethur:
> If this allows us to run tox in parallel on CI, why wouldn't we want to do it?

I don't see a reason why not. Trying it out in https://github.com/ni/nimi-python/pull/1471.

PR #1467, opened 2020-05-26
Update prepare_python_system_tests.py' and 'run_python_system_tests.py'

### LINKED ISSUE #1440: Cannot use 'prepare_python_system_tests.py' and 'run_python_system_tests.py' as standalone script.
### Description of issue

While using 'prepare_python_system_tests.py' as standalone script, we are getting 'FileNotFoundError' error. 
Since we are downloading only 'prepare_python_system_tests.py', there won't be any 'src' folder.
Script we are using: [Here](https://github.com/ni/nimi-python/pull/1269#discussion_r383402003)

If we run 'prepare_python_system_tests.py' by removing os.listdir('src'), then it will fail to run 'run_python_system_tests.py', since we are changing the directory to generated/driver folder and there won't be any 'src' folder.

### System report

python prepare_python_system_tests.py --driver nidigital --python-version py36

    ```
      File "prepare_python_system_tests.py", line 14, in <module>
        choices=os.listdir('src'))
    FileNotFoundError: [WinError 3] The system cannot find the path specified: 'src'
    ```

After removing os.listdir('src'):

    ```
    .......
    ****Invoking build specific script****
    Running command:
    ['python', '../../tools/run_python_system_tests.py', '-d', 'niscope', '--python-version', 'py36']
      File "../../tools/run_python_system_tests.py", line 9, in <module>
        choices=os.listdir('src'))
    FileNotFoundError: [WinError 3] The system cannot find the path specified: 'src'
      File "prepare_python_system_tests.py", line 76, in <module>
        results = subprocess.run(command, check=True)
      File "C:\Program Files\Python36\lib\subprocess.py", line 438, in run
        output=stdout, stderr=stderr)
    subprocess.CalledProcessError: Command '['python', '../../tools/run_python_system_tests.py', '-d', 'nidigital', '--python-version', 'py36']' returned non-zero exit status 1.
    ```

### Steps to reproduce issue

Run the following commands:
1. mkdir %temp%\nimi-python
2. chdir %temp%\nimi-python
3. curl -o prepare_python_system_tests.py https://raw.githubusercontent.com/ni/nimi-python/master/tools/prepare_python_system_tests.py
4. python prepare_python_system_tests.py --driver niscope --python-version py36

Note:
On some machines python is pointing to python2.7, we can use 'py -3' or directly 'py' to call into python 3 in above command.

pmitra30:

Changed the choices parameter to be a list of all the supported drivers instead of generating it through the 'src' folder

* Fix #1440

Ran the test locally and verified it passes. CI will run all tests.

### DIFF
--- a/tools/prepare_python_system_tests.py
+++ b/tools/prepare_python_system_tests.py
@@ -44,7 +44,7 @@ zip_ref.extractall(zip_folder)
 zip_ref.close()
 print(zip_folder)
 
-src = os.path.join(zip_folder, os.listdir(zip_folder)[0], 'generated')
+src = os.path.join(zip_folder, os.listdir(zip_folder)[0], 'src')
 
 parser = argparse.ArgumentParser(description='Downloads the latest release artifacts from nimi-python and runs system tests on the specified python package.',
                                  formatter_class=argparse.ArgumentDefaultsHelpFormatter)
--- a/tools/prepare_python_system_tests.py
+++ b/tools/prepare_python_system_tests.py
@@ -10,7 +10,8 @@ import zipfile
 parser = argparse.ArgumentParser(description='Downloads the latest release artifacts from nimi-python and runs system tests on the specified python package.',
                                  formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 parser.add_argument('-d', '--driver', required=True, type=str,
-                    help='Python package name.')
+                    help='Python package name.',
+                    choices=['nidigital', 'nidmm', 'nidcpower', 'niscope', 'nifgen', 'nimodinst', 'nise', 'niswitch', 'nitclk'])
 parser.add_argument('-pv', '--python-version', required=False, type=str,
                     help='Python version to be run. This is used to invoke the appropriate tox environment.',
                     choices=['py35', 'py36', 'py37', 'py38', ], default='py38')
--- a/tools/prepare_python_system_tests.py
+++ b/tools/prepare_python_system_tests.py
@@ -7,20 +7,6 @@ import time
 import urllib.request
 import zipfile
 
-parser = argparse.ArgumentParser(description='Downloads the latest release artifacts from nimi-python and runs system tests on the specified python package.',
-                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-parser.add_argument('-d', '--driver', required=True, type=str,
-                    help='Python package name.',
-                    choices=['nidigital', 'nidmm', 'nidcpower', 'niscope', 'nifgen', 'nimodinst', 'nise', 'niswitch', 'nitclk'])
-parser.add_argument('-pv', '--python-version', required=False, type=str,
-                    help='Python version to be run. This is used to invoke the appropriate tox environment.',
-                    choices=['py35', 'py36', 'py37', 'py38', ], default='py38')
-parser.add_argument('-pb', '--python-bitness', required=False, type=str,
-                    help='Python bitness to be run. "32" means pass "--32" to tox, which will force 32 bit. "any" does not pass anything to tox, so it will used whatever bitness is installed, preferring 64 if available',
-                    choices=['32', 'any'], default=None)
-args = parser.parse_args()
-
-
 print('****Installing tox to Python.****')
 results = subprocess.run(["python", '-m', 'pip', 'install', '--disable-pip-version-check', '--upgrade', 'pip', 'tox'], check=True)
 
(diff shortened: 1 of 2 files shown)


tools/prepare_python_system_tests.py
```diff
@@ -11,7 +11,7 @@
                                  formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 parser.add_argument('-d', '--driver', required=True, type=str,
                     help='Python package name.',
-                    choices=os.listdir('src'))
+                    choices=['nidigital', 'nidmm', 'nidcpower', 'niscope', 'nifgen', 'nimodinst', 'nise', 'niswitch', 'nitclk'])
```
sbethur, line 14: Listing out the modules here would become a maintenance burden (when we add or remove modules). Can you obtain this list dynamically by parsing the data got from `https://api.github.com/repos/ni/nimi-python/releases`?

sbethur:
ok to test

tools/prepare_python_system_tests.py
pmitra30: Modified the 'prepare_python_system_test' to download the latest package first and then generating the choices list from the 'src' directory. Did the same in 'run_python_system_test'

tools/run_python_system_tests.py
```diff
@@ -2,11 +2,13 @@
 import os
 import subprocess
 
+src = r'..\..\src'
```
marcoskirsch, line 5: backslashes are Windows-specific.

pmitra30: Fixed that

tools/run_python_system_tests.py
```diff
@@ -2,11 +2,13 @@
 import os
 import subprocess
 
+src = r'../../src'
```
sbethur, line 5: The pythonic way of doing this is:
```
src = pathlib.Path().resolve().parents[1] / 'src'
```

pmitra30: Made the requested changes

tools/prepare_python_system_tests.py
```diff
@@ -7,20 +7,6 @@
...
-parser.add_argument('-pv', '--python-version', required=False, type=str,
-                    help='Python version to be run. This is used to invoke the appropriate tox environment.',
-                    choices=['py35', 'py36', 'py37', 'py38', ], default='py38')
-parser.add_argument('-pb', '--python-bitness', required=False, type=str,
-                    help='Python bitness to be run. "32" means pass "--32" to tox, which will force 32 bit. "any" does not pass anything to tox, so it will used whatever bitness is installed, preferring 64 if available',
-                    choices=['32', 'any'], default=None)
-args = parser.parse_args()
-
-
 print('****Installing tox to Python.****')
```
marcoskirsch, line 10: Doesn't this mean that this script will now print and do a bunch of stuff before processing command line arguments?

You want to validate all inputs before doing anything. Why was it moved down in the script?

tools/run_python_system_tests.py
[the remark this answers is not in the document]
pmitra30: Have changed it to `src = pathlib.Path().resolve().parents[1] / 'src'`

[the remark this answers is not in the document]
sbethur:
```
src = r'../../src'
```
There is a newer version of this line, where it's replaced by:
```
src = pathlib.Path().resolve().parents[1] / 'src'
```

`pathlib` provides a nicer interface for getting parent directories.

tools/prepare_python_system_tests.py
pmitra30: Without this change, the script would throw FileNotFoundError as it would only download the 'prepare_python_system_test'. There won't be any 'src' folder. Due to this, the package has to be downloaded first

tools/prepare_python_system_tests.py
```diff
@@ -58,6 +44,22 @@
 zip_ref.close()
 print(zip_folder)
 
+src = os.path.join(zip_folder, os.listdir(zip_folder)[0], 'src')
```
marcoskirsch, line 47: Avoid being inconsistent. You use `src = pathlib.Path().resolve().parents[1] / 'src'` in the other script.

tools/prepare_python_system_tests.py
sbethur: As part of providing help for `-d` option, script lists the available choices:
```
parser.add_argument('-d', '--driver', required=True, type=str,
                    help='Python package name.',
                    choices=os.listdir(src))
```

I suggested reordering so we can avoid hard-coding the choices. We could remove the prints but still do the processing upfront. If you think that's overkill, we could just get rid of the choices.

marcoskirsch: That is an unfortunate chicken and egg problem. I think the root cause is that the scripts aren't decoupled from each other.

There are two scripts:

1. prepare_python_system_tests.py
2. run_python_system_tests.py

The first one prepares for, then runs the selected system test. This is not what the name of the script conveys. It calls on the second script in order to do it.

What if you decoupled the two scripts so that they are more specific and do just the one thing?

The first script prepares to run tests, the second one runs the specified test. It means the test suites need to call two scripts, but it also means that you could in theory test several drivers (for example, NI-SCOPE and NI-TClk) on the same Test Suite which would be desirable.

sbethur: I like this idea. @pmitra30, I'm working on a new PR to implement Marcos' idea.

sbethur:
I'm implementing [this](https://github.com/ni/nimi-python/pull/1467#discussion_r432673308) solution in a new PR.

tools/prepare_python_system_tests.py
pmitra30: We can do it this way:

1. We can install the necessary modules and the latest package using `prepare_python_system_tests.py`. To do that we do not need to restrict the driver argument. Actually we do not need to pass any arguments to install the modules and download the nimi-python package.

2. In `run_python_system_tests.py` we can have the logic to pass the driver name and python version to run the tests for any driver

Issue #1475, opened 2020-06-02
Test code should be separate from module code


thejcannon:
(Copied from #1473)

Having the tests be in the same file as the source code is not something I'd expect when installing a Python module.

- It pollutes the API of the module (attribute discovery will show a bunch of unrelated functions)
- It can pollute the module's dependencies
- It bloats the installed modules (more LOC to download and sift through)
- It makes features such as `pytest`'s `conftest.py` harder to use.

If you want people to be able to run tests after installation you should allow installation using an extra (like `nidmm[test]`)

sbethur:
#1477 should fix this issue for `converters`. With that, none of the "shipping" modules should have this problem.
But there are other modules in [helper folder](https://github.com/ni/nimi-python/tree/master/build/helper), which are used during codegen, that we need to fix.

thejcannon:
:rocket:

PR #1477, opened 2020-06-03
Split the converter functions and its unit tests into two files

### LINKED ISSUE #1473: Unintentional dependency on pytest?
### Description of issue

nidcpower has a dependency on pytest (I assume unintentional since it is not listed as a dependency).

### System report

python -c "import nidcpower; nidcpower.print_diagnostic_information()
  File "<string>", line 1, in <module>
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\__init__.py", line 10, in <module>
    from nidcpower.session import Session  # noqa: F401
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\session.py", line 8, in <module>
    import nidcpower._attributes as _attributes
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\_attributes.py", line 3, in <module>
    import nidcpower._converters as _converters
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\_converters.py", line 10, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'

### Steps to reproduce issue

`pip install nidcpower==1.3.0`
`python`
`import nidcpower`
> Traceback (most recent call last):
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\__init__.py", line 10, in <module>
    from nidcpower.session import Session  # noqa: F401
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\session.py", line 8, in <module>
    import nidcpower._attributes as _attributes
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\_attributes.py", line 3, in <module>
    import nidcpower._converters as _converters
  File "C:\Users\nitest\.virtualenvs\nidcpower_test-yBwmuPo5\lib\site-packages\nidcpower\_converters.py", line 10, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'

sbethur:

Converter functions and unit tests for them were in the same file. This is Unpythonic as explained in #1475. It's also a maintenance headache as seen in #1473.

I've split the functions and unit tests into two separate files. The unit test gets codegen'd into `unit_tests` folder, which is not included in installer.

* Fix #1473 
* Partial fix for #1475 

Ran the unit tests locally and verified that the number of unit tests is the same as before.


sbethur:
test this please

sbethur:
> I think this is more complex than it needs to be.
> 
> We should run the unit tests for _converters.py as part of the NI-FAKE unit tests and that's it. There's no need to copy them over for each generated module.

I mistakenly thought it would be more complex to avoid copying `test_converters.py` in modules other than nifake. But I was able to do it by modifying `nifake.mak`. I've updated the PR.

> This assumes the NI-FAKE version of _converters.py is the superset of all which I believe to be the case.

It is a superset except for this converter and the one below it:
https://github.com/ni/nimi-python/blob/2447b23a8c38b6ca8d3da3026aa92019037533b0/generated/nitclk/nitclk/_converters.py#L274

Although there aren't any unit tests for the nitclk-specific converters (maybe there should be). I think we should do something about this exception, I'm not sure what that something is.

build/templates/test_converters.py.mako
```diff
@@ -0,0 +1,297 @@
+# This file was generated
```
marcoskirsch, line 1: I don't think this needs to remain a .mako file.
Just turn it into a regular .py file and source it in https://github.com/ni/nimi-python/tree/master/src/nifake/unit_tests

sbethur: Hmm, you're right. No need to use mako at all 🤦‍♂️ 

Updated the PR.

PR #1498, opened 2020-08-05
NI-DMM: Make getting the last calibration temp and datetime consistent with other drivers.

### LINKED ISSUE #1462: `nidmm.Session.get_cal_date_and_time` and `get_last_cal_temp` is different from other nimi-python drivers
### Description of issue

`nidmm.Session.get_cal_date_and_time` and `get_last_cal_temp` are inconsistent with the rest of `nimi-python`.
Instead of one method that takes in an enum, we should have two methods for each: 
* `get_ext_cal_last_date_and_time` and `get_self_cal_last_date_and_time`.
* `get_ext_cal_last_temp` and `get_self_cal_last_temp`
Note - this inconsistency was inherited from the C/LabVIEW API.

shilohbeeler-NI:

Adds external and self calibration variations of get last cal temp and get last cal date and time, in order to be consistent with other drivers with similar methods. Removes the previously existing get last cal date time method since it was an added function rather than part of base metadata, and is now redundant with the newly added methods.

* Fix #1462 

Branch has successfully built and passed all system tests on a local testing machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,14 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### `nidmm` (NI-DMM)
     * #### Added
+        * New methods for getting calibration information. - [#1462](https://github.com/ni/nimi-python/issues/1462)
+            * `get_ext_cal_last_date_and_time`
+            * `get_ext_cal_last_temp`
+            * `get_self_cal_last_date_and_time`
+            * `get_self_cal_last_temp`
     * #### Changed
     * #### Removed
+        * Outdated `get_cal_date_and_time` method removed in favor of more consistent new methods. - [#1462](https://github.com/ni/nimi-python/issues/1462)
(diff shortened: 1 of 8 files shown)


sbethur:
ok to test

shilohbeeler-NI:
I'll leave the branch on my fork of the repo so that any code I wrote doesn't just disappear. If I'd had a week more, my plan was to look into making the cal temp/date and time stuff into single-sourced templates, but I don't think that's feasible in the span of approximately a day since I've never looked into that part of the codebase at all before.

marcoskirsch:
I think it's worth adding these methods, and we worry about obsoleting the old inconsistent ones separately. Do you agree @sbethur ?

sbethur:
> I think it's worth adding these methods, and we worry about obsoleting the old inconsistent ones separately. Do you agree @sbethur ?

I agree. But I'd want to add it without duplicating the mako templates between `niscope` and `nidmm`.

PR #1506, opened 2020-08-11
Add support for properties of type `str` that represents a repeated capability

### LINKED ISSUE #1497: meas_other_channel property should accept all types supported by channel repeated capability
### Description of issue

Property in question:
https://github.com/ni/nimi-python/blob/df494361119cab7bc27191cef77925661cfe03c7/generated/niscope/niscope/session.py#L902

Currently, the property only supports `str` type when setting. But it should support other types (like int) that are supported when channel is passed in as repeated capability. Code like this should be made to work:
```
my_session.meas_other_channel = 2
```

sbethur:

- Add support for properties of type `str` that represents a repeated capability
- It will be used to fix #1497 

Contribute towards fixing #1497 

Added and ran new unit tests in nifake

### DIFF
--- a/generated/nifake/nifake/session.py
+++ b/generated/nifake/nifake/session.py
@@ -181,6 +181,11 @@ class _SessionBase(object):
 
     A property of type string with read/write access.
     '''
+    read_write_string_repeated_capability = _attributes.AttributeViStringRepeatedCapability(1000010)
+    '''Type: basic sequence types or str or int
+
+    A property of type string with read/write access, that represents a repeated capability
+    '''
 
     def __init__(self, repeated_capability_list, vi, library, encoding, freeze_it=False):
         self._repeated_capability_list = repeated_capability_list
--- a/generated/nifake/nifake/session.py
+++ b/generated/nifake/nifake/session.py
@@ -182,9 +182,14 @@ class _SessionBase(object):
     A property of type string with read/write access.
     '''
     read_write_string_repeated_capability = _attributes.AttributeViStringRepeatedCapability(1000010)
-    '''Type: basic sequence types or str or int
+    '''Type: Any repeated capability type, as defined in nimi-python:
+        - str
+        - str - Comma delimited list
+        - str - Range (using '-' or ':')
+        - int
+        - Basic sequence types (list, tuple, range, slice) of other supported types
 
-    A property of type string with read/write access, that represents a repeated capability
+    A property with read/write access, that represents a repeated capability
     '''
 
     def __init__(self, repeated_capability_list, vi, library, encoding, freeze_it=False):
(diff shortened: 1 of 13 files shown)


generated/nifake/nifake/session.py
```diff
@@ -181,6 +181,11 @@ class _SessionBase(object):
 
     A property of type string with read/write access.
     '''
+    read_write_string_repeated_capability = _attributes.AttributeViStringRepeatedCapability(1000010)
+    '''Type: basic sequence types or str or int
+
+    A property of type string with read/write access, that represents a repeated capability
```
marcoskirsch, line 187: I know this is internal, but we support and even encourage int, lists of int... etc
Look at other places where we explain it and consider using that string, or instead say something like:

> Anything nimi-python generally allows as a repeated capability (int, str, sequence of these...)

generated/nifake/nifake/unit_tests/test_session.py
```diff
@@ -966,6 +966,25 @@ def test_set_attribute_string(self):
...
+            from mock import call
+            calls = [call(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), _matchers.ViStringMatcher(''), _matchers.ViAttrMatcher(1000010), _matchers.ViInt32Matcher(0), None), call(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), _matchers.ViStringMatcher(''), _matchers.ViAttrMatcher(1000010), _matchers.ViInt32Matcher(20), _matchers.ViCharBufferMatcher(len(string)))]
+            self.patched_library.niFake_GetAttributeViString.assert_has_calls(calls)
+            assert self.patched_library.niFake_GetAttributeViString.call_count == 2
+
+    def test_set_attribute_string_with_converter(self):
+        self.patched_library.niFake_SetAttributeViString.side_effect = self.side_effects_helper.niFake_SetAttributeViString
+        attribute_id = 1000010
+        with nifake.Session('dev1') as session:
+            session.read_write_string_repeated_capability = 42
```
marcoskirsch, line 985: I think the `nifake` test covering a single conversion type is sufficient - since the converter function is heavily tested. However it would be best if you can mock the conversion function, then we know for sure we're good.

Is this a reasonable thing to do?

sbethur: I added this unit test to keep the coverage up. So the intent is to just exercise the API method.

What would mocking the converter get us?

marcoskirsch: >What would mocking the converter get us?
 
Our Python APIs accept all sorts of things for repeated capabilities:
* `0`
* `[0,1]`
* `"0,1"`
* etc... 

We know the conversion is done by a single function so this test only tests one type of conversion. In essence, the test assumes that if one conversion worked, all of them will because they go through the one function. Which is true today but there is guarantee that it will be true in the future.

So we could either stop assuming it goes through that one function by checking for it explicitly through a mock (test the implementation, not great) or we can test more types (test the behavior, arguably more correct).

sbethur: > test assumes that if one conversion worked, all of them will because they go through the one function. Which is true today but there is guarantee that it will be true in the future.

I think we can or should trust ourselves to not break this in the future.

As you mentioned, mocking isn't great since we'll testing the implementation. Testing more types is better but this is not the right place to do it. Unit tests for `_converters.convert_repeated_capabilities()` is the more appropriate place and we do test all the supported types there.

I'm inclined to the leave the tests as is, unless you feel strongly against it.

PR #1511, opened 2020-08-20
Make meas_other_channel in niscope work with channel repeated capability types

### LINKED ISSUE #1497: meas_other_channel property should accept all types supported by channel repeated capability

sbethur:

Make meas_other_channel in niscope work with channel repeated capability types. Users can now pass in `str` or `int` or any other type supported by channel repeated capability.

Metadata changes come from internal metadata source.

* Fix #1497 

None. The generated code is tested by nifake unit tests.

### DIFF
--- a/docs/niscope/class.rst
+++ b/docs/niscope/class.rst
@@ -4902,7 +4902,7 @@ meas_other_channel
             +----------------+------------+
             | Characteristic | Value      |
             +================+============+
-            | Datatype       | str        |
+            | Datatype       | str or int |
             +----------------+------------+
             | Permissions    | read-write |
             +----------------+------------+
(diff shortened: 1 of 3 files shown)


docs/niscope/class.rst
```diff
@@ -4902,7 +4902,7 @@ meas_other_channel
             +----------------+------------+
             | Characteristic | Value      |
             +================+============+
-            | Datatype       | str        |
+            | Datatype       | str or int |
```
marcoskirsch, line 4905: list too, no?

marcoskirsch: never mind, in this case it's one channel only

PR #1514, opened 2020-08-24
Single source the code for populating sample data in niscope fetch methods

sbethur:

- Single source the code for populating sample data in niscope fetch methods by creating a helper method - `_populate_samples_info`
- With this change, the array of `waveform_info` instances is populated with data in two passes (one for samples data and one for waveform metadata) rather than one. I think the performance degradation from this is within the "acceptable" range and worth the simplification of code.

None

Ran existing system tests

### DIFF
--- a/generated/niscope/niscope/session.py
+++ b/generated/niscope/niscope/session.py
@@ -1974,14 +1974,14 @@ class _SessionBase(object):
 
         mv = memoryview(wfm)
 
-        self._populate_samples_info(wfm_info, mv, num_samples)
+        waveform_info._populate_samples_info(wfm_info, mv, num_samples)
 
         lwfm_i = len(wfm_info)
         lrcl = len(self._repeated_capability_list)
         # Should this raise instead? If this asserts, is it the users fault?
         assert lwfm_i % lrcl == 0, 'Number of waveforms should be evenly divisible by the number of channels: len(wfm_info) == {0}, len(self._repeated_capability_list) == {1}'.format(lwfm_i, lrcl)
         actual_num_records = int(lwfm_i / lrcl)
-        self._populate_channel_and_record_info(wfm_info, self._repeated_capability_list, range(offset, offset + actual_num_records))
+        waveform_info._populate_channel_and_record_info(wfm_info, self._repeated_capability_list, range(offset, offset + actual_num_records))
 
         return wfm_info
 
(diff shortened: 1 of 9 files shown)


generated/niscope/niscope/session.py
```diff
@@ -2046,19 +2041,24 @@ def fetch_array_measurement(self, array_meas_function, timeout=hightime.timedelt
...
-            wfm_info[i].samples = meas_wfm[start:end]
+        record_length = int(len(meas_wfm) / len(wfm_info))
+        self._populate_samples_info(wfm_info, meas_wfm, record_length)
 
         num_records = int(len(wfm_info) / len(self._repeated_capability_list))
         self._populate_channel_and_record_info(wfm_info, self._repeated_capability_list, range(num_records))
 
         return wfm_info
 
+    def _populate_samples_info(self, waveform_info, sample_data, num_samples_per_item):
```
marcoskirsch, line 2053: `self` is never used.
Is that true for `_populate_channel_and_record_info` as well?

I wonder if they should both move to be free-standing functions in `waveform_info.py`.

sbethur: Good point. `self` isn't used there either. I like moving them to `waveform_info.py`. Only issue is that `_populate_channel_and_record_info` is needed for `measurement_stats` too. Do we create two copies of it, one for each type? Maybe that's ok.

marcoskirsch: Have measurement_stats import waveform_info.py or viceversa? These are both fully handcoded. It also allows for adding unit tests to these functions.

sbethur: Yup, that works. Updated the PR.

generated/niscope/niscope/waveform_info.py
```diff
@@ -111,4 +111,20 @@ def __str__(self):
         return string_representation
 
 
-
+def _populate_samples_info(waveform_infos, sample_data, num_samples_per_waveform):
```
marcoskirsch, line 114: add docstring to both these functions

src/niscope/custom_types/measurement_stats.py
```diff
@@ -41,3 +44,7 @@ def __str__(self):
         string_representation += row_format_d.format('Number in Stats', self.num_in_stats)
 
         return string_representation
+
+
+def _populate_channel_and_record_info(measurement_stats, channels, records):
```
marcoskirsch, line 49: Why do you need this at all? it is just a straight up wrapper.

sbethur: Because I felt it would be weird to use a helper defined in `waveform_info` module to populate members of `measurement_stats` class. But I don't feel strongly either way. I can get rid of it, if you prefer.

generated/niscope/niscope/waveform_info.py
sbethur: done

src/niscope/custom_types/measurement_stats.py
sbethur: Removed the wrapper function in `measurement_stats.py`

PR #1530, opened 2020-11-10
Add type hints in the public API

texasaggie97-zz:
* Add type hints to public API
    * Uses `type_in_documentation` metadata since it mostly has the information needed - we need the complete list of possible types in the API, not just the type that is passed into the DLL
    * Several special cases to deal with free form test
    * Not sure how to put this information into the metadata
* Most type hints are calculated
* A few in fancy functions are hardcoded
* The preferred method for type hinting namedtuples is to create a class that inherits from `typing.NamedTuple`. This would require the couple of namedtuples we have to be moved to custom types. I did not do that work for this PR.
* Unit tests and flake8 still pass with type hinting added
* Need to add mypy tests, maybe using examples?

### DIFF
--- a/generated/nidcpower/nidcpower/session.py
+++ b/generated/nidcpower/nidcpower/session.py
@@ -2,8 +2,10 @@
 # This file was generated
 import array  # noqa: F401
 import ctypes
+import datetime
 # Used by @ivi_synchronized
 from functools import wraps
+import typing
 
 import nidcpower._attributes as _attributes
 import nidcpower._converters as _converters
@@ -2556,7 +2558,7 @@ class _SessionBase(object):
     ''' These are code-generated '''
 
     @ivi_synchronized
-    def self_cal(self):
+    def self_cal(self) -> None:
         r'''self_cal
 
         Performs a self-calibration upon the specified channel(s).
@@ -2600,7 +2602,7 @@ class _SessionBase(object):
         return
 
     @ivi_synchronized
-    def configure_aperture_time(self, aperture_time, units=enums.ApertureTimeUnits.SECONDS):
+    def configure_aperture_time(self, aperture_time: float, units: 'enums.ApertureTimeUnits' = enums.ApertureTimeUnits.SECONDS) -> None:
         r'''configure_aperture_time
 
         Configures the aperture time on the specified channel(s).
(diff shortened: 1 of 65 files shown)


generated/nidcpower/nidcpower/session.py
```diff
@@ -2734,15 +2735,14 @@ def measure_multiple(self):
                 - **in_compliance** (bool) - Always None
 
         '''
-        import collections
-        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance'])
+        measurement = typing.NamedTuple('Measurement', [('voltage', 'float'), ('current', 'float'), ('in_compliance', 'bool')])
```
bkeryan, line 2738: > The preferred method for type hinting namedtuples is to create a class that inherits from typing.NamedTuple. This would require the couple of namedtuples we have to be moved to custom types. I did not do that work for this PR.

FYI, there is a separate GitHub issue about that: https://github.com/ni/nimi-python/issues/1885

generated/nimodinst/nimodinst/session.py
```diff
@@ -350,7 +351,7 @@ def _get_installed_device_attribute_vi_string(self, index, attribute_id):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return attribute_value_ctype.value.decode(self._encoding)
 
-    def _open_installed_devices_session(self, driver):
+    def _open_installed_devices_session(self, driver: str) -> typing.Tuple[int, int]:
```
bkeryan, line 354: `typing.Tuple`, `typing.List`, `typing.Dict`, etc. are deprecated in Python 3.9 and later. Use `tuple[T]`, `list[T]`, `dict[K,V]`, etc.

generated/nise/nise/session.py
```diff
@@ -104,7 +106,7 @@ def _get_error_description(self, error_code):
 
     ''' These are code-generated '''
 
-    def _get_error(self, error_description_size=[1024]):
+    def _get_error(self, error_description_size: typing.Iterable[int] = [1024]) -> typing.Tuple[int, str]:
```
bkeryan, line 109: `collections.abc.Iterable[T]`

generated/nise/nise/session.py
```diff
@@ -763,7 +765,7 @@ def _open_session(self, virtual_device_name, option_string=""):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return int(vi_ctype.value)
 
-    def wait_for_debounce(self, maximum_time_ms=hightime.timedelta(milliseconds=-1)):
+    def wait_for_debounce(self, maximum_time_ms: typing.Union['hightime.timedelta', 'datetime.timedelta', int] = hightime.timedelta(milliseconds=-1)) -> None:
```
bkeryan, line 768: FYI, if you use `from __future__ import annotations`, you can write this as `maximum_time_ms: hightime.timedelta | datetime.timedelta | int` even though Python 3.9 does not support the new union syntax.

Issue #1073, opened 2019-10-30
Need nidigital examples


sbethur:
* Description of issue

Need nidigital examples

rriola:
Hi nimi-python Team,

Are there any updates on this? 
We'll be looking forward to your most favorable response, and Happy New Year!

sbethur:
We are currently finalizing [nidigital API](https://github.com/ni/nimi-python/milestone/20). After that's done, we'll be adding examples, which is tied to [nidigital 1.0](https://github.com/ni/nimi-python/milestone/22) milestone.

As we're working on the API, our goal is to make it as user-friendly and "Pythonic" as possible. As a result, we are envisioning adding examples only for specific parts of the API, where they are warranted.

For now, you can refer to the [system tests](https://github.com/ni/nimi-python/tree/master/src/nidigital/system_tests) we've written and continue to write. Something like the [fetch_capture_waveform test](https://github.com/ni/nimi-python/blob/3e5b0e12c0d6070c36a4c142e7544cb84095f365/src/nidigital/system_tests/test_system_nidigital.py#L182) performs most of the core digital operations (load pattern, load pin map and other files; configure source and capture waveforms; and burst a pattern). It would be awesome if you can start using the API and provide feedback on API usability and need for examples.

PR #1544, opened 2021-01-13
Create nidigital_burst_with_start_trigger example

ni-jxie:

- [x ] I've added tests applicable for this pull request.

Add nidigital_burst_with_start_trigger.py example and its accompanying DPE generated files to show how to use an external start trigger to burst a digital pattern 

- #1073

System test of example passed with simulated device.


ni-jxie:
- The LV example has a case structure to select between no start trigger, software start trigger, and external start trigger. To simplify the example, it only supports external start trigger now.
- nidigital was the only MI driver where the example has built-in support for simulation. So I built that into the Python example as well.
- Looking at other MI driver Python example, there is very little comment. I put a bit more comments compare to other examples, not sure what the expectation is.

sbethur:
ok to test

src/nidigital/examples/burst_with_start_trigger_files/burst_with_start_trigger.digiproj
```diff
@@ -0,0 +1,38 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sbethur, line 1: I tried to open the project in DPE and got this error:
image

sbethur: I suspect you were trying to follow the snake_case for file names used in `do_nothing` example (which BTW should be deleted since we now have real examples soon), but I'm starting to think that's a mistake. I think we're better off following the file name format provided by DPE, for support files. The actual example script file can still use snake_case.

sbethur:
test this please

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,69 @@
...
+        # Apply the settings from the levels and timing files we just loaded on the instrument.
+        session.apply_levels_and_timing(levels_filename, timing_filename)
+
+        # Configure the session by loading the pattern file (.digipat) created in the Digital Pattern Editor on the instrument.
+        pattern_filename = os.path.join(dir, 'burst_with_start_trigger_files', 'pattern.digipat')
+        session.load_pattern(pattern_filename)
+
+        # Specify a source and edge for the external start trigger
+        session.digital_edge_start_trigger_source = source
+        session.digital_edge_start_trigger_edge = nidigital.DigitalEdge.RISING if rising_edge else nidigital.DigitalEdge.FALLING
```
sbethur, line 34: Ideally, we want the examples to run "out of the box". With digital edge triggering, the example would hang waiting for the trigger, right?

sbethur (changes requested):
(Not fully done with the review)

sbethur:
test this please

src/nidigital/examples/burst_with_start_trigger_files/burst_with_start_trigger.digiproj
ni-jxie: Yeah, I renamed the file to follow the snake_case because everything else seems to follow the snake_case.

If I stick with the original name for the project, I have to fix something in the build process. Because it cannot deal with file names with spaces in it.

sbethur:
test this please

src/nidigital/examples/nidigital_burst_with_start_trigger.py
ni-jxie: If the customer is using a real device, then yes, it would wait for a trigger. But with simulation enabled by default, it would pass on simulated device.

But if that is a concern, we can convert the example into using a software start trigger instead.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,69 @@
...
+        # Waiting for the trigger to start bursting and then blocks until the pattern is done bursting
+        session.burst_pattern('new_pattern')
+
+        # Disconnect all channels using programmable onboard switching
+        session.selected_function = nidigital.SelectedFunction.DISCONNECT
+
+
+def _main(argsv):
+    parser = argparse.ArgumentParser(description='Demonstrates how to create and configure an instrument session and to burst a pattern on the digital pattern instrument using a start trigger.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-n', '--resource-name', default='PXI1Slot2,PXI1Slot3', help='Resource name of a National Instruments digital pattern instrument')
```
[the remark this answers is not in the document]
ni-jxie: Done

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,69 @@
...
+def main():
+    _main(sys.argv[1:])
+
+
+def test_main():
+    _main([])
+
+
+def test_example():
+    options = {'simulate': True, 'driver_setup': {'Model': '6570'}, }
```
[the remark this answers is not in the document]
ni-jxie: Done

sbethur:
test this please

sbethur:
test this please

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,69 @@
...
+        # Configure the session by loading the pattern file (.digipat) created in the Digital Pattern Editor on the instrument.
+        pattern_filename = os.path.join(dir, 'burst_with_start_trigger_files', 'Pattern.digipat')
+        session.load_pattern(pattern_filename)
+
+        # Specify a source and edge for the external start trigger
+        session.digital_edge_start_trigger_source = source
+        session.digital_edge_start_trigger_edge = nidigital.DigitalEdge.RISING if rising_edge else nidigital.DigitalEdge.FALLING
+
+        # Waiting for the trigger to start bursting and then blocks until the pattern is done bursting
+        session.burst_pattern('new_pattern')
```
sbethur, line 37: `start_label='new_pattern'`

(Makes it more readable)

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # Disconnect all channels using programmable onboard switching
+        session.selected_function = nidigital.SelectedFunction.DISCONNECT
+
```
sbethur, line 41: When the example is run there is no indication of success/fail. Consider logging a couple of lines to the console.

sbethur (changes requested):

src/nidigital/examples/nidigital_burst_with_start_trigger.py
sbethur: I like the "out of the box" experience to be good on both real and simulated h/w. At the same time, I also like demonstrating how to configure and use external triggering. You could do what we do in LabVIEW and .NET examples - allow both (and also no-trigger case) via input args.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,69 @@
...
+
+def _main(argsv):
+    parser = argparse.ArgumentParser(description='Demonstrates how to create and configure an instrument session and to burst a pattern on the digital pattern instrument using a start trigger.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-n', '--resource-name', default='PXI1Slot2,PXI1Slot3', help='Resource name of a NI digital pattern instrument')
+    parser.add_argument('-s', '--simulated', default=True, type=bool, help='Whether to run on hardware or run on software simulation')
+    parser.add_argument('-src', '--source', default='/PXI1Slot2/PXI_Trig0', help='Source terminal where the external signal is connected')
+    parser.add_argument('-re', '--rising-edge', default=True, type=bool, help='Trigger on rising edge or falling edge of the signal')
```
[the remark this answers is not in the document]
ni-jxie: What do you mean? Do you mean --trigger-edge {rising|falling} or just convert the bool into choices of {True|False}?

src/nidigital/examples/nidigital_burst_with_start_trigger.py
ni-jxie: Okay, I opt for a simpler example but yeah, I can add input args to switch between start trigger and no start trigger.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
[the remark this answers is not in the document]
sbethur: `--trigger-edge {rising|falling}`

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Disconnect all channels using programmable onboard switching
+        session.selected_function = nidigital.SelectedFunction.DISCONNECT
+    print('Done bursting pattern')
+
+
+def _main(argsv):
+    parser = argparse.ArgumentParser(description='Demonstrates how to create and configure an instrument session and to burst a pattern on the digital pattern instrument using a start trigger.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-n', '--resource-name', default='PXI1Slot2,PXI1Slot3', help='Resource name of a NI digital pattern instrument. Ensure the resource name matches the instrument name in the pinmap file.')
+    parser.add_argument('-s', '--simulate', default=True, type=bool, help='Whether to run on hardware or run on software simulation')
```
sbethur, line 53: I don't think this will allow passing `-s False`, with the way it's currently coded. Using `bool` type in argparse is [tricky](https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse). There are some workarounds to make it work if we want default to be `False`. But here the default is 'True'.

I think the best option we have is to use `str` type: `'-s', '--simulate', default=True, choices=['True', 'False'],`

If we go with `choices=['True', 'False']`, you might want to switch the order in `help` to `Run on simulated hardware or real hardware`.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,86 @@
...
+    subparser = parser.add_subparsers(dest='command', help='Sub-command help')
+    start_trigger = subparser.add_parser('start-trigger', help='Configure start trigger')
+    start_trigger.add_argument('-src', '--trigger-source', default='/PXI1Slot2/PXI_Trig0', help='Source terminal for the start trigger')
+    start_trigger.add_argument('-edge', '--trigger-edge', default='Rising', choices=['Rising', 'Falling'], help='Trigger on rising edge or falling edge of start trigger')
```
sbethur, line 57: The convention is for short args to be one or two chars. So maybe `-ts` and `te`?

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,86 @@
...
+        pattern_filename = os.path.join(dir, 'Pattern.digipat')
+        session.load_pattern(pattern_filename)
+
+        if trigger_source is not None:
+            # Specify a source and edge for the external start trigger
+            session.start_trigger_type = nidigital.TriggerType.DIGITAL_EDGE
+            session.digital_edge_start_trigger_source = trigger_source
+            session.digital_edge_start_trigger_edge = nidigital.DigitalEdge.RISING if trigger_edge == 'Rising' else nidigital.DigitalEdge.FALLING
+
+        # Waiting for the trigger to start bursting and then blocks until the pattern is done bursting
```
sbethur, line 38: Prefix `If Start trigger is not None, `

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,86 @@
...
+        session.burst_pattern(start_label='new_pattern')
+        if trigger_source is None:
+            print('Start bursting pattern')
+        else:
+            print('Waiting for start trigger and then start bursting pattern')
```
sbethur, line 43: `Wait for start trigger and then start bursting pattern`

sbethur (approved):

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -56,13 +69,17 @@ def main():
...
 def test_example():
+    resource_name = 'PXI1Slot2,PXI1Slot3'
     options = {'simulate': True, 'driver_setup': {'Model': '6571'}, }
-    source = '/PXI1Slot2/PXI_Trig0'
-    rising_edge = True
-    example('PXI1Slot2,PXI1Slot3', options, source, rising_edge)
+    example(resource_name, options)
+
+    trigger_source = '/PXI1Slot2/PXI_Trig0'
+    trigger_edge = True
```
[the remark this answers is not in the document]
ni-jxie: Fixed.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
ni-jxie: Interesting, if the argument type is bool, it is actually expecting a bool object. Weird.

Fixed.

marcoskirsch:
`src/nidigital/examples/burst_with_start_trigger_files/PinLevels.digilevels`

Does this layout make sense? I think we originally had all our examples directly under `examples/` with the assumption that examples only needed one file. But it may be better to reorganize:

`src/nifoo/examples/do_foo/do_foo.py`
`src/nifoo/examples/do_foo/<other files>`

or something like that. What do you think?

src/nidigital/examples/burst_with_start_trigger_files/burst_with_start_trigger.digiproj
marcoskirsch: >The actual example script file can still use snake_case.

Here you left the `.digiproj` in snake case form. Is that because the original name had spaces?
Everything is in a folder and the folder has no spaces so I'd assume we can copy that fine at build time.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,87 @@
...
+        if trigger_source is not None:
+            # Specify a source and edge for the external start trigger
+            session.start_trigger_type = nidigital.TriggerType.DIGITAL_EDGE
+            session.digital_edge_start_trigger_source = trigger_source
+            session.digital_edge_start_trigger_edge = nidigital.DigitalEdge.RISING if trigger_edge == 'Rising' else nidigital.DigitalEdge.FALLING
+
+        # If start trigger is configured, waiting for the trigger to start bursting and then blocks until the pattern is done bursting
+        # Else just start bursting and block until the pattern is done bursting
+        session.burst_pattern(start_label='new_pattern')
+        if trigger_source is None:
```
marcoskirsch, line 41: seems like you can combine this if/else with the one in line 32

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,87 @@
...
+            print('Start bursting pattern')
+        else:
+            print('Wait for start trigger and then start bursting pattern')
+
+        # Disconnect all channels using programmable onboard switching
+        session.selected_function = nidigital.SelectedFunction.DISCONNECT
```
marcoskirsch, line 47: I am no NI-Digital expert so bear with me...
```
        session.burst_pattern(start_label='new_pattern')
```
This is what makes the hardware begin generation.
But we don't wait for the pattern to finish before closing the session and printing `Done bursting pattern`. Shouldn't we wait?

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,87 @@
...
+    print('Done bursting pattern')
+
+
+def _main(argsv):
+    parser = argparse.ArgumentParser(description='Demonstrates how to create and configure an instrument session and to burst a pattern on the digital pattern instrument using a start trigger.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
marcoskirsch, line 52: > Demonstrates how to create and configure a session and to burst a pattern on the digital pattern instrument using a start trigger

(remove instrument, redundant)

src/nidigital/examples/nidigital_burst_with_start_trigger.py
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_argument('-n', '--resource-name', default='PXI1Slot2,PXI1Slot3', help='Resource name of a NI digital pattern instrument. Ensure the resource name matches the instrument name in the pinmap file.')
```
marcoskirsch, line 53: why default to multi-instrument? just to show it's possible?

marcoskirsch (changes requested):

ni-jxie:
> `src/nidigital/examples/burst_with_start_trigger_files/PinLevels.digilevels`
> 
> Does this layout make sense? I think we originally had all our examples directly under `examples/` with the assumption that examples only needed one file. But it may be better to reorganize:
> 
> `src/nifoo/examples/do_foo/do_foo.py`
> `src/nifoo/examples/do_foo/<other files>`
> 
> or something like that. What do you think?

I don't have a strong preference either way. I was mimicking the existing layout for the do_nothing example. @sbethur do you have a preference?

src/nidigital/examples/nidigital_burst_with_start_trigger.py
ni-jxie: [Link to burst_pattern documentation](https://nimi-python.readthedocs.io/en/master/nidigital/class.html#nidigital.Session.burst_pattern)

There is an optional parameter wait_until_done that is set to true by default and will block until the pattern is done.

src/nidigital/examples/burst_with_start_trigger_files/burst_with_start_trigger.digiproj
ni-jxie: I left the digiproj in snake case in part to match the name of the example and in part because a makefile in the build process is not friendly to file names with spaces in it.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
ni-jxie: @sbethur requested that the Python examples show usage multiple instruments to show off channel expansion just like their counterparts in LV and .NET.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
ni-jxie: Done.

src/nidigital/examples/nidigital_burst_with_start_trigger.py
ni-jxie: Combined

sbethur:
> @marcoskirsch I don't have a strong preference either way. I was mimicking the existing layout for the do_nothing example. @sbethur do you have a preference?

I don't see a reason why we need another folder for the support files. I like this layout:
```
src/nifoo/examples/do_foo/do_foo.py
src/nifoo/examples/do_foo/<other files>
```

Let's go with this.

ni-jxie:
> > @marcoskirsch I don't have a strong preference either way. I was mimicking the existing layout for the do_nothing example. @sbethur do you have a preference?
> 
> I don't see a reason why we need another folder for the support files. I like this layout:
> 
> ```
> src/nifoo/examples/do_foo/do_foo.py
> src/nifoo/examples/do_foo/<other files>
> ```
> 
> Let's go with this.

Okay, so instead of the current layout of 

src/nifoo/examples/do_foo.py
src/nifoo/examples/do_foo_files/<other files>

we want to go with
src/nifoo/examples/do_foo/<do_foo.py + other files>

I will update all the examples to follow this layout.

ni-jxie:
@marcoskirsch @sbethur The directory layout has been updated. Please take a look at this PR.

sbethur:
> @marcoskirsch @sbethur The directory layout has been updated. Please take a look at this PR.

LGTM. @marcoskirsch, do you plan to review this PR?

docs/nidigital/examples.rst
```diff
@@ -3,6 +3,15 @@ Examples
 
 `You can download all nidigital examples for latest version here <https://github.com/ni/nimi-python/releases/download/1.3.2/nidigital_examples.zip>`_
 
+nidigital_burst_with_start_trigger.py
+-------------------------------------
+
+.. literalinclude:: ../../src/nidigital/examples/nidigital_burst_with_start_trigger.py
```
marcoskirsch, line 9: This doesn't match the location of the file, right?
`src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger.py`

marcoskirsch (changes requested):

docs/nidigital/examples.rst
ni-jxie: Yeah, you are right. examples.rst does not get regenerated unless a "tox -e clean" is performed. Fixed the generation mako to recursively look through all the folders under examples.

marcoskirsch (approved):

PR #1581, opened 2021-04-14
Improve error message when function not found in driver runtime

### LINKED ISSUE #1585: Terrible error when nimi-python module tries to call a function not supported by driver runtime
### Description of issue

If the system has an old driver runtime installed and a newer nimi-python module, and the client tries to use a method that was added later to the driver runtime, then the error is not helpful: You get an `AttributeError` with no indication of what the problem is.

We should improve the error so that clear guidance is given to the user for correcting it.

ni-erikcrank:
- [ ] ~I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~
- [ ] ~I've added tests applicable for this pull request~

When a required function (e.g. CommitWithChannels) is missing from the driver runtime, nimi-python adds an error description that is more descriptive than "function not found" to instruct the user to upgrade to a newer version of the driver.

- Fix #1585 

Called abort and commit methods on a system with NI-DCPower 20.5 instrument driver and confirmed that the error message is displayed in the exception information at runtime.

### DIFF
--- a/build/templates/_library.py.mako
+++ b/build/templates/_library.py.mako
@@ -15,6 +15,7 @@ functions = helper.filter_library_functions(functions)
 %>\
 
 import ctypes
+import ${module_name}.errors as errors
 import threading
 
 from ${module_name}._visatype import *  # noqa: F403,H303
@@ -38,6 +39,13 @@ class Library(object):
 % for func_name in sorted(functions):
         self.${c_function_prefix}${func_name}_cfunc = None
 % endfor
+
+    def _get_library_function(self, name):
+        try:
+            function = getattr(self._library, name)
+        except AttributeError as e:
+            raise errors.DriverTooOldError() from e
+        return function
 % for func_name in sorted(functions):
 <%
     f = functions[func_name]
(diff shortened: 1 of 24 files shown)


build/templates/_library.py.mako
```diff
@@ -51,7 +51,12 @@ class Library(object):
     def ${c_func_name}(${param_names_method}):  # noqa: N802
         with self._func_lock:
             if self.${c_func_name}_cfunc is None:
-                self.${c_func_name}_cfunc = self._library.${c_func_name}
+                try:
+                    self.${c_func_name}_cfunc = self._library.${c_func_name}
+                except AttributeError as e:
+                    raise AttributeError("A required function was not found in the instrument driver DLL. This might "
+                                         "be an indication that the version of the instrument driver is too old for "
+                                         "this version of the python API. Upgrade your instrument driver.") from e
```
ni-erikcrank, line 59: Please provide the preferred message text.

sbethur:
ok to test

marcoskirsch:
We use the term "driver runtime" to refer to the driver API called by nimi-python, as opposed to synonyms like "instrument driver DLL". Please correct description and other instances to be consistent.

marcoskirsch:
Please referrence #1585 in Description.

build/templates/_library.py.mako
marcoskirsch: I think you have enough context in the codegenerator to give a really specific error. Here's a suggestion of what that should look like based on the error string in `DriverNotInstalledError`:

> Function `niDCPower_InitializeWithChannels` was not found in the NI-DCPower runtime. Please visit http://www.ni.com/downloads/drivers/ to download a newer version and install it.' 

An additional question: do we want a `DriverTooOld` class for this or is `AttributeError` better?

marcoskirsch (changes requested):

marcoskirsch:
@sbethur @ni-erikcrank we should be able to unit test in `nifake` by mocking ctypes, no?

build/templates/_library.py.mako
ni-erikcrank: I don't have an opinion on the DriverTooOld class.
Do you know of an existing way to obtain the driver name (e.g. NI-DCPower) for this message?

sbethur:
test this please

build/templates/_library.py.mako
ni-erikcrank: I added a DriverTooOldError class.

ni-erikcrank:
Added a unit test for the new LibraryTooOldError.

sbethur (approved):

marcoskirsch (approved):

PR #1599, opened 2021-05-07
US1198692: Instrument-based repeated capabilities for attributes

luisgomes252:

1. Add support for instrument-based repeated capabilities in attributes
2. Rename the instrument-based capability from 'devices' to 'instruments'
3. Update metadata properties to be more consistent across the internal code generation tools.

Unit test for internal function, plus visual inspection of code-generated API elements


sbethur:
FYI - @luisgomes252, There are merge conflicts.

sbethur:
ok to test

sbethur:
test this please

src/nidcpower/metadata/attributes.py
```diff
@@ -36,7 +33,6 @@
     },
     1050203: {
         'access': 'read only',
-        'channel_based': False,
         'documentation': {
             'description': 'Indicates the number of channels that NI-DCPower supports for the instrument that was chosen when  the current session was opened. For channel-based attributes, the IVI engine maintains a separate  cache value for each channel.'
```
sbethur, line 37: This doesn't support `instruments` rep cap?

More broadly, are the supported rep cap changes already reviewed in internal metadata?

build/helper/documentation_helper.py
```diff
@@ -872,6 +872,19 @@ def add_notes_re_links(config):
             _check_documentation(v['documentation'], config)
 
 
+def get_attr_repeated_caps(attr):
```
marcoskirsch, line 875: please add docstring

docs/nidcpower/rep_caps.rst
```diff
@@ -35,4 +35,16 @@ channels
...
 
 
+instruments
+-----------
+
+    .. py:attribute:: nidcpower.Session.instruments[]
+
+        .. code:: python
+
+            session.instruments['0-2'].channel_enabled = True
```
marcoskirsch, line 45: I think we need to revisit this:

1. '0-2' is not a goo example for `instruments` rep cap
2. `channel_enabled` makes no sense to be set on `instruments`
3. "passes a string of :python:`'0, 1, 2'` to the set attribute function" is irrelevant to Python programmers.

I do believe it should be done separately. Once this is merged please open an issue about this documentation problem.

src/nidcpower/metadata/attributes.py
```diff
@@ -80,7 +73,6 @@
     },
     1050327: {
         'access': 'read only',
-        'channel_based': False,
```
marcoskirsch, line 83: Is the nimi-python metadata expansion defaulting lack of `supported_rep_caps` to mean `channels`?
I think that's a bad idea. We should make it so that the nimi-python metadata is explicit. 
Lack of `supported_rep_caps` should mean "None".

marcoskirsch (changes requested):

build/helper/documentation_helper.py
```diff
@@ -123,7 +123,7 @@ def get_rst_admonition_snippet(node, admonition, d, config, indent=0):
 
 def add_attribute_rep_cap_tip_rst(attr, config):
     '''Add the appropriate (r/w/rw/none) rst formatted tip for an attribute'''
-    if 'repeated_capability_type' in attr:
```
sbethur, line 126: This key is used in [nidigital and nifgen](https://github.com/search?q=repeated_capability_type+repo%3Ani%2Fnimi-python+filename%3Aattributes.py&type=Code&ref=advsearch&l=&l=). I'm surprised changing this didn't update their generated code and docs.

build/helper/documentation_helper.py
```diff
@@ -195,7 +195,7 @@ def get_docstring_admonition_snippet(node, admonition, d, config, indent=0, extr
 
 def add_attribute_rep_cap_tip_docstring(attr, config):
     '''Add the appropriate (r/w/rw/none) docstring formatted tip for an attribute'''
```
sbethur, line 197: Can you please fix this existing mistake - the method docstring is not accurate.

src/nidcpower/metadata/attributes.py
sbethur: > Is the nimi-python metadata expansion defaulting lack of supported_rep_caps to mean channels?

I don't think this is true, but I'll let Luis confirm.

build/helper/documentation_helper.py
```diff
@@ -123,7 +123,7 @@ def get_rst_admonition_snippet(node, admonition, d, config, indent=0):
 
 def add_attribute_rep_cap_tip_rst(attr, config):
     '''Add the appropriate (r/w/rw/none) rst formatted tip for an attribute'''
-    if 'repeated_capability_type' in attr:
+    if 'supported_rep_caps' in attr:
         if 'documentation' not in attr:
             attr['documentation'] = {}
```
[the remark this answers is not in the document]
luisgomes252: Good idea. I've done that now.

build/helper/documentation_helper.py
```diff
@@ -872,6 +872,19 @@ def add_notes_re_links(config):
             _check_documentation(v['documentation'], config)
 
 
+def get_attr_repeated_caps(attr):
+    if 'supported_rep_caps' in attr and len(attr['supported_rep_caps']) > 0:
+        caps = attr['supported_rep_caps'][0]
```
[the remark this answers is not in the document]
luisgomes252: Yeah, you're right of course. I'll change it.

src/nidcpower/metadata/attributes.py
luisgomes252: Shreyas is correct. That's not true. The lack of `supported_rep_caps` means "None". That's why all those `'channel_based': False` attributes simply disappeared without having to be replaced by anything.

docs/nidcpower/rep_caps.rst
luisgomes252: #1602

build/helper/documentation_helper.py
luisgomes252: It did. I manually removed the generated files of the other drivers from the commit.

Even though I've already updated the metadata for the other drivers, I didn't include the output from the metadata exports of the other drivers (e.g. attributes.py) as inputs to the code generation in this repo, and so the code generated files would have been incorrect.

sbethur: Ok, Please have a work item in your backlog to update the metadata and generated files for those drivers.

sbethur:
test this please

sbethur (changes requested):
I don't have any further comments. Once you address all my comments, please re-request review.

sbethur:
Fix for the Travis CI failure: https://github.com/ni/nimi-python/pull/1603

sbethur:
test this please

generated/nidcpower/nidcpower/session.py
```diff
@@ -132,7 +132,7 @@ class _SessionBase(object):
     Note: This property is not supported by all devices. Refer to Supported Properties by Device topic.
 
     Tip:
-    This property can use repeated capabilities. If set or get directly on the
+    This property can use repeated capabilities for channels. If set or get directly on the
     nidcpower.Session object, then the set/get will use all repeated capabilities in the session.
     You can specify a subset of repeated capabilities using the Python index notation on an
     nidcpower.Session repeated capabilities container, and calling set/get value on the result.
```
marcoskirsch, line 138: This text is pretty old and not great. Here's a stab at improving it to better reflect the state of the art in our Python APIs and to be clearer to customers.

```
This property can be set/get to specific channels within you nidcpower.Session object. Use Python index notation on the repeated capabilities container `channels` to specify a subset.
Example: `my_session.channels[ ... ].active_advanced_sequence_step`,
To set/get on all, you can call the property directly on the nidcpower.Session:
`my_session.active_advanced_sequence_step`.
```

marcoskirsch: #1604

marcoskirsch (changes requested):

build/helper/documentation_helper.py
marcoskirsch: Does this mean that should someone clone the repo and run codegen would see generated files that are different from what was committed? That's not a state we want to leave the repo in. (unresolved until clarification)

generated/nidcpower/nidcpower/session.py
luisgomes252: 1. @marcoskirsch, does the fact that you created an issue imply that you'd rather I not switch to using your text in this PR?
2. The current text is parameterized, so that the "...for channels" part, in the first line, can change with the particular capability type(s). Does your new text also allow for that? Specifically, if I were to replace "... to specific **channels**" with, say, "... to specific **channels, instruments and pins**", would that still be okay?
3. As for the example in the second line, I could replace it with the first item in the types list (allowing for cases where an attribute is valid for more than one type), since I can't use multiple types in the example. But I suppose I'd also have to pick an attribute that makes sense for that capability type. That's going to make it a bit harder to parameterize this text, as @sbethur [requested](https://github.com/ni/nimi-python/pull/1599#discussion_r630524419) earlier in this PR.

build/helper/documentation_helper.py
```diff
@@ -1297,6 +1312,28 @@ def test_get_function_rst_numpy():
     assert_rst_strings_are_equal(expected_fuction_rst, actual_function_rst)
 
 
+def test_get_attribute_repeated_caps():
+    attr = {'supported_rep_caps': ['channels', 'instruments', 'pins']}
+    expected_caps = 'channels, instruments, pins'
```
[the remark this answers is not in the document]
luisgomes252: We were. I've added one now.

build/helper/documentation_helper.py
luisgomes252: Yes, @marcoskirsch, you are correct. Your scenario would happen as you described.

I left out the digital, dmm, fgen, modinst, scope, switch and tclk files from this PR because I assumed we wouldn't want to include diffs in this PR belonging to drivers, unrelated to the issue being addressed by this PR, and also because the internal nidigital build has been broken for a while (not sure it if still is) and so it wouldn't have been possible for me to use the metadata changes that I made for those drivers in the codegen build for this PR. Until that happens, I can't generate correct versions of at least the digital and scope files.

So now I see 3 possibilities. Let me know which one you prefer:

1. Block this PR until the nidigital build is fixed, and then include all the other updated driver files as part of this PR
2. Submit this PR as is, wait for the nidigital build to be fixed, and then immediately implement the user story that @sbethur ask that I create, which captured the task of updating those other files. This will result in the repo having latent, pending changes as you pointed out, at least for a couple of days, until the next PR goes through.
3. Include in this PR the version of the files that are currently being generated, which are incorrect for digital and scope. Similarly to item 2, this would be fixed in a few days. It's only the generated documentation that is incorrect, so maybe that's not so terrible.

I had implicitly chosen option 2, based on the fact that changes have gone into this repo in past that consisted of doing direct edits on the generated files, which also triggers the same scenario that you described. And in this case, we'd be in that state for a shorter amount of time. But if you guys prefer one of the other options, or know of a different option, let me know, and I'll be happy to comply.

marcoskirsch: > based on the fact that changes have gone into this repo in past that consisted of doing direct edits on the generated files

Really? That is pretty bad. Or are you talking about the metadata files generated internally at NI and uploaded to GitHub?

The problem with committing like this is that the files will show up as noise to every other developer.
@sbethur discussed and we think option 3 is best. Documentation will be temporarily wrong but is possibly the lesser of 3 evils.

generated/nidcpower/nidcpower/session.py
marcoskirsch: 1. own PR, that's why I created the issue to track
2. yes
3. 👍🏻

build/helper/documentation_helper.py
luisgomes252: Sorry, yes, I was referring to editing the metadata files (functions.py, etc...), rather than changing their source hapigen files. So I guess cloning the repo and rebuilding wouldn't have resulted in changes. Sorry for conflating the two.

sbethur:
test this please

docs/nitclk/class.rst
```diff
@@ -461,17 +461,17 @@ exported_sync_pulse_output_terminal
...
-            | Resettable     | No         |
-            +----------------+------------+
+            +-----------------------+------------+
+            | Characteristic        | Value      |
+            +=======================+============+
+            | Datatype              | str        |
+            +-----------------------+------------+
+            | Permissions           | read-write |
+            +-----------------------+------------+
+            | Repeated Capabilities | None       |
```
marcoskirsch, line 471: There is no concept of Repeated Capabilities in `nitclk`.
Or `nimodinst` for that matter.
NI-DMM has the concept but it isn't applicable to the `nidmm` Python module.

I'm thinking we might be better off not even displaying this row at all for APIs without repeated capabilities.

I think Resettable, while a valid thing, should not be here either because the Python APIs have not exposed it (most driver runtimes don't implement it I believe)

#1606

marcoskirsch (approved):

sbethur (approved):

PR #1578, opened 2021-04-07
Us1198694/nidcpower get channel names

luisgomes252:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

TODO: Check the above box with an 'x' indicating you've read and followed [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

- [x] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.

TODO: Check the above box with an 'x' if considered if there were any and then documented client facing changes in [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md). Strike through if this is not a relevant client facing change.

- [x] I've added tests applicable for this pull request

TODO: Check the above box with an 'x' indicating you have considered and added any applicable system or unit tests. Strike through if you considered and decided additional tests were not warranted.

### What does this Pull Request accomplish?

* Add get_channel_names()
* Remove get_channel_name() and get_channel_name_from_string()
* Update generated code from earlier, unrelated metadata changes:
  * Remove attribute IDs from attribute names
  * Add ouput cutoff functions, attributes and enums

### List issues fixed by this Pull Request below, if any.

#1386 (for NI-DCPower)

### What testing has been done?

Added new system test and ran all unit tests. System tests will be run by CI.

PR #1601, opened 2021-05-10
Use explicit metadata attribute to avoid adding repeated capability

luisgomes252:

Uses an explicit metadata attribute (`is_repeated_capability`) to avoid adding repeated capability in a parameter for which its `name` would otherwise require it, but for which its `python_name` might not (e.g. get_channel_names)

This is a (lower priority) continuation of #1578 

Ensured unit test passes. Performed code generation locally to verify that the repeated capability is correctly removed from the `get_channel_names` function.

### DIFF
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -691,6 +691,8 @@ functions_input = {
                 'direction': 'in',
                 'enum': None,
                 'name': 'channelName',
+                'python_name': 'name',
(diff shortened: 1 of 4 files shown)


sbethur:
ok to test

marcoskirsch (approved):

marcoskirsch:
This will be added to internal NI metadata, right @luisgomes252 ?
Please note CI failed.

Issue #1627, opened 2021-06-11
nifgen API reference does not mention which properties and methods can be called on a repeated capabilities container


sbethur:
### Description of issue

According to the documentation, [these ](https://nimi-python.readthedocs.io/en/1.3.3/nifgen/rep_caps.html#)repeated capabilities are supported in nifgen. But the API reference for properties and methods doesn't mention which of them can be called on a repeated capabilities container.

sbethur:
@ni-jfitzger, I remember you looking into this issue; can you please provide the latest status on this?

ni-jfitzger:
The PR lined above addressed this issue for properties, but not methods.
But looking at [the latest documentation](https://nimi-python.readthedocs.io/en/master/nifgen/class.html#methods), several methods have tips that mention they can be called on specific channels and show the syntax. I'm not sure if we have any methods that support any repeated capabilities other than channels. Maybe this can be closed?

marcoskirsch:
Maybe not in `nifgen`, but in `nidigital` we do have functions that take in pin names.

Issue #1351, opened 2020-03-25
Redundant metadata about repeated capabilities


marcoskirsch:
### Description of issue

We have parameter metadata that looks like:
```
                'is_repeated_capability': True,
                 'repeated_capability_type': 'sites',
```

We should eliminate `is_repeated_capability` and simply set `repeated_capability_type` to None when applicable.

This might alos apply to NI's internal metadata.

Issue #1338, opened 2020-03-20
Repeated capabilities "tip" is not useful to mere mortals


marcoskirsch:
Whenever there is a method/property that supports repeated capabilities we add the following text to the documentation:

>This property can use repeated capabilities. If set or get directly on the nidigital.Session object, then the set/get will use all repeated capabilities in the session. You can specify a subset of repeated capabilities using the Python index notation on an niscope.Session repeated capabilities container, and calling set/get value on the result.

>This method requires repeated capabilities. If called directly on the niscope.Session object, then the method will use all repeated capabilities in the session. You can specify a subset of repeated capabilities using the Python index notation on an niscope.Session repeated capabilities container, and calling this method on the result.

* We don't say which repeated capability (`channels`? `instruments`? `script_triggers`? `sites`?) applies
* We don't give usage example

ni-jfitzger:
@marcoskirsch is #1610 not sufficient to address this?

It seems to me like it addressed the 2 concerns you mentioned for attributes and methods.
The examples given leave the iterable to the user's imagination, but an example is provided.

>This property can be set/get on specific channels or pins within your nidigital.Session instance. Use Python index notation on the repeated capabilities container channels or pins to specify a subset.
>
>Example: my_session.channels[ ... ].active_load_ioh
>
>To set/get on all channels or pins, you can call the property directly on the nidigital.Session.
>
>Example: my_session.active_load_ioh

>This method can be called on specific channels within your niscope.Session instance. Use Python index notation on the repeated capabilities container channels to specify a subset, and then call this method on the result.
>
>Example: my_session.channels[ ... ].add_waveform_processing()
>
>To call the method on all channels, you can call it directly on the niscope.Session.
>
>Example: my_session.add_waveform_processing()

PR #1681, opened 2022-01-10
nifgen repeated capability documentation and test updates 

### LINKED ISSUE #1668: NI-FGEN API has no repeated capability for Data Markers
### Description of issue
[Marker Events](https://zone.ni.com/reference/en-XX/help/370524V-01/siggenhelp/ni_5421_markers/) and [Data Marker Events](https://zone.ni.com/reference/en-XX/help/370524V-01/siggenhelp/events_data_markers/) have different repeated capabilities that they are accessible through.

Markers are accessible through 'MarkerN' (where N is a decimal) repeated capabilities.
Data Markers are accessible through 'DataMarkerN' repeated capabilities.

The Python API currently provides no DataMarker repeated capability, so the only way to access data marker attributes is with the channels repeated capability.

This happened because [markers and data markers were conflated in the metadata.](https://github.com/ni/nimi-python/blob/master/src/nifgen/metadata/attributes.py#L723) 'markers' is listed as the repeated capability type for data marker event attributes.

### System report

ni-fgen package release: 1.4.1.
System info isn't really relevant (in my opinion).

### Steps to reproduce issue
Example 1 (errors)
```python
with nifgen.Session("Dev1", options= {'simulate': True, 'driver_setup': {'Model': '5441', 'BoardType': 'PXI'}}) as session:
    session.markers[0].data_marker_event_data_bit_number=0
```
Example 2 (errors)
```python
with nifgen.Session("Dev1", options= {'simulate': True, 'driver_setup': {'Model': '5441', 'BoardType': 'PXI'}}) as session:
    session.data_markers[0].data_marker_event_data_bit_number=0
```
Example 3 (errors)
```python
with nifgen.Session("Dev1", options= {'simulate': True, 'driver_setup': {'Model': '5441', 'BoardType': 'PXI'}}) as session:
    session.data_marker_event_data_bit_number=0
```
Example 4 (no error)
```python
with nifgen.Session("Dev1", options= {'simulate': True, 'driver_setup': {'Model': '5441', 'BoardType': 'PXI'}}) as session:
    session.channels['DataMarker0'].data_marker_event_data_bit_number=0
```

ni-jfitzger:

* Updates nifgen metadata to properly document and support accessing of attributes with the following repeated capabilities:
   * channels
   * script_triggers
   * markers
   * data_markers
 * Side effect: removed mention of `CreateWaveformFromFileHWS` from documentation. Support for this entrypoint is removed in NI-FGEN 21.3.
 * Regenerates nifgen docs to display the changes mentioned above
 * Adds nifgen system tests for the repeated capabilities mentioned above

* Fix #1668
* Partially addresses #1627

* Ran tox locally.
* Ran the NI-FGEN systems tests on a local system.

### DIFF
--- a/docs/nifgen/rep_caps.rst
+++ b/docs/nifgen/rep_caps.rst
@@ -87,4 +87,30 @@ markers
         passes a string of :python:`'Marker0, Marker1, Marker2'` to the set attribute function.
 
 
+data_markers
+------------
+
+    .. py:attribute:: nifgen.Session.data_markers[]
+
+        If no prefix is added to the items in the parameter, the correct prefix will be added when
+        the driver function call is made.
+
+        .. code:: python
+
+            session.data_markers['0-2'].channel_enabled = True
+
+        passes a string of :python:`'DataMarker0, DataMarker1, DataMarker2'` to the set attribute function.
+
+        If an invalid repeated capability is passed to the driver, the driver will return an error.
+
+        You can also explicitly use the prefix as part of the parameter, but it must be the correct prefix
+        for the specific repeated capability.
+
+        .. code:: python
+
+            session.data_markers['DataMarker0-DataMarker2'].channel_enabled = True
+
+        passes a string of :python:`'DataMarker0, DataMarker1, DataMarker2'` to the set attribute function.
+
+
 
--- a/src/nifgen/system_tests/test_system_nifgen.py
+++ b/src/nifgen/system_tests/test_system_nifgen.py
@@ -88,6 +88,39 @@ def test_self_cal(session):
     session.self_cal()
 
 
+def test_channels_rep_cap(session):
+    session.func_amplitude = 0.5
+    assert session.channels['0-1'].func_amplitude == 0.5
+
+    session.channels[0].func_amplitude = 1
+    assert session.channels['0'].func_amplitude == 1
+    assert session.channels[1].func_amplitude == 0.5
+
+
+def test_markers_rep_cap(session):
+    assert '' == session.markers['Marker0'].marker_event_output_terminal
+
+    requested_terminal_name = '/Dev1/PXI_Trig0'
+    session.markers['Marker0'].marker_event_output_terminal = requested_terminal_name
+    assert requested_terminal_name == session.markers[0].marker_event_output_terminal
+
+
+def test_data_markers_rep_cap(session):
+    assert '' == session.data_markers['DataMarker0'].marker_event_output_terminal
+
+    requested_terminal_name = '/Dev1/PXI_Trig0'
+    session.data_markers['DataMarker0'].data_marker_event_output_terminal = requested_terminal_name
+    assert requested_terminal_name == session.data_markers[0].data_marker_event_output_terminal
+
+
+def test_script_triggers_rep_cap(session):
+    assert '' == session.script_triggers['ScriptTrigger0'].exported_script_trigger_output_terminal
+
+    requested_terminal_name = '/Dev1/PXI_Trig0'
+    session.script_triggers['ScriptTrigger0'].exported_script_trigger_output_terminal = requested_terminal_name
+    assert requested_terminal_name == session.script_triggers[0].exported_script_trigger_output_terminal
+
+
 def test_standard_waveform(session):
     session.output_mode = nifgen.OutputMode.FUNC
     session.configure_standard_waveform(nifgen.Waveform.SINE, 2.0, 2000000, 1.0, 0.0)
--- a/src/nifgen/system_tests/test_system_nifgen.py
+++ b/src/nifgen/system_tests/test_system_nifgen.py
@@ -88,13 +88,14 @@ def test_self_cal(session):
     session.self_cal()
 
 
-def test_channels_rep_cap(session):
-    session.func_amplitude = 0.5
-    assert session.channels['0-1'].func_amplitude == 0.5
+def test_channels_rep_cap():
+    with nifgen.Session('', '', False, 'Simulate=1, DriverSetup=Model:5433 (2CH);BoardType:PXIe') as session:
+        session.func_amplitude = 0.5
+        assert session.channels['0-1'].func_amplitude == 0.5
 
-    session.channels[0].func_amplitude = 1
-    assert session.channels['0'].func_amplitude == 1
-    assert session.channels[1].func_amplitude == 0.5
+        session.channels[0].func_amplitude = 1
+        assert session.channels['0'].func_amplitude == 1
+        assert session.channels[1].func_amplitude == 0.5
 
 
 def test_markers_rep_cap(session):
@@ -106,11 +107,11 @@ def test_markers_rep_cap(session):
 
 
 def test_data_markers_rep_cap(session):
-    assert '' == session.data_markers['DataMarker0'].marker_event_output_terminal
+    assert nifgen.DataMarkerEventLevelPolarity.HIGH == session.data_markers['DataMarker0'].data_marker_event_level_polarity
 
-    requested_terminal_name = '/Dev1/PXI_Trig0'
-    session.data_markers['DataMarker0'].data_marker_event_output_terminal = requested_terminal_name
-    assert requested_terminal_name == session.data_markers[0].data_marker_event_output_terminal
+    requested_polarity = nifgen.DataMarkerEventLevelPolarity.LOW
+    session.data_markers['DataMarker0'].data_marker_event_level_polarity = requested_polarity
+    assert requested_polarity == session.data_markers[0].data_marker_event_level_polarity
 
 
 def test_script_triggers_rep_cap(session):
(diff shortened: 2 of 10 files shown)


src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -88,6 +88,40 @@ def test_self_cal(session):
...
+    with nifgen.Session('', '', False, 'Simulate=1, DriverSetup=Model:5433 (2CH);BoardType:PXIe') as session:
+        session.func_amplitude = 0.5
+        assert session.channels['0-1'].func_amplitude == 0.5
+
+        session.channels[0].func_amplitude = 1
+        assert session.channels['0'].func_amplitude == 1
+        assert session.channels[1].func_amplitude == 0.5
+
+
+def test_markers_rep_cap(session):
```
ni-jfitzger, line 101: Should I just use verbose strings everywhere (as an example for how good code should be written) in lieu of coverage for more inputs to the repeated capability?
If so, I'll apply the answer to all of the new tests.

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -88,6 +88,40 @@ def test_self_cal(session):
...
+    assert '' == session.markers['Marker0'].marker_event_output_terminal
+
+    requested_terminal_name = '/Dev1/PXI_Trig0'
+    session.markers['Marker0'].marker_event_output_terminal = requested_terminal_name
+    assert requested_terminal_name == session.markers[0].marker_event_output_terminal
```
ni-jfitzger, line 106: Should I just use verbose strings everywhere (as an example for how good code should be written) in lieu of coverage for more inputs to the repeated capability?
If so, I'll apply the answer to all of the new tests.

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -88,6 +88,40 @@ def test_self_cal(session):
...
+
+
+def test_data_markers_rep_cap(session):
+    assert nifgen.DataMarkerEventLevelPolarity.HIGH == session.data_markers['DataMarker0'].data_marker_event_level_polarity
+
+    requested_polarity = nifgen.DataMarkerEventLevelPolarity.LOW
+    session.data_markers['DataMarker0'].data_marker_event_level_polarity = requested_polarity
+    assert requested_polarity == session.data_markers[0].data_marker_event_level_polarity
```
ni-jfitzger, line 114: I originally wrote this test to use `data_marker_event_output_terminal` but reading back the value after setting it to '/Dev1/PXI_Trig0' returned 
```
>           raise DriverError(code, description)
E           nifgen.errors.DriverError: -1074135024: IVI: (Hex 0xBFFA0010) Invalid value for parameter or property.
E
E           Property: Data Marker Event Output Terminal
E           Invalid Value: /Dev1/PXI_Trig0
E           Valid Values:

C:\Program Files\Python36\lib\site-packages\nifgen\errors.py:102: DriverError
```

ni-jfitzger:
Key items to review:
* new tests
* changelog

Metadata changes came from an internal export of nifgen_nimi_python that contained relevant metadata changes implemented and reviewed internally.

Other changes are the result of running `tox -e codegen`.

src/nifgen/system_tests/test_system_nifgen.py
marcoskirsch: Do you mean `["Marker0"]` instead of `[0]`.
No, as a matter of fact, the more pythonic thing is to pass the number or an iterable, as opposed to a fully spelled out string.

I am of the opinion that the spelled-out string shouldn't even be documented behavior.

src/nifgen/system_tests/test_system_nifgen.py
marcoskirsch: It's possible that Munich doesn't support data markers. Can you do a quick experiment with a 5421 on your end?

Keep the system test as-is.

docs/nifgen/rep_caps.rst
```diff
@@ -87,4 +87,30 @@ markers
...
+    .. py:attribute:: nifgen.Session.data_markers[]
+
+        If no prefix is added to the items in the parameter, the correct prefix will be added when
+        the driver function call is made.
+
+        .. code:: python
+
+            session.data_markers['0-2'].channel_enabled = True
+
+        passes a string of :python:`'DataMarker0, DataMarker1, DataMarker2'` to the set attribute function.
```
marcoskirsch, line 102: This is very bad documentation, because it documents what the implementation does under the hood which is irrelevant to the Python programmer.

We should (separate PR naturally) rewrite this to be relevant to the user. @ni-jfitzger  want to tackle it? I can help wordsmith.

marcoskirsch (changes requested):

src/nifgen/system_tests/test_system_nifgen.py
ni-jfitzger: Hmm ... we're definitely spelling out our rep_caps in the nidigital rep_caps tests, which is why I asked.
https://github.com/ni/nimi-python/blob/master/src/nidigital/system_tests/test_system_nidigital.py
Maybe it's different though, because those aren't numbers.

Yeah, if we don't document the spelled-out string, then it should be fine to only test passing the number.
I'll apply this answer to all of the tests.

src/nifgen/system_tests/test_system_nifgen.py
ni-jfitzger: I think we should avoid using codenames in public forums.
I'll check the 5421 behavior and get back to you.

docs/nifgen/rep_caps.rst
ni-jfitzger: Yeah, I'm not thrilled with this documentation easier, but for a different reason. I don't like the use of channel_enabled in the example code. I understand that it just an example attribute which isn't necessarily supported. It might be confusing to users, though.

I don't mind tackling the improvement of this documentation, but I'll have to do some digging to figure out how, since it's generated.

src/nifgen/system_tests/test_system_nifgen.py
ni-jfitzger: Updated all of the new tests to use integers and iterables, instead of strings. I even got rid of the '0-1' channel string.

docs/nifgen/rep_caps.rst
ni-jfitzger: It should just be a matter of updating [rep_caps.rest.mako](https://github.com/ni/nimi-python/blob/5be94c69ca4677d8062f2b84e5f3b2f1136edbe9/build/templates/rep_caps.rst.mako#L42)

src/nifgen/system_tests/test_system_nifgen.py
ni-jfitzger: I didn't have access to a 5421 but I tried to exporting the data marker event with a simulated 5421 and a real PXI-5441. Neither one errored.

I also looked through the NI-FGEN documentation and found [this page](https://zone.ni.com/reference/en-XX/help/370524V-01/pxie-5433/routes/) which seems to confirm that the PXIe-5433 (2CH) that we simulate with by default, does not support Data Marker Events. Marker Events are listed, but Data Marker Events are not.

I'll leave the test code basically as is, since you've already reviewed and okayed it, but I'm going to change this test to simulate with the 5421, instead. I don't want us to rely on the behavior for a property related to an unsupported event remaining unchanged.

marcoskirsch: > I'll leave the test code basically as is, since you've already reviewed and okayed it, but I'm going to change this test to simulate with the 5421, instead.

Please don't do this, as simulating DAQmx based instruments has its own set of problems* and we're aggressively moving away from it in order to simplify things and speed up testing even if it means a bit worse test coverage.

I'd rather stick to 54x3 instruments. We could look at the error. Unfortunately it does not appear to tell you _which_ Data Marker Event is being configured.

src/nifgen/system_tests/test_system_nifgen.py
marcoskirsch: Well, I think we need to be a bit careful here.
Our documentation and examples should teach customers to use the API in a "pythonic" way.
But if using strings work, we should have test coverage to ensure we don't accidentally break it. It should be tested _in addition_ to the recommended usage.

docs/nifgen/rep_caps.rst
marcoskirsch: Related issues, for which there appears to be a fair bit of duplication:
#1388
#1338
#1334
#1126

#1351 would be a nice one to fix too.

src/nifgen/system_tests/test_system_nifgen.py
marcoskirsch: All this said, we have unit tests that test the conversion for repeated capability into strings that the driver accepts, so I don't think we need to test every permutation in system tests. This is testing what we care about which is that the correct prefix is used.

PR #1686, opened 2022-01-21
Repeated Capabilites documentation improvements

### LINKED ISSUE #1126: Repeated capability documentation explains implementation details
Our public documentation for repeated capabilities contains text like:

`        passes a string of :python:`'0, 1, 2'` to the set attribute function.`

nimi-python customer should not care nor need to care how the API interacts with the driver runtime. We should document visible behavior only.

### LINKED ISSUE #1334: Poorly worded repeated capabilities documentation
### Description of issue

While working on #1290, a comment about how the repeated capabilities text in documentation was poor:

```
.. py:attribute:: nidigital.Session.sites[]
        If no prefix is added to the items in the parameter, the correct prefix will be added when
        the driver function call is made.
```
This phrase implies that the user knows about the underlying C API and describes an implementation detail.

Python users should not need to know nor care nor have us confuse them with hthis. I'd even argue that if documentation just told them how to use the basic tuple of numbers way we'd be fine.

### LINKED ISSUE #1388: For properties and method parameters that accept multiple types, documentation should promote the recommended type

ni-jfitzger:

There are several issues open against the way we document usage of repeated capabilities in our rep_caps.rst pages.

Summary of Issues
* We present implementation details that the user doesn't care about
  - These may be useful to contributors, but to everyone else they're just distracting.
* We list all of the possible types that can be passed to the indexing operator but don't make it clear what the recommended or "pythonic" types are.
  - This information is worth keeping, but we need to provide better guidance.
* We use channel_enabled in all of our examples, though it may not actually support the repeated capability in the example code.

This PR addresses these by doing the following:
* Turn documentation related to implementation details into comments only viewable in the raw source
* Explicitly recommend what types to pass to the indexing operator.
* Present examples which demonstrate accessing the repeated capability with our recommended types.
  - Do not present examples using other types.
* Add a note that channel_enabled may not actually support the repeated capability and to check individual property and method documentation.

TODO: List of issues.

* Fix #1388
* Fix #1334
* Fix #1126

Local code generation with documentation inspection.

### DIFF
--- a/build/templates/rep_caps.rst.mako
+++ b/build/templates/rep_caps.rst.mako
@@ -25,10 +25,15 @@ ${helper.get_rst_header_snippet('Repeated Capabilities', '=')}
 
     Repeated capabilities attributes use the indexing operator :python:`[]` to indicate the repeated capabilities.
     The parameter can be a string, list, tuple, or slice (range). Each element of those can be a string or
-    an integer. If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
-    :python:`'0:2'`
+    an integer.
 
-    Some repeated capabilities use a prefix before the number and this is optional
+    ..
+        If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
+        :python:`'0:2'`
+
+        Some repeated capabilities use a prefix before the number and this is optional.
+
+    The recommended way of accessing repeated capabilities is with an integer :python:`[0]` or range :python:`[0:2]`.
 
 % for rep_cap in config['repeated_capabilities']:
 <%
@@ -40,27 +45,43 @@ ${helper.get_rst_header_snippet(name, '-')}
     .. py:attribute:: ${module_name}.Session.${name}[]
 
 % if len(prefix) > 0:
-        If no prefix is added to the items in the parameter, the correct prefix will be added when
-        the driver function call is made.
+        ..
+            If no prefix is added to the items in the parameter, the correct prefix will be added when
+            the driver function call is made.
 
-        .. code:: python
+            .. code:: python
 
-            session.${name}['0-2'].channel_enabled = True
+                session.${name}['0-2'].channel_enabled = True
 
-        passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
+            passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
 
-        If an invalid repeated capability is passed to the driver, the driver will return an error.
+            If an invalid repeated capability is passed to the driver, the driver will return an error.
 
-        You can also explicitly use the prefix as part of the parameter, but it must be the correct prefix
-        for the specific repeated capability.
+            You can also explicitly use the prefix as part of the parameter, but it must be the correct prefix
+            for the specific repeated capability.
+
+            .. code:: python
+
+                session.${name}['${prefix}0-${prefix}2'].channel_enabled = True
+
+            passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
 
 % endif
         .. code:: python
 
-            session.${name}['${prefix}0-${prefix}2'].channel_enabled = True
+            session.${name}[0].channel_enabled = True
 
-        passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
+        sets :py:attr:`channel_enabled` to :python:`True` for ${name} 0.
 
+        .. code:: python
+
+            session.${name}[0:2].channel_enabled = True
+        
+        sets :py:attr:`channel_enabled` to :python:`True` for ${name} 0, 1, 2.
+
+        Note that :py:attr:`channel_enabled` is only used as an example and is not necessarily a property which
+        supports this repeated capability. See documentation for individual properties and methods to
+        learn what repeated capabilites they support, if any.
 
 % endfor
 


build/templates/rep_caps.rst.mako
```diff
@@ -25,10 +25,15 @@ ${helper.get_rst_header_snippet('Repeated Capabilities', '=')}
...
+    an integer.
 
-    Some repeated capabilities use a prefix before the number and this is optional
+    ..
+        If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
+        :python:`'0:2'`
+
+        Some repeated capabilities use a prefix before the number and this is optional.
+
+    The recommended way of accessing repeated capabilities is with an integer :python:`[0]` or range :python:`[0:2]`.
```
ni-jfitzger, line 36: I think I'm leaving out something important here.
:python;`[0,1,2]`

How do I refer to that? List? Tuple?

build/templates/rep_caps.rst.mako
```diff
@@ -40,27 +45,43 @@ ${helper.get_rst_header_snippet(name, '-')}
...
+            session.${name}[0].channel_enabled = True
+
+        sets :py:attr:`channel_enabled` to :python:`True` for ${name} 0.
+
+        .. code:: python
 
-        passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
+            session.${name}[0:2].channel_enabled = True
+        
+        sets :py:attr:`channel_enabled` to :python:`True` for ${name} 0, 1, 2.
```
ni-jfitzger, line 80: If I add anything to the recommend types above, I should add an example usage, here.

ni-jfitzger:
I haven't committed the results of `tox -e codegen`.
I can do that when there's agreement on the contents of rep_caps.rst.mako.

marcoskirsch:
> I haven't committed the results of tox -e codegen.
> I can do that when there's agreement on the contents of rep_caps.rst.mako.

No, please commit them before the PR is ready for review so we can see the effects and merge without having to track anything.

ni-jfitzger:
committed the results of tox-e codegen as requested.

docs/nidcpower/rep_caps.rst
```diff
@@ -18,10 +18,15 @@ Repeated Capabilities
```
marcoskirsch, line 18: >    Repeated capabilities attributes are used to set the `channel_string` parameter to the
    underlying driver function call. This can be the actual function based on the :py:class:`Session`
    method being called, or it can be the appropriate Get/Set Attribute function, such as :c:`niDCPower_SetAttributeViInt32()`.

The above is gibberish to a Python developer not familiar with our C API. Knowing our C API is not a requirement for using our Python APIs.

This is tricky word-smithing, so here goes something based on https://www.ivifoundation.org/resources/IVI%20Repeated%20Capabilities.pdf

> `nidcpower.Session` supports "Repeated Capabilities", which are multiple instances of the same type of
> functionality. The repeated capabilities supported by `nidcpower.Session` are:
> 1. `channels`
> 2. `instruments`
>
> Use the indexing operator `[]` to indicate which repeated capability you are trying to access. The parameter can 
> be an integer, a string, a list, a tuple, or slice (range).

docs/nidcpower/rep_caps.rst
```diff
@@ -18,10 +18,15 @@ Repeated Capabilities
...
     Repeated capabilities attributes use the indexing operator :python:`[]` to indicate the repeated capabilities.
     The parameter can be a string, list, tuple, or slice (range). Each element of those can be a string or
-    an integer. If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
-    :python:`'0:2'`
+    an integer.
 
-    Some repeated capabilities use a prefix before the number and this is optional
+    ..
+        If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
+        :python:`'0:2'`
```
marcoskirsch, line 25: I would remove this. We are discouraging Python users from using "0-2" and instead should use [0,1,2] or some more pythonic thing

docs/nidcpower/rep_caps.rst
```diff
@@ -18,10 +18,15 @@ Repeated Capabilities
...
+
+        Some repeated capabilities use a prefix before the number and this is optional.
```
marcoskirsch, line 27: remove

docs/nidcpower/rep_caps.rst
```diff
@@ -42,9 +56,18 @@ instruments
 
         .. code:: python
 
-            session.instruments['0-2'].channel_enabled = True
+            session.instruments[0].channel_enabled = True
+
+        sets :py:attr:`channel_enabled` to :python:`True` for instruments 0.
```
marcoskirsch, line 61: If we're going to show an example, we should make the documentation smarter.
How?

a) use a property that actually supports the repeated capability, and use that in the example. This can be done by us telling the generator what to use via metadata or by searching for a property that supports it. Probably easier to explicitly add to metadata.
b) pass in `"Dev1"` instead of `0`. This is not something we can do without additional metadata though.

Then we can remove the silly "Note that our codegenerator isn't smart enough to do this well"

marcoskirsch (changes requested):

docs/nidcpower/rep_caps.rst
marcoskirsch: To better illustrate what I'm saying, we'd need additional Python-specific metadata, like:

```
    'repeated_capabilities': [
        {
            'prefix': '',
            'python_name': 'channels',
            'property_for_docs':'channel enabled',
            'value_for_docs':'0',
        },
```

docs/nidcpower/rep_caps.rst
ni-jfitzger: Github does a terrible job of highlighting what we want to talk about.
Are you referring to the

"If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or :python:`'0:2'`"? That's a comment that won't show up except in the raw .rst, by the way.

ni-jfitzger: I see better in the conversation that is what you're referring to.
Yeah, I can see how that wouldn't even be useful for contributors. It's not an implementation detail, but rather a tip for how to do something that we don't want to do in the first place. I'll remove it.

docs/nidcpower/rep_caps.rst
ni-jfitzger: Yeah, I guess this comment is redundant with the commented example.

docs/nidcpower/rep_caps.rst
ni-jfitzger: I think I may have time on Monday to experiment with updating the metadata to achieve this.

>b) pass in "Dev1" instead of 0. This is not something we can do without additional metadata though.

There's a repeated capability that won't accept [0]? I'll need to change my edits higher up for recommended types to pass.

docs/nidcpower/rep_caps.rst
ni-jfitzger: Done. With the list items linking to sections of the page!
Let me know if you don't like it.

docs/nidcpower/rep_caps.rst
marcoskirsch: yes, for example the `instruments` repeated capability takes in instrument names (string)

build/templates/rep_caps.rst.mako
```diff
@@ -19,16 +19,20 @@
...
-    :python:`'0:2'`
+% for rep_cap in config['repeated_capabilities']:
+<%
+name = rep_cap['python_name']
+%>\
+    #. ${name}_
+% endfor
+
+    Use the indexing operator :python:`[]` to indicate which repeated capability you are trying to access.
+    The parameter can be an integer, a string, a list, a tuple, or slice (range).
```
sbethur, line 33: Few comments:
1) `slice` and `range` are two different things. `slice` is a built-in [function](https://docs.python.org/3/library/functions.html#slice) that returns a slice object. `range` is a [type](https://docs.python.org/3/library/functions.html#func-range).
2) `list`, `tuple`, and `range` are collectively known as [basic sequence types](https://docs.python.org/3/library/stdtypes.html#typesseq).
3) We support a range-based string using a dash or a colon. e.g. `session.channels['1-4']` and `session.channels['1:4']`. These are just strings as far as the language is concerned. But we have [logic](https://github.com/ni/nimi-python/blob/f074be747bf32054fce5ffe13047a9412806030b/generated/nifake/nifake/_converters.py#L66) to handle these.

We can be explicit about the types we support, along with an example for each:
- `int`
  - `session.foo[42]`
- `str`
  - Single instance: `session.foo['42']`
  - A range of instances: `session.foo['0-2']` or `session.foo['0:2']`
- `list`
  - `session.foo[0,1,2]`
- `tuple`
  - `session.foo[(0,1,2)]`
- `range`
  - `session.foo[range(10)]`
- `slice`
  - `session.foo[range(10)[3:5]]`

build/templates/rep_caps.rst.mako
```diff
@@ -19,16 +19,20 @@
...
 
-    Some repeated capabilities use a prefix before the number and this is optional
+    The recommended way of accessing repeated capabilities is with an integer :python:`[0]` or range :python:`[0:2]`.
```
sbethur, line 35: > range :python:`[0:2]`

I don't think this will work since this is not the correct syntax for Python `range`. But `['0:2']` will work because of the custom logic we have for handling range-based strings.

Having said that, I don't think any of the ways I've enumerated above can be recommended over others. `int` is obviously the go-to for single instances. Basic sequence types and `slice` are good to use because they're part of the language, Range-based strings provides syntactic sugar for defining ranges. e.g. `['0:2']` rather than `[range(0,2)]`.

build/templates/rep_caps.rst.mako
```diff
@@ -40,27 +44,43 @@ ${helper.get_rst_header_snippet(name, '-')}
     .. py:attribute:: ${module_name}.Session.${name}[]
 
 % if len(prefix) > 0:
-        If no prefix is added to the items in the parameter, the correct prefix will be added when
-        the driver function call is made.
+        ..
+            If no prefix is added to the items in the parameter, the correct prefix will be added when
+            the driver function call is made.
```
sbethur, line 49: I think this is implementation detail that is not useful to the end user.

build/templates/rep_caps.rst.mako
```diff
@@ -40,27 +44,43 @@ ${helper.get_rst_header_snippet(name, '-')}
...
 
-        .. code:: python
+            .. code:: python
 
-            session.${name}['0-2'].channel_enabled = True
+                session.${name}['0-2'].channel_enabled = True
 
-        passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
+            passes a string of :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to the set attribute function.
```
sbethur, line 55:
```suggestion
            passes repeated capability instances represented by :python:`'${prefix}0, ${prefix}1, ${prefix}2'` to `channel_enabled` property.
```

build/templates/rep_caps.rst.mako
```diff
@@ -40,27 +44,43 @@ ${helper.get_rst_header_snippet(name, '-')}
...
 
-        If an invalid repeated capability is passed to the driver, the driver will return an error.
+            If an invalid repeated capability is passed to the driver, the driver will return an error.
```
sbethur, line 57: This is usually implied and can be omitted. For method parameters, we don't say something like this explicitly.

sbethur (changes requested):

build/templates/rep_caps.rst.mako
sbethur: I see @marcoskirsch doesn't like us to recommend the range-based strings. Since I saw we had logic specifically for handling it, I thought it was important. But it sounds like it was ill-conceived. I don't mind not documenting it.

Once we land on recommended ways, we should go over the examples to ensure only recommended types are used.

docs/nidcpower/rep_caps.rst
sbethur: > use a property that actually supports the repeated capability, and use that in the example. 


Please check that there is at least one property for every supported repeated capability type. Else, you'll have plumb the logic for using methods.

build/templates/rep_caps.rst.mako
marcoskirsch: Note that the APIs will work with any iterable.

ni-jfitzger: >Note that the APIs will work with any iterable.
Yes, the APIs will, though it seems like there are certain repeated capabilities that will only accept strings?
instruments?
maybe pins?

>We can be explicit about the types we support, along with an example for each:
@sbethur are you talking about the example code shown in the individual sections for each repeated capability or new examples that should be placed before the individual sections for each repeated capability type? If it's the former, that will further complicate an already difficult task.

sbethur: Well, not all iterables - `dict` and `set` won't work.

build/templates/rep_caps.rst.mako
marcoskirsch: >  It's only visible in the raw source, not in the html documentation.

remove it, it's only causing confusion

build/templates/rep_caps.rst.mako
sbethur: > new examples that should be placed before the individual sections 

This

sbethur: Elsewhere Marcos suggested not documenting range-based strings, which I think is a good idea. If we leave that out, then the supported types are all Python-native and we don't need to elaborate the support with examples. Here's my new suggestion:

> The parameter can be a single element or an iterable that implements sequence semantics, like list, tuple, range, and slice.

build/templates/rep_caps.rst.mako
```diff
@@ -19,16 +19,20 @@
...
-    an integer. If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
-    :python:`'0:2'`
+% for rep_cap in config['repeated_capabilities']:
+<%
+name = rep_cap['python_name']
+%>\
+    #. ${name}_
+% endfor
+
+    Use the indexing operator :python:`[]` to indicate which repeated capability you are trying to access.
```
sbethur, line 32:
```suggestion
    Use the indexing operator :python:`[]` to indicate which repeated capability instance you are trying to access.
```

build/templates/rep_caps.rst.mako
ni-jfitzger: It's actually a tuple. The '[]' is intended to be the indexing operator.

build/templates/rep_caps.rst.mako
```diff
@@ -19,48 +19,47 @@
...
+% endfor
+
+    Use the indexing operator :python:`[]` to indicate which repeated capability instance you are trying to access.
+    The parameter can be a single element or an iterable that implements sequence semantics, like list, tuple, range and slice.
+
+    The recommended way of accessing a single repeated capability is with an integer :python:`[0]` for capabilities that support it and a string :python:`['Dev1']`
+    for those that don't support integers.
 
-    Some repeated capabilities use a prefix before the number and this is optional
+    The recommended way of accessing multiple repeated capabilites at once is with a tuple (:python:`[0, 1]` or :python:`['Dev1', 'Dev2']`) or slice :python:`[0:2]`.
```
ni-jfitzger, line 38: I'm not currently providing slice usage examples. Let me know if you want me to add them.

docs/nidigital/rep_caps.rst
ni-jfitzger, line 1: Still undocumented: rep cap chaining.

docs/nidcpower/rep_caps.rst
```diff
@@ -42,9 +50,14 @@ instruments
...
+            print(session.instruments['Dev1'].serial_number)
+
+        prints :py:attr:`serial_number` for instruments 'Dev1'.
+
+        .. code:: python
 
-        passes a string of :python:`'0, 1, 2'` to the set attribute function.
+            print(session.instruments['Dev1', 'Dev2', '3rdDevice'].serial_number)
 
+        prints :py:attr:`serial_number` for instruments 'Dev1', 'Dev2', '3rdDevice' or errors if the value is not the same for all.
```
ni-jfitzger, line 61: I included the part about the error behavior on a whim. It should hold true for any multi-capability property read, but let me know if you think I should leave it out.

marcoskirsch: Whether the driver allows passing multiple rep caps or not varies. The behavior you describe for serial_number happens to be accurate for serial_number but no guaranteee for other things. I think it's wiser to leave out.

build/templates/rep_caps.rst.mako
marcoskirsch: I think that would be nice.
I wonder if we should change wording to say that it should be an iterable rather than provide examples of iterables.

docs/nidcpower/rep_caps.rst
```diff
@@ -12,16 +12,19 @@
...
-    Repeated capabilities attributes use the indexing operator :python:`[]` to indicate the repeated capabilities.
-    The parameter can be a string, list, tuple, or slice (range). Each element of those can be a string or
-    an integer. If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
-    :python:`'0:2'`
+    #. channels_
+    #. instruments_
 
-    Some repeated capabilities use a prefix before the number and this is optional
+    Use the indexing operator :python:`[]` to indicate which repeated capability instance you are trying to access.
+    The parameter can be a single element or an iterable that implements sequence semantics, like list, tuple, range and slice.
```
marcoskirsch, line 22: "like" -> "such as"

docs/nidcpower/rep_caps.rst
```diff
@@ -30,10 +33,15 @@ channels
...
 
-            session.channels['0-2'].channel_enabled = True
+            session.channels[0].output_function = nidcpower.OutputFunction.DC_CURRENT
+
+        sets :py:attr:`output_function` to :py:data:`~nidcpower.OutputFunction.DC_CURRENT` for channels 0.
+
+        .. code:: python
 
-        passes a string of :python:`'0, 1, 2'` to the set attribute function.
+            session.channels[0, 2].output_function = nidcpower.OutputFunction.DC_CURRENT
```
marcoskirsch, line 43: here is where we could add an example using an iterable, below line 42?

session.channles[range(3)].output_function = nidcpower.OutputFunction.DC_CURRENT

docs/nidigital/rep_caps.rst
```diff
@@ -12,16 +12,25 @@
...
-    The parameter can be a string, list, tuple, or slice (range). Each element of those can be a string or
-    an integer. If it is a string, you can indicate a range using the same format as the driver: :python:`'0-2'` or
-    :python:`'0:2'`
+    #. channels_
+    #. pins_
+    #. instruments_
+    #. pattern_opcode_events_
+    #. conditional_jump_triggers_
+    #. sites_
+    #. rio_events_
```
marcoskirsch, line 24: I'm not sure about this... are the RIO triggers and events usable from Python at all in an application?

src/niscope/metadata/config_addon.py
```diff
@@ -15,5 +15,22 @@
             'file_name': 'measurement_stats',
             'python_name': 'MeasurementStats'
         }
-    ]
+    ],
+    'repeated_capabilities': [
+        {
+            'attr_for_docs_example': 'channel_enabled',
```
marcoskirsch, line 21: does the generator die if you put a non-existen attribute here? or does it use this string to look up additional info about the property?

marcoskirsch (changes requested):

build/templates/rep_caps.rst.mako
ni-jfitzger: Well, we already mention supported values above:

>Use the indexing operator [] to indicate which repeated capability instance you are trying to access. The parameter can be a single element or an iterable that implements sequence semantics, like list, tuple, range and slice.

I guess it wouldn't hurt, to keep this part more generic, though, especially given our examples below.

docs/nidigital/rep_caps.rst
ni-jfitzger: They are. We test it: https://github.com/ni/nimi-python/blob/2986696c11dd4b64b50b3e41744a7dc55ecc1f6b/src/nidigital/system_tests/test_system_nidigital.py#L146

docs/nidcpower/rep_caps.rst
ni-jfitzger: I was actually referring to the inability to retrieve a property value for multiple capabilities if the value is not the same for all.

docs/nidigital/rep_caps.rst
marcoskirsch: They work. But are they actually useful in an application? Given that there is no LVFPGA support in Python.

docs/nidcpower/rep_caps.rst
marcoskirsch: I understand. I'm saying that the ability to retrieve a property value for multiple capabilities if the value is the same for all is not a universally true thing. It can vary from attribute to attribute. It just happens that we implemented the described behavior for _most_ attributes in our driver runtimes.

docs/nidigital/rep_caps.rst
ni-jfitzger: Oh, maybe not.
I can create an issue to remove them.

src/niscope/metadata/config_addon.py
ni-jfitzger: We don't use it to look anything up. We only use metadata from config.py, config_addon.py to build the rep cap examples. The only thing that happens if the attribute doesn't exist is the link to the attribute in the documentation won't work.

docs/nidcpower/rep_caps.rst
```diff
@@ -42,9 +50,14 @@ instruments
...
 
-            session.instruments['0-2'].channel_enabled = True
+            print(session.instruments['Dev1'].serial_number)
+
+        prints :py:attr:`serial_number` for instruments 'Dev1'.
+
+        .. code:: python
 
-        passes a string of :python:`'0, 1, 2'` to the set attribute function.
+            print(session.instruments['Dev1', 'Dev2', '3rdDevice'].serial_number)
```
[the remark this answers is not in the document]
ni-jfitzger: >I found 3rdDevice confusing.

I just didn't want users to think it had to start with "Dev". I have the same concern with the examples you gave.
Maybe my blurb about the basic element should specify "device alias string" for instruments?

docs/nidigital/rep_caps.rst
ni-jfitzger: >They work. But are they actually useful in an application? Given that there is no LVFPGA support in Python.

I think there is support.
https://github.com/ni/nifpga-python

docs/nidcpower/rep_caps.rst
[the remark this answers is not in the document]
bkeryan: `PXI1Slot2` is a good alternative.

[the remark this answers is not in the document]
marcoskirsch: Right, just use something that matches defaults. This will be informative enough and not confused anyone.

> Maybe my blurb about the basic element should specify "device alias string" for instruments?

I question how familiar customers are with term "device alias". It's certainly not in our APIs. Let's avoid,

PR #1702, opened 2022-03-31
NI-DCPower: Create custom types for LCR

olsl21:

- Update config_addon.py to enable custom types codegen
- Enable `LCRReferenceValueType` enum that is used by `LCRLoadCompensationSpot` custom type
- Implement lcr_load_compensation_spot and lcr_measurement custom types
- Update nidcpower.mak to copy the custom types into the generated package
- Create unit tests for the new custom types
- Update coverage_unit_tests.rc to omit code-generated files and include the new custom type implementations
- Update tox.ini and tox-travis.ini to run the unit tests and generate the code coverage report for NI-DCPower unit tests
- Update CHANGELOG

- All the new unit tests passed.
- Code coverage percentage is 97% (the only lines that are not covered are those lines in `lcr_load_compensation_spot` that deals with the byte packing alignment of the `struct_NILCRLoadCompensationSpot` ctypes struct)


generated/nidcpower/nidcpower/ni_lcr_measurement.py
```diff
@@ -0,0 +1,242 @@
...
+        lp=0.0,
+        cp=0.0,
+        rp=0.0,
+        d=0.0,
+        q=0.0,
+        measurement_mode=enums.InstrumentMode.LCR,
+        dc_in_compliance=False,
+        ac_in_compliance=False,
+        unbalanced=False,
+        reserved1=0.0,
```
marcoskirsch, line 175: I don't see the need to reserve anything in the public API in Python

generated/nidcpower/nidcpower/ni_lcr_measurement.py
```diff
@@ -0,0 +1,242 @@
...
+        stimulus_frequency=0.0,
+        ac_voltage=complex(),
+        ac_current=complex(),
+        z=complex(),
+        z_magnitude=0.0,
+        z_phase=0.0,
+        y=complex(),
+        y_magnitude=0.0,
+        y_phase=0.0,
+        ls=0.0,
```
marcoskirsch, line 163: ls, cs, etc... are those obvious to users?

There is definitely a lack of documentation for this class that you need to address.

marcoskirsch (changes requested):

generated/nidcpower/nidcpower/ni_lcr_measurement.py
olsl21: `ls`, `cs`, `rs`, etc. are named based on the `nidcpower.h` header file:
```
// Z values
NIComplexNumber Z;
ViReal64 ZMagnitude;
ViReal64 ZPhase; // degrees
// Y values
NIComplexNumber Y;
ViReal64 YMagnitude;
ViReal64 YPhase; // degrees
```

I do agree that this class requires documentation. I have added documentation for all its non-reserved properties (similar documentation will also be added to the new LCR-related functions that use this custom type in #1704)

generated/nidcpower/nidcpower/ni_lcr_measurement.py
olsl21: I agree, I have removed the reserved properties from the constructor parameter of `NILCRMeasurement`. Do note that I still retain them in the ctypes class `struct_NILCRMeasurement` as they are needed to preserve the layout of the struct, and I still set these reserved properties in `NILCRMeasurement` so that it remains compatible with `struct_NILCRMeasurement`, i.e. they can be converted to each other easily.

marcoskirsch:
A lot of my comments apply globally even beyond `nidcpower` structs.
Let's go over them, and consider making some of the adjustments in separate PRs that go in.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,145 @@
...
+import numbers
+import platform
+
+import nidcpower._visatype
+import nidcpower.enums as enums
+
+
+# This class is an internal implementation detail
+# ctypes definition
+# Name must match exactly what the name of the structure type is named in the C API.
```
marcoskirsch, line 11: This comment is inaccurate. This is the struct:

```
   typedef struct NILCRLoadCompensationSpot_struct
   {
      ViReal64 frequency;
      ViInt32 referenceValueType;
      ViReal64 referenceValueA;
      ViReal64 referenceValueB;
   } NILCRLoadCompensationSpot;
```

marcoskirsch: I also don't understand why they **must match exactly**. Can you explain this?

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,144 @@
...
+        ("reference_value_a", nidcpower._visatype.ViReal64),
+        ("reference_value_b", nidcpower._visatype.ViReal64),
+    ]
+
+    def __init__(self, data=None):
+        super(ctypes.Structure, self).__init__()
+        if data is not None:
+            self.frequency = data.frequency
+            self.reference_value_type = enums.LCRReferenceValueType(data.reference_value_type).value
+            if isinstance(data, LCRLoadCompensationSpot):
```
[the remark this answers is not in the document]
marcoskirsch: waiting on this

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,145 @@
...
+                    self.reference_value_a = data.reference_value.real
+                    self.reference_value_b = data.reference_value.imag
+                else:
+                    self.reference_value_a = data.reference_value
+                    self.reference_value_b = 0.0
+            else:
+                self.reference_value_a = data.reference_value_a
+                self.reference_value_b = data.reference_value_b
+        else:
+            # Assign default value for all fields
```
marcoskirsch, line 42: I think Python/ctypes may get you this for free.

marcoskirsch (changes requested):
There are still a lot of unresolved questions in this PR. I'll look at it once they are all resolved.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,144 @@
...
+
+# This class is an internal implementation detail
+# ctypes definition
+# Name must match exactly what the name of the structure type is named in the C API.
+class struct_NILCRLoadCompensationSpot(ctypes.Structure):  # noqa N801
+    if platform.system() == "Windows" and platform.architecture()[0] == "64bit":
+        _pack_ = 8
+    elif platform.system() == "Linux" or platform.system() == "Darwin":
+        _pack_ = 4
+    else:
```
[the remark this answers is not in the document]
olsl21: This code for determining packing is really following the exact logic in the public "nidcpower.h" header file:
```
#if defined(_WIN64)
#pragma pack(push,8)
#elif defined(__linux__) || defined(__APPLE__)
#pragma pack(push,4)
#else
#pragma pack(push,1)
#endif
```
I don't think we should add explicit constraints here that would make it inconsistent with the public header file.

> Also, this code for determining packing is probably duplicated with other similar structs. Single source it please. Consider doing it in its own PR.

There is currently only one instance of usage of `_pack_` in the entire nimi-python, i.e. in [`niscope.struct_niScope_wfmInfo`](https://github.com/ni/nimi-python/blob/master/src/niscope/custom_types/waveform_info.py), and it is using a different logic (i.e. its packing is fixed at 8 bytes, same as what have been defined in "niscope.h" public header file).

I have created #1777 for single-sourcing such code, but I don't think it should be implemented now because:
- There isn't any driver using this at the moment
- Different drivers may require different logic for their structs

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,144 @@
...
+import platform
+
+import nidcpower._visatype
+import nidcpower.enums as enums
+
+
+# This class is an internal implementation detail
+# ctypes definition
+# Name must match exactly what the name of the structure type is named in the C API.
+class struct_NILCRLoadCompensationSpot(ctypes.Structure):  # noqa N801
```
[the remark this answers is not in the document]
olsl21: >  Do the other structs not do this?

They don't, e.g. [`struct_niScope_wfmInfo`](https://github.com/ni/nimi-python/blob/master/src/niscope/custom_types/waveform_info.py#L9). 

In fact, the existing codegen logic assumes that the name of the internal ctype begins with `struct_`, i.e. https://github.com/ni/nimi-python/blob/5e4ee740b893c017f9f53da98c965b77f86159d0/build/helper/metadata_add_all.py#L306.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,145 @@
...
+    if platform.system() == "Windows" and platform.architecture()[0] == "64bit":
```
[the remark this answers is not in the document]
olsl21: As mentioned in [my reply to your other comment](https://github.com/ni/nimi-python/pull/1702#discussion_r919946197), the Python code for figuring out the struct packing *is not* the same for all the structs across nimi-python.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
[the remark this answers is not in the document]
marcoskirsch: 🤦

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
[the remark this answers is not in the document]
olsl21: Constructing from an `LCRLoadCompensationSpot` is the only necessary option as it would be used by the `perform_lcr_load_compensation()` method to be added as part of #1772 (https://github.com/ni/nimi-python/pull/1772/files#diff-31582e14c636470681a52cfa45ebf6f28dbd68afde8827d6058a43022b3231c4R5716). I came up with the other code paths based on https://github.com/ni/nimi-python/blob/master/src/niscope/custom_types/waveform_info.py

I have removed the other unnecessary code paths in my latest commit.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,144 @@
...
+        data=None,
+        frequency=0.0,
+        impedance=complex(),
+        ideal_capacitance=None,
+        ideal_inductance=None,
+        ideal_resistance=None,
+    ):
+        """LCRLoadCompensationSpot
+
+        Creates and returns an LCRLoadCompensationSpot object. When data is None, exactly one of
```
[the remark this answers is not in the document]
olsl21: This docstring is for the `__init__()` method, I have added a docstring for the class itself in my latest commit.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
olsl21: This comment was actually copied from https://github.com/ni/nimi-python/blob/master/src/niscope/custom_types/waveform_info.py#L8.

I have removed the comment "Name must match exactly what the name of the structure type is named in the C API" in my latest commit as I don't think the names must match exactly either.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
olsl21: As part of removing the unnecessary code paths in https://github.com/ni/nimi-python/pull/1702#discussion_r920326129, these default settings are no longer needed in my latest commit.

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,305 @@
...
+        ("z"                 , "Z"                                     ),  # noqa: E202,E203
+        ("series_lcr"        , ("Ls", "Cs", "Rs")                      ),  # noqa: E202,E203
+        ("parallel_lcr"      , ("Lp", "Cp", "Rp")                      ),  # noqa: E202,E203
+        ("d"                 , "D"                                     ),  # noqa: E202,E203
+        ("measurement_mode"  , "Measurement mode"                      ),  # noqa: E202,E203
+        ("in_compliances"    , ("DC in compliance", "AC in compliance")),  # noqa: E202,E203
+        ("unbalanced"        , "Unbalanced"                            ),  # noqa: E202,E203
+    ]
+
+    def __init__(
```
marcoskirsch, line 109: There are two ways of constructing an LCRMeasurement here.
This is an object returned to the user by the driver exclusively, right?
Why do we need more than one way to construct this?

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,305 @@
...
+    def get_z_magnitude_and_phase(self):
+        """get_z_magnitude_and_phase
+
+        Returns a tuple of (z_magnitude, z_phase) of this LCRMeasurement object.
+
+        Returns:
+            z_magnitude_and_phase (tuple of (float, float)):
+
+                - **z_magnitude** (float): The magnitude of the complex impedance, in ohms.
+                - **z_phase** (float): The impedance phase angle, in degrees.
```
marcoskirsch, line 220: Are you sure `polar` doesn't return radians?
If this is wrong, how did it pass tests? Might be missing coverage.

marcoskirsch (changes requested):

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,116 @@
...
+            self.reference_value_b = data.reference_value.imag
+        else:
+            self.reference_value_a = data.reference_value
+            self.reference_value_b = 0.0
+
+
+class LCRLoadCompensationSpot(object):
+    """Constructs a DUT specification for a given frequency to use in LCR load compensation."""
+
+    def __init__(
```
sbethur, line 39: Did you consider the ctor to take two params - `reference_value_type` (of type LCRReferenceValueType) and `value` (complex or float)?

generated/nidcpower/nidcpower/unit_tests/test_nidcpower.py
```diff
@@ -0,0 +1,434 @@
...
+    ctype_object = nidcpower.struct_NILCRLoadCompensationSpot(python_object)
+    for member in expected_ctype_members:
+        assert getattr(ctype_object, member) == pytest.approx(
+            expected_ctype_members[member]
+        )
+
+
+def test_lcr_load_compensation_spot_repr_and_str():
+    python_object = nidcpower.LCRLoadCompensationSpot(frequency=100.0, impedance=complex(1.0, 2.0))
+    assert repr(python_object) == "LCRLoadCompensationSpot(frequency=100.0, impedance=(1+2j))"
```
sbethur, line 401: You should be able to create a valid python object using the output of `repr`, by calling `eval` on it. Verify this is doable by something like:
```
recreated_object = eval(repr(python_object))
assert str(recreated_object) == str(python_object)
```

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,116 @@
...
+                    tuple(
+                        uppercase_parameter_name.lower()
+                        for uppercase_parameter_name in enums.LCRReferenceValueType.__members__
+                    )
+                )
+            )
+
+    def __repr__(self):
+        return "{0}(frequency={1}, {2}={3})".format(
+            self.__class__.__name__,
```
sbethur, line 103:
```suggestion
        return "{0}.{1}(frequency={2}, {3}={4})".format(
            self.__class__.__module__,
            self.__class__.__qualname__,
```

I think you need to do like above to be able to recreate class instances based on `repr` output.

sbethur (changes requested):

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,305 @@
...
+                self.y_magnitude, self.y_phase = data.get_y_magnitude_and_phase()
+                self.ls, self.cs, self.rs = data.series_lcr
+                self.lp, self.cp, self.rp = data.parallel_lcr
+                self.d = data.d
+                self.q = data.get_q()
+                self.measurement_mode = enums.InstrumentMode(data.measurement_mode).value
+                self.dc_in_compliance, self.ac_in_compliance = data.in_compliances
+                self.unbalanced = data.unbalanced
+            else:
+                for field_name, _ in struct_NILCRMeasurement._fields_:
```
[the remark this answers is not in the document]
sbethur: Hmm.. it's the other way around - LCRMeasurement is constructed using struct_NILCRMeasurement, right?

Do we need the `if data is not None:` case?

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,305 @@
...
+import ctypes
+import math
+
+import nidcpower._visatype
+import nidcpower.enums as enums
+
+
+# This class is an internal implementation detail
+# ctypes definition
+# Name must match exactly what the name of the structure type is named in the C API.
```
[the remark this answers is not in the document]
olsl21: Ya, it isn't strictly true. Similar to the [comment for `struct_NILCRLoadCompensationSpot`](https://github.com/ni/nimi-python/pull/1702#discussion_r920335031), I have removed the statement "Name must match exactly what the name of the structure type is named in the C API." and added a mention of the corresponding C API type in my latest commit.

src/nidcpower/custom_types/lcr_measurement.py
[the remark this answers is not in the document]
olsl21: > Hmm.. it's the other way around - LCRMeasurement is constructed using struct_NILCRMeasurement, right?

Yes, the driver will return `struct_NILCRMeasurement` which will then be converted into `LCRMeasurement` for Python user. Similar to `lcr_load_compensation_spot`, I have removed all the unnecessary code paths for the ctype-Python type conversions in my latest commit. `struct_NILCRMeasurement` now has no custom constructor.

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,305 @@
...
+                self.d = data.d
+                self.q = data.get_q()
+                self.measurement_mode = enums.InstrumentMode(data.measurement_mode).value
+                self.dc_in_compliance, self.ac_in_compliance = data.in_compliances
+                self.unbalanced = data.unbalanced
+            else:
+                for field_name, _ in struct_NILCRMeasurement._fields_:
+                    setattr(self, field_name, getattr(data, field_name))
+        else:
+            # Assign default value for all fields
```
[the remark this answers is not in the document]
olsl21: It is not needed, and as mentioned in https://github.com/ni/nimi-python/pull/1702#discussion_r920879292, I have removed all the unnecessary code paths and `struct_NILCRMeasurement` now has no custom constructor.

src/nidcpower/custom_types/lcr_measurement.py
olsl21: Yes, this is an object returned to the user by the driver exclusively. As mentioned in https://github.com/ni/nimi-python/pull/1702#discussion_r920879292, I have removed all the unnecessary code paths, and now `LCRMeasurement` can only be constructed from an instance of `struct_NILCRMeasurement` in my latest commit.

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,305 @@
...
+        Returns:
+            y_magnitude_and_phase (tuple of (float, float)):
+
+                - **y_magnitude** (float): The magnitude of the complex admittance, in siemens.
+                - **y_phase** (float): The admittance phase angle, in degrees.
+
+        """
+        return cmath.polar(self.get_y())
+
+    def get_q(self):
```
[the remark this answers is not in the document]
olsl21: I have renamed these getters by removing the "get_" prefix and added `@property` decorator to them so that they behave like read-only members of this class in my latest commit.

src/nidcpower/custom_types/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,116 @@
...
+            impedance (complex): Specifies the actual impedance of your DUT to be used as the basis
+                for load compensation, or None to use another type of DUT specification value.
+
+            ideal_capacitance (float): Specifies the ideal capacitance of your DUT to be used as the
+                basis for load compensation, or None to use another type of DUT specification value.
+
+            ideal_inductance (complex): Specifies the ideal inductance of your DUT to be used as the
+                basis for load compensation, or None to use another type of DUT specification value.
+
+            ideal_resistance (complex): Specifies the ideal inductance of your DUT to be used as the
```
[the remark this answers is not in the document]
olsl21: Thanks for pointing this out, I have fixed the docstring in my latest update.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
olsl21: Nope, based on our LCR Python API review meeting, the conclusion was that we don't want Python user to set `LCRReferenceValueType` explicitly, hence we went for such a design (that uses keyword parameters).

generated/nidcpower/nidcpower/unit_tests/test_nidcpower.py
olsl21: Thanks for the suggestion, I have this test to `test_lcr_load_compensation_spot_repr_and_str()` in my latest commit.

Note that this cannot be accomplished for LCRMeasurement which can now only be constructed from an instance of  `struct_NILCRMeasurement`, hence I have removed the custom `__repr__()` method for `LCRMeasurement`.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
olsl21: Thanks for the suggestion, I have updated it in my latest commit.

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+
+class LCRMeasurement(object):
+    """Specifies an LCR measurement.
+
+    Fields:
+        channel (str): The channel name associated with this LCRMeasurement.
+
+        vdc (float): Specifies the measured DC voltage, in volts.
+
+        idc (float): Specifies the measured DC current, in amps.
```
marcoskirsch, line 61: amperes

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+
+        stimulus_frequency (float): Specifies the frequency of the LCR test signal, in Hz.
```
marcoskirsch, line 63: hertz

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+
+        ac_voltage (complex): Specifies the measured AC voltage, in volts RMS.
+
+        ac_current (complex): Specifies the measured AC current, in amps RMS.
```
marcoskirsch, line 67: amperes

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+
+        z (complex): Specifies the complex impedance.
```
marcoskirsch, line 69: For consistency:

Specifies the impledance, in ohms.

(I believe impedance is understood to be complex)

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+
+        series_lcr (LCR): Specifies the inductance, in henrys, the capacitance, in farads, and
+            the resistance, in ohms, as measured using a series circuit model.
+
+        parallel_lcr (LCR): Specifies the inductance, in henrys, the capacitance, in farads, and
+            the resistance, in ohms, as measured using a parallel circuit model.
+
+        d (float): The dissipation factor of the circuit.
```
marcoskirsch, line 77: `dissipation_factor` ?

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+        measurement_mode (enums.InstrumentMode): Specifies the measurement mode.
+            **Defined Values**:
+
+            +-----------------------+-----------------------------------------------------+
+            | InstrumentMode.SMU_PS | The channel(s) are operating as a power supply/SMU. |
+            +-----------------------+-----------------------------------------------------+
+            | InstrumentMode.LCR    | The channel(s) are operating as an LCR meter.       |
+            +-----------------------+-----------------------------------------------------+
+
+        in_compliances (InCompliances): Indicates whether the output was in DC compliance and/or
```
marcoskirsch, line 88: consider having:

dc_in_compliance
ac_in_compliance

as bools

marcoskirsch: Zen of Python:

> Flat is better than nested.

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def __init__(self, data):
+        """LCRMeasurement
+
+        Creates and returns an LCRMeasurement object.
+
+        Args:
+            data (struct_NILCRMeasurement): The LCR measurement ctypes object returned by the driver.
+        """
+        self.channel = ""
```
marcoskirsch, line 122: consider making channel a parameter to __init__

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+        self.ac_voltage = complex(data.ac_voltage_real, data.ac_voltage_imaginary)
+        self.ac_current = complex(data.ac_current_real, data.ac_current_imaginary)
+        self.z = complex(data.z_real, data.z_imaginary)
+        self.series_lcr = LCRMeasurement.LCR(
+            inductance=data.ls, capacitance=data.cs, resistance=data.rs
+        )
+        self.parallel_lcr = LCRMeasurement.LCR(
+            inductance=data.lp, capacitance=data.cp, resistance=data.rp
+        )
+        self.d = data.d
```
marcoskirsch, line 135: since you are keeping this around as a member, consider not storing any of the others (self.vdc, self.idc, etc) since they can be extracted or computed on demand from self.d.

Also, avoid abbreviations and call it self.data

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+        self.measurement_mode = enums.InstrumentMode(data.measurement_mode)
+        self.in_compliances = LCRMeasurement.InCompliances(
+            dc=bool(data.dc_in_compliance), ac=bool(data.ac_in_compliance)
+        )
+        self.unbalanced = bool(data.unbalanced)
+
+    @property
+    def z_magnitude_and_phase(self):
```
marcoskirsch, line 143: Cstomers can get the impedance as complex and do the math themselves. I don't think there's much point in providing these utility properties.

Otherwise, why not have properties to return these as rectangular values?

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,229 @@
...
+
+        Returns the quality factor of this LCRMeasurement object.
+
+        Returns:
+            q (float): The quality factor of the circuit.
+
+        """
+        if self.d == 0:
+            return math.nan
+        return 1.0 / self.d
```
marcoskirsch, line 200: If you have self.d and that has q as computed by the driver... why compute here again?

generated/nidcpower/nidcpower/lcr_measurement.py
marcoskirsch: I'd rather err on the side of putting too little functionality than too much. See [wiki](https://github.com/ni/nimi-python/wiki/API-design-guidelines#deviate-from-c-apis-in-the-following-ways)

>If there's debate on whether a function should be included or not, then don't include it
Trivial to add if needed
Hard to remove once added

marcoskirsch (changes requested):
I think we're better off breaking this up into two separate PRs.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+# This class is an internal ctypes implementation detail that corresponds to
+# NILCRLoadCompensationSpot in the C API
+class struct_NILCRLoadCompensationSpot(ctypes.Structure):  # noqa N801
+    if platform.system() == "Windows" and platform.architecture()[0] == "64bit":
+        _pack_ = 8
+    elif platform.system() == "Linux" or platform.system() == "Darwin":
+        _pack_ = 4
+    else:
```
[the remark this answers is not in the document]
olsl21: As explained earlier in https://github.com/ni/nimi-python/pull/1702#discussion_r919946197, this is following the exact logic in the public "nidcpower.h" header file for NI-DCPower 21.8, which has already been released, so I don't think it can be fixed by a quick PR.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+        none_count = 0
+        for uppercase_parameter_name in enums.LCRReferenceValueType.__members__:
+            parameter_name = uppercase_parameter_name.lower()
+            parameter = locals()[parameter_name]
+            if parameter is None:
+                none_count += 1
+            elif parameter_name == "impedance" and not isinstance(parameter, numbers.Complex):
+                raise TypeError("Parameter impedance must be of type complex")
+            elif parameter_name != "impedance" and not isinstance(parameter, numbers.Real):
+                raise TypeError("Parameter {} must be a real number".format(parameter_name))
```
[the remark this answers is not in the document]
olsl21: This type checking is needed because the parameter would be assigned directly to `self.reference_value`, and the constructor of `struct_NILCRLoadCompensationSpot` assumes that `self.reference_value` is a complex number or a real number depending on `self.reference_value_type`. If we really want to support duck-typing in this case, I can remove the type validation and just let the constructor of `struct_NILCRLoadCompensationSpot` errors out when the user passes in the wrong data type while trying to call the `perform_lcr_load_compensation_spot()` method.

>  So what if the user passes impedance as integer and not numbers.Real??

Impedance need to be a complex number instead of a real number. Also, passing in an integer when a `numbers.Real` instance is expected would pass the type validation, since `numbers.Real` isn't a concrete type, and `isinstance()` checks for subtype relationship as well instead of an exact type match:
```
>>> isinstance(1, numbers.Real)
True
```

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+            self.reference_value_b = data.reference_value.imag
+        else:
+            self.reference_value_a = data.reference_value
+            self.reference_value_b = 0.0
+
+
+class LCRLoadCompensationSpot(object):
+    """Constructs a DUT specification for a given frequency to use in LCR load compensation."""
+
+    def __init__(
```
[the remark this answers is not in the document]
olsl21: As mentioned in https://github.com/ni/nimi-python/pull/1702#discussion_r920916779, such a class constructor design was actually the *result of our previous LCR Python API review meeting*, the rationale mentioned during that meeting was that we want to provide Python users with an easy way to construct an `LCRLoadCompensationSpot` based on the `referenceValueType`, instead of having the users populating the `referenceValueA` and `referenceValueB` by themselves.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+            raise ValueError(
+                "At most one of {0} parameters can be set and the remaining parameters must be None".format(
+                    tuple(
+                        uppercase_parameter_name.lower()
+                        for uppercase_parameter_name in enums.LCRReferenceValueType.__members__
+                    )
+                )
+            )
+
+    def __repr__(self):
```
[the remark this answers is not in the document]
olsl21: Yes, it was mentioned by Shreyas in https://github.com/ni/nimi-python/pull/1702#discussion_r920476978 and there is now a unit test case which checks that the reconstruction of the `LCRLoadCompensationSpot` object from its repr string is possible.

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: `amps` is actually the convention used across the API documentations for this custom struct, e.g. in LabVIEW:
image

In fact, the documentations for the fields in LCRMeasurement are largely based on the corresponding LabVIEW documentations for these fields if applicable. For consistency, I believe we should stick to the same wordings across different APIs whenever possible.

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: As mentioned in https://github.com/ni/nimi-python/pull/1702#discussion_r921843495, `Hz` is actually the convention used across the API documentations for this custom struct (e.g. LabVIEW).

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: As mentioned in https://github.com/ni/nimi-python/pull/1702#discussion_r921843495, `amps` is actually the convention used across the API documentations for this custom struct (e.g. LabVIEW).

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: As mentioned in https://github.com/ni/nimi-python/pull/1702#discussion_r921843495, `complex impedance` is actually the wording used across the API documentations for this custom struct (e.g. LabVIEW).

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: The `channel` field would not be populated through the constructor as `LCRMeasurement` objects would be converted directly from the `struct_NILCRMeasurement` objects returned by the driver before the additional Python code retrieves the channel names separately from the driver and populate the `channel` field of the already-constructed `LCRMeasurement` objects accordingly. Kindly refer to https://github.com/ni/nimi-python/pull/1771/files#diff-31582e14c636470681a52cfa45ebf6f28dbd68afde8827d6058a43022b3231c4R4921 for more details of how this is done.

Based on your previous comments, we should remove code paths that would not be used, so I believe `channel` should not be made a parameter to `__init__()`.

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: Nope, `data` is not stored as a member. `self.d` is not `data`, instead it is referring to dissipation factor, and it cannot be used to derive the other fields.

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: These properties were specifically requested by Kunal:

"The helper functions also need to be part of the API as our specs are really tied to the 'derived measurements' not the complex versions."

> Otherwise, why not have properties to return these as rectangular values?

The `z` member itself is a rectangular value (it is stored as a complex number). `y` is created as a property that returns a rectangular value (complex number).

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: Are you mistaking `self.d` as `data`? `self.d` here is referring to dissipation factor, which is the reciprocal of `q`.

`q` is not stored as a member of `LCRMeasurement` as decided during our previous LCR Python API review meeting, so it has to be calculated / derived on-demand here.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+        if self.reference_value_type == enums.LCRReferenceValueType.IMPEDANCE.value:
+            self.reference_value_a = data.reference_value.real
+            self.reference_value_b = data.reference_value.imag
+        else:
+            self.reference_value_a = data.reference_value
+            self.reference_value_b = 0.0
+
+
+class LCRLoadCompensationSpot(object):
+    """Constructs a DUT specification for a given frequency to use in LCR load compensation."""
```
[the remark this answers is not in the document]
olsl21: This is meant for the class. I have updated the docstring to use the word "Specifies" instead of "Constructs" so that it sounds less like a docstring for constructor.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+        ideal_inductance=None,
+        ideal_resistance=None,
+    ):
+        """LCRLoadCompensationSpot
+
+        Creates and returns an LCRLoadCompensationSpot object. At most one of impedance,
+            ideal_capacitance, ideal_inductance and ideal_resistance can be set, and the remaining
+            parameters (excluding frequency) must be None. The parameter that is not None specifies
+            the known specification value of the DUT to be used as the basis for load compensation.
+            If all of them are None, then the default value of `impedance=complex()` will be used.
```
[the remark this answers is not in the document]
olsl21: I have updated this docstring based on your suggestion in my latest commit. Note that I've removed the statement related to default value since the code no longer sets a default value when all the parameters are None.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+            self.reference_value_a = data.reference_value
+            self.reference_value_b = 0.0
+
+
+class LCRLoadCompensationSpot(object):
+    """Constructs a DUT specification for a given frequency to use in LCR load compensation."""
+
+    def __init__(
+        self,
+        frequency=0.0,
```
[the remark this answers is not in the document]
olsl21: Yes, it's required. I've removed the default value for it in my latest commit.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+        Creates and returns an LCRLoadCompensationSpot object. At most one of impedance,
+            ideal_capacitance, ideal_inductance and ideal_resistance can be set, and the remaining
+            parameters (excluding frequency) must be None. The parameter that is not None specifies
+            the known specification value of the DUT to be used as the basis for load compensation.
+            If all of them are None, then the default value of `impedance=complex()` will be used.
+
+        Args:
+            frequency (float): Specifies the spot frequency, in Hz.
+
+            impedance (complex): Specifies the actual impedance of your DUT to be used as the basis
```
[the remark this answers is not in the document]
olsl21: Yes, I have updated it in my latest commit.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+                for load compensation, or None to use another type of DUT specification value.
+
+            ideal_capacitance (float): Specifies the ideal capacitance of your DUT to be used as the
+                basis for load compensation, or None to use another type of DUT specification value.
+
+            ideal_inductance (float): Specifies the ideal inductance of your DUT to be used as the
```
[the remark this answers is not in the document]
olsl21: Yes, I have updated it in my latest commit.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+                basis for load compensation, or None to use another type of DUT specification value.
+
+            ideal_resistance (float): Specifies the ideal resistance of your DUT to be used as the
```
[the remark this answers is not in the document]
olsl21: Yes, I have updated it in my latest commit.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,118 @@
...
+                basis for load compensation, or None to use another type of DUT specification value.
+        """
+        self.frequency = frequency
+        # Set default values
+        self.reference_value_type = enums.LCRReferenceValueType.IMPEDANCE
+        self.reference_value = complex()
```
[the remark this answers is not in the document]
olsl21: Giving default values to custom type members is actually based on what the other existing custom types in other drivers do (e.g. https://github.com/ni/nimi-python/blob/master/src/niscope/custom_types/measurement_stats.py#L2).

Nevertheless, since we are going with the approach of having no unnecessary default values here, I have removed the setting of default values in my latest commit. I have also updated the input validation code to throw an error if all the 4 input parameters (excluding `frequency`) are None.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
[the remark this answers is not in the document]
olsl21: I have removed the type validation in my latest commit.

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: The field name should be "d" to be consistent with the public nidcpower.h header
```
ViReal64 D;
```

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: > Flat is better than nested.

This seems to contradict with our discussions during our previous LCR Python API review meeting (based on what I recall), as the initial design for this custom struct had all the members laid out flatly, but then we decided to group related fields into namedtuples (e.g. `series_lcr` and `parallel_lcr`). 

Anyway, I have split this into two bools in my latest commit.

src/nidcpower/unit_tests/test_nidcpower.py
```diff
@@ -0,0 +1,319 @@
+import cmath
+import pytest
+
+import nidcpower
+
+
```
sbethur, line 6: Add unit tests to validate the math in the properties of `LCRMeasurement`.

sbethur (approved):

generated/nidcpower/nidcpower/lcr_measurement.py
marcoskirsch: I've spoken to Kunal and Sien. I understand the concern. I don't like to have redundant functionality but apparently portions of the user base cannot be trusted to know to extract the magnitude of a complex number.

However, there is redundant implementation here - the driver runtime already populates all these things, no need to recompute.

Consider chaning the implementation of this class so that there is minimal Python code getting executed and you rely more heavily in the values already returned by the driver runtime. That way there's less codepaths and less opportunities for bugs.

marcoskirsch (changes requested):

src/nidcpower/unit_tests/test_nidcpower.py
marcoskirsch: I suggested elsewhere to remove the math and that way we don't need to validate it.
All the math is done by the driver runtime and passed in in `data`.

I suggest just keeping `data` as a private member and implementing the exposed properties in term of that. That way members aren't creating until and only if needed.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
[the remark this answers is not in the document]
marcoskirsch: The people with the design context of the Python APIs (Shreyas, Mark, myself) were not involved in that meeting nor understand the design goals and tradeoffs of nimi-python.

I don't think this is worth it. `__init__` should take in `frequency`, `reference_value_type` and `reference_value`.

generated/nidcpower/nidcpower/lcr_measurement.py
marcoskirsch: This is going to get a bit philosophical. Per OOP best practices, an object should be in an internally consistent state after construction.

I argue that a channel of "" puts the object in an internally inconsistent state so it'd be nice to avoid.

If it is reasonable to do so, see if the client code can delay construction of the LCRMeasurement until it has the channel. If it is not reasonable to do so because it complicates the code, then reply and resolve this again.

generated/nidcpower/nidcpower/lcr_measurement.py
marcoskirsch: I did mistake it. But I do think it's a good idea to store _data and remove calculations done in this code.
As much as possible, this is a thin wrapper over the C API and the math should be kept to an absolute minimum. In this case, we can have the same desired API without doing the math because the info is already in `data` as calculated by the driver runtime.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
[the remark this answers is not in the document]
olsl21: I have updated the constructor design to the latest suggested version in my latest commit.

generated/nidcpower/nidcpower/lcr_measurement.py
olsl21: The current implementation for `fetch_multiple_lcr` and `measure_multiple_lcr` has the public method calls into the private method (which is codegen-ed) to retrieve and construct the LCRMeasurement instances before populating the `channel` fields with the respective channel names in the public method. The codegen logic does not currently support passing in additional parameter(s) in the conversion from the ctypes instances to the Python instances of `LCRMeasurement`, so it will be very complicated to delay the construction of the LCRMeasurement until it has the channel.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,114 @@
...
+            self.reference_value_type.name.lower(),
+            self.reference_value,
+        )
+
+    def __str__(self):
+        return "".join(
+            [
+                "Frequency           : {:,.6g}\n".format(self.frequency),
+                "Reference Value Type: {:}\n".format(self.reference_value_type.name),
+                "Reference Value     : {:,.6g}\n".format(self.reference_value),
```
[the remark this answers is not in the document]
olsl21: I have added a private dict `_lcr_reference_value_type_to_label_and_units` to the `LCRLoadCompensationSpot` class and updated the `__str__()` method to show something like `"Impedance        : 3+4j ohms"`

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,237 @@
...
+        ("measurement_mode"  , "Measurement mode"                                                    ),  # noqa: E202,E203
+        ("dc_in_compliance"  , "DC in compliance"                                                    ),  # noqa: E202,E203
+        ("ac_in_compliance"  , "AC in compliance"                                                    ),  # noqa: E202,E203
+        ("unbalanced"        , "Unbalanced"                                                          ),  # noqa: E202,E203
+    ]
+
+    def __init__(self, data):
+        """LCRMeasurement
+
+        Creates and returns an LCRMeasurement object.
```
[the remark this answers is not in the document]
olsl21: I have updated it in my latest commit.

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,237 @@
...
+        ("reserved5"           , nidcpower._visatype.ViReal64 ),  # noqa: E202,E203
+        ("reserved6"           , nidcpower._visatype.ViReal64 ),  # noqa: E202,E203
+        ("reserved7"           , nidcpower._visatype.ViReal64 ),  # noqa: E202,E203
+    ]
+
+
+class LCRMeasurement(object):
+    """Specifies an LCR measurement.
+
+    Fields:
```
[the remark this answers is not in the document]
olsl21: I have updated it in my latest commit.

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,237 @@
...
+        """z_magnitude_and_phase
+
+        Returns a tuple of (z_magnitude, z_phase) of this LCRMeasurement object.
+
+        Returns:
+            z_magnitude_and_phase (tuple of (float, float)):
+
+                - **z_magnitude** (float): The magnitude of the complex impedance, in ohms.
+                - **z_phase** (float): The impedance phase angle, in degrees.
+
```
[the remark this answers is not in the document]
olsl21: I have updated it according to your suggestion in my latest commit.

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,237 @@
...
+    @property
+    def y(self):
+        """y
+
+        Returns the admittance of this LCRMeasurement object.
+
+        Returns:
+            y (complex): The complex admittance.
+
```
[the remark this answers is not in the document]
olsl21: I have updated it according to your suggestion in my latest commit.

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        if self.z == 0:
```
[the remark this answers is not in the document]
olsl21: I have removed the calculations in this class based on Marcos' suggestion https://github.com/ni/nimi-python/pull/1702#discussion_r922560478, so this is no longer relevant.

src/nidcpower/custom_types/lcr_measurement.py
```diff
@@ -0,0 +1,237 @@
...
+        """y_magnitude_and_phase
+
+        Returns a tuple of (y_magnitude, y_phase) of this LCRMeasurement object.
+
+        Returns:
+            y_magnitude_and_phase (tuple of (float, float)):
+
+                - **y_magnitude** (float): The magnitude of the complex admittance, in siemens.
+                - **y_phase** (float): The admittance phase angle, in degrees.
+
```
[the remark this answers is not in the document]
olsl21: I have updated it according to your suggestion in my latest commit.

src/nidcpower/unit_tests/test_nidcpower.py
olsl21: I have followed Marcos' suggestion to store `data` as a private member and removed all the calculations in this class.

src/nidcpower/custom_types/lcr_load_compensation_spot.py
```diff
@@ -0,0 +1,114 @@
...
+        self,
+        frequency,
+        impedance=None,
+        ideal_capacitance=None,
+        ideal_inductance=None,
+        ideal_resistance=None,
+    ):
+        """LCRLoadCompensationSpot
+
+        Creates and returns an LCRLoadCompensationSpot object. One and only one of impedance,
```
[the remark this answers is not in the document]
olsl21: I have updated it according to your suggestion in my latest commit.

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,237 @@
...
+    ]
+
+
+class LCRMeasurement(object):
+    """Specifies an LCR measurement.
+
+    Fields:
+        channel (str): The channel name associated with this LCRMeasurement.
+
+        vdc (float): Specifies the measured DC voltage, in volts.
```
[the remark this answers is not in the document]
olsl21: I have removed the word "Specifies" from the docstring for all the data attributes in this class.

src/nidcpower/custom_types/lcr_load_compensation_spot.py
```diff
@@ -36,79 +36,66 @@ def __init__(self, data):
...
 
-    def __init__(
-        self,
-        frequency,
-        impedance=None,
-        ideal_capacitance=None,
-        ideal_inductance=None,
-        ideal_resistance=None,
-    ):
+    _lcr_reference_value_type_to_label_and_units = {
```
sbethur, line 39: If we're adding units to string representation of these, then we should do the same for `frequency` attribute of this struct and also for all the attributes of `LCRMeasurement` struct.

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+
+    @property
+    def q(self):
+        """q
+
+        Get the quality factor (float) of the circuit. The dimensionless quality factor is inversely
+        proportional to the degree of damping in a system. Q is the reciprocal of D, the dissipation
+        factor.
+        """
+        return self._data.q
```
marcoskirsch, line 183: Why are some properties implemented this way (`@property` decorator) and others by setting an internal member during `__init__()` (like `self.ac_voltage`)?

The inconsistency is cofusing if needless.

generated/nidcpower/nidcpower/lcr_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+        ("reserved5"           , nidcpower._visatype.ViReal64 ),  # noqa: E202,E203
+        ("reserved6"           , nidcpower._visatype.ViReal64 ),  # noqa: E202,E203
+        ("reserved7"           , nidcpower._visatype.ViReal64 ),  # noqa: E202,E203
+    ]
+
+
+class LCRMeasurement(object):
+    """Specifies an LCR measurement.
+
+    Data attributes:
```
marcoskirsch, line 54: Not all the supported properties are documented here.
It looks like only those that are implemented by setting `_self.foo` are.
But that's an implementation detail as far as I understand and irrelevant to the user.

marcoskirsch (changes requested):

src/nidcpower/custom_types/lcr_load_compensation_spot.py
```diff
@@ -36,79 +36,66 @@ def __init__(self, data):
...
-                specification value.
-
-            ideal_inductance (float): Specifies the ideal inductance, in henrys, of your DUT to be
-                used as the basis for load compensation, or None to use another type of DUT
-                specification value.
-
-            ideal_resistance (float): Specifies the ideal resistance, in ohms, of your DUT to be
-                used as the basis for load compensation, or None to use another type of DUT
-                specification value.
+            reference_value (complex or float): A value that describes the referenceValueType
```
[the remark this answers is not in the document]
sbethur: Tracked by #1783.

src/nidcpower/custom_types/lcr_load_compensation_spot.py
sbethur: Tracked by #1782

generated/nidcpower/nidcpower/lcr_measurement.py
sbethur: Tracked by #1781.

generated/nidcpower/nidcpower/lcr_measurement.py
sbethur: Tracked by #1781.

PR #1703, opened 2022-03-31
Implement `_fix_custom_type()`

olsl21:

- Implement `_fix_custom_type()` to add a "struct_" prefix to a parameter type name if necessary to match a custom type's ctypes_type
  - This allows custom type to be specified as the type of a function parameter without adding a "struct " prefix to the type name

- All the build tests (including the updated ones) passed


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ All notable changes to this project will be documented in this file.
 * ### ALL
     * #### Added
     * #### Changed
+        * Support specifying custom type for function parameter without "struct " prefix
```
marcoskirsch, line 35: only publicly facing changes should be in this file
In this case you can remove

marcoskirsch: #1713

build/helper/metadata_add_all.py
```diff
@@ -297,6 +297,15 @@ def _fix_type(parameter):
     parameter['type'] = parameter['type'].replace('[ ]', '[]').replace(' []', '[]').replace(' ', '_')
 
 
+def _fix_custom_type(parameter, config):
+    '''Add "struct_" prefix to custom type if necessary to match its ctypes_type.'''
```
marcoskirsch, line 301: This function is "monkey patching" the metadata. It's hacky. We have full control of the metadata.

Do we want to enforce a convention that a custom type and the internal ctypes type must have the same name except for the prefix? Then let's add a check and fix the metadata instead. It's more work but more correct in the long term.

If so, then let's add a strict fix

marcoskirsch (changes requested):
I'd like to discuss the approach before proceeding.

CHANGELOG.md
olsl21: Thanks for pointing this out, I have reverted the changes on CHANGELOG.md with the latest commit.

build/helper/metadata_add_all.py
olsl21: Following our offline discussion, we have agreed to proceed with the current approach as a temporary workaround to unblock the LCR-related PRs. We shall update the codegen logic for handling custom types once the metadata have been fixed from the upstream.

build/helper/metadata_add_all.py
```diff
@@ -297,6 +297,15 @@ def _fix_type(parameter):
     parameter['type'] = parameter['type'].replace('[ ]', '[]').replace(' []', '[]').replace(' ', '_')
 
 
+def _fix_custom_type(parameter, config):
```
marcoskirsch, line 300: Please add a comment here, along the lines of:

> TODO(olsl21): Metadata is inconsistent with regards to how structs are treated.
> This is a hack to workaround that inconsistency. The real root caused is tracked
> by internal NI bug 1918101. Once that is addressed, this method can be removed.

marcoskirsch (changes requested):

build/helper/metadata_add_all.py
olsl21: I've added the TODO comment in the latest commit.

marcoskirsch (approved):

sbethur (approved):

PR #1653, opened 2021-09-08
Adding azure pipeline for nimi-bot pytests

SukruthGrandhiNI:
- [ ] ~~This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).~~

- [ ] ~~I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~

- [ ] ~~I've added tests applicable for this pull request~~

### What does this Pull Request accomplish?

In this Pull Request added a functionality to run system tests using azure pipelines.
This azure pipeline will be run for every new PR raised in this repository.
Example- https://ni.visualstudio.com/Users/_build/results?buildId=1312088&view=results

Note: we are using Azure pipelines instead of github actions for the security reasons. In future we will change to github actions after the vms are created on the AWS cloud(Non NI network). Security constraints are captured as comments in https://ni.visualstudio.com/DevCentral/_sprints/backlog/MTO%20NIB/DevCentral/21Q3/09?workitem=1534568

Feature specifications.
1. Total 9 modules system tests are run in the pipeline. Please refer the above link for the list of modules.
2. Systems tests for each OS will run parallelly using staging. Also, each module system tests are ran parallel in a given OS using Jobs. 
2.  Each module system tests are run for python 36,37,38 and 39 versions.
3.  Each operating system are defined as a separate stage in the pipeline, this PR has changes only for testing on windows 10 OS.
4.  To reduce pipeline build time, in a single machine there are 4 agents created. That means 4 modules system tests can run parallel.
5. Module system tests are run only after validating the codegen.
6. Added new agent pool called "nimi-bot"(https://ni.visualstudio.com/_settings/agentpools?poolId=474&view=agents), all the agents related to nimi-bot are hosted here.
7. Created two stages, first stage will verify codegen is up-to date, in the second stage we are testing system tests for win64.

Below tasks are addressed after submitting the changes
1. Need to create a official azure pipeline which is linked to master branch of nimi-bot.
2. If the PR is raised by external contributor then pipeline build should start only after approval from the owner of this repo. This is done for the sake of security.
3. Adding new stage for Linux OS - OS Flavor To be defined.(Expected within this iteration).
4. Adding new stage for win32(python 32).
5. Currently agent machines in NI domain are created manually- this need to automated using IaC(Expectation for windows agents within this iteration).
6. Adding the pipeline build status badge to the readme file.

[...]

PR #1719, opened 2022-04-08
Add new script to check if codegen is up to date

### LINKED ISSUE #1674: Automated check missing for ensuring PRs include "generated" changes to go with "source" changes
### Description of issue

For reasons (mainly to make code review and repo maintenance easier), generated code and docs are part of source control. When an author makes changes to metadata or code generator logic, they are also supposed to run the code generator and include the updated generated files in the PR. The only check we have for this is to rely on code reviewers to catch any oversights, which is not 100% foolproof.

We need a step (preferably as part of Travis CI) that automatically checks this.

ayesham-729:

- ~~[ ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~

- ~~[ ] I've added tests applicable for this pull request~~

This update adds a Python script to check if contributors who have changed files related to codegen have run code generation and committed the updated files. The Python script was originally written by Sukruth Grandhi in this [pull request](https://github.com/ni/nimi-python/pull/1653/commits/5219ed497a67426440c048cd22098adbe8dace95), with some additional updates added.

Travis.yml was updated to call this script before running the tox command. This adds the codegen check to TravisCI.

* Fix #1674 

Tested the script locally by committing a file without the corresponding codegen changes and checking if the script caught the error.


marcoskirsch:
Description has a bunch of TODOs, please address and delete.

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,97 @@
...
+    Add the codegen files to git staging area.
+    No files will be added to git staging area if
+      1) No new files are generated
+      2) No existing files are modified/deleted.
+    """
+    if os.system('git add .') != 0:
+        print("Unable to add files to git", file=sys.stderr)
+        sys.exit(-1)
+
+def commit_git_staging_area_files():
```
marcoskirsch, line 62: don't add/commit/reset files! That's crazy.

`git status -s -uno | wc -l`
Will tell you how many dirty files there are. Just check that it's 0.

marcoskirsch (changes requested):

tools/ensure_codegen_up_to_date.py
ayesham-729: Updated to use this command instead

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,64 @@
...
+
+def create_codegen_files():
+    """
+    create codegen files
+    """
+    if os.system("tox -e codegen") != 0:
+        print("Unable to generate code using \"tox -e codegen\"", file=sys.stderr)
+        sys.exit(-1)
+
+def check_file_status():
```
marcoskirsch, line 43: The name of this function does not describe what the function does. That's bad. 
I myself aren't sure what it does. It appears to "return the number of modified uncommitted files".
If that's the case, then you should rename it. Something like `get_number_of_modified_files` or something.

But I think it's better to instead change the function to `check_no_dirty_files` and have it return the error message and error inside the body of the function - because that's what you do in all the other functions.

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,64 @@
...
+    Checks if there are any modified files
+    '''
+    return int(subprocess.check_output("git status -s -uno | wc -l", shell=True).decode().strip("b'\\n'"))
+
+if __name__ == "__main__":
+    """
+    Main
+    """
+    # Configure git credentials
+    configure_git_user_email()
```
marcoskirsch, line 54: consider making `configure_git_user_email` and `configure_git_user_name` one function `configure_git_credentials`.

That makes the comment obsolete and makes reading `__main__` nicer.

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,64 @@
...
+    configure_git_user_name()
+
+    clean_codegen_files()
+    create_codegen_files()
+
+    if check_file_status() != 0:
+        print("Run codegen and include the generated files in the PR", file=sys.stderr)
```
marcoskirsch, line 61: Expand on this message to follow format: "X problem was found. Do Y to fix it"; rather than just "Do Y".

Suggestion:

> Code-generated files do not match what's committed to Git. Please run code-generation and commit all changed files to Git.

.travis.yml
```diff
@@ -13,6 +13,9 @@ install:
   - travis_retry pip install --upgrade pip
   - travis_retry pip install --upgrade tox tox-travis codecov
 
+before_script:
+  - python tools//ensure_codegen_up_to_date.py
```
marcoskirsch, line 17: i'm surprised you need 2 forward slashes.

marcoskirsch (changes requested):

.travis.yml
ayesham-729: The build failed previously when I used one backslash, so Shreyas suggested using forward slashes and escaping.

marcoskirsch: Backslash fails because this does not run on a Windows system. Backslashes need escaping in some circumstances.

This is a forward slash.

Can you verify you need it twice?

tools/ensure_codegen_up_to_date.py
ayesham-729: Do you mean that the check_no_dirty_files function would have this section in its body instead of in main?
`print("Run codegen and include the generated files in the PR", file=sys.stderr)
        sys.exit(-1)`

.travis.yml
ayesham-729: I can try a run with one forward slash

tools/ensure_codegen_up_to_date.py
marcoskirsch: Yes, so that `__main__` is straightforward and readable and all the helper methods are where the `sys.exit` occurs.

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,64 @@
...
+
+def clean_codegen_files():
+    """
+    Before code generation clean the existing codegen files
+    """
+    if os.system("tox -e clean") != 0:
+        print("Unable to clean the repo using \"tox -e clean\"", file=sys.stderr)
+        sys.exit(-1)
+
+def create_codegen_files():
```
marcoskirsch, line 35: prepend all helper functions with `_` which is a convention for "don't call this from the outside it's private".

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+import sys
+import subprocess
+
+def configure_git_user_email():
+    """
+    Configures git user email with dummy email id
+    """
+    if os.system('git config user.email "dummy@ni.com"') != 0:
+        print("Unable to configure \"user email\" using git", file=sys.stderr)
```
marcoskirsch, line 16: Consider adding `Error: ` to all these prints to make them stand out better in the log.

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,64 @@
...
+        print("Unable to configure \"user name\" using git", file=sys.stderr)
+        sys.exit(-1)
+
+def clean_codegen_files():
+    """
+    Before code generation clean the existing codegen files
+    """
+    if os.system("tox -e clean") != 0:
+        print("Unable to clean the repo using \"tox -e clean\"", file=sys.stderr)
+        sys.exit(-1)
```
marcoskirsch, line 33: See https://docs.python.org/3/library/sys.html#sys.exit

Pass the string instead of the number to `sys.exit`. This is:
* briefer
* returns 1 not -1 which is a convention for unix
* prints to stderr not stdout which is more correct

marcoskirsch (changes requested):

tools/ensure_codegen_up_to_date.py
ayesham-729: Updated

tools/ensure_codegen_up_to_date.py
ayesham-729: Updated

tools/ensure_codegen_up_to_date.py
ayesham-729: Updated

tools/ensure_codegen_up_to_date.py
```diff
@@ -0,0 +1,70 @@
...
+if __name__ == "__main__":
+    """
+    Main
+    """
+    configure_git_credentials()
+
+    clean_codegen_files()
+    create_codegen_files()
+
+    convert_all_files_to_unix_line_endings()
```
marcoskirsch, line 69: Can you explain why this is needed?
I suspect the correct fix to this issue is in `git config`.
Simply cloning the repo and running codegen should not leave you with files different than what is committed but that appears to be what you are correcting for in this function.

marcoskirsch (changes requested):

tools/ensure_codegen_up_to_date.py
ayesham-729: When codegen runs in the TravisCI build, nidigital/session.py's line endings change. For some reason, it only happens to that file. It's causing the CI to fail, so Shreyas suggested working around it by normalizing the line endings to LF and running the command on all the files for simplicity.

sbethur: Since the workaround was proving difficult, I went ahead and root-caused the issue and put up a fix: https://github.com/ni/nimi-python/pull/1728

With that, you don't need the workaround.

tools/ensure_codegen_up_to_date.py
ayesham-729: Updated

tools/ensure_codegen_up_to_date.py
ayesham-729: TravisCI is now passing

marcoskirsch (approved):

sbethur (approved):

PR #1733, opened 2022-04-18
Update return type of NI-FGEN datetime-related functions

### LINKED ISSUE #1716: The return value of `get_date_and_time()` related function(s) across drivers should be consistently named
### Description of issue

Currently, the return value of `get_date_and_time()` related function(s) is correctly named "last_cal_datetime" in NI-SCOPE, but incorrectly named "month" in NI-DCPower, NI-DMM, NI-FAKE and NI-FGEN. They should be made consistent.

This issue is mentioned in #1707.

ayesham-729:

This pull request fixes a documentation issue where some functions in NI-FGEN that returned calibration dates returned `month` instead of `last_cal_datetime`. This has been updated to be consistent with SCOPE and DC-Power. 

Part of fix for issue "The return value of get_date_and_time() related function(s) across drivers should be consistently named"

Confirmed that generated files show `last_cal_datetime` as the return value instead of `month`.

### DIFF
--- a/generated/nifgen/nifgen/_library.py
+++ b/generated/nifgen/nifgen/_library.py
@@ -387,21 +387,21 @@ class Library(object):
                 self.niFgen_GetHardwareState_cfunc.restype = ViStatus  # noqa: F405
         return self.niFgen_GetHardwareState_cfunc(vi, state)
 
-    def niFgen_GetLastExtCalLastDateAndTime(self, vi, month):  # noqa: N802
+    def niFgen_GetLastExtCalLastDateAndTime(self, vi, last_cal_datetime):  # noqa: N802
         with self._func_lock:
             if self.niFgen_GetLastExtCalLastDateAndTime_cfunc is None:
                 self.niFgen_GetLastExtCalLastDateAndTime_cfunc = self._get_library_function('niFgen_GetLastExtCalLastDateAndTime')
                 self.niFgen_GetLastExtCalLastDateAndTime_cfunc.argtypes = [ViSession, ctypes.POINTER(hightime.datetime)]  # noqa: F405
                 self.niFgen_GetLastExtCalLastDateAndTime_cfunc.restype = ViStatus  # noqa: F405
-        return self.niFgen_GetLastExtCalLastDateAndTime_cfunc(vi, month)
+        return self.niFgen_GetLastExtCalLastDateAndTime_cfunc(vi, last_cal_datetime)
 
-    def niFgen_GetLastSelfCalLastDateAndTime(self, vi, month):  # noqa: N802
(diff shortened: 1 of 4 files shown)


sbethur:
@ayesham-729, please update the PR title

ayesham-729:
@sbethur do I need to update the changelog?

src/nifgen/metadata/functions.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-FGEN API metadata version 21.5.0d106
+# This file is generated from NI-FGEN API metadata version 22.0.0d9999
```
marcoskirsch, line 2: Let's try to grab this from a real internal NI build artifact.
If you cannot, then let's temporarily override the name through the add-on.

generated/nifgen/nifgen/_library.py
```diff
@@ -387,21 +387,21 @@ def niFgen_GetHardwareState(self, vi, state):  # noqa: N802
                 self.niFgen_GetHardwareState_cfunc.restype = ViStatus  # noqa: F405
         return self.niFgen_GetHardwareState_cfunc(vi, state)
 
-    def niFgen_GetLastExtCalLastDateAndTime(self, vi, month):  # noqa: N802
+    def niFgen_GetLastExtCalLastDateAndTime(self, vi, last_cal_datetime):  # noqa: N802
```
marcoskirsch, line 390: There is something wrong in our code generator that it creates this wrapper function for something that doesn't have a C function. @sbethur do we have any easy way to tell that the functions shouldn't be getting generated? `'python-only'`?

marcoskirsch (changes requested):

generated/nifgen/nifgen/_library.py
sbethur: Yes, we should use 'python-only`, I'll fix it.

sbethur: Fixed it in internal metadata. @ayesham-729, please update this PR with the latest export of internal metadata.

sbethur:
@ayesham-729, GitHub auto-closes an issue if it can find a string like "Fix #xyz" in the PR description. This is pretty cool, but in this case, we don't want that to happen since the referenced issue also talks about needing fixes in other nimi-python modules.

src/nifgen/metadata/functions.py
ayesham-729: Updated

generated/nifgen/nifgen/_library.py
ayesham-729: Updated

marcoskirsch (approved):

PR #1738, opened 2022-05-06
NI-DCPower: Create AttributeEnumWithConverters, enable `isolation_state`

olsl21:
- [X] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~

- Fix `_add_enum_codegen_method()` to respect user-configured enum `codegen_method`
  - If the 'codegen_method' of an enum has been explicitly set to a value other than 'public' and it is used by a 'public'/'python-only' function or 'public' attribute, a ValueError would be thrown
- Add `get_enum_value_snippet()` and `enum_uses_converter()` helper functions to build helper
  - Add unit tests for them
- Add `AttributeEnumWithConverter` to _attributes.py.mako
  - Throw `DriverTooNewError` if there is a conversion error in the `getter_converter` (due to driver returning invalid value)
  - Add NI-FAKE unit tests for the `enum_with_converter` attribute
- Add codegen logic to _converters.py.mako to codegen the converters for enums with converter
- Update session.py.mako to handle enums with converter
- Update metadata for IsolationState and the generated files accordingly (update NI-DCPower metadata to 22.5.0d71)

Manually tested with locally updated metadata, the "isolation state" attribute can be set/get correctly using bool data type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,12 @@ All notable changes to this project will be documented in this file.
                 * `lcr_dc_bias_source`
                 * `lcr_dc_bias_voltage_level`
                 * `lcr_frequency`
+                * `lcr_impedance_range`
+                * `lcr_impedance_range_source`
+                * `lcr_load_capacitance`
                 * `lcr_load_compensation_enabled`
+                * `lcr_load_inductance`
+                * `lcr_load_resistance`
                 * `lcr_measured_load_reactance`
                 * `lcr_measured_load_resistance`
                 * `lcr_measurement_time`
@@ -73,6 +78,7 @@ All notable changes to this project will be documented in this file.
                 * `InstrumentMode`
                 * `LCRCompensationType`
                 * `LCRDCBiasSource`
+                * `LCRImpedanceRangeSource`
                 * `LCRMeasurementTime`
                 * `LCROpenShortLoadCompensationDataSource`
                 * `LCRSourceDelayMode`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,7 +83,6 @@ All notable changes to this project will be documented in this file.
                 * `LCROpenShortLoadCompensationDataSource`
                 * `LCRSourceDelayMode`
                 * `LCRStimulusFunction`
-                * `_IsolationState`
             * Methods added:
                 * `configure_lcr_custom_cable_compensation`
                 * `get_lcr_compensation_last_date_and_time`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ All notable changes to this project will be documented in this file.
                 * `autorange_maximum_delay_after_range_change`
                 * `cable_length`
                 * `instrument_mode`
+                * `isolation_state`
                 * `lcr_actual_load_reactance`
                 * `lcr_actual_load_resistance`
                 * `lcr_automatic_level_control`
@@ -76,6 +77,7 @@ All notable changes to this project will be documented in this file.
                 * `LCROpenShortLoadCompensationDataSource`
                 * `LCRSourceDelayMode`
                 * `LCRStimulusFunction`
+                * `_IsolationState`
             * Methods added:
                 * `configure_lcr_custom_cable_compensation`
                 * `get_lcr_compensation_last_date_and_time`
--- a/build/helper/codegen_helper.py
+++ b/build/helper/codegen_helper.py
@@ -485,6 +485,11 @@ def get_dictionary_snippet(d, indent=4):
     return ('\n' + (' ' * indent)).join(d_lines)
 
 
+def get_enum_value_snippet(value):
+    '''Returns value formatted into string, surrounding it with single quotes if it is of str type'''
+    return ("'{}'" if type(value) is str else "{}").format(value)
+
+
 #
 # Let's do some unit testing!
 #
@@ -1675,3 +1680,9 @@ def test_get_ctype_variable_declaration_snippet_bad_ivi_dance_step():
 # TODO(marcoskirsch): unit tests for reamining cases of get_ctype_variable_declaration_snippet(): parameter is a buffer.
 
 
+def test_get_enum_value_snippet():
+    assert get_enum_value_snippet(None) == 'None'
+    assert get_enum_value_snippet(True) == 'True'
+    assert get_enum_value_snippet(0) == '0'
+    assert get_enum_value_snippet('True') == "'True'"
+    assert get_enum_value_snippet('0') == "'0'"
--- a/build/helper/helper.py
+++ b/build/helper/helper.py
@@ -137,9 +137,19 @@ def enum_uses_converter(enum):
     '''Returns True if enum uses converter, False otherwise.
 
     An enum uses converter if it has both 'enum_to_converted_value_function_name' and
-    'converted_value_to_enum_function_name' defined which are not None.
+    'converted_value_to_enum_function_name' defined which are not None. If one of them is defined
+    and not None but not the other, an AssertionError would be thrown.
     '''
-    return enum.get('enum_to_converted_value_function_name', None) is not None and enum.get('converted_value_to_enum_function_name', None) is not None
+    has_enum_to_converted_value_function_name = enum.get(
+        'enum_to_converted_value_function_name', None
+    ) is not None
+    has_converted_value_to_enum_function_name = enum.get(
+        'converted_value_to_enum_function_name', None
+    ) is not None
+    assert has_enum_to_converted_value_function_name == has_converted_value_to_enum_function_name, (
+        "An enum must either define both 'converted_value_to_enum_function_name' and 'converted_value_to_enum_function_name', or define none of them"
+    )
+    return has_enum_to_converted_value_function_name and has_converted_value_to_enum_function_name
 
 
 # Tests
@@ -244,13 +254,9 @@ def test_get_development_status():
 
 
 def test_enum_uses_converter():
+    import pytest
+
     assert not enum_uses_converter({})
-    assert not enum_uses_converter({
-        'enum_to_converted_value_function_name': lambda x: x
-    })
-    assert not enum_uses_converter({
-        'converted_value_to_enum_function_name': lambda x: x
-    })
     assert not enum_uses_converter({
         'enum_to_converted_value_function_name': None,
         'converted_value_to_enum_function_name': None
@@ -260,3 +266,22 @@ def test_enum_uses_converter():
         'enum_to_converted_value_function_name': lambda x: x,
         'converted_value_to_enum_function_name': lambda x: x
     })
+
+    with pytest.raises(AssertionError):
+        enum_uses_converter({
+            'enum_to_converted_value_function_name': lambda x: x
+        })
+    with pytest.raises(AssertionError):
+        enum_uses_converter({
+            'converted_value_to_enum_function_name': lambda x: x
+        })
+    with pytest.raises(AssertionError):
+        enum_uses_converter({
+            'enum_to_converted_value_function_name': None,
+            'converted_value_to_enum_function_name': lambda x: x
+        })
+    with pytest.raises(AssertionError):
+        enum_uses_converter({
+            'enum_to_converted_value_function_name': lambda x: x,
+            'converted_value_to_enum_function_name': None
+        })
--- a/build/helper/helper.py
+++ b/build/helper/helper.py
@@ -133,6 +133,15 @@ def get_development_status(config):
     return dev_status
 
 
+def enum_uses_converter(enum):
+    '''Returns True if enum uses converter, False otherwise.
+
+    An enum uses converter if it has both 'enum_to_converted_value_function_name' and
+    'converted_value_to_enum_function_name' defined which are not None.
+    '''
+    return enum.get('enum_to_converted_value_function_name', None) is not None and enum.get('converted_value_to_enum_function_name', None) is not None
+
+
 # Tests
 def test_get_development_status():
     config = {}
@@ -233,3 +242,21 @@ def test_get_development_status():
     config['module_version'] = '19.9.9'
     assert get_development_status(config) == '5 - Production/Stable'
 
+
+def test_enum_uses_converter():
+    assert not enum_uses_converter({})
+    assert not enum_uses_converter({
+        'enum_to_converted_value_function_name': lambda x: x
+    })
+    assert not enum_uses_converter({
+        'converted_value_to_enum_function_name': lambda x: x
+    })
+    assert not enum_uses_converter({
+        'enum_to_converted_value_function_name': None,
+        'converted_value_to_enum_function_name': None
+    })
+
+    assert enum_uses_converter({
+        'enum_to_converted_value_function_name': lambda x: x,
+        'converted_value_to_enum_function_name': lambda x: x
+    })
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -442,8 +442,8 @@ def add_all_attribute_metadata(attributes, config):
 def _add_enum_codegen_method(enums, config):
     '''Adds 'codegen_method' that will determine whether and how the enum is code-generated.
 
-    Set the 'codegen_method' of all enums without any explicit 'codegen_method' to 'no', then go
-    through all functions and attributes and set to the least restrictive use.
+    Set the 'codegen_method' of all enums that do not explicitly specify it in metadata to 'no',
+    then go through all functions and attributes and set to the least restrictive use.
     If an enum without converter has an explicit non-public 'codegen_method' but it is used by a
     'public'/'python-only' function or 'public' attribute, a ValueError would be thrown.
     '''
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -7,6 +7,7 @@ from .documentation_snippets import options_table_header
 from .documentation_snippets import options_text
 from .documentation_snippets import session_return_text
 from .helper import camelcase_to_snakecase
+from .helper import enum_uses_converter
 from .helper import get_numpy_type_for_api_type
 from .helper import get_python_type_for_api_type
 from .metadata_filters import filter_codegen_attributes
@@ -466,7 +467,7 @@ def _add_enum_codegen_method(enums, config):
                         enums[e]['codegen_method'] = f_codegen_method
                     elif f_codegen_method in ('public', 'python-only') and enums[e]['codegen_method'] != 'public':
                         if enum_to_has_explicit_codegen_method[e]:
-                            if not enums[e].get('use_converter', False):
+                            if not enum_uses_converter(enums[e]):
                                 raise ValueError(
                                     "Public / python-only function {} cannot use non-public enum {} (without converter)".format(
                                         f, e
@@ -487,7 +488,7 @@ def _add_enum_codegen_method(enums, config):
                     enums[e]['codegen_method'] = a_codegen_method
                 elif a_codegen_method == 'public' and enums[e]['codegen_method'] != 'public':
                     if enum_to_has_explicit_codegen_method[e]:
-                        if not enums[e].get('use_converter', False):
+                        if not enum_uses_converter(enums[e]):
                             raise ValueError(
                                 "Public attribute {} cannot use non-public enum {} (without converter)".format(
                                     config['attributes'][a]['name'], e
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -440,62 +440,57 @@ def add_all_attribute_metadata(attributes, config):
 
 
 def _add_enum_codegen_method(enums, config):
-    '''Adds 'codegen_method' that will determine whether and how the enum is code genned. Default is public
+    '''Adds 'codegen_method' that will determine whether and how the enum is code-generated.
 
-    Set all to 'no', then go through all functions and attributes and set to least restrictive use.
-    If the 'codegen_method' of an enum has been explicitly set to a value other than 'public' and it
-    is used by a 'public'/'python-only' function or 'public' attribute, a ValueError would be thrown
+    Set the 'codegen_method' of all enums without any explicit 'codegen_method' to 'no', then go
+    through all functions and attributes and set to the least restrictive use.
+    If an enum without converter has an explicit non-public 'codegen_method' but it is used by a
+    'public'/'python-only' function or 'public' attribute, a ValueError would be thrown.
     '''
-    enum_to_has_explicit_codegen_method = {}
+    enum_has_explicit_codegen_method = {}
     for e in enums:
         if 'codegen_method' not in enums[e]:
             enums[e]['codegen_method'] = 'no'
-            enum_to_has_explicit_codegen_method[e] = False
+            enum_has_explicit_codegen_method[e] = False
         else:
-            enum_to_has_explicit_codegen_method[e] = True
+            enum_has_explicit_codegen_method[e] = True
 
-    # Iterate through all codegen functions and set any enum parameters to the same level
+    # Iterate through all codegen functions and set any enum parameters to the same level of codegen_method
     for f in filter_codegen_functions(config['functions']):
         f_codegen_method = config['functions'][f]['codegen_method']
-        if f_codegen_method != 'no':
-            for p in config['functions'][f]['parameters']:
-                e = p['enum']
-                if e is not None and e not in enums:
-                    print('Missing enum {0} referenced by function {1}'.format(e, f))
-                elif e is not None:
-                    if f_codegen_method == 'private' and enums[e]['codegen_method'] == 'no':
-                        enums[e]['codegen_method'] = f_codegen_method
-                    elif f_codegen_method in ('public', 'python-only') and enums[e]['codegen_method'] != 'public':
-                        if enum_to_has_explicit_codegen_method[e]:
-                            if not enum_uses_converter(enums[e]):
-                                raise ValueError(
-                                    "Public / python-only function {} cannot use non-public enum {} (without converter)".format(
-                                        f, e
-                                    )
-                                )
-                        else:
-                            enums[e]['codegen_method'] = 'public'
-
-    # Iterate through all codegen attributes and set any enum parameters to the same level
-    for a in filter_codegen_attributes(config['attributes']):
-        a_codegen_method = config['attributes'][a]['codegen_method']
-        if a_codegen_method != 'no':
-            e = config['attributes'][a]['enum']
+        for p in config['functions'][f]['parameters']:
+            e = p['enum']
             if e is not None and e not in enums:
-                print('Missing enum {0} referenced by attribute {1}'.format(e, a))
+                print('Missing enum {0} referenced by function {1}'.format(e, f))
             elif e is not None:
-                if a_codegen_method == 'private' and enums[e]['codegen_method'] == 'no':
-                    enums[e]['codegen_method'] = a_codegen_method
-                elif a_codegen_method == 'public' and enums[e]['codegen_method'] != 'public':
-                    if enum_to_has_explicit_codegen_method[e]:
+                if f_codegen_method == 'private' and enums[e]['codegen_method'] == 'no':
+                    enums[e]['codegen_method'] = f_codegen_method
+                elif f_codegen_method in ('public', 'python-only') and enums[e]['codegen_method'] != 'public':
+                    if enum_has_explicit_codegen_method[e]:
                         if not enum_uses_converter(enums[e]):
                             raise ValueError(
-                                "Public attribute {} cannot use non-public enum {} (without converter)".format(
-                                    config['attributes'][a]['name'], e
-                                )
+                                f"Enum {e} used by public function {f} must be public, is private"
                             )
                     else:
-                        enums[e]['codegen_method'] = a_codegen_method
+                        enums[e]['codegen_method'] = 'public'
+
+    # Iterate through all codegen attributes and set any enum parameters to the same level of codegen_method
+    for a in filter_codegen_attributes(config['attributes']):
+        a_codegen_method = config['attributes'][a]['codegen_method']
+        e = config['attributes'][a]['enum']
+        if e is not None and e not in enums:
+            print('Missing enum {0} referenced by attribute {1}'.format(e, a))
+        elif e is not None:
+            if a_codegen_method == 'private' and enums[e]['codegen_method'] == 'no':
+                enums[e]['codegen_method'] = a_codegen_method
+            elif a_codegen_method == 'public' and enums[e]['codegen_method'] != 'public':
+                if enum_has_explicit_codegen_method[e]:
+                    if not enum_uses_converter(enums[e]):
+                        raise ValueError(
+                            f"Enum {e} used by public attribute {config['attributes'][a]['name']} must be public, is private"
+                        )
+                else:
+                    enums[e]['codegen_method'] = a_codegen_method
 
 
 def _add_enum_value_python_name(enum_info, config):
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -441,11 +441,17 @@ def add_all_attribute_metadata(attributes, config):
 def _add_enum_codegen_method(enums, config):
     '''Adds 'codegen_method' that will determine whether and how the enum is code genned. Default is public
 
-    Set all to 'no', then go through all functions and attributes and set to least restrictive use
+    Set all to 'no', then go through all functions and attributes and set to least restrictive use.
+    If the 'codegen_method' of an enum has been explicitly set to a value other than 'public' and it
+    is used by a 'public'/'python-only' function or 'public' attribute, a ValueError would be thrown
     '''
+    enum_to_has_explicit_codegen_method = {}
     for e in enums:
         if 'codegen_method' not in enums[e]:
             enums[e]['codegen_method'] = 'no'
+            enum_to_has_explicit_codegen_method[e] = False
+        else:
+            enum_to_has_explicit_codegen_method[e] = True
 
     # Iterate through all codegen functions and set any enum parameters to the same level
     for f in filter_codegen_functions(config['functions']):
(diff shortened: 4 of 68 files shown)


build/templates/_attributes.py.mako
```diff
@@ -104,6 +104,20 @@ class AttributeEnum(object):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+# nidcpower specific attribute type
+class AttributeIsolationState(AttributeViInt32):
```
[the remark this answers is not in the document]
marcoskirsch: It seems like we should be able to solve this generically for attributes of type enum, no?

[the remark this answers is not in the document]
marcoskirsch: or create an Attribute class that takes in a pair of converter functions ?

marcoskirsch (changes requested):

sbethur (commented):
Design looks good to me.

marcoskirsch:
fyi merge conflicts

build/templates/_attributes.py.mako
```diff
@@ -104,6 +104,29 @@ class AttributeEnum(object):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+class AttributeBooleanEnum(object):
```
marcoskirsch, line 107: note the missing newline compared to other classes

build/templates/_attributes.py.mako
```diff
@@ -104,6 +104,29 @@ class AttributeEnum(object):
...
+    def __init__(
+        self,
+        underlying_attribute_meta_class,
```
marcoskirsch, line 110: Why "underlying meta class" and not just "underlying enum"?

build/templates/_attributes.py.mako
```diff
@@ -104,6 +104,29 @@ class AttributeEnum(object):
...
+        attribute_id,
+        enum_value_to_bool_map
+    ):
+        self._underlying_attribute = underlying_attribute_meta_class(attribute_id)
+        self._attribute_id = attribute_id
```
marcoskirsch, line 115: seems you're storing things that aren't needed

src/nidcpower/metadata/attributes.py
```diff
@@ -2305,11 +2305,14 @@
     },
     1150302: {
         'access': 'read-write',
-        'attribute_class': 'AttributeIsolationState',
         'documentation': {
             'description': '\nDefines whether the channel is isolated.\n',
             'note': '\nThis attribute is not supported on all devices. For more information about supported devices, search ni.com for Supported Attributes by Device.\n'
         },
+        'enum_value_to_bool_map': {
+            1128: True,
```
marcoskirsch, line 2313: This is a not a good way to define things, in my opinion.
The enum is presumably already in metadata. I'd like for the metadata _of the enum_ to specify which value corresponds to True and which value corresponds to False.

Let's break this up in several steps / PRs:

1. Add the attribute as an enum.
2. Mark the enum as private. This may or may not be a build error (it should be). The enum should start with `_` and not be included in documentation because it's private.
3. Add a way to specify, in enum metadata, what values map to.
5. In nimi-python world, codegen "converter" functions for those enums. The name of the converter function should be in "expanded" metadata.
6. Add a new class "AttributeWithEnumConverter" or something like that.
7. Modify nimi-python so that the attribute, if enum, uses the converter.

marcoskirsch (changes requested):

build/templates/_attributes.py.mako
```diff
@@ -104,6 +104,26 @@ class AttributeEnum(object):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+class AttributeEnumWithConverter(AttributeEnum):
+    def __init__(
```
marcoskirsch, line 108: needs docstring

build/templates/_converters.py.mako
```diff
@@ -324,3 +326,50 @@ def convert_double_each_element(numbers):
 
 
 % endif
+% for enum_name in sorted(helper.filter_codegen_enums(enums)):
+    % if enums[enum_name].get('use_converter', False):
+<%
+    enum_name_in_snakecase = helper.camelcase_to_snakecase(enums[enum_name]['python_name'])
+%>\
+# Getter converter for ${enums[enum_name]['python_name']}
```
marcoskirsch, line 334: These generated converters could apply to methods, right?

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -275,3 +275,25 @@ def convert_chained_repeated_capability_to_parts(chained_repeated_capability):
     return [','.join(collections.OrderedDict.fromkeys(x)) for x in repeated_capability_lists]
 
 
+# Getter converter for _IsolationState
+___isolation_state_enum_value_to_converted_value_dict = {
+    1128: True,
```
marcoskirsch, line 280: you have an enum declared, use that not the hardcoded values

marcoskirsch (changes requested):

build/templates/_converters.py.mako
olsl21: Technically yes, but the codegen logic doesn't handle the usage of enums with converter for function parameters currently.

CHANGELOG.md
```diff
@@ -76,6 +77,7 @@ All notable changes to this project will be documented in this file.
                 * `LCROpenShortLoadCompensationDataSource`
                 * `LCRSourceDelayMode`
                 * `LCRStimulusFunction`
+                * `_IsolationState`
```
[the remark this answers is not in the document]
olsl21: I've removed the `_IsolationState` private enum from CHANGELOG.md in my latest commit

build/templates/_attributes.py.mako
olsl21: I have added docstrings for `AttributeEnumWithConverter` class and its constructor in the latest commit

build/templates/_converters.py.mako
```diff
@@ -324,3 +326,50 @@ def convert_double_each_element(numbers):
 
 
 % endif
+% for enum_name in sorted(helper.filter_codegen_enums(enums)):
+    % if enums[enum_name].get('use_converter', False):
+<%
+    enum_name_in_snakecase = helper.camelcase_to_snakecase(enums[enum_name]['python_name'])
```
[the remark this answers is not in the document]
olsl21: I have modified the internal metadata expansion logic to generate the `enum_to_converted_value_function_name` and `converted_value_to_enum_function_name` keys instead of `use_converter`. 

Nevertheless, I have set the expansion logic to **not** generate these keys if the enum doesn't use converter instead of expanding them to None by default (the codegen logic here would treat these keys as optional) in order to avoid breaking the existing metadata for all other drivers and requiring metadata updates for all of them.

I have added unit tests for the internal metadata expansion logic but not here since this code lives in the mako template and can only be tested after the codegen is performed.

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -275,3 +275,25 @@ def convert_chained_repeated_capability_to_parts(chained_repeated_capability):
     return [','.join(collections.OrderedDict.fromkeys(x)) for x in repeated_capability_lists]
 
 
+# Getter converter for _IsolationState
+___isolation_state_enum_value_to_converted_value_dict = {
```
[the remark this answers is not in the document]
olsl21: These additional underscores were inadvertently generated due to the fact that the enum `python_name` had already been prefixed with an underscore.

Anyway, I have updated the mako template to use the converter function names that are generated and stored in the metadata, which no longer have these additional underscores.

generated/nidcpower/nidcpower/_converters.py
olsl21: My initial rationale of using the enum values instead of the enums themselves for conversion was to avoid the dependency of `_converters.py` on `enums.py`, since some drivers may opt to exclude the codegen of some of these files.

Anyway, I have updated the converter codegen logic to work with enums instead of enum values, and added a logic to conditionally `import enums` if the driver uses enum(s) with converter.

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -275,3 +275,25 @@ def convert_chained_repeated_capability_to_parts(chained_repeated_capability):
...
+    1129: False,
+}
+
+
+def convert_from___isolation_state_enum_value(enum_value):
+    return ___isolation_state_enum_value_to_converted_value_dict[enum_value]
+
+
+# Setter converter for _IsolationState
+___isolation_state_converted_value_to_enum_value_dict = {
```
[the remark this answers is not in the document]
olsl21: As far as I'm aware, there are ways to reverse a dict in Python, e.g. https://stackoverflow.com/a/61102213, but these methods still require a separate variable to be declared. To avoid declaring an additional dict, we can either reverse the dict on-the-fly inside the converter function, or loop through the first dict to search for the corresponding key for a value (e.g. https://stackoverflow.com/a/2569076) inside the converter function. Both of these approaches seem to be more less efficient in terms of execution time compared to declaring an additional dict outside the converter function if this converter function need to be called repeatedly. 

@marcoskirsch Should I go with one of these approaches?

src/nidcpower/metadata/attributes.py
olsl21: This has been addressed through the latest changes.

build/templates/_attributes.py.mako
olsl21: This class has been updated in my latest commits.

build/templates/_attributes.py.mako
olsl21: This class has been removed in my latest commit

build/templates/_attributes.py.mako
olsl21: The missing newline has been added in my latest commit

src/nidcpower/metadata/attributes.py
```diff
@@ -2295,20 +2295,25 @@
...
     1150302: {
         'access': 'read-write',
+        'attribute_class': 'AttributeIsolationState',
         'documentation': {
             'description': '\nDefines whether the channel is isolated.\n',
             'note': '\nThis attribute is not supported on all devices. For more information about supported devices, search ni.com for Supported Attributes by Device.\n'
         },
-        'enum': 'IsolationState',
         'lv_property': 'Advanced:Isolation State',
         'name': 'ISOLATION_STATE',
```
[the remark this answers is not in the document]
olsl21: This attribute is now showing

build/templates/_attributes.py.mako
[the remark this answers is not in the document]
olsl21: I have created an Attribute class that takes in a pair of converter functions in my latest design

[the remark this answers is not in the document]
olsl21: This has been removed in my latest commit

generated/nidcpower/nidcpower/_converters.py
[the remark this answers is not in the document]
marcoskirsch: After thinking about it, I think the approach I favor the most is to codegen inside the converter function.

```
    return {
        1 : RED,
        2 : BLUE,
        3: GREEN,
    }[value]
```

I think this is cleanest, polutes namespaces least, and is more Pythonic.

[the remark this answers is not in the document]
olsl21: I have updated the converter codegen logic to codegen the dicts inside the converter functions directly in my latest commit.

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -275,3 +276,19 @@ def convert_chained_repeated_capability_to_parts(chained_repeated_capability):
     return [','.join(collections.OrderedDict.fromkeys(x)) for x in repeated_capability_lists]
 
 
+def convert_from_isolation_state_enum(enum):
+    if type(enum) is not enums._IsolationState:
```
marcoskirsch, line 280: * TypeError doesn't seem like the right thing to raise. Probably ValueError. This would only happen if the driver runtime returns something unexpected, right?
* Ask for forgiveness, not for permission, in other words, do a try/except. Don't check first.

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -275,3 +276,19 @@ def convert_chained_repeated_capability_to_parts(chained_repeated_capability):
...
+        raise TypeError('must be _IsolationState not {}'.format(type(enum).__name__))
+    return {
+        enums._IsolationState.ISOLATED: True,
+        enums._IsolationState.NON_ISOLATED: False,
+    }[enum]
+
+
+def convert_to_isolation_state_enum(value):
```
marcoskirsch, line 288: If the user passes something that doesn't convert cleanly, you will get a KeyError. Probably should be a ValueError.

generated/nidmm/nidmm/_attributes.py
```diff
@@ -100,6 +100,44 @@ def __set__(self, session, value):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+class AttributeEnumWithConverter(AttributeEnum):
+    '''Attribute metaclass for enums that use converters.'''
+
+    def __init__(
+        self,
+        underlying_attribute_meta_class,
```
marcoskirsch, line 108: What if instead of passing 

1. `underlying_attribute_meta_class`
2. `enum_meta_class`
3. `attribute_id`

_and_ deriving from `AttributeEnum`...

...you instead took in an instance of the underlying attribute?

I think that would make the code simpler to follow and simpler to unit test.

marcoskirsch (changes requested):

marcoskirsch:
You are adding new capabilities to the nimi-python code generator.
These should be covered via NI-FAKE. Please

* add an attribute with built-in conversion to NI-FAKE metadata
* add unit tests for the new attribute

build/helper/helper.py
```diff
@@ -133,6 +133,15 @@ def get_development_status(config):
     return dev_status
 
 
+def enum_uses_converter(enum):
+    '''Returns True if enum uses converter, False otherwise.
+
+    An enum uses converter if it has both 'enum_to_converted_value_function_name' and
+    'converted_value_to_enum_function_name' defined which are not None.
+    '''
+    return enum.get('enum_to_converted_value_function_name', None) is not None and enum.get('converted_value_to_enum_function_name', None) is not None
```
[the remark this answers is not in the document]
olsl21: No, it's not valid for the metadata to have one but not both keys defined. I have added an assertion in my latest commit.

build/helper/metadata_add_all.py
```diff
@@ -441,11 +442,17 @@ def add_all_attribute_metadata(attributes, config):
 def _add_enum_codegen_method(enums, config):
     '''Adds 'codegen_method' that will determine whether and how the enum is code genned. Default is public
 
-    Set all to 'no', then go through all functions and attributes and set to least restrictive use
+    Set all to 'no', then go through all functions and attributes and set to least restrictive use.
+    If the 'codegen_method' of an enum has been explicitly set to a value other than 'public' and it
+    is used by a 'public'/'python-only' function or 'public' attribute, a ValueError would be thrown
     '''
+    enum_to_has_explicit_codegen_method = {}
```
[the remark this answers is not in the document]
olsl21: My initial rationale of naming it as such was to indicate that it's a dict using a `<key>_to_<value>` naming convention, since it's a dict that maps an `enum` instance to `has_explicit_codegen_method` boolean.

Nevertheless, I agree that this name is awkward and I have renamed it to `enum_has_explicit_codegen_method` in my latest commit.

build/helper/metadata_add_all.py
```diff
@@ -458,8 +465,16 @@ def _add_enum_codegen_method(enums, config):
...
+                            if not enum_uses_converter(enums[e]):
+                                raise ValueError(
+                                    "Public / python-only function {} cannot use non-public enum {} (without converter)".format(
+                                        f, e
+                                    )
+                                )
+                        else:
+                            enums[e]['codegen_method'] = 'public'
 
     # Iterate through all codegen attributes and set any enum parameters to the same level
```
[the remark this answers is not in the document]
olsl21: I believe it refers to the same level of codegen method. I have updated it to `the same level of codegen_method` in my latest commit.

build/helper/metadata_add_all.py
```diff
@@ -471,8 +486,16 @@ def _add_enum_codegen_method(enums, config):
...
+                elif a_codegen_method == 'public' and enums[e]['codegen_method'] != 'public':
+                    if enum_to_has_explicit_codegen_method[e]:
+                        if not enum_uses_converter(enums[e]):
+                            raise ValueError(
+                                "Public attribute {} cannot use non-public enum {} (without converter)".format(
+                                    config['attributes'][a]['name'], e
+                                )
+                            )
+                    else:
+                        enums[e]['codegen_method'] = a_codegen_method
```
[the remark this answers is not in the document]
olsl21: I haven't really found a nice way to refactor it as some parts of the function are similar but not exactly the same, e.g. the wording of 'attribute' vs. 'function'. Nevertheless, I did discover that the `if f_codegen_method != 'no':` and `if a_codegen_method != 'no':` checks were redundant since `filter_codegen_functions()` and `filter_codegen_attributes()` had already performed these checks, so I removed them from the code.

build/helper/metadata_add_all.py
```diff
@@ -441,11 +442,17 @@ def add_all_attribute_metadata(attributes, config):
 def _add_enum_codegen_method(enums, config):
     '''Adds 'codegen_method' that will determine whether and how the enum is code genned. Default is public
```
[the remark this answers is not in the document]
olsl21: I have removed the 'Default is public' wording as I believe the default is actually 'no' if it is not used anywhere, although if it is used in any of the functions or attributes, its 'codegen_method' would be set to the least restrictive use.

I'm not really sure what do you mean by "explaining the rules" vs. "explaining / paraphrasing the algorithm", but I have updated the docstring a bit to make it closer to what the function is actually doing. Feel free to reopen this issue or open another issue if you think it's still unclear.

build/templates/_attributes.py.mako
```diff
@@ -104,6 +104,44 @@ class AttributeEnum(object):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+class AttributeEnumWithConverter(AttributeEnum):
+    '''Attribute metaclass for enums that use converters.'''
```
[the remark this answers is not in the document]
olsl21: I have updated the docstring in the latest commit.

build/templates/_attributes.py.mako
```diff
@@ -104,6 +104,44 @@ class AttributeEnum(object):
...
+
+    def __init__(
+        self,
+        underlying_attribute_meta_class,
+        enum_meta_class,
+        attribute_id,
+        getter_converter,
+        setter_converter
+    ):
+        '''Creates and returns an AttributeEnumWithConverter attribute meta class.
```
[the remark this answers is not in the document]
olsl21: I have updated the docstring in the latest commit.

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -275,3 +276,19 @@ def convert_chained_repeated_capability_to_parts(chained_repeated_capability):
     return [','.join(collections.OrderedDict.fromkeys(x)) for x in repeated_capability_lists]
 
 
+def convert_from_isolation_state_enum(enum):
```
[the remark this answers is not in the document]
olsl21: I have renamed the variable to `value` in the latest commit.

generated/nidcpower/nidcpower/_converters.py
olsl21: > This would only happen if the driver runtime returns something unexpected, right?

Yes, you're right.

This explicit `type()` check was actually based on the [existing implementation of the setter method of `AttributeEnum` class](https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/_attributes.py#L98). Nevertheless, I have changed it to use try/except in my latest commit.

generated/nidcpower/nidcpower/_converters.py
olsl21: With the try/except block in my latest commit, the converter function will now through a ValueError instead of KeyError if the user input value cannot be converted.

generated/nidmm/nidmm/_attributes.py
olsl21: Thanks for the suggestion, I have updated `AttributeEnumWithConverter` to use this design in the latest commit.

src/nidcpower/metadata/attributes.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-DCPower API metadata version 22.0.0d216
+# This file is generated from NI-DCPower API metadata version 22.5.0d9999
```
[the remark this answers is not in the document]
olsl21: I have updated the metadata version to 22.5.0d71 in my latest commit.

generated/nifake/nifake/_converters.py
```diff
@@ -280,3 +281,27 @@ def convert_double_each_element(numbers):
...
 
+def convert_from_enum_with_converter_enum(value):
+    try:
+        return {
+            enums.EnumWithConverter.RED: True,
+            enums.EnumWithConverter.BLUE: False,
+            enums.EnumWithConverter.YELLOW: 'yellow',
+            enums.EnumWithConverter.BLACK: 42,
+        }[value]
+    except Exception:
```
marcoskirsch, line 292: Use KeyError.

generated/nifake/nifake/unit_tests/test_session.py
```diff
@@ -1042,6 +1042,25 @@ def test_set_attribute_enum_real64(self):
...
+
+    def test_set_attribute_enum_with_converter(self):
+        self.patched_library.niFake_SetAttributeViInt32.side_effect = self.side_effects_helper.niFake_SetAttributeViInt32
+        enum_value = nifake.EnumWithConverter.RED
+        converted_value = True
+        with nifake.Session('dev1') as session:
+            session.read_write_enum_with_converter = converted_value
+            attribute_id = 1000011
+            self.patched_library.niFake_SetAttributeViInt32.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), _matchers.ViStringMatcher(''), _matchers.ViAttrMatcher(attribute_id), _matchers.ViInt32Matcher(enum_value.value))
+
```
marcoskirsch, line 1063: add tests for the wrong value

generated/nifake/nifake/_converters.py
```diff
@@ -280,3 +281,27 @@ def convert_double_each_element(numbers):
...
+
+def convert_to_enum_with_converter_enum(value):
+    try:
+        return {
+            True: enums.EnumWithConverter.RED,
+            False: enums.EnumWithConverter.BLUE,
+            'yellow': enums.EnumWithConverter.YELLOW,
+            42: enums.EnumWithConverter.BLACK,
+        }[value]
+    except Exception:
```
marcoskirsch, line 304: shouldn't this be key error?

generated/nifake/nifake/_converters.py
```diff
@@ -280,3 +281,27 @@ def convert_double_each_element(numbers):
...
+        raise ValueError(f"'{value}' cannot be converted to an EnumWithConverter enum value")
```
marcoskirsch, line 305: Under what circumstance would this occur?
I think that'd be when the driver returns an unexpected enum value.

When would that happen?
Maybe if the driver runtime is too new for the Python API. In which case `f"'{value}' cannot be converted to an EnumWithConverter enum value")` is useless to the customer. I think for this case we need a `DriverTooNewError` with a reasonable helpful message like:

> The driver runtime returned an unexpected value. This can occur if the `nidcpower` Python module is too old. Upgrade `nidcpower`.

marcoskirsch: But perhaps this is not the place for that as this function doesn't have the context. This might be somethign to be changed in the Attribute class.

generated/nifake/nifake/_converters.py
```diff
@@ -280,3 +281,27 @@ def convert_double_each_element(numbers):
...
+def convert_from_enum_with_converter_enum(value):
+    try:
+        return {
+            enums.EnumWithConverter.RED: True,
+            enums.EnumWithConverter.BLUE: False,
+            enums.EnumWithConverter.YELLOW: 'yellow',
+            enums.EnumWithConverter.BLACK: 42,
+        }[value]
+    except Exception:
+        raise ValueError(f"EnumWithConverter enum value '{value}' cannot be converted")
```
marcoskirsch, line 293: What is a customer supposed to do with this error message?

But perhaps changing this to a better error belongs in the attribute class, where there is more context.

marcoskirsch (changes requested):

generated/nifake/nifake/_converters.py
olsl21: I have updated this to use KeyError in my latest commit

generated/nifake/nifake/_converters.py
olsl21: I have updated this to KeyError in my latest commit

generated/nifake/nifake/_converters.py
olsl21: I believe this comment applies to `convert_from_enum_with_converter_enum()` (the getter converter) instead of `convert_to_enum_with_converter_enum()` (the setter converter), since the value returned from the driver will only be passed to the getter converter.

I have created #1761 to add the new `DriverTooNewError`, and updated this PR to throw it from the `AttributeEnumWithConverter` attribute class (I have to throw it from there instead of the getter converter since the invalid return value from the driver will actually be caught by the [enum conversion there ](https://github.com/ni/nimi-python/pull/1738/files#diff-0ba5ea21d30086cc356b2093b1d46113f7bcc0c51c014c6b83115bde4f66b53fR127) before it gets passed into the getter converter)
```
            return self._getter_converter(
                self._underlying_enum_attribute.__get__(session, session_type)
            )
```

src/nifake/unit_tests/test_session.py
```diff
@@ -1042,6 +1042,25 @@ def test_set_attribute_enum_real64(self):
             attribute_id = 1000005
             self.patched_library.niFake_SetAttributeViReal64.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), _matchers.ViStringMatcher(''), _matchers.ViAttrMatcher(attribute_id), _matchers.ViReal64Matcher(enum_value.value))
 
+    def test_get_attribute_enum_with_converter(self):
```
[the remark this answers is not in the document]
olsl21: I have added `test_get_attribute_enum_with_converter_invalid_value_from_driver()` and `test_set_attribute_enum_with_converter_invalid_input()` to test for the wrong values and their corresponding error messages in my latest commit.

generated/nifake/nifake/_converters.py
olsl21: This is related to another comment on the error message. 

For `convert_from_enum_with_converter_enum()` (the getter converter), the error message here would not be displayed even when there's driver runtime mismatch as it will be caught by the enum conversion in the attribute class, which I have added a try-catch clause there to handle it.

For convert_to_enum_with_converter_enum() (the setter converter), I have improved the error message to include the valid values that can be passed in by the user.

build/helper/codegen_helper.py
```diff
@@ -485,6 +485,11 @@ def get_dictionary_snippet(d, indent=4):
     return ('\n' + (' ' * indent)).join(d_lines)
 
 
+def get_enum_value_snippet(value):
+    '''Returns value formatted into string, surrounding it with single quotes if it is of str type'''
+    return ("'{}'" if type(value) is str else "{}").format(value)
```
marcoskirsch, line 490: let's hope we never have a string value of `"True"` 🤣

build/helper/metadata_add_all.py
```diff
@@ -439,39 +440,56 @@ def add_all_attribute_metadata(attributes, config):
 
 
 def _add_enum_codegen_method(enums, config):
-    '''Adds 'codegen_method' that will determine whether and how the enum is code genned. Default is public
+    '''Adds 'codegen_method' that will determine whether and how the enum is code-generated.
```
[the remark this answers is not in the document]
marcoskirsch: I think the rule is that enums default to not getting generated unless they are used by a function or attribute, in which case 'codegen_method' is set to match to that of whatever uses it.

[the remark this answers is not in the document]
marcoskirsch: If what I said above is correct, then you can hopefully this unreadable function...

you can do it in term of functions like:

`def get_functions_that_use_enum(enum_name)`
`def get_attributes_that_use_enum(attribute_name)`
`def get_least_restrictive_codegen_method(codegen_method`

You can add a unit test

build/templates/_attributes.py.mako
```diff
@@ -104,6 +105,38 @@ class AttributeEnum(object):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+class AttributeEnumWithConverter(AttributeEnum):
+    '''Class for attributes that use enums internally but are exposed in nimi-python API as something else, thus need conversion.'''
+
+    def __init__(self, underlying_enum_attribute, getter_converter, setter_converter):
```
marcoskirsch, line 111: since the name of the class you pass in as `underlying_enum_attribute` is `AttributeEnum`, consider changing variable name to `underlying_attribute_enum`

build/templates/_converters.py.mako
```diff
@@ -324,3 +329,29 @@ def convert_double_each_element(numbers):
...
+% for enum_name in sorted(helper.filter_codegen_enums(enums)):
+    % if helper.enum_uses_converter(enums[enum_name]):
+def ${enums[enum_name]['enum_to_converted_value_function_name']}(value):
+    try:
+        return {
+        % for enum_value in enums[enum_name]['values']:
+            enums.${enums[enum_name]['python_name']}.${enum_value['python_name']}: ${helper.get_enum_value_snippet(enum_value['converts_to_value'])},
+        % endfor
+        }[value]
+    except KeyError:
```
marcoskirsch, line 341: after thinking about it a bit much, I think it's ok for the converter functions to return KeyError (simpler).

What to do about it is really the repsonsibility of the attribute class.

build/templates/_attributes.py.mako
```diff
@@ -104,6 +105,38 @@ class AttributeEnum(object):
...
+        self._underlying_enum_attribute = underlying_enum_attribute
+        self._getter_converter = getter_converter
+        self._setter_converter = setter_converter
+
+    def __get__(self, session, session_type):
+        try:
+            return self._getter_converter(
+                self._underlying_enum_attribute.__get__(session, session_type)
+            )
+        except ValueError:
```
marcoskirsch, line 133: change to KeyError after simplifying converter functions

build/templates/_attributes.py.mako
```diff
@@ -104,6 +105,38 @@ class AttributeEnum(object):
...
+            raise errors.DriverTooNewError('The driver runtime returned an unexpected value. ')
+
+    def __set__(self, session, value):
+        return self._underlying_enum_attribute.__set__(session, self._setter_converter(value))
```
marcoskirsch, line 137: I think you need a try/except in here.

@sbethur this is a place where being restrictive can come back and bite us. For example if an external or internal client is using the Python API with unreleased driver or using a private value... we are going to prevent it.

Are we better off in those cases just passing the unknown values up and down to the driver runtime and letting it error? If that's the case then the new DriverTooNewError is not needed after all. Let's discuss.

generated/nidcpower/nidcpower/_attributes.py
```diff
@@ -100,6 +101,38 @@ def __set__(self, session, value):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+class AttributeEnumWithConverter(AttributeEnum):
+    '''Class for attributes that use enums internally but are exposed in nimi-python API as something else, thus need conversion.'''
```
marcoskirsch, line 105: consider changing nimi-python for the module name.

generated/nidigital/nidigital/_attributes.py
```diff
@@ -100,6 +101,38 @@ def __set__(self, session, value):
         return self._underlying_attribute.__set__(session, value.value)
 
 
+class AttributeEnumWithConverter(AttributeEnum):
```
marcoskirsch, line 104: wondering if we should make the generator smarter so that only the Attribute classes that are used get generated.... it would of course be a separate PR

marcoskirsch (changes requested):

marcoskirsch:
FYI merge conflict

build/helper/metadata_add_all.py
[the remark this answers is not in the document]
olsl21: > It adds codegen method only if not explictly specified in metadata.

Not really, it might change an explicitly specified `codegen_method` of  'no' to 'private' if there is a private method / attribute using it.

> I think the rule is that enums default to not getting generated unless they are used by a function or attribute, in which case 'codegen_method' is set to match to that of whatever uses it.

Yes, this is right. I believe this is what my docstring is trying to explain: 
```
    Set the 'codegen_method' of all enums that do not explicitly specify it in metadata to 'no',
    then go through all functions and attributes and set to the least restrictive use.
```

> you can do it in term of functions like:
>
> def get_functions_that_use_enum(enum_name)
> def get_attributes_that_use_enum(attribute_name)
> def get_least_restrictive_codegen_method(codegen_method

I could not really find a way to break this function further down to your suggested functions as they are really intertwined, e.g. `get_least_restrictive_codegen_method()` would have to iterate through all the functions and attributes before arriving at the final results, which is essentially what this `_add_enum_codegen_method()` function is doing.

build/helper/metadata_add_all.py
```diff
@@ -439,39 +440,56 @@ def add_all_attribute_metadata(attributes, config):
...
 
-    Set all to 'no', then go through all functions and attributes and set to least restrictive use
+    Set the 'codegen_method' of all enums without any explicit 'codegen_method' to 'no', then go
```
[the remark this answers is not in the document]
olsl21: I have updated this in my latest commit.

build/templates/_attributes.py.mako
olsl21: I have updated this in my latest commit.

build/templates/_converters.py.mako
olsl21: I have removed the KeyError try-catch for the getter converter (enum_to_converted_value_function), but not the setter converter (converted_value_to_enum_function), since the Attribute class has no access to the enum conversion dictionary and would not be able to provide any helpful error message if it encounters a KeyError.

generated/nidcpower/nidcpower/_attributes.py
olsl21: I have updated this in my latest commit.

generated/nidigital/nidigital/_attributes.py
olsl21: I have filed an issue for this #1762.

marcoskirsch (approved):

build/helper/metadata_add_all.py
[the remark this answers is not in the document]
marcoskirsch: > Not really, it might change an explicitly specified codegen_method of 'no' to 'private' if there is a private method / attribute using it.

So the codegenerator will override the explicit instructions from metadata? That's not a good idea. It should instead error out with a good message, something like:

> Enum `Colors` specifies `'codegen_method': 'no'`, but attribute `color` uses it and has `'codegen_method': 'private'` which is incompatible. Please fix your metadata.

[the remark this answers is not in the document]
marcoskirsch: > Yes, this is right. I believe this is what my docstring is trying to explain:

Well, the comment describes the algorithm, not what it does at a high level which is more helpful.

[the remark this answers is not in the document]
marcoskirsch: > I could not really find a way to break this function further down to your suggested functions as they are really intertwined,

Maybe once we stop overriding explicit metadata this is easier to do?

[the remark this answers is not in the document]
olsl21: This is being resolved by #1767.

PR #1750, opened 2022-06-09
NI-FAKE: Update metadata to 22.5.0d9, fix _mock_helper

olsl21:

- Update NI-FAKE metadata to 22.5.0d9
  - New additions:
    - `custom_struct_typedef` and `custom_struct_nested_typedef` custom types
    - `AltColor` enum
    - `FunctionWithRepeatedCapabilityType` and `GetCustomTypeTypedef`
  - Create custom type implementation files for the new `custom_struct_typedef` and `custom_struct_nested_typedef` custom types
  - Update nifake.mak to copy the custom type implementation files so that the build can succeed
  - Create unit test for the new `get_custom_type_typedef()` function that uses the new custom types
- Fix `_mock_helper.py.mako` codegen logic for functions with custom type parameter to use the actual parameter name instead of hardcoded "cs"

Local build succeeded.

### DIFF
--- a/src/nifake/custom_types/custom_struct_nested_typedef.py
+++ b/src/nifake/custom_types/custom_struct_nested_typedef.py
@@ -1,11 +1,50 @@
 import ctypes
 
+import nifake.custom_struct as custom_struct
+import nifake.custom_struct_typedef as custom_struct_typedef
+
 
 # These custom type classes are created as placeholders
 class struct_CustomStructNestedTypedef(ctypes.Structure):  # noqa N801
-    _fields_ = []
+    _fields_ = [
+        ('struct_custom_struct', custom_struct.struct_CustomStruct),
+        ('struct_custom_struct_typedef', custom_struct_typedef.struct_CustomStructTypedef),
+    ]
+
+    def __init__(self, data=None):
+        super(ctypes.Structure, self).__init__()
+        if data is not None:
+            self.struct_custom_struct = custom_struct.struct_CustomStruct(data.struct_custom_struct)
+            self.struct_custom_struct_typedef = custom_struct_typedef.struct_CustomStructTypedef(
+                data.struct_custom_struct_typedef
+            )
+        else:
+            self.struct_custom_struct = custom_struct.struct_CustomStruct()
+            self.struct_custom_struct_typedef = custom_struct_typedef.struct_CustomStructTypedef()
 
 
 class CustomStructNestedTypedef(object):
-    pass
+    def __init__(
+        self,
+        data=None,
+        struct_custom_struct=custom_struct.CustomStruct(),
+        struct_custom_struct_typedef=custom_struct_typedef.CustomStructTypedef()
+    ):
+        if data is not None:
+            self.struct_custom_struct = custom_struct.CustomStruct(data.struct_custom_struct)
+            self.struct_custom_struct_typedef = custom_struct_typedef.CustomStructTypedef(
+                data.struct_custom_struct_typedef
+            )
+        else:
+            self.struct_custom_struct = struct_custom_struct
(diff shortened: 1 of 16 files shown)


src/nifake/metadata/attributes.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-FAKE API metadata version 21.0.0d11
+# This file is generated from NI-FAKE API metadata version 22.5.0d9999
```
olsl21, line 2: I'll update this to an official export once the internal PR gets approved and merged.

marcoskirsch:
What is this PR really doing?
It's not just "updating metadata", it's adding new things to it. This is the "what" and "updating metadata" is the how.
Please improve the description and call the fake driver "NI-FAKE" (similar to NI-DMM or NI-FGEN).

marcoskirsch:
Is it possible to do this in smaller, distinct, single purpose PRs?

First one would be, it seems, to update NI-FAKE metadata to something newer but that doesn't attempt to exercise the new capabilities you are adding to nimi-python (namely the ability for enums to specify what values they convert to when exposed in nimi-python APIs).

src/nifake/metadata/attributes.py
olsl21: I will split this PR into smaller PRs and this PR will now updata the NI-FAKE metadata to 22.5.0d9

olsl21:
> Is it possible to do this in smaller, distinct, single purpose PRs?
> 
> First one would be, it seems, to update NI-FAKE metadata to something newer but that doesn't attempt to exercise the new capabilities you are adding to nimi-python (namely the ability for enums to specify what values they convert to when exposed in nimi-python APIs).

I've split it into two PRs (i.e. this PR which updates the metadata to 22.5.0d9 (without the new enum with converter and its related attribute) and #1751 which adds the new enum with converter and its related attribute).

> What is this PR really doing? It's not just "updating metadata", it's adding new things to it. This is the "what" and "updating metadata" is the how. Please improve the description and call the fake driver "NI-FAKE" (similar to NI-DMM or NI-FGEN).

This PR is now updating the metadata to 22.5.0d9 with new custom types, enum and functions, as well as some fixes. I have updated the PR description to list out the changes in the metadata more explicitly.

marcoskirsch (changes requested):
There are two methods being added to `nifake` in this PR but they aren't unit tested.
I think we need a more realistic struct so that it can be properly exercised.

olsl21:
> There are two methods being added to `nifake` in this PR but they aren't unit tested. I think we need a more realistic struct so that it can be properly exercised.

I have completed the implementations for the new custom types and added unit tests for the two new methods in my latest commit.

marcoskirsch:
[CI Error](https://github.com/ni/nimi-python/pull/1750/checks?check_run_id=6917155578)
```
The following code generated files do not match what is commited to Git. Run codegen and include the generated files in the PR.
 M generated/nifake/nifake/unit_tests/test_session.py
```

olsl21:
Sorry for the CI error, I have added in the missing code generated file.

sbethur (approved):

marcoskirsch (approved):

src/nifake/custom_types/custom_struct_nested_typedef.py
```diff
@@ -0,0 +1,50 @@
+import ctypes
+
+import nifake.custom_struct as custom_struct
+import nifake.custom_struct_typedef as custom_struct_typedef
+
+
+# These custom type classes are created as placeholders
```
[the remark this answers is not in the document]
olsl21: Oh, this comment was actually from before I added the full implementation for these new custom types. I've removed it in my latest commit.

PR #1754, opened 2022-06-15
NI-DCPower: Add channel to Measurement namedtuple

### LINKED ISSUE #1752: `Measurement` named tuple should include a `channel`
### Description of issue

`nidcpower.Session.fetch_multiple()` and `nidcpower.Session.fetch_multiple()` returned a named tuple `Measurement` that contains:

```
        - **voltage** (float)
        - **current** (float)
        - **in_compliance** (bool)
```

For usability purposes, this should be like `niscope`'s `WaveformInfo` and include the `channel`.

Future API (like the LCR functionality in the PXIe-4190) should also include the `channel` in the information it returns.

olsl21:

- Create `expand_channel_string()` converter function to convert channel string (from repeated capabilities) to a list of channel names in their canonical form based on what `get_channel_names()` returns (to be returned to the user when `fetch_multiple()` and `measure_multiple()` are called)
  - Add unit tests for `expand_channel_string()`
- Update `fancy_fetch` method template to generate the relevant channel names for the Measurement namedtuple returned by `fetch_multiple()` and `measure_multiple()`
  - Add system tests for the new implementations of `fetch_multiple()` and `measure_multiple()`
- Update CHANGELOG

* Fix #1752

Ran the newly added system tests and unit tests, all of them passed.


marcoskirsch:
@olsl21 please resolve merge conflicts

generated/nidcpower/nidcpower/session.py
```diff
@@ -4801,11 +4801,39 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
 
         voltage_measurements, current_measurements, in_compliance = self._fetch_multiple(timeout, count)
 
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=in_compliance[i]) for i in range(count)]
+        if self._repeated_capability == '':
+            channel_names = self._get_channel_names(range(self.channel_count))
+        else:
+            # Check if the session was opened with independent_channels set to True by checking if
+            #  _get_channel_name() returns channel names with prefix
+            first_channel_name = self._get_channel_name(1)
```
sbethur, line 4813: This seems brittle. You can get the value of `independent_channels` from `self._param_list`. Instead of parsing `self._param_list`, modify the constructor to store the session params as a dict and convert it to the list when needed (inside `__repr__`).

This should be its own PR.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4801,11 +4801,39 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
+                default_prefix = first_channel_name[:first_channel_name.find('/') + 1]
+                channel_names = _converters.convert_independent_channels_repeated_capabilities(
+                    self._repeated_capability,
+                    default_prefix
+                )
+            else:
+                channel_names = _converters.convert_repeated_capabilities(
+                    self._repeated_capability,
+                    ''
+                )
```
sbethur, line 4827: Move all of this logic into `_converters`. Better than having some channel parsing logic there and some here.

sbethur: This will also avoid code duplication between this method and `measure_multiple`.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4801,11 +4801,39 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
+        if self._repeated_capability == '':
+            channel_names = self._get_channel_names(range(self.channel_count))
+        else:
+            # Check if the session was opened with independent_channels set to True by checking if
+            #  _get_channel_name() returns channel names with prefix
+            first_channel_name = self._get_channel_name(1)
+            if '/' in first_channel_name:
+                # If there is any repeated capabilities without prefix, the session must have only
+                # one instrument, so just get the prefix from the first channel and add to all of
+                # the repeated capabilities that are without prefix
```
sbethur, line 4817: > If there is any repeated capabilities without prefix, the session must have only one instrument

This isn't always true. A user can open a session to `dev1, dev2/1-3`.

sbethur (changes requested):

generated/nidcpower/nidcpower/session.py
olsl21: This logic is parsing `self._repeated_capability`, which is only populated through `session.channels`, instead of `resource_name`. It should contain only channel names or ranges (either with or without prefix. Calling `session.channels['dev1, dev2/1-3']` would result in "Error -1074135008: IVI: Unknown channel or repeated capability name".

generated/nidcpower/nidcpower/session.py
olsl21: I've created #1755 to address this.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4801,11 +4801,39 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
+                    self._repeated_capability,
+                    ''
+                )
+
+        return [
+            Measurement(
+                voltage=voltage_measurements[i],
+                current=current_measurements[i],
+                in_compliance=in_compliance[i],
+                channel=channel_names[0],
```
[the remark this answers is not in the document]
olsl21: > fetch_multiple must be called on only a single channel. Is this correct?

Yes, this is correct.

> Can't we just get the channel name from self._repeated_capability when it isn't an empty string?

The user may still pass in the channels (which eventually get stored as `self._repeated_capability`) in many forms, e.g. "0", "dev1/0", or even "" if the session has only one channel, so we still need the converter logic here.

generated/nidcpower/nidcpower/session.py
olsl21: Thanks for the suggestion. I have moved the logic to `convert_channels_repeated_capabilities()` inside the converters.py in my latest commit.

generated/nidcpower/nidcpower/session.py
olsl21: @sbethur It seems that `independent_channels` is only available in `Session` and not `_SessionBase`, hence this approach of getting `independent_channels` from `self._param_list` would not work.

Also, even though the assumption that session opened with `independent_channels=True` would return channel name with prefix seems brittle, the actual conversion logic wouldn't work if this assumption is not valid, since it need to extract the prefix from the channel name and add it to channels repeated capability groups that are without prefix (again, the check here relies on the assumption that the prefix ends with a '/'). Hence, I propose keeping the current logic (it now lives in `convert_channels_repeated_capabilities()` following my latest commit).

marcoskirsch:
> The following code generated files do not match what is commited to Git. Run codegen and include the generated files in the PR.
 M generated/nidcpower/nidcpower/session.py

Are you running the tests locally?

build/templates/_converters.py.mako
```diff
@@ -153,6 +153,59 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        prefix = repeated_capability[:split_index]
+        return convert_repeated_capabilities(repeated_capability[split_index:], prefix)
+    return convert_repeated_capabilities(repeated_capability)
+
+
+def convert_channels_repeated_capabilities(
+    channels_repeated_capability,
+    first_channel_name
+):
+    '''Convert a channels repeated capabilities string, possibly with no or multiple prefixes (each ends with '/'), to a list
```
marcoskirsch, line 178: maybe add example to docstring

generated/nidcpower/nidcpower/session.py
```diff
@@ -4846,11 +4862,27 @@ def measure_multiple(self):
...
+                first_channel_name
+            )
+
+        return [
+            Measurement(
+                voltage=voltage_measurements[i],
+                current=current_measurements[i],
+                in_compliance=None,
+                channel=channel_names[i],
+            ) for i in range(self._parse_channel_count())
```
marcoskirsch, line 4884: Is it true that the result of _parse_channel_count must be the same as the length of `channel_names`?
I wonder if there are  more pythonic ways of doing this (zip?) that don't require calling into the driver runtime again.

generated/nidcpower/nidcpower/session.py
```diff
@@ -5233,6 +5265,67 @@ def _get_attribute_vi_string(self, attribute_id):
         errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
         return attribute_value_ctype.value.decode(self._encoding)
 
+    @ivi_synchronized
+    def _get_channel_name(self, index):
```
marcoskirsch, line 5269: don't put this in SessionBase

marcoskirsch (changes requested):
Please consider doing a self-contained PR for making _get_channel_names accessible in SessionBase

build/templates/_converters.py.mako
```diff
@@ -153,6 +153,59 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
     return ','.join(convert_repeated_capabilities(repeated_capability, ''))
 
 
+def convert_single_group_repeated_capabilities(repeated_capability):
+    '''Convert a repeated capabilities string, with at most one prefix (that ends with '/') and without any comma, to a list
```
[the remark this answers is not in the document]
olsl21: Sure, I have added examples for `convert_single_group_repeated_capabilities()` in my latest commit.

build/templates/_converters.py.mako
```diff
@@ -153,6 +153,59 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        repeated_capability (str) - Supported types (with an optional prefix):
+            - range (using '-' or ':')
+            - single item
+
+    Returns:
+        rep_cap (str) - comma delimited list of strings of the expanded repeated capability items
+    '''
+    if '/' in repeated_capability:
+        split_index = repeated_capability.find('/') + 1
+        prefix = repeated_capability[:split_index]
```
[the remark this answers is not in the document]
olsl21: I'm not sure about the other existing converters that deal with repeated capabilities, but my newly added converter functions assume that the prefix is an instrument prefix (that ends with '/'), hence I agree that a more specific term should be used. I have renamed `prefix` to `instrument_prefix` in both the documentation and the code in my latest commit.

build/templates/_converters.py.mako
olsl21: I have added examples to the docstring of `convert_channels_repeated_capabilities()` in my latest commit.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4801,11 +4801,27 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
 
         '''
         import collections
-        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance'])
+        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance', 'channel'])
```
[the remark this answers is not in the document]
olsl21: I have moved 'channel' to be the first member of `Measurement` namedtuple in my latest commit.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4801,11 +4801,27 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
 
         voltage_measurements, current_measurements, in_compliance = self._fetch_multiple(timeout, count)
 
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=in_compliance[i]) for i in range(count)]
+        if self._repeated_capability == '':
+            channel_names = self._get_channel_names(range(self.channel_count))
+        else:
+            first_channel_name = self._get_channel_name(1)
```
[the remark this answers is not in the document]
olsl21: I have updated the code to use `first_channel_name = self._get_channel_names([0])[0]` instead in my latest commit. I have also removed `_get_channel_name` from `_SessionBase`.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4846,11 +4862,27 @@ def measure_multiple(self):
 
         '''
         import collections
-        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance'])
+        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance', 'channel'])
```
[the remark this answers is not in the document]
olsl21: I have updated the docstring for these functions to mention the new member `channel` in my latest commit.

generated/nidcpower/nidcpower/session.py
olsl21: I have removed it in my latest commit.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4846,11 +4862,27 @@ def measure_multiple(self):
...
 
         voltage_measurements, current_measurements = self._measure_multiple()
 
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=None) for i in range(self._parse_channel_count())]
+        if self._repeated_capability == '':
```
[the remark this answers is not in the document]
olsl21: > This if/else statement should probably be factored out into own function.

@marcoskirsch Where should this "if/else" statement live? In `_SessionBase`? In the `_converters`?

If it is to be added into `_SessionBase`, how should I make sure that it is generated for nidcpower only? Should I add an if/else to the `session.py.mako` template? Or add it as a function in the metadata and use another template (e.g. `fancy_fetch.py.mako`) to store its implementation?

>  What is it doing exactly? Setting a list to channel_names?

Yes, it is creating a list of channel names based on the repeated capabilities so that the list can be used to populate the `channel` member of the `Measurement` namedtuple.

olsl21:
> > The following code generated files do not match what is commited to Git. Run codegen and include the generated files in the PR.
> > M generated/nidcpower/nidcpower/session.py
> 
> Are you running the tests locally?

Sorry for the error message. It was actually caused by me removing the deprecated `get_channel_name()` function in `src` without regenerating the generated files. 

I have rerun the tests locally after my latest changes to make sure that they all pass.

olsl21:
> Please consider doing a self-contained PR for making _get_channel_names accessible in SessionBase

It has already been created at #1757.

generated/nidcpower/nidcpower/session.py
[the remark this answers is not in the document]
sbethur: `namedtuple` allows indexing to access members. Putting `channels` first could break existing clients.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4795,17 +4796,36 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
 
-        voltage_measurements, current_measurements, in_compliance = self._fetch_multiple(timeout, count)
+        voltage_measurements, current_measurements, in_compliances = self._fetch_multiple(timeout, count)
 
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=in_compliance[i]) for i in range(count)]
+        if self._repeated_capability == '':
+            channel_names = self._get_channel_names(range(self.channel_count))
+        else:
+            first_channel_name = self._get_channel_names([0])[0]
+            channel_names = _converters.convert_channels_repeated_capabilities(
```
sbethur, line 4814: I _think_ `_get_channel_names` returns channel numbers when "independent channels" is False and FQCNs when it's True.

If the above is true, we need to make sure the format we return is the same for both `if self._repeated_capability == ''` and for the `else` case.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4795,17 +4796,36 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
+        Measurement = collections.namedtuple('Measurement', ['channel', 'voltage', 'current', 'in_compliance'])
 
-        voltage_measurements, current_measurements, in_compliance = self._fetch_multiple(timeout, count)
+        voltage_measurements, current_measurements, in_compliances = self._fetch_multiple(timeout, count)
 
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=in_compliance[i]) for i in range(count)]
+        if self._repeated_capability == '':
+            channel_names = self._get_channel_names(range(self.channel_count))
+        else:
+            first_channel_name = self._get_channel_names([0])[0]
```
sbethur, line 4813: Need a comment on why `first_channel_name` is needed; which as I understand is to handle the case where `channels` is something like `Dev1/1:2,0`.

(Commentary: I think we made a mistake by allowing users to pass channels with mixed formatting (channel indices and FQCNs).

sbethur: I guess there is also the case where the resource_name is `dev1/0:3` and repeated capability is `0`.

sbethur (changes requested):

generated/nidcpower/nidcpower/session.py
olsl21: I believe these have been covered by my newly added [system tests](https://github.com/ni/nimi-python/pull/1754/files/9d606e85894e6c80b5ed7e87431a2feb2d1d21a0#diff-588aa0a2d97cf60012b577567bec4ae6bad13d2d7201507a924b8aa150bbe328):

```
@pytest.mark.parametrize(
    'resource_name,channels,independent_channels,measurement_channels,expected_measured_channels',
    [
        ('Dev1', None, False, None, [str(x) for x in range(12)]), # "independent channels" is False, self._repeated_capability == '', expected non-FQCNs
        ('Dev1', '0,2-3,6', False, '2-3,0', ['2', '3', '0']), # "independent channels" is False, self._repeated_capability != '', expected non-FQCNs
        ('Dev1', None, True, None, [f'Dev1/{x}' for x in range(12)]), # "independent channels" is True, self._repeated_capability == '', expected FQCNs
        ('Dev1/1:4', ' ', True, '2:3', ['Dev1/2', 'Dev1/3']), # "independent channels" is True, self._repeated_capability != '', expected FQCNs
        ...
    ]
)
```

build/templates/_converters.py.mako
```diff
@@ -153,6 +153,59 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        return convert_repeated_capabilities(repeated_capability[split_index:], prefix)
+    return convert_repeated_capabilities(repeated_capability)
+
+
+def convert_channels_repeated_capabilities(
```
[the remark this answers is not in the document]
olsl21: > I think we have bigger problems than the function signature being weird. The below code snippet should tell you the issue:

I believe this is not an issue as the converter function works on "channels repeated capabilities" (as suggested by the function name) instead of "repeated capabilities" in general. That means, if there are multiple instruments in the session, the "channels repeated capabilities" must be fully-qualified, otherwise the driver runtime would throw an error:
```
>>> with nidcpower.Session(resource_name="dev1,dev2", channels="", options="Simulate=1, DriverSetup=Model:4147; BoardType:PXIe", independent_channels=True) as smu:
...     with smu.initiate():
...             smu.channels[6].measure_multiple()
...
nidcpower.errors.DriverError: -1074135008: IVI: Unknown channel or repeated capability name.

Channel Name: 6
```

That also implies that if the "channels repeated capabilities" has no instrument prefix and it still works, the session has only one instrument in it, and `get_channel_names()` would return the same instrument prefix (if it does return an instrument prefix) regardless of which channel it is called with.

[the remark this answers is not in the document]
olsl21: Thanks for pointing it out. I believe that bug doesn't affect my implementation as it involves `independent_channels = False`, which means `first_channel_name` will not include an instrument prefix and thus there will not be a need to add any instrument info to the final list of channel names.

generated/nidcpower/nidcpower/session.py
olsl21: I have added a comment to explain why `first_channel_name` is needed in my latest commit.

```
            # first_channel_name is used to deduce if the session was opened with independent_channels
            #  set to True (in that case, the channel name would be fully-qualified) or False (in
            #  that case, the channel name would not have any instrument prefix). It is also used
            #  to extract the instrument prefix and populate the expanded channels repeated
            #  capabilities with it if needed
```

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,81 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+):
+    '''Convert a channels repeated capabilities string, possibly with no or multiple prefixes (each ends with '/'), to a list
+
+    Examples:
+        - convert_channels_repeated_capabilities('1', '0') --> ['1']
+        - convert_channels_repeated_capabilities('0-2', '0') --> ['0', '1', '2']
+        - convert_channels_repeated_capabilities('0:2', '0') --> ['0', '1', '2']
+        - convert_channels_repeated_capabilities('0:2,4', '0') --> ['0', '1', '2', '4']
+        - convert_channels_repeated_capabilities('4,1:2', '1') --> ['4', '1', '2']
+        - convert_channels_repeated_capabilities('2:3,0', 'Dev1/0') --> ['Dev1/2', 'Dev1/3', 'Dev1/0']
```
marcoskirsch, line 199: Is this valid syntax for the driver? We definitely don't want to be more permissive than the driver.

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,81 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        - convert_channels_repeated_capabilities('Dev1/1', 'Dev1/0') --> ['Dev1/1']
+        - convert_channels_repeated_capabilities('Dev1/0-2', 'Dev1/0') --> ['Dev1/0', 'Dev1/1', 'Dev1/2']
```
marcoskirsch, line 201: is it valid in the driver to specify the same channel twice?

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,81 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        - convert_channels_repeated_capabilities('Dev1/0:2', 'Dev1/0') --> ['Dev1/0', 'Dev1/1', 'Dev1/2']
+        - convert_channels_repeated_capabilities('Dev1/0:2,4', 'Dev1/0') --> ['Dev1/0', 'Dev1/1', 'Dev1/2', 'Dev1/4']
```
marcoskirsch, line 203: Is this valid syntax for the driver?

build/templates/_converters.py.mako
olsl21: Yes, '2:3,0' is a valid syntax for the driver.
```
>>> import nidcpower
>>> with nidcpower.Session(resource_name="Dev1", channels="", options="Simulate=1, DriverSetup=Model:4162; BoardType:PXIe") as smu:
...     with smu.initiate():
...         print(",".join(measurement.channel for measurement in smu.channels["2:3,0"].measure_multiple()))
...
Dev1/2,Dev1/3,Dev1/0
```

In fact, this example is tested in the newly added NI-DCPower system test [`test_measure_multiple_channels()` ](https://github.com/ni/nimi-python/pull/1754/files#diff-588aa0a2d97cf60012b577567bec4ae6bad13d2d7201507a924b8aa150bbe328R284)

build/templates/_converters.py.mako
olsl21: Yes, 'Dev1/0:2,4' is a valid syntax for the driver:
```
>>> import nidcpower
>>> with nidcpower.Session(resource_name="Dev1", channels="", options="Simulate=1, DriverSetup=Model:4162; BoardType:PXIe") as smu:
...     with smu.initiate():
...         print(",".join(measurement.channel for measurement in smu.channels["Dev1/0:2,4"].measure_multiple()))
...
Dev1/0,Dev1/1,Dev1/2,Dev1/4
```

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,81 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+    return convert_repeated_capabilities(repeated_capability)
+
+
+def convert_channels_repeated_capabilities(
+    channels_repeated_capability,
+    first_channel_name
+):
+    '''Convert a channels repeated capabilities string, possibly with no or multiple prefixes (each ends with '/'), to a list
+
+    Examples:
```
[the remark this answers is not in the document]
olsl21: I have added unit tests for these newly created converter functions (i.e. `convert_channels_repeated_capabilities()` and `convert_single_group_repeated_capabilities()` in my latest commit.

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,81 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        - convert_channels_repeated_capabilities('1', '0') --> ['1']
+        - convert_channels_repeated_capabilities('0-2', '0') --> ['0', '1', '2']
+        - convert_channels_repeated_capabilities('0:2', '0') --> ['0', '1', '2']
+        - convert_channels_repeated_capabilities('0:2,4', '0') --> ['0', '1', '2', '4']
+        - convert_channels_repeated_capabilities('4,1:2', '1') --> ['4', '1', '2']
+        - convert_channels_repeated_capabilities('2:3,0', 'Dev1/0') --> ['Dev1/2', 'Dev1/3', 'Dev1/0']
+        - convert_channels_repeated_capabilities('Dev1/1', 'Dev1/0') --> ['Dev1/1']
```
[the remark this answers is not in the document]
olsl21: It is a valid example:
```
>>> import nidcpower
>>> with nidcpower.Session(resource_name="Dev1", channels="", options="Simulate=1, DriverSetup=Model:4162; BoardType:PXIe") as smu:
...     print(smu.get_channel_names([0])[0])
...     with smu.initiate():
...         print(",".join(measurement.channel for measurement in smu.channels["Dev1/1"].measure_multiple()))
...
Dev1/0
Dev1/1
```

The second parameter of `convert_channels_repeated_capabilities()` is actually `first_channel_name`, which may or may not appear in the `channels_repeated_capabilities` string since user can choose to open a session to all channels but only perform some operations on certain channels.

I have updated the examples by making the second parameter a keyword parameter (e.g. `convert_channels_repeated_capabilities('Dev1/1', first_channel_name='Dev1/0') --> ['Dev1/1']` so that it is clearer that it is not part of the `channels_repeated_capabilities` string input.

I have also included all the examples as test cases for the newly created unit tests in "test_converters.py".

build/templates/_converters.py.mako
olsl21: The driver would throw an error if the same channel is specified twice, but here none of the channels (i.e. 'Dev1/0', 'Dev1/1', 'Dev1/2') are specified twice. `('Dev1/0-2', 'Dev1/0')` are actually two separate parameters: 'Dev1/0' is actually `first_channel_name`, the second parameter of `convert_channels_repeated_capabilities()`, and not part of the `channels_repeated_capabilities` string input.

I have updated the examples by making the second parameter a keyword parameter (e.g. `convert_channels_repeated_capabilities('Dev1/0-2', first_channel_name='Dev1/0') --> ['Dev1/0', 'Dev1/1', 'Dev1/2']` so that it is clearer that it is a separate parameter.

build/templates/_converters.py.mako
marcoskirsch: I'm surprised that doesn't error! And I say this as the tech lead for the NI-DCPower Independent Channels tech lead.
Screen Shot 2022-06-30 at 2 09 51 PM

sbethur (changes requested):

generated/nidcpower/nidcpower/session.py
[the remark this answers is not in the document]
sbethur: We could push the if/else logic down to `convert_channels_repeated_capabilities`, especially given the function would need access to "all channels in the session" to handle the 2 TBDs mentioned [here](https://github.com/ni/nimi-python/pull/1754#discussion_r912223170).

Regarding getting and storing "all channels in the session", I was thinking we add a function to `_SessionBase` called `_get_session_channels`, in where we call `_get_channel_names`. We would need to surround the driver runtime function call with `try/catch` since not all driver runtimes support it.
Additionally we could use the `@lru_cache` decorator to ensure we do the actual work only once.

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -149,6 +149,81 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+    if '/' in repeated_capability:
+        split_index = repeated_capability.find('/') + 1
+        instrument_prefix = repeated_capability[:split_index]
+        return convert_repeated_capabilities(repeated_capability[split_index:], instrument_prefix)
+    return convert_repeated_capabilities(repeated_capability)
+
+
+def convert_channels_repeated_capabilities(
+    channels_repeated_capability,
+    first_channel_name
```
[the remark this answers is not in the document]
olsl21: > `first_channel_name` is confusing here since the function only needs the instrument name.

`first_channel_name` is not only used to extract the instrument name prefix, it is also used to determine whether the session is an independent-channels session (only independent-channels session would return `first_channel_name` with instrument prefix). I could rename it to `example_channel_name` if you prefer.

> But the more important point is, you need the instrument name because you're embedding FQCNs (Fully-qualified channel name like `dev1/4`) in the returned struct/namedtuple, in the case of single-instrument sessions.

Not really, I'm embedding FQCNs in the case of **independent-channels** sessions, which may be single-instrument or multi-instrument. This is also why I need `first_channel_name` to determine whether the session is an independent-channels session as mentioned earlier.

> I think you decided to embed FQCNs in single-instrument sessions because you need to take care of mixing i.e. an input like 4,Dev1/1:2'.

Nope, I'm embedding FQCNs in single-instrument session only when the session is an independent-channels session. In fact, non-FQCNs will be embedded in single-instrument sessions that are non-independent-channels:

```
>>> with nidcpower.Session(resource_name="dev1", channels="", options="Simulate=1, DriverSetup=Model:4162; BoardType:PXIe", independent_channels=False) as smu:
...     with smu.initiate():
...         smu.channels["4,1:2"].measure_multiple()
...
[Measurement(voltage=0.0, current=1e-05, in_compliance=None, channel='4'), Measurement(voltage=0.0, current=1e-05, in_compliance=None, channel='1'), Measurement(voltage=0.0, current=1e-05, in_compliance=None, channel='2')]
```

This **decision** was made based on the convention of the existing `get_channel_names()`, where **FQCNs are returned for independent-channels session** (regardless of whether it is single-instrument or multi-instrument) and **non-FQCNs are returned for non-independent-channels session** (this will always be single-instrument).

image

@sbethur @marcoskirsch With these being said, I believe the first 3 points of what you're proposing are not entirely valid since they are not consistent with the behavior of `get_channel_names()` in driver runtime. Point 4 is quite a different issue, and doing it would mean that it would no longer be consistent with the driver behavior. Nevertheless, I can file a GitHub issue for it if it's the desired behavior.

generated/nidcpower/nidcpower/session.py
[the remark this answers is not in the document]
olsl21: > We could push the if/else logic down to `convert_channels_repeated_capabilities`, especially given the function would need access to "all channels in the session" to handle the 2 TBDs mentioned https://github.com/ni/nimi-python/pull/1754#discussion_r912223170.

@sbethur Do you mean I should change the function signature of `convert_channels_repeated_capabilities()` to something like
```
def convert_channels_repeated_capabilities(
    channels_repeated_capability,
    session_channel_names
):
```
and call if from `measure_multiple()` like
```
channel_names = _converters.convert_channels_repeated_capabilities(
    self._repeated_capability,
    self._get_channel_names(range(self.channel_count))
)
```
?

Please note that I've replied to your comment regarding the 2 TBDs at https://github.com/ni/nimi-python/pull/1754#discussion_r912875369.

> Regarding getting and storing "all channels in the session"

Can you clarify what do you mean by "all channels in the session"? Do you mean channels that are opened as part of the Session constructor? Isn't that what `get_channel_names()` is currently doing?

If you mean channels that are specified by `session.channels[...]`, isn't that what my `convert_channels_repeated_capabilities()` is trying to achieve here?

generated/nidcpower/nidcpower/_converters.py
```diff
@@ -149,6 +149,81 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        - convert_channels_repeated_capabilities('0:2,4', first_channel_name='0') --> ['0', '1', '2', '4']
+        - convert_channels_repeated_capabilities('4,1:2', first_channel_name='1') --> ['4', '1', '2']
+        - convert_channels_repeated_capabilities('2:3,0', first_channel_name='Dev1/0') --> ['Dev1/2', 'Dev1/3', 'Dev1/0']
+        - convert_channels_repeated_capabilities('Dev1/1', first_channel_name='Dev1/0') --> ['Dev1/1']
+        - convert_channels_repeated_capabilities('Dev1/0-2', first_channel_name='Dev1/0') --> ['Dev1/0', 'Dev1/1', 'Dev1/2']
+        - convert_channels_repeated_capabilities('Dev1/0:2', first_channel_name='Dev1/0') --> ['Dev1/0', 'Dev1/1', 'Dev1/2']
+        - convert_channels_repeated_capabilities('Dev1/0:2,4', first_channel_name='Dev1/0') --> ['Dev1/0', 'Dev1/1', 'Dev1/2', 'Dev1/4']
+        - convert_channels_repeated_capabilities('4,Dev1/1:2', first_channel_name='Dev1/1') --> ['Dev1/4', 'Dev1/1', 'Dev1/2']
+        - convert_channels_repeated_capabilities('Dev1/4,Dev1/2,Dev1/3', first_channel_name='Dev1/2') --> ['Dev1/4', 'Dev1/2', 'Dev1/3']
+
```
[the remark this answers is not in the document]
olsl21: Yes it does, I've added it to the examples here as well as the `test_convert_channels_repeated_capabilities_with_prefix()` test case in nifake `test_converters.py` in my latest commit.

generated/nidcpower/nidcpower/_converters.py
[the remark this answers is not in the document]
olsl21: After discussing with @marcoskirsch, it seems like "single-instrument sessions" and "multi-instrument sessions" mentioned in the original comment were actually terms used in NI-SCOPE that are equivalent to "non-independent-channels sessions" and "independent-channels sessions" in NI-DCPower.

> you need the instrument name because you're embedding FQCNs (Fully-qualified channel name like dev1/4) in the returned struct/namedtuple, in the case of single-instrument sessions.

Given the new interpretation of "single-instrument sessions", nope, I'm not embedding FQCNs in the case of single-instrument sessions / non-independent-channels sessions, as shown in the test cases of `test_fetch_multiple_channels()` where `independent_channels` is set to False and the `expected_measured_channel` is non-FQCN:
https://github.com/ni/nimi-python/pull/1754/files#diff-588aa0a2d97cf60012b577567bec4ae6bad13d2d7201507a924b8aa150bbe328R239

As for the proposed points, the first 2 points are already fulfilled, while the 3rd proposed point would never really occur, since user cannot pass in FQCNs for an NI-DCPower's "single-instrument session" / "non-independent-channels session":

```
>>> with nidcpower.Session(resource_name="Dev1", channels="", options="Simulate=1, DriverSetup=Model:4162; BoardType:PXIe", independent_channels=False) as smu:
...     smu.channels["Dev1/1"].voltage_level
...

nidcpower.errors.DriverError: -1074135008: IVI: Unknown channel or repeated capability name.

Channel Name: Dev1
```

With this, I believe all of the points have been addressed. I shall mark this conversation as resolved for now, feel free to reopen it if there's any further action item.

generated/nidcpower/nidcpower/session.py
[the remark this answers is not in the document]
olsl21: I have made the change for `convert_channels_repeated_capabilities()` to allow pushing the if/else logic from the "fancy fetch" body to the converter function in my latest commit:
https://github.com/ni/nimi-python/pull/1754/commits/8641fc8cd4031be0a04f86f97f575dc470fea3a3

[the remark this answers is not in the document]
sbethur: > Can you clarify what do you mean by "all channels in the session"? Do you mean channels that are opened as part of the Session constructor? Isn't that what get_channel_names() is currently doing?

Yes, "channels that are opened as part of the Session constructor". 

>  Isn't that what get_channel_names() is currently doing?
It is. This data is constant throughout the session. So we could fetch it from the driver runtime once and store it as class state.

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,98 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+    if '/' in repeated_capability:
+        split_index = repeated_capability.find('/') + 1
+        instrument_prefix = repeated_capability[:split_index]
+        return convert_repeated_capabilities(repeated_capability[split_index:], instrument_prefix)
+    return convert_repeated_capabilities(repeated_capability)
+
+
+def convert_channels_repeated_capabilities(
+    channels_repeated_capability,
+    session_channel_names
```
marcoskirsch, line 189: call this `all_channels_in_session`

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,98 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+):
+    '''Convert a channels repeated capabilities string, possibly with no or multiple prefixes (each ends with '/'), to a list
```
marcoskirsch, line 191: ```Expands a channel_string to a list of individual channel names. The individual channel names may or may not be fully qualified channel names as applicable for the session.```

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,98 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+
+    Examples:
+        - convert_channels_repeated_capabilities('1', ['0', '1', '2', '3']) --> ['1']
+        - convert_channels_repeated_capabilities('0-2', ['0', '1', '2', '3']) --> ['0', '1', '2']
```
marcoskirsch, line 195: Suggest having less examples here, just enough to figure out what the function does.
The unit tests should be comprehensive of course.

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,98 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+            repeated_capabilities.extend(convert_single_group_repeated_capabilities(rep_cap))
+        # If there is any channels repeated capabilities without prefix, the session must have only
+        # one instrument, so just get the prefix from session_channel_name and add to all of the
+        # repeated capabilities that are without prefix
+        default_prefix = session_channel_name[:session_channel_name.find('/') + 1]
+        return [
+            rep_cap if '/' in rep_cap else default_prefix + rep_cap
+            for rep_cap in repeated_capabilities
+        ]
+    return convert_repeated_capabilities(channels_repeated_capability, '')
```
marcoskirsch, line 250: this should be inside an `else` for readability

generated/nidcpower/nidcpower/session.py
```diff
@@ -4869,11 +4882,24 @@ def measure_multiple(self):
...
         import collections
-        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance'])
+        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance', 'channel'])
 
         voltage_measurements, current_measurements = self._measure_multiple()
 
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=None) for i in range(self._parse_channel_count())]
+        channel_names = _converters.convert_channels_repeated_capabilities(
+            self._repeated_capability,
+            session_channel_names=self._get_channel_names(range(self._parse_channel_count()))
```
marcoskirsch, line 4891: It'd be nice if we stored this in the Session (and pass it to SessionBase) objects so we're not creating this list on every function call.

`self._all_channels_in_session`

marcoskirsch: if we do this... do we still `_get_channel_names` to be accessible from SessionBase? I think no!

marcoskirsch: Populating `self._all_channels_in_session` can be done in the shared template for all modules, surrounded by `try`/`except`.

generated/nifake/nifake/unit_tests/test_converters.py
```diff
@@ -294,6 +294,92 @@ def test_repeated_capabilities_invalid_resource_names():
...
+    test_result_list = _converters.convert_single_group_repeated_capabilities('0-2')
+    assert test_result_list == ['0', '1', '2']
+    test_result_list = _converters.convert_single_group_repeated_capabilities('0:2')
+    assert test_result_list == ['0', '1', '2']
+    test_result_list = _converters.convert_single_group_repeated_capabilities('Dev1/0')
+    assert test_result_list == ['Dev1/0']
+    test_result_list = _converters.convert_single_group_repeated_capabilities('Dev1/0-2')
+    assert test_result_list == ['Dev1/0', 'Dev1/1', 'Dev1/2']
+    test_result_list = _converters.convert_single_group_repeated_capabilities('Dev1/0:2')
+    assert test_result_list == ['Dev1/0', 'Dev1/1', 'Dev1/2']
```
marcoskirsch, line 309: we're missing tests with random but valid whitespace.
we're missing tests with different capitalization.

generated/nifake/nifake/unit_tests/test_converters.py
```diff
@@ -294,6 +294,92 @@ def test_repeated_capabilities_invalid_resource_names():
     assert test_result_list == '0/1:2,'
 
 
+def test_convert_single_group_repeated_capabilities():
```
marcoskirsch, line 297: remember to rename the test per the function

generated/nifake/nifake/unit_tests/test_converters.py
```diff
@@ -294,6 +294,92 @@ def test_repeated_capabilities_invalid_resource_names():
...
+    assert test_result_list == ['0', '1', '2']
+    test_result_list = _converters.convert_single_group_repeated_capabilities('Dev1/0')
+    assert test_result_list == ['Dev1/0']
+    test_result_list = _converters.convert_single_group_repeated_capabilities('Dev1/0-2')
+    assert test_result_list == ['Dev1/0', 'Dev1/1', 'Dev1/2']
+    test_result_list = _converters.convert_single_group_repeated_capabilities('Dev1/0:2')
+    assert test_result_list == ['Dev1/0', 'Dev1/1', 'Dev1/2']
+
+
+def test_convert_channels_repeated_capabilities_without_prefix():
```
marcoskirsch, line 312: Instead of "with/without prefix" use term "fully_qualified_channel_name" and "not_fully_qualified_channel_name".
This is a widely used concept by MI driver developers and public driver documentation to distinguish "0" from "Dev1/0".

generated/nifake/nifake/unit_tests/test_converters.py
```diff
@@ -294,6 +294,92 @@ def test_repeated_capabilities_invalid_resource_names():
...
+        ['0', '1', '2', '3']
+    )
+    assert test_result_list == ['0', '1', '2']
+    test_result_list = _converters.convert_channels_repeated_capabilities(
+        '0:2',
+        ['0', '1', '2', '3']
+    )
+    assert test_result_list == ['0', '1', '2']
+    test_result_list = _converters.convert_channels_repeated_capabilities(
+        '0:2,4',
```
marcoskirsch, line 329: you need to test random but valid whitespace

generated/nifake/nifake/unit_tests/test_converters.py
```diff
@@ -294,6 +294,92 @@ def test_repeated_capabilities_invalid_resource_names():
...
+        ['Dev1/1', 'Dev1/2', 'Dev1/4']
+    )
+    assert test_result_list == ['Dev1/4', 'Dev1/1', 'Dev1/2']
+    test_result_list = _converters.convert_channels_repeated_capabilities(
+        'Dev1/4,Dev1/2,Dev1/3',
+        ['Dev1/2', 'Dev1/3', 'Dev1/4']
+    )
+    assert test_result_list == ['Dev1/4', 'Dev1/2', 'Dev1/3']
+    test_result_list = _converters.convert_channels_repeated_capabilities(
+        'Dev1/1,Dev2/2',
```
marcoskirsch, line 377: need tests for random and valid whitespace

generated/nifake/nifake/unit_tests/test_converters.py
```diff
@@ -294,6 +294,92 @@ def test_repeated_capabilities_invalid_resource_names():
...
+        ['Dev1/0', 'Dev1/1', 'Dev1/2', 'Dev1/3', 'Dev2/0', 'Dev2/1', 'Dev2/2', 'Dev2/3']
+    )
+    assert test_result_list == ['Dev1/1', 'Dev2/2']
+
```
marcoskirsch, line 381: need failing but disabled tests for different capitalization

marcoskirsch (changes requested):

build/templates/_converters.py.mako
sbethur, line 250: I think we can simplify all this logic and get rid of the helper function using `str.rpartition`..

Here's what I'm thinking:
```
# Rule 1: If `all_channels_in_session` is fully-qualified then returned channel names should be
#         fully-qualified. Else returned channel names should not be fully-qualified.
# Rule 2: If any channel in the input is not fully-qualified, but we need to return fully-qualified
#         channels because of Rule 1, then use the channel qualifier obtained from
#         `all_channels_in_session`. This can only happen on a single-instrument session, so all
#         the channel qualifiers are the same and we pick the first one.

instrument, separator, channel = all_channels_in_session[0].rpartition('/')
default_channel_qualifier = instrument + separator

expanded_channel_list = []
for token in channels_string.split(','):
    instrument, separator, channel = token.rpartition('/')
    channel_qualifier = instrument + separator if instrument else default_channel_qualifier
    expanded_channel_list.extend(convert_repeated_capabilities(channel, instrument + separator))

return expanded_channel_list
```

sbethur (changes requested):

build/templates/_converters.py.mako
sbethur: Note the use of `strip` and `lower` to make FQCN be in canonical form.

@marcoskirsch, to make the channel names canonical, we can always return all-lower-case or do we need to match what is returned by `get_channel-names`? (which follows the casing passed into resource name during init?)

generated/nidcpower/nidcpower/session.py
```diff
@@ -4822,11 +4822,24 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
+        )
+        return [
+            Measurement(
+                voltage=voltage,
+                current=current,
+                in_compliance=in_compliance,
+                channel=channel_names[0]
```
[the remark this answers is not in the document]
olsl21: I have added the assertion in my latest commit.

build/templates/_converters.py.mako
olsl21: Thanks for the suggestion. I have combined the two converter functions into one single function `expand_channel_string()` in my latest commit. This function takes care of removing the "valid whitespaces" and returning canonical channel names (based on what are returned by `get_channel_names()`; in fact, the channel names returned by this new function would be a subset of the channel names returned by `get_channel_names()`).

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,98 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+    '''
+    if '/' in repeated_capability:
+        split_index = repeated_capability.find('/') + 1
+        instrument_prefix = repeated_capability[:split_index]
+        return convert_repeated_capabilities(repeated_capability[split_index:], instrument_prefix)
+    return convert_repeated_capabilities(repeated_capability)
+
+
+def convert_channels_repeated_capabilities(
+    channels_repeated_capability,
```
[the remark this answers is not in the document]
olsl21: I have used this parameter name for the new `expand_channel_string()` converter function.

build/templates/_converters.py.mako
olsl21: I have integrated this suggested text into the docstring for the new converter function `expand_channel_string()`.

build/templates/_converters.py.mako
olsl21: I have removed some unnecessary examples for the new converter function `expand_channel_string()`.

generated/nifake/nifake/unit_tests/test_converters.py
olsl21: I have renamed the test function based on the new converter function name.

generated/nifake/nifake/unit_tests/test_converters.py
olsl21: The new unit tests are now named `test_expand_channel_string_non_fully_qualified_channel_names()` and `test_expand_channel_string_fully_qualified_channel_names()` respectively.

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -233,6 +233,88 @@ def test_measure_multiple(session):
...
+        ('Dev1', '0,2-3,6', False, '2-3,0', ['2', '3', '0']),
+        ('Dev1', None, True, None, [f'Dev1/{x}' for x in range(12)]),
+        ('Dev1', '', True, '', [f'Dev1/{x}' for x in range(12)]),
+        ('Dev1', ' ', True, ' ', [f'Dev1/{x}' for x in range(12)]),
+        ('Dev1', '0:2,Dev1/4', True, 'Dev1/1:2,0', ['Dev1/1', 'Dev1/2', 'Dev1/0']),
+        ('Dev1', 'Dev1/0:2,Dev1/4', True, 'Dev1/1:2,Dev1/0', ['Dev1/1', 'Dev1/2', 'Dev1/0']),
+        ('Dev1/1', '', True, None, ['Dev1/1']),
+        ('Dev1/1:4', ' ', True, '2:3', ['Dev1/2', 'Dev1/3']),
+        ('Dev1/0-1,Dev2/0:2', ' ', True, 'Dev2/0-1,Dev1/0', ['Dev2/0', 'Dev2/1', 'Dev1/0']),
+        ('Dev1/1:4,Dev1/6', ' ', True, '6,2-3', ['Dev1/6', 'Dev1/2', 'Dev1/3']),
```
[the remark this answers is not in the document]
olsl21: The first and third test cases have been added to the new `test_measure_multiple_error_invalid_channels()` test function in my latest commit. 

The second test case would fail during session initialization as what you expected, and it is already tested as part of [`test_init_raises_value_error_for_multi_instrument_resource_name_and_channels_argument()`](https://github.com/ni/nimi-python/blob/master/src/nidcpower/system_tests/test_system_nidcpower.py#L491), so I did not create a new test case for it.

build/templates/_converters.py.mako
marcoskirsch: > or do we need to match what is returned by get_channel-names? 

This is the "correct" behavior in my opinion.

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,75 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+            supported formats (this string is expected to be used as the index of session.channels)
+
+        all_channels_in_session (list of str) - names of all the channels in the session as returned
+            by get_channel_names()
+
+    Returns:
+        channel_names (list of str) - list of strings of the expanded channels repeated capability
+            items (essentially the channel names)
+    '''
+    channel_string = channel_string.strip()
```
marcoskirsch, line 197: I don't think you need to mutate the user's passed parameter.
Consider removing the assignment and simply doing

```
if channel_string.strip() == '':
```

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,75 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        )
+
+    # Convert the expanded channel names to their canonical form based on all_channels_in_session
+    lowercase_channel_name_to_session_channel_name_dict = {
+        channel_name.lower(): channel_name for channel_name in all_channels_in_session
+    }
+    return [
+        lowercase_channel_name_to_session_channel_name_dict[channel_name.lower()]
+        for channel_name in expanded_channel_list
+    ]
```
marcoskirsch, line 227: I found the implementation of this function to be easy to follow and well documented. Not an easy thing to accomplish! Good job.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4822,11 +4822,28 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
-
-        voltage_measurements, current_measurements, in_compliance = self._fetch_multiple(timeout, count)
-
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=in_compliance[i]) for i in range(count)]
+        Measurement = collections.namedtuple('Measurement', ['voltage', 'current', 'in_compliance', 'channel'])
+
+        voltage_measurements, current_measurements, in_compliances = self._fetch_multiple(timeout, count)
+
+        with _NoChannel(session=self):
+            all_channels_in_session = self._get_channel_names(range(self.channel_count))
```
marcoskirsch, line 4830: Add a comment here:

```
#TODO(olsl21): retrieving the list of channels in the session on every function call is silly because they never change #12345
```

Where 12345 is the issue number for fixing this.

generated/nidcpower/nidcpower/session.py
```diff
@@ -4822,11 +4822,28 @@ def fetch_multiple(self, count, timeout=hightime.timedelta(seconds=1.0)):
...
+
+        channel_names = _converters.expand_channel_string(
+            self._repeated_capability,
+            all_channels_in_session
+        )
+        assert len(channel_names) == 1
```
marcoskirsch, line 4836: `assert len(channel_names) == 1, "fetch_multiple only supports one channel at a time"`

generated/nidcpower/nidcpower/session.py
```diff
@@ -4869,11 +4886,27 @@ def measure_multiple(self):
...
         voltage_measurements, current_measurements = self._measure_multiple()
 
-        return [Measurement(voltage=voltage_measurements[i], current=current_measurements[i], in_compliance=None) for i in range(self._parse_channel_count())]
+        with _NoChannel(session=self):
+            all_channels_in_session = self._get_channel_names(range(self.channel_count))
+
+        channel_names = _converters.expand_channel_string(
+            self._repeated_capability,
+            all_channels_in_session
+        )
```
marcoskirsch, line 4899: assert lenght of channel_names is the same as voltage_measurements and current_measurements

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -233,6 +233,115 @@ def test_measure_multiple(session):
         assert measurements[1].current == 0.00001
 
 
+@pytest.mark.parametrize(
+    'resource_name,channels,independent_channels,measurement_channels,expected_measured_channel',
+    [
+        ('Dev1', '1', False, None, '1'),
```
marcoskirsch, line 239: I'm not sure how much runtime all these test cases add. But given that the bulk of the logic is dealing with channel names and that is thoroughly tested in unit tests, I think you can do a handful of test cases here and that should suffice. No need to stress things that were already stressed (redundant coverage), it's even detrimental as it can be confusing and harder to maintain.

We're more concerned with testing integration here.

This applies to the all the new tests, and perhaps the need to add more checks to the existing tests for the two modified methods.

I am not giving you specific feedback on what to add / remove, rather guideline on how to approach it.

marcoskirsch (changes requested):

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,75 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+    all_channels_in_session.
+
+    Examples:
+        - expand_channel_string('1', ['0', '1', '2', '3']) --> ['1']
+        - expand_channel_string('4,1:2', ['1', '2', '4']) --> ['4', '1', '2']
+        - expand_channel_string('2:3,0', ['Dev1/0', 'Dev1/1', 'Dev1/2', 'Dev1/3'])
+            --> ['Dev1/2', 'Dev1/3', 'Dev1/0']
+        - expand_channel_string('Dev1/1', ['Dev1/0', 'Dev1/1', 'Dev1/2', 'Dev1/3'])
+            --> ['Dev1/1']
+        - expand_channel_string('4,Dev1/1:2', ['Dev1/1', 'Dev1/2', 'Dev1/4'])
```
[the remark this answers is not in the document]
olsl21: Yes, you're right that `"4,Dev1/1:2"` is a valid string if:
- The session is "independent channels" _and_
- There is only one instrument

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,75 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+
+    Args:
+        channel_string (str) - refer to _convert_repeated_capabilities() for the
+            supported formats (this string is expected to be used as the index of session.channels)
+
+        all_channels_in_session (list of str) - names of all the channels in the session as returned
+            by get_channel_names()
+
+    Returns:
+        channel_names (list of str) - list of strings of the expanded channels repeated capability
```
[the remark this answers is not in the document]
olsl21: I have updated the docstring in my latest commit.

build/templates/_converters.py.mako
olsl21: I have updated it in my latest commit.

build/templates/_converters.py.mako
```diff
@@ -158,6 +158,75 @@ def convert_repeated_capabilities_without_prefix(repeated_capability):
...
+        return all_channels_in_session
+
+    # Rule 1: If all_channels_in_session is fully-qualified then returned channel names should be
+    #         fully-qualified, otherwise returned channel names should not be fully-qualified.
+    # Rule 2: If any channel in the input is not fully-qualified, but we need to return
+    #         fully-qualified channels because of Rule 1, then use the channel qualifier obtained
+    #         from all_channels_in_session. This can only happen on a single-instrument session,
+    #         so all the channel qualifiers are the same and we pick the first one.
+
+    instrument, separator, channel = all_channels_in_session[0].rpartition('/')
```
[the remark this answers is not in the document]
olsl21: Yes, the driver runtime will return an error. I have added this as a test case for `test_measure_multiple_error_invalid_channels()` in https://github.com/ni/nimi-python/pull/1754/files#diff-588aa0a2d97cf60012b577567bec4ae6bad13d2d7201507a924b8aa150bbe328R328

generated/nidcpower/nidcpower/session.py
olsl21: I have created #1776 and added the TODO comment in my latest commit.

generated/nidcpower/nidcpower/session.py
olsl21: I have added this assertion for `measure_multiple()` in my latest commit.

src/nidcpower/system_tests/test_system_nidcpower.py
olsl21: Thanks for the info. I have gone through the test cases once more and I think they are essential for covering the various integration scenarios, e.g. mixing different types of `resource_name` and `channels` during session initialization with different `measurement_channels` string (including passing no string, i.e. calling `session.fetch_multiple()` directly without `channels`) as well as independent vs. non-independent channels session.

I shall retain the test cases as is for now, feel free to reopen this conversation if there are any particular test cases which you think are redundant.

generated/nidcpower/nidcpower/session.py
olsl21: I have added the assertion error message in my latest commit.

sbethur (approved):

marcoskirsch (approved):

PR #1757, opened 2022-06-17
NI-DCPower: Make get_channel_names() accessible in _SessionBase

### LINKED ISSUE #1756: NI-DCPower: `get_channel_names()` should be accessible privately in `_SessionBase`
### Description of issue

A user can use `get_channel_names` as follows: `my_session.channels[[0,1,2]].get_channel_names("0-2")`.

This is equivalent to  no sense. It's equivalent in functionality to `my_session.get_channel_names("0-2")` but the `channels[]` part is ignored. It should not be allowed to cal lthe method in this manner.

The reason this happens is that currently, `get_channel_names()` is declared as a method of `Session` as it does not require passing `_repeated_capability` to the corresponding C API function, but it need to be accessible by `fetch_multiple()` and `measure_multiple()` in order for them to populate their Measurement namedtuple instances with channel names (kindly refer to #1752). One way of solving this is to create private method for it and put it in `_SessionBase`.

olsl21:

- Create `get_channel_names` method template under /build/templates
- Update NI-DCPower metadata to 22.5.0d135
  - Create private method for `get_channel_names()` and move its core logic to the private version
  - Make the current public method acts like a wrapper method that calls into its corresponding private method
    - Make use of the `get_channel_names` method template
  - Minor doc updates

* Fix #1756

Manually tested that `session.channels[""]._get_channel_names()` works as expected.


src/nidcpower/metadata/functions.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-DCPower API metadata version 22.5.0d65
+# This file is generated from NI-DCPower API metadata version 22.5.0d9999
```
olsl21, line 2: I've update this when an official export is available.

generated/nidcpower/nidcpower/session.py
```diff
@@ -7551,6 +7551,28 @@ def _initialize_with_independent_channels(self, resource_name, reset, option_str
...
+    @ivi_synchronized
+    def get_channel_names(self, indices):
+        '''get_channel_names
+
+        Returns a list of channel names for the given channel indices.
+
+        Args:
+            indices (basic sequence types or str or int): Index list for the channels in the session. Valid values are from zero to the total number of channels in the session minus one. The index string can be one of the following formats:
+
+                -   A comma-separated list—for example, "0,2,3,1"
```
marcoskirsch, line 7563: We've had similar debates over repeated capabilities documentation in the past.
Essentially, passing strings and our insane syntax (this function is not as insane but still...) is the only way to use this in C, CVI, C#, LabVIEW.

But in Python we support (and it needs to be in the tests) passing any iterable. This is the most pythonic way of using this so it's the one we should be promoting.

Since this is a change in metadata that might have to be done internally at NI, I'm ok if you follow up with an issue so we can merge this sooner and unblock further work.

marcoskirsch: Upon more detailed reader, I think it's fine. I don't like the emphasis given to the string stuff but I can live with it.

src/nidcpower/metadata/functions.py
```diff
@@ -1815,7 +1817,8 @@
                 'type_in_documentation': 'list of str'
             }
         ],
-        'python_name': 'get_channel_names',
+        'python_name': '_get_channel_names',
```
marcoskirsch, line 1820: is this necessary? won't the nimi-python generator expand it this way?

src/nidcpower/templates/session.py/delegate_to_private_method.py.mako
```diff
@@ -0,0 +1,13 @@
...
+    import build.helper as helper
+
+    suffix = method_template['method_python_name_suffix']
+%>\
+    def ${f['python_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.SESSION_METHOD_DECLARATION)}):
+        '''${f['python_name']}
+
+        ${helper.get_function_docstring(f, False, config, indent=8)}
+        '''
+        return self._${f['python_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.SESSION_METHOD_PASSTHROUGH_CALL)})
```
marcoskirsch, line 12: Here there is a hidden contract that the private method's name will be identical to the public's method name; even though there's nothing enforcing that elsewhere. The method names can be specified to be arbitrary via metadata.

There is another less evil hidden contract that the parameter names are the same. This I think is more defensible.

I like that you are trying to solve this in a generic manner (but notice the template is specific to nidcpower based on its location) but I don't know that it's worth this design compromise. Are there other opportunities for us to simplify nimi-python and use this template? If there isn't I think we're better off not prematurely generalizing and making this template specific to `get_channel_names()`.

Now, there is a need to do the same in `nidigital` per #1758 so what I think you need to do is the following:

1. Create a `get_channel_names` template. Don't make it `nidcpower`-specific. This should be added and unit-tested in `nifake`.
2. Apply it to `nidcpower`
3. Apply it to `nidigital`

In separate PRs.

marcoskirsch (changes requested):

src/nidcpower/metadata/functions.py
olsl21: I believe it is necessary, otherwise the generator would expand the method name to '_get_channel_name_from_string' as the key is 'GetChannelNameFromString'.

generated/nidcpower/nidcpower/session.py
olsl21: Ya, I think the method supports any iterable of int as-is. I shall mark this comment as resolved, feel free to reopen it if there is any further action item on it.

src/nidcpower/templates/session.py/delegate_to_private_method.py.mako
olsl21: > Are there other opportunities for us to simplify nimi-python and use this template?

I'm not aware of any currently.

> Create a get_channel_names template. Don't make it nidcpower-specific. This should be added and unit-tested in nifake.

I've created #1760 for this.

> Apply it to nidcpower

I'll apply the change from #1760 to this PR once it's submitted.

> Apply it to nidigital

I've updated the description of #1758 to also address this comment when that issue is being worked on.

marcoskirsch (changes requested):

build/templates/session.py/get_channel_names.py.mako
```diff
@@ -0,0 +1,13 @@
+<%page args="f, config, method_template"/>\
+<%
```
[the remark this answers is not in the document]
olsl21: I have added a comment to the template based on your suggestion in the latest commit.

marcoskirsch:
merge conflicts

sbethur (approved):

marcoskirsch (approved):

marcoskirsch:
merge conflicts

src/nidcpower/metadata/functions.py
olsl21: The NI-DCPower metadata has been updated to 22.5.0d135 in my latest commit.

PR #1763, opened 2022-06-24
Update DriverTooNewError, converters for enum

olsl21:

- Add KeyError try-catch for setter converter to raise a ValueError with a list of valid values instead
- Update `test_get_attribute_enum_with_converter_invalid_value_from_driver()` to expect an updated error message

Local build was successful.

### DIFF
--- a/build/templates/_attributes.py.mako
+++ b/build/templates/_attributes.py.mako
@@ -134,7 +134,10 @@ class AttributeEnumWithConverter(AttributeEnum):
             raise errors.DriverTooNewError('The driver runtime returned an unexpected value. ')
 
     def __set__(self, session, value):
-        return self._underlying_attribute_enum.__set__(session, self._setter_converter(value))
+        try:
+            return self._underlying_attribute_enum.__set__(session, self._setter_converter(value))
+        except KeyError:
+            raise ValueError(f'{value} cannot be converted to an {str(self._underlying_attribute_enum._attribute_type.__name__)} enum value')
 
 
 # nitclk specific attribute type
--- a/build/templates/_converters.py.mako
+++ b/build/templates/_converters.py.mako
@@ -340,11 +340,14 @@ def ${enums[enum_name]['enum_to_converted_value_function_name']}(value):
 
 
 def ${enums[enum_name]['converted_value_to_enum_function_name']}(value):
-    return {
+    try:
+        return {
         % for enum_value in enums[enum_name]['values']:
-        ${helper.get_enum_value_snippet(enum_value['converts_to_value'])}: enums.${enums[enum_name]['python_name']}.${enum_value['python_name']},
+            ${helper.get_enum_value_snippet(enum_value['converts_to_value'])}: enums.${enums[enum_name]['python_name']}.${enum_value['python_name']},
         % endfor
-    }[value]
+        }[value]
+    except KeyError:
+        raise ValueError(f"{value} cannot be converted to an ${enums[enum_name]['python_name']} enum value, valid values: ${', '.join([helper.get_enum_value_snippet(enum_value['converts_to_value']) for enum_value in enums[enum_name]['values']])}")
 
 
     % endif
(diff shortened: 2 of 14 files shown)


build/templates/_converters.py.mako
```diff
@@ -340,11 +340,14 @@ def ${enums[enum_name]['enum_to_converted_value_function_name']}(value):
...
-    return {
+    try:
+        return {
         % for enum_value in enums[enum_name]['values']:
-        ${helper.get_enum_value_snippet(enum_value['converts_to_value'])}: enums.${enums[enum_name]['python_name']}.${enum_value['python_name']},
+            ${helper.get_enum_value_snippet(enum_value['converts_to_value'])}: enums.${enums[enum_name]['python_name']}.${enum_value['python_name']},
         % endfor
-    }[value]
+        }[value]
+    except KeyError:
```
marcoskirsch, line 349: I don't think it makes sense for function that goes from "converted value" to "enum" ro raise `ValueError` while the function that converts from "enum" to "converted value" raises `KeyError`.

They should both raise the same exception on unknown values.

This is mostly academic since calling code will do its own `except`+`raise` but might as well keep it consistent so nobody is confused about it.

I think `KeyError` is fine. It's simpler.

marcoskirsch (changes requested):

build/templates/_converters.py.mako
olsl21: As explained in https://github.com/ni/nimi-python/pull/1738#discussion_r905256369 and https://github.com/ni/nimi-python/pull/1738#discussion_r905261737, the calling code (i.e. the `AttributeEnumWithConverter class) has no context info about the supported values for conversion, so it cannot provide more detailed info after it catches the KeyError.

The data flows look like below:

**Getting attribute value:**
Driver runtime --> (convert to `Enum` with enum instantiation, which throws `ValueError`) --> Converter (which throws `KeyError`) --> User

**Setting attribute value:**
User --> Converter (which throws `KeyError`) --> (passes `enum,value` into C API) --> Driver runtime

If we want consistency across both converters, we can either:
1. Throw `ValueError` from both converters, then in the `AttributeEnumWithConverter` class:
    - Don't do anything for `__set__()` (let the `ValueError` with a list of valid values passes through to user)
    - Catch `ValueError` in `__get__()` and re-raise as `DriverTooNewError`
2. Throw `KeyError` from both converters, then in the `AttributeEnumWithConverter` class:
    - Catch `KeyError` in `__set__()` and re-raise as `ValueError` with generic error message
    - Catch `KeyError` and `ValueError` in `__get__()` and re-raise as `DriverTooNewError`

I'm going with option 2 (throwing `KeyError` from both converters now. Feel free to let me know if you prefer option 1.

build/templates/_attributes.py.mako
```diff
@@ -134,7 +134,10 @@ class AttributeEnumWithConverter(AttributeEnum):
             raise errors.DriverTooNewError('The driver runtime returned an unexpected value. ')
 
     def __set__(self, session, value):
-        return self._underlying_attribute_enum.__set__(session, self._setter_converter(value))
+        try:
+            return self._underlying_attribute_enum.__set__(session, self._setter_converter(value))
+        except KeyError:
+            raise ValueError(f'{value} cannot be converted to an {str(self._underlying_attribute_enum._attribute_type.__name__)} enum value')
```
marcoskirsch, line 140: Imagine you are a Python using customer.
You set an invalid value of 3.

The fact that the 3 gets converted under the hood to some private enum is an irrelevant (to the customer) implementation detail. What are they supposed to do with that information?

Let's make this simple, useful, and not confusing.

> Invalid value: 3

marcoskirsch (changes requested):

build/templates/_converters.py.mako
marcoskirsch: It makes sense to me to _raise_ KeyError from the converter functions. This is the simplest thing to do. Client code (in this case Attribute class) has more context and can raise something more suitable for the situation.

PR #1790, opened 2022-07-22
Add `_all_channels_in_session` to `_SessionBase`

### LINKED ISSUE #1776: NI-DCPower: Store all the channel names in the session to avoid having to call `_get_channel_names()` repeatedly during fetch/measure multiple
### Description of issue

The current implementation of #1754 calls `_get_channel_names()` as part of `fetch_multiple()` and `measure_multiple()`, which is inefficient as the channel names in a session won't change. A suggestion is to store the result of calling `get_channel_names()` as a member variable of `Session` (i.e. `_all_channels_in_session`) and pass it to `_SessionBase` as part of its constructor so that it can be reused.

Related discussions:
https://github.com/ni/nimi-python/pull/1754#discussion_r912288175
https://github.com/ni/nimi-python/pull/1754#discussion_r917113448
https://github.com/ni/nimi-python/pull/1754#discussion_r918402450

olsl21:

- Add `_all_channels_in_session` state member variable to `_SessionBase` to store the result of calling `get_channel_names()` in `Session`
  - Populate `_all_channels_in_session` in the constructor of `Session` by trying to call `get_channel_names()` on all the channels of the session
    - If `get_channel_names()` or `channel_count` is not defined (e.g. in some driver), set it to a default value of None
  - Propagate this value to each `_SessionBase` instance created from `_RepeatedCapabilities`
- Update test_nidigital to expect the additional function calls during session init
- Update `fancy_fetch_measure.py.mako` and `fancy_fetch_measure_lcr.py.mako` to leverage this new member to avoid calling `_get_channel_names()` repeatedly to populate the channel names in the measurement structs

* Fix #1776

Local build succeeded, all the existing tests still passed. Manually tested that calling `measure_multiple()` multiple times using the same session no longer results in multiple calls to `_get_channel_names()`:

image
image


build/templates/session.py.mako
```diff
@@ -191,7 +198,7 @@ constructor_params = helper.filter_parameters(init_function, helper.ParameterUsa
 % if attributes:
 
 % endif
-    def __init__(self, repeated_capability_list, ${config['session_handle_parameter_name']}, library, encoding, freeze_it=False):
+    def __init__(self, repeated_capability_list, ${config['session_handle_parameter_name']}, library, encoding, freeze_it=False, all_channels_in_session=None):
```
marcoskirsch, line 201: From a logic perspective, it makes more sense for the new parameter `all_channels_in_session` to be adjacent to `repeated_capability_list` than to append it at the end.

build/templates/session.py.mako
```diff
@@ -273,7 +283,15 @@ class Session(_SessionBase):
 
         ${helper.get_function_docstring(init_function, False, config, indent=8)}
         '''
-        super(Session, self).__init__(repeated_capability_list=[], ${config['session_handle_parameter_name']}=None, library=None, encoding=None, freeze_it=False)
+        # Initialize the superclass with default values first, populate them later
+        super(Session, self).__init__(
+            repeated_capability_list=[],
```
marcoskirsch, line 288: If we're doing this, then why make them parameters of `__init__()` at all?
Alternatively, and probably cleaner, can we invoke super's `__init__()` later?

build/templates/session.py.mako
```diff
@@ -297,6 +315,15 @@ class Session(_SessionBase):
 %       endfor
         self._param_list = ', '.join(param_list)
 
+        # self.get_channel_names() and self.channel_count can only be called after the session
```
marcoskirsch, line 318: suggestion - 

Store the list of channels in the Session which is needed by some `nimi-python` modules.
Use try/except because not all the modules support channels.

build/templates/session.py.mako
```diff
@@ -297,6 +315,15 @@ class Session(_SessionBase):
...
+        # handle `self._${config['session_handle_parameter_name']}` is set
+        try:
+            self._all_channels_in_session = self.get_channel_names(range(self.channel_count))
+        except AttributeError:
+            self._all_channels_in_session = None
+
+        # _is_frozen must be set last to prevent __setattr__ from raising an exception while
```
marcoskirsch, line 325: suggestion:

Finally, set _is_frozen to True which is used to prevent clients from accidentally adding members when trying to set a property with a typo.

(describe the true reason, not the symptom of not doing it correctly)

generated/nidigital/nidigital/unit_tests/test_nidigital.py
```diff
@@ -29,6 +29,12 @@ def setup_method(self, method):
         self.patched_library.niDigital_UnlockSession.side_effect = self.side_effects_helper.niDigital_UnlockSession
         self.side_effects_helper['UnlockSession']['callerHasLock'] = True
 
+        # For _try_to_get_channel_names()
```
marcoskirsch, line 32: there is no function by this name...

But I'm curious why you added this to `nidigital`'s test but not the other modules. If anything this should be in `nifake` no?

marcoskirsch (changes requested):

marcoskirsch:
From the screenshot in the description, it appears you instrumented your local build to print() things out.
FYI you can use NI I/O Trace to get a log of function calls into the driver.

sbethur:
No comments beyond Marcos' feedback.

build/templates/session.py.mako
olsl21: This is actually the approach used by the existing code, so I'm not too sure about the actual rationale behind.

Based on my understanding, these have to be parameters of `_SessionBase.__init__()` because it may be called from here (i.e. `Session.__init__()`) and `_RepeatedCapabilities.__getitem__()` (which will actually call it with the final values instead of default values).

We couldn't invoke super's `__init__()` later because that requires calling `self.get_channel_names()` and `self.channel_count` first to get the final `all_channels_in_session`, but calling these method and attribute requires repeated capabilities related properties to be populated first, which is currently done in `_SessionBase.__init__()`, so it kinda creates a circular dependency. If we really want to invoke super's `__init__()` later, we'll need to duplicate the logic in `_SessionBase.__init__()` to `Session.__init__()`, which is not desirable in my opinion.

Hence, I suggest that we retain the current approach (my added comment should have made it clear enough that the initial values passed to the super class are just default values that would be overridden).

build/templates/session.py.mako
olsl21: Appending it at the end has the benefit of preserving backward compatibility as well as the ability to give the parameter a default value (i.e. None). However, I figured that backward compatibility might not matter much here since `_SessionBase` is a private class, so I've moved the new parameter to after `repeated_capability_list` in my latest commit.

build/templates/session.py.mako
olsl21: I've added your suggested comment on top of my existing comment in my latest commit. Since they are explaining about different things, I decided to keep them both (mine is explaining about the need to put the try/except code after `self._vi` is populated).

generated/nidigital/nidigital/unit_tests/test_nidigital.py
olsl21: Thanks for spotting this, this function had been removed in my previous commit. I've updated the comment accordingly in my latest commit.

I ran the unit tests and found that only `nidigital`'s test failed, so I only updated it. Apparently, `nifake` has no `channel_count` property, so its `_all_channels_in_session` property is actually not populated.

sbethur (approved):

build/templates/session.py.mako
marcoskirsch: > This is actually the approach used by the existing code

The fact that the existing code does things in some way does not mean it is the most correct way or that it doesn't have room for improvement.

If anything this further highlights how tightly coupled the implementations of `_SessionBase` and `Session` are, which is not great architecturally but also not something we can fix cheaply.

generated/nidigital/nidigital/unit_tests/test_nidigital.py
```diff
@@ -29,6 +29,12 @@ def setup_method(self, method):
         self.patched_library.niDigital_UnlockSession.side_effect = self.side_effects_helper.niDigital_UnlockSession
         self.side_effects_helper['UnlockSession']['callerHasLock'] = True
 
+        # For trying to set `_all_channels_in_session` in the Session constructor
+        self.patched_library.niDigital_GetAttributeViInt32.side_effect = self.side_effects_helper.niDigital_GetAttributeViInt32
+        self.side_effects_helper['GetAttributeViInt32']['value'] = 1  # channel_count
+        self.patched_library.niDigital_GetChannelNameFromString.side_effect = self.side_effects_helper.niDigital_GetChannelNameFromString
+        self.side_effects_helper['GetChannelNameFromString']['names'] = '0'  # get_channel_names()
+
```
marcoskirsch, line 37: Is there anywhere in our testing where we confirm that `_all_channels_in_session` is set correctly?
Naturally it would belong in `nifake` but I understand that would require metadata changes.

marcoskirsch (approved):

marcoskirsch (approved):

generated/nidigital/nidigital/unit_tests/test_nidigital.py
marcoskirsch: Thinking about this a bit more, I think it's fine. We have tests in place for the public facing behavior in `nidcpower` and this is what matters.

PR #1796, opened 2022-07-27
Fix AttributeEnumWithConverter to work with AdvSeq

olsl21:

- Make `AttributeEnumWithConverter` inherits from `object` instead of `AttributeEnum`
- Add `_attribute_id` property to `AttributeEnumWithConverter` to be accessed by `create_advanced_sequence` method in NI-DCPower
- Update `fancy_advanced_sequence.py.mako` method template to consider `AttributeEnumWithConverter` a valid attribute class
- Add `test_create_and_delete_advanced_sequence_attribute_enum_with_converter()` system test to NI-DCPower to ensure that attribute that uses enum with converter can be used in an advanced sequence

The newly added system test passed.

### DIFF
--- a/build/templates/_attributes.py.mako
+++ b/build/templates/_attributes.py.mako
@@ -89,12 +89,12 @@ class AttributeViBoolean(Attribute):
         session._set_attribute_vi_boolean(self._attribute_id, value)
 
 
-class AttributeEnum(object):
+class AttributeEnum(Attribute):
 
     def __init__(self, underlying_attribute_meta_class, enum_meta_class, attribute_id):
+        super(AttributeEnum, self).__init__(attribute_id)
         self._underlying_attribute = underlying_attribute_meta_class(attribute_id)
         self._attribute_type = enum_meta_class
-        self._attribute_id = attribute_id
 
     def __get__(self, session, session_type):
         return self._attribute_type(self._underlying_attribute.__get__(session, session_type))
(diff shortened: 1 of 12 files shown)


build/templates/_attributes.py.mako
```diff
@@ -122,6 +122,7 @@ class AttributeEnumWithConverter(AttributeEnum):
                 enum value
         '''
         self._underlying_attribute_enum = underlying_attribute_enum
+        self._attribute_id = underlying_attribute_enum._attribute_id
```
sbethur, line 125: Can you make this class and `AttributeEnum` class be derived from `Attribute` class, which has the `_attribute_id` attribute?

sbethur (changes requested):

generated/nidcpower/nidcpower/session.py
```diff
@@ -4880,7 +4880,7 @@ def create_advanced_sequence(self, sequence_name, property_names, set_as_active_
         for prop in property_names:
             if prop not in Session.__base__.__dict__:
                 raise KeyError('{0} is not an property on the nidcpower.Session'.format(prop))
-            if not isinstance(Session.__base__.__dict__[prop], _attributes.Attribute) and not isinstance(Session.__base__.__dict__[prop], _attributes.AttributeEnum):
+            if not isinstance(Session.__base__.__dict__[prop], _attributes.Attribute) and not isinstance(Session.__base__.__dict__[prop], _attributes.AttributeEnum) and not isinstance(Session.__base__.__dict__[prop], _attributes.AttributeEnumWithConverter):
```
marcoskirsch, line 4883: Shreyas' suggestion I think would get rid of this, which to be honest is a bit of codesmell.

marcoskirsch (changes requested):

build/templates/_attributes.py.mako
olsl21: Thanks for the suggestion, I've updated it in my latest commit.

generated/nidcpower/nidcpower/session.py
olsl21: Yes, I have updated it in my latest commit.

sbethur (approved):

marcoskirsch (approved):

PR #1817, opened 2022-08-27
Simplify NI-FGEN example and remove its use of numpy and scipy

### LINKED ISSUE #1811: nimi-python uses scipy which keeps us from adding support to Python 3.10
### Description of issue

`nimi-python` uses scipy in

1. [nifgen_script.py](https://github.com/ni/nimi-python/blob/master/src/nifgen/examples/nifgen_script.py)
2. tox-system_tests.ini install it just for `nifgen` (see above)

This is both silly on its own, but also an impediment to support Python 3.10 officially because [scipy dropped support for 32-bit Windows](https://github.com/scipy/scipy/issues/15836) which nimi-python still supports.

We should remove the need for scipy altogether.
Alternatively we could look into selectively doing things differently for Python 3.10 + Windows 32-bit but I don't think this is worth it.

### Steps to reproduce issue

#1810 see nimibot 32-bit system test results.

ni-jfitzger:

* Eliminates our use of numpy and scipy from the NI-FGEN example, and slightly simplifies the waveform creation, to keep the focus on the usage of NI-FGEN.
   *  The new waveforms will only be a single cycle. Previous waveforms were an arbitrary 1.x cycles, because the phase ended at 10, which is greater than 2*PI and not even a multiple of 2*PI.
* The Triangle waveform in the example was actually a sawtooth. We will now refer to it as such.
  * I could have added an actual triangle wave, but doing so with 0 offset would have been complicated enough that I worried I felt it would distract from the focus on the API usage.  3 or 4 lines, probably, but still the most complicated of the waveforms.
* [Scipy has stopped releasing Windows 32-wheels for scipy 1.8.0+ and for Python 3.10 and later](https://github.com/scipy/scipy/issues/15836). This is breaking nimi-python testing for PRs, when we try to add Python 3.10 support. This change fixes that.
  * Side note: building the  scipy Windows 32 wheel would be non-trivial.

* Fix #1811 

- Hand-tested updated functions, because they are part of the example code and not unit-tested.
  - Examined the output (in list form) from the updated functions and they all look correct to me.
  - I plotted the noise waveform and it looked good.
  - I generated a 1000 point noise waveform with the same algorithm and I think the max absolute value was around 0.8. It never exceed the bounds of [-1, 1].

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ All notable changes to this project will be documented in this file.
 * ### ALL
     * #### Added
     * #### Changed
+        * Fix [#1811](https://github.com/ni/nimi-python/issues/1811): nimi-python uses scipy which keeps us from adding support to Python 3.10
+          * Replaced scipy usage in NI-FGEN examples with hand-crafted functions and dropped scipy installation system-test setup. 
     * #### Removed
 * ### `nidcpower` (NI-DCPower)
     * #### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,6 @@ All notable changes to this project will be documented in this file.
 * ### ALL
     * #### Added
     * #### Changed
-        * Fix [#1811](https://github.com/ni/nimi-python/issues/1811): nimi-python uses scipy which keeps us from adding support to Python 3.10
-          * Replaced scipy usage in NI-FGEN examples with hand-crafted functions and dropped scipy installation system-test setup. 
     * #### Removed
 * ### `nidcpower` (NI-DCPower)
     * #### Added
--- a/src/nifgen/examples/nifgen_script.py
+++ b/src/nifgen/examples/nifgen_script.py
@@ -9,9 +9,9 @@ import sys
 import time
 
 # waveform size should be a multiple of the quantum, which is 4, 2 or 1, for all devices
-# minimum waveform size varies with sample rate and device.
-# 512 should be enough for any device and sample rate, except certain 5450, 5451 configurations.
-NUMBER_OF_SAMPLES = 512
+# minimum waveform size needed to prevent underflow varies with sample rate and device.
+# If underflow occurs, increase this value.
+NUMBER_OF_SAMPLES = 2096
 
 
 # waveforms finish just short of 360 degrees, so that we don't repeat the first point
--- a/src/nifgen/examples/nifgen_script.py
+++ b/src/nifgen/examples/nifgen_script.py
@@ -2,89 +2,42 @@
 
 import argparse
 import nifgen
-import numpy as np
 
 import math
 import sys
 import time
 
+# waveform size should be a multiple of the quantum, which is 4 or 8, for some devices
+# some older devices also have a 256 sample minimum waveform size
 number_of_points = 256
 
 
-def equivalent_phase(phase):
-    if phase < 0:
-        phase += (abs(phase) // (2 * math.pi) + 1) * 2 * math.pi
-    elif phase > (2 * math.pi):
-        phase = phase % (2 * math.pi)
-    return phase  # phase in range [0, 2 * PI]
-
-
-def duty_cycle_is_on(phase, duty_cycle):
-    # duty_cycle is a decimal in the range [0, 1]
-    phase = equivalent_phase(phase)
-    if duty_cycle == 0:
-        return False
-    elif phase == 2 * math.pi:
-        return True
-    cycle_percentage = phase / (2 * math.pi)
-    return cycle_percentage <= duty_cycle
-
-
-def square_wave(t, duty_cycle):
-    wfm = []
-    for phase in t:  # treat time as phase
-        if duty_cycle_is_on(phase, duty_cycle):
-            wfm.append(1.0)
-        else:
-            wfm.append(-1.0)
-    return wfm
-
-
-def sawtooth_wave(t):
-    wfm = []
-    for phase in t:  # treat time as phase
-        phase = equivalent_phase(phase)
-        cycle_percentage = phase / (2 * math.pi)
-        wfm.append(-1.0 + (2 * cycle_percentage))
-    return wfm
-
-
 def calculate_sinewave():
-    time = np.linspace(start=0, stop=10, num=number_of_points)    # np.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)
-    amplitude = np.sin(time)
-    sinewave = amplitude.tolist()                               # List of Float
-    return sinewave
+    # waveforms finish just short of 360 degrees, so that we don't repeat the first point
+    # if we repeat the waveform
+    return [math.sin(math.pi * 2 * x / (number_of_points) for x in range(number_of_points)]
 
 
 def calculate_rampup():
-    ramp = np.linspace(start=0, stop=0.5, num=number_of_points)   # np.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)
-    ramp_up = ramp.tolist()                                     # List of Float
-    return ramp_up
+    return [x / (number_of_points) for x in range(number_of_points)]
 
 
 def calculate_rampdown():
-    ramp = np.linspace(start=0, stop=0.5, num=number_of_points)   # np.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)
-    ramp_down = ramp.tolist()                                   # List of Float
-    ramp_down.reverse()                                         # Reverse list to get a ramp down
-    return ramp_down
+    return [-1.0 * x for x in calculate_rampup()]
 
 
 def calculate_square():
-    time = np.linspace(start=0, stop=10, num=number_of_points)    # np.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)
-    square = square_wave(t=time, duty_cycle=0.5)
-    return square
+    return [1.0 if x < (number_of_points / 2) else -1.0 for x in range(number_of_points)]
 
 
-def calculate_triangle():
-    time = np.linspace(start=0, stop=1, num=number_of_points)     # np.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)
-    triangle = sawtooth_wave(t=time)
-    return triangle
+def calculate_sawtooth():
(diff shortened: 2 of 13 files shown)


src/nifgen/examples/nifgen_script.py
```diff
@@ -32,15 +71,13 @@ def calculate_rampdown():
 
 def calculate_square():
     time = np.linspace(start=0, stop=10, num=number_of_points)    # np.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)
-    square_build = signal.square(t=time, duty=0.5)              # signal.square(t, duty=0.5)
-    square = square_build.tolist()                              # List of Float
+    square = square_wave(t=time, duty_cycle=0.5)
     return square
 
 
 def calculate_triangle():
```
ni-jfitzger, line 78: A triangle wave and sawtooth wave are not really the same thing, but I'm going to let it go because I don't want to change example behavior as part of this PR.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ All notable changes to this project will be documented in this file.
 * ### ALL
     * #### Added
     * #### Changed
+        * Fix [#1811](https://github.com/ni/nimi-python/issues/1811): nimi-python uses scipy which keeps us from adding support to Python 3.10
```
marcoskirsch, line 36: Examples are referenced in released documentation but aren't themselves part of the release, so I think it's ok to leave out of the CHANGELOG.md

You are welcome to leave it in though.

src/nifgen/examples/nifgen_script.py
```diff
@@ -3,13 +3,52 @@
...
-from scipy import signal
+
+import math
 import sys
 import time
 
 number_of_points = 256
 
 
+def equivalent_phase(phase):
```
marcoskirsch, line 14: Lines 14 - 87 is dedicated to creating waveforms. It uses numpy which is not a requirement for using `nifgen`.
That's excessive, the point of the examples is not to teach customers how to create waveforms but how to use `nifgen`.

The following code snippet creates all the waveforms using only built in modules `math` and `random`, each in one line. I _think_ they are correct :) at least they looked correct in Excel.
One difference is that the noise is no longer gaussian distribution, just uniform distribution. That's fine just need to rename things in the example.

```
import math
import random

number_of_samples = 50;

sine = [math.sin(math.pi * 2 * x / number_of_samples) for x in range(number_of_samples)]
square = [1.0 if x < number_of_samples/2 else -1.0 for x in range(number_of_samples)]
ramp_up = [2.0 * x / number_of_samples - 1.0 for x in range(number_of_samples)]
ramp_down = [-1.0 * x for x in ramp_up]
triangle = ramp_up[::2] + ramp_down[::2]
sawtooth = [(x / number_of_samples * 2) if x < number_of_samples/2 else (x / number_of_samples * 2) - 1 for x in range(number_of_samples)]
noise = [random.uniform(-1.0, 1.0) for x in range(number_of_samples)]
```

I much prefer the above as it's simpler and less verbose - not detracting from the point of the example.

marcoskirsch (changes requested):

CHANGELOG.md
ni-jfitzger: Reverted.

src/nifgen/examples/nifgen_script.py
ni-jfitzger: It seems a shame to throw away such a nice square function, but you have a point. I can certainly see how all this other code distracts from the actual example, especially with me adding to it. A single cycle (as opposed to an arbitrary number of cycles) for each waveform is probably also better for scripting.

A few things about the code you posted: 
* Ramps should start at 0
* Your triangle wave has more samples than the other wfms (which is okay) and has a -90 degree phase offset for no apparent reason
* The script doesn't actually call for both a triangle and sawtooth, but I don't mind adding it

ni-jfitzger: I have taken your feedback and removed the use of both scipy and numpy from the example.
I updated the wfm functions to be similar to what you have above, though not exactly the same in every case. I'm not going to add an actual triangle wfm. I'm just renaming the triangle wfm to sawtooth. As you more or less stated, it's not important to show users how to create every basic signal and I'd prefer all of our signals have 0 offset , 1 cycle (excluding the point at 2pi in case it's repeated) and be relatively simple.

I'm also sticking with 256 points, though I added a comment about it.

src/nifgen/examples/nifgen_script.py
```diff
@@ -2,52 +2,43 @@
...
-    return square
+    return [1.0 if x < (number_of_points / 2) else -1.0 for x in range(number_of_points)]
 
 
-def calculate_triangle():
-    time = np.linspace(start=0, stop=1, num=number_of_points)     # np.linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None)
-    triangle_build = signal.sawtooth(t=time)                    # signal.sawtooth(t, width=1)
-    triangle = triangle_build.tolist()                          # List of Float
-    return triangle
+def calculate_sawtooth():
```
marcoskirsch, line 34: It's weird / surprising that this function returns something other than `number_of_points`.
Surprising is bad.

src/nifgen/examples/nifgen_script.py
ni-jfitzger: What size do you recommend?

src/nifgen/examples/nifgen_script.py
```diff
@@ -2,52 +2,43 @@
...
-from scipy import signal
+
+import math
+import random
 import sys
 import time
 
+# waveform size should be a multiple of the quantum, which is 4 or 8, for some devices
+# some older devices also have a 256 sample minimum waveform size
 number_of_points = 256
```
marcoskirsch, line 13: the proper term is "samples" so let's rename to `number_of_samples`

src/nifgen/examples/nifgen_script.py
```diff
@@ -2,52 +2,43 @@
...
 
 
 def calculate_sinewave():
```
marcoskirsch, line 16: TBH I think having functions for creating the waveforms is overkill.
A well named variable to hold it is self-documenting enough.

marcoskirsch (changes requested):

src/nifgen/examples/nifgen_script.py
ni-jfitzger: replaced the functions with CONSTANTS.

src/nifgen/examples/nifgen_script.py
ni-jfitzger: Replaced the functions with CONSTANTS.

src/nifgen/examples/nifgen_script.py
ni-jfitzger: I checked the specs for all currently supported devices. 512 samples should be enough to avoid underflow for anything except certain 5450/51 configurations. The 5450/51 is a major outlier with minimum waveform sizes above 2300 samples for some configurations.

It's also a major outlier in usage, since it's primarily an RF device, so I'm setting 512 as the waveform size. Let me know if you want me to increase it.

marcoskirsch: The number 2096 popped in my mind.

src/nifgen/examples/nifgen_script.py
```diff
@@ -2,52 +2,26 @@
...
-    random_noise = np.random.normal(loc=0, scale=0.1, size=number_of_points)  # random.normal(loc=0.0, scale=1.0, size=None)
-    noise = random_noise.tolist()                                           # List of Float
-    return noise
+# waveforms finish just short of 360 degrees, so that we don't repeat the first point
+# if we repeat the waveform
+SINE_WAVE = [math.sin(math.pi * 2 * x / (NUMBER_OF_SAMPLES)) for x in range(NUMBER_OF_SAMPLES)]
+RAMP_UP = [x / (NUMBER_OF_SAMPLES) for x in range(NUMBER_OF_SAMPLES)]
+RAMP_DOWN = [-1.0 * x for x in RAMP_UP]
+SQUARE_WAVE = [1.0 if x < (NUMBER_OF_SAMPLES / 2) else -1.0 for x in range(NUMBER_OF_SAMPLES)]
+SAWTOOTH_WAVE = RAMP_UP + [(-1 + x) for x in RAMP_UP]
```
marcoskirsch, line 23: this constant does not have NUMBER_OF_SAMPLES samples.

```suggestion
SAWTOOTH_WAVE = RAMP_UP[::2] + [(-1 + x) for x in RAMP_UP][::2]
```

src/nifgen/examples/nifgen_script.py
ni-jfitzger: Yeah, I can set it to that. The 5450/51 has a burst trigger mode that has an even higher minimum waveform size, but I don't believe we're using that. https://www.ni.com/docs/en-US/bundle/pxie-5450-specs/page/specs.html

Really, it's hard for me to interpret, because I don't see anything for Script Trigger. Maybe script trigger would count as a single trigger? In which case the minimum size would be 2.

My one concern about increasing the size that much is that it might make the waveform noisy, when it's actually generated, due to a low delta.

src/nifgen/examples/nifgen_script.py
ni-jfitzger: Ah that's what you were doing with the slicing. Yeah, makes sense.

src/nifgen/examples/nifgen_script.py
ni-jfitzger: Increased to 2096 and marking this resolved.

marcoskirsch: > My one concern about increasing the size that much is that it might make the waveform noisy, when it's actually generated, due to a low delta.

I don't follow.

marcoskirsch (approved):

src/nifgen/examples/nifgen_script.py
ni-jfitzger: The signal is limited to the -1 to 1 range and as we increase the number of samples, the sine, sawtooth and ramps will decrease phase change between points, which decreases the amplitude change. So I worry about hardware precision limits causing this to make the signals appear noisy.

Perhaps it's an unfounded concern. I'm not an expert on how FGENs work.

PR #1819, opened 2022-09-06
Move ctypes code to new LibraryInterpreter layer

DavidCurtiss:

Moves ctypes-specific code out of the Session layer to a new LibraryInterpreter layer, which sits between the Session and the Library.

[AB#2139776](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2139776)

py39-test


src/nidcpower/metadata/functions.py
```diff
@@ -2009,7 +2009,7 @@
                 'name': 'measurements',
                 'size': {
                     'mechanism': 'python-code',
-                    'value': 'self._parse_channel_count()'
+                    'value': 'self._parse_channel_count(session, channel_name)'
```
DavidCurtiss, line 2012: FYI, I'll copy the metadata changes to the original sources after this PR is accepted.

ni-jfitzger:
Disregarding the generated code, this is a very large change.
Is there some kind of enumeration of changes that you could do to ease the review burden?
Is there some kind of search that can easily detect where changes were needed / confirm that you haven't missed something?

DavidCurtiss:
> Is there some kind of enumeration of changes that you could do to ease the review burden?

Yeah, sorry about that, but it's mostly an all-or-nothing sort of deal. Once you start moving some of the ctypes stuff, it sort of snowballs. I'd be happy to have a chat & screen share to talk through how it all interconnects, if that'd be helpful.

> Is there some kind of search that can easily detect where changes were needed / confirm that you haven't missed something?

I'm mostly relying on the tests to catch anything I may have missed.

marcoskirsch:
In a video call, @ni-jfitzger @DavidCurtiss and I decided to leave the Library class as a direct, exact mapping to the C DLL; and move IVI dance, error handling, ctypes to native type conversion to a new class that sits between Session and Library.

build/generate_template.py
```diff
@@ -15,7 +15,7 @@ def generate_template(template_name, template_params, dest_file, in_zip_file=Fal
     try:
         template_params['encoding_tag'] = '# -*- coding: utf-8 -*-'
         module_name = template_params['metadata'].config['module_name']
-        lookup = TemplateLookup(directories=['build/templates', 'src/{0}/templates'.format(module_name)])
+        lookup = TemplateLookup(directories=['src/{0}/templates'.format(module_name), 'build/templates'])
```
dmondrik, line 18: This seems to be that `templates` would appear twice in the string. Intentional?

generated/nidcpower/nidcpower/_library_interpreter.py
```diff
@@ -0,0 +1,701 @@
...
+
+    def export_attribute_configuration_buffer(self):  # noqa: N802
+        vi_ctype = _visatype.ViSession(self._vi)  # case S110
+        size_ctype = _visatype.ViInt32()  # case S170
+        configuration_ctype = None  # case B580
+        error_code = self._library.niDCPower_ExportAttributeConfigurationBuffer(vi_ctype, size_ctype, configuration_ctype)
+        errors.handle_error(self, error_code, ignore_warnings=True, is_error_handling=False)
+        size_ctype = _visatype.ViInt32(error_code)  # case S180
+        configuration_size = size_ctype.value  # case B590
+        configuration_array = array.array("b", [0] * configuration_size)  # case B590
```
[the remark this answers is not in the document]
ni-jfitzger: The [0] is a list with a value of 0. It's actually the '*' operator that has a special meaning here.
Example:
```
print([0] * 3)
# [0, 0, 0]
```

https://note.nkmk.me/en/python-list-initialize/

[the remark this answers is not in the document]
dmondrik: Oh. Just to make it confusing with the index operator such as I was reading it: `array[0] * size`.

build/generate_template.py
DavidCurtiss: I think you misread the code. It's building up a list of two elements wherein the first one will look like 'src/nifake/templates' and the second one will be 'build/templates'.

build/helper/codegen_helper.py
```diff
@@ -117,7 +133,7 @@ def _get_output_param_return_snippet(output_parameter, parameters, config):
         snippet = '[' + return_type_snippet + output_parameter['ctypes_variable_name'] + '[i]) for i in range(' + size + ')]'
     else:
         if output_parameter['is_string']:
-            # 'self._encoding' is a variable on the session object
+            # '_encoding' is a variable on the LibraryInterpreter object
```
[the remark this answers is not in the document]
DavidCurtiss: Yes, this helper is used across all the files. The .mako file that calls this particular function used to be the Session object, but now it's the LibraryInterpreter object.

build/helper/codegen_helper.py
```diff
@@ -877,7 +909,7 @@ def get_enum_value_snippet(value):
         'is_string': True,
         'is_repeated_capability': True,
         'is_session_handle': False,
-        'library_method_call_snippet': 'channel_list_ctype',
+        'library_method_call_snippet': 'self._repeated_capability',
```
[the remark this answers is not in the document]
DavidCurtiss: There are certain "repeated capability" parameters that are special. The `channel_list` is one of them. Rather than being passed along as a function argument, they're stored in `self._repeated_capability`, so that's where we need to pull them from when calling down from the Session to the next layer.

build/generate_template.py
dmondrik: You're right. It appeared to me that `build/templates` was an argument to the format method. Sigh.

build/templates/_library_interpreter.py/_get_error_description.py.mako
```diff
@@ -0,0 +1,21 @@
+    def get_error_description(self, error_code):
+        '''get_error_description
+
+        Returns the error description.
+        '''
+        try:
+            _, error_string = self.get_error()
```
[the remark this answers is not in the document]
DavidCurtiss: The comment at that link says "Since another thread has changed the status, fall back to the static message lookup." That's not a concern here, since python will use the same thread to call back down to the (local) driver.

build/templates/_library_interpreter.py/default_method.py.mako
```diff
@@ -0,0 +1,46 @@
...
+    import build.helper as helper
+
+    parameters = f['parameters']
+    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_METHOD_DECLARATION)
+    param_names_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.CTYPES_CALL)
+
+    ivi_dance_parameters = helper.filter_ivi_dance_parameters(f)
+    ivi_dance_size_parameter = helper.find_size_parameter(ivi_dance_parameters, parameters)
+    len_parameters = helper.filter_len_parameters(f)
+    len_size_parameter = helper.find_size_parameter(len_parameters, parameters)
```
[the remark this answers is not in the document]
DavidCurtiss: That's right, the `len_size_parameter` variable is only used in the `assert` that's on the subsequent line.

docs/niscope/class.rst
```diff
@@ -1677,7 +1677,7 @@ fetch_into
 
                 .. code-block:: python
 
-                    waveform = numpy.ndarray(num_samples * session.actual_num_wfms(), dtype=numpy.float64)
+                    waveform = numpy.ndarray(num_samples * session.actual_num_wfms(channel_list), dtype=numpy.float64)
```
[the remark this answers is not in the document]
DavidCurtiss: My mistake. I did a global search/replace, but shouldn't have updated the docstring (which this .rst is sourced from). The `actual_num_wfms` call on the `session` object does not take a `channel_list`. Fixed.

src/nidigital/templates/_library_interpreter.py/fancy_fetch_capture_waveform.py.mako
```diff
@@ -0,0 +1,25 @@
+
+    def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):
+        # This is slightly modified codegen from the function
+        # We cannot use codegen without major modifications to the code generator
+        # This function uses two 'ivi-dance' parameters and then multiplies them together - see
+        # the (modified) line below
+        # Also, we want to return the two sized that normally wouldn't be returned
```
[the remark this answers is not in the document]
DavidCurtiss: Looks like a typo. I moved this function from another location, but I think it's supposed to be "sizes". Updated.

build/templates/_library_interpreter.py/numpy_write_method.py.mako
```diff
@@ -5,8 +5,8 @@
     import build.helper as helper
 
     parameters = f['parameters']
-    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_METHOD_DECLARATION)
-    param_names_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.CTYPES_CALL)
+    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_DECLARATION)
```
dmondrik, line 8: In the `numpy_read` script above this was `LIBRARY_INTERPRETER_NUMPY_INTO_METHOD_DECLARATION`

dmondrik (approved):

build/templates/_library_interpreter.py/numpy_write_method.py.mako
DavidCurtiss: Yep. The [existing version](https://github.com/ni/nimi-python/blob/master/build/templates/session.py/numpy_write_method.py.mako) of the numpy_write file also uses non-NUMPY declaration/call parameters.

build/helper/codegen_helper.py
```diff
@@ -37,21 +45,29 @@
...
     },
-    ParameterUsageOptions.CTYPES_CALL: {
-        'skip_self': True,
+    ParameterUsageOptions.LIBRARY_METHOD_DECLARATION: {
+        'skip_self': False,
         'name_to_use': 'python_name',
     },
     ParameterUsageOptions.LIBRARY_METHOD_CALL: {
+        'skip_self': True,
+        'name_to_use': 'ctypes_method_call_snippet',
```
ni-jfitzger, line 54: Although, I've looked at the code this is used to generate (a bunch of ctypes parameters), I feel like this key name doesn't fit in with the ENUM's description:
`'''For calling into a Library method (from LibraryInterpreter).'''`.

DavidCurtiss: I think I prefer the accuracy of 'ctypes_method_call_snippet' here.

However, after you brought it up, now I'm wondering if maybe something like LazyCDLL would be a more appropriate name than Library, since it's really just a CDLL wrapper that lazy-loads the functions. Doing so would also have the side effect of making 'ctypes_method_call_snippet' make more sense.

build/templates/_library_interpreter.py/default_method.py.mako
```diff
@@ -0,0 +1,46 @@
...
+
+    parameters = f['parameters']
+    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_DECLARATION)
+    param_names_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_METHOD_CALL)
+
+    ivi_dance_parameters = helper.filter_ivi_dance_parameters(f)
+    ivi_dance_size_parameter = helper.find_size_parameter(ivi_dance_parameters, parameters)
+    len_parameters = helper.filter_len_parameters(f)
+    len_size_parameter = helper.find_size_parameter(len_parameters, parameters)
+    assert ivi_dance_size_parameter is None or len_size_parameter is None, str(f)
```
ni-jfitzger, line 15: This `, str(f)` wasn't there in the old session.py/default_method.py.mako code and `find_size_parameter` doesn't appear to have changed. Why was this added?

src/nidcpower/metadata/functions.py
```diff
@@ -2949,6 +2956,7 @@
         'method_templates': [
             {
                 'documentation_filename': 'default_method',
+                'library_interpreter_filename': 'none',
```
ni-jfitzger, line 2959: I must be missing something. Shouldn't this prevent us from adding a method to the LibraryInterpreter?
But it looks like we added one: https://github.com/ni/nimi-python/blob/7f9d9993ef60f7315f36ee08385b4b2e35114ef6/generated/nidcpower/nidcpower/_library_interpreter.py#L380

build/templates/session.py/lock.py.mako
```diff
@@ -44,8 +39,4 @@
 
         Actual call to driver
```
ni-jfitzger, line 40: @marcoskirsch does this comment need to change?

build/templates/_library_interpreter.py.mako
```diff
@@ -0,0 +1,85 @@
...
+def get_ctypes_pointer_for_buffer(value=None, library_type=None, size=None):
+    if isinstance(value, array.array):
+        assert library_type is not None, 'library_type is required for array.array'
+        addr, _ = value.buffer_info()
+        return ctypes.cast(addr, ctypes.POINTER(library_type))
+    elif str(type(value)).find("'numpy.ndarray'") != -1:
+        import numpy
+        return numpy.ctypeslib.as_ctypes(value)
+    elif isinstance(value, bytes):
+        return ctypes.cast(value, ctypes.POINTER(library_type))
```
ni-jfitzger, line 43: I'm sure we already discussed this, but this case is new. Is it a bug fix?

src/nidcpower/metadata/functions.py
DavidCurtiss: Different function, though named almost the same:
GetLastSelfCalLastDateAndTime
GetSelfCalLastDateAndTime
(The former is an invented python-only wrapper around the latter, used for datetime conversion.)

build/templates/_library_interpreter.py.mako
DavidCurtiss: It's not new; you must've overlooked it. I just verified that this function is identical to what was removed from session.py.mako.

ni-jfitzger: I see what I did. I was comparing it with the nitclk/session.py.mako.

ni-jfitzger: I guess that means it is new for nitclk.

DavidCurtiss: Agreed, it's new for tclk. I don't see anything in tclk that would be of type `bytes`, though, so I don't think it'll have any effect there. (Note that even for other drivers, there's only bytes in the places where we use `convert_to_bytes`.)

build/templates/_library_interpreter.py/default_method.py.mako
DavidCurtiss: Oops, I probably just added that for debugging and forgot to take it back out. Removed.

build/helper/codegen_helper.py
```diff
@@ -25,6 +25,14 @@
         'skip_self': False,
         'name_to_use': 'python_name_with_default',
     },
+    ParameterUsageOptions.LIBRARY_INTERPRETER_NUMPY_INTO_METHOD_DECLARATION: {
+        'skip_self': False,
+        'name_to_use': 'python_name',
+    },
+    ParameterUsageOptions.LIBRARY_INTERPRETER_NUMPY_INTO_METHOD_CALL: {
+        'skip_self': True,
+        'name_to_use': 'library_method_call_snippet',
```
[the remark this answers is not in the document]
DavidCurtiss: Renamed.

build/templates/_library_interpreter.py/numpy_write_method.py.mako
```diff
@@ -0,0 +1,44 @@
...
+    parameters = f['parameters']
+    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_DECLARATION)
+    param_names_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_METHOD_CALL)
+    param_ctypes_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.CTYPES_ARGTYPES)
+
+    ivi_dance_parameters = helper.filter_ivi_dance_parameters(f)
+    ivi_dance_size_parameter = helper.find_size_parameter(ivi_dance_parameters, parameters)
+    len_parameters = helper.filter_len_parameters(f)
+    len_size_parameter = helper.find_size_parameter(len_parameters, parameters)
+    assert ivi_dance_size_parameter is None or len_size_parameter is None, str(f)
```
ni-jfitzger, line 16: I'm guessing this was for debugging?

ni-jfitzger: (The str(f))

build/templates/_library_interpreter.py/numpy_read_method.py.mako
```diff
@@ -0,0 +1,50 @@
...
+    parameters = f['parameters']
+    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_NUMPY_INTO_METHOD_DECLARATION)
+    param_names_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_METHOD_CALL)
+    param_ctypes_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.CTYPES_ARGTYPES)
+
+    ivi_dance_parameters = helper.filter_ivi_dance_parameters(f)
+    ivi_dance_size_parameter = helper.find_size_parameter(ivi_dance_parameters, parameters)
+    len_parameters = helper.filter_len_parameters(f)
+    len_size_parameter = helper.find_size_parameter(len_parameters, parameters)
+    assert ivi_dance_size_parameter is None or len_size_parameter is None, str(f)
```
ni-jfitzger, line 16: I'm guessing this `, str(f)` was also for debugging?

generated/nitclk/nitclk/session.py
```diff
@@ -303,18 +246,7 @@ def _get_attribute_vi_string(self, attribute_id):
...
-        error_code = self._library.niTClk_GetAttributeViString(session_ctype, channel_name_ctype, attribute_id_ctype, buf_size_ctype, value_ctype)
-        errors.handle_error(self, error_code, ignore_warnings=True, is_error_handling=False)
-        buf_size_ctype = _visatype.ViInt32(error_code)  # case S180
-        value_ctype = (_visatype.ViChar * buf_size_ctype.value)()  # case C060
-        error_code = self._library.niTClk_GetAttributeViString(session_ctype, channel_name_ctype, attribute_id_ctype, buf_size_ctype, value_ctype)
-        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
-        return value_ctype.value.decode(self._encoding)
+        return self._library_interpreter.get_attribute_vi_string(self._repeated_capability, attribute_id)
 
     def _get_extended_error_info(self):
```
ni-jfitzger, line 251: Should we do something to prevent this from generating ins session.py? It's not called by anything.

generated/niscope/niscope/session.py
```diff
@@ -3732,14 +3508,7 @@ def _get_equalization_filter_coefficients(self, number_of_coefficients):
...
-        channel_ctype = ctypes.create_string_buffer(self._repeated_capability.encode(self._encoding))  # case C010
-        number_of_coefficients_ctype = _visatype.ViInt32(number_of_coefficients)  # case S210
-        coefficients_size = number_of_coefficients  # case B600
-        coefficients_ctype = get_ctypes_pointer_for_buffer(library_type=_visatype.ViReal64, size=coefficients_size)  # case B600
-        error_code = self._library.niScope_GetEqualizationFilterCoefficients(vi_ctype, channel_ctype, number_of_coefficients_ctype, coefficients_ctype)
-        errors.handle_error(self, error_code, ignore_warnings=False, is_error_handling=False)
-        return [float(coefficients_ctype[i]) for i in range(number_of_coefficients_ctype.value)]
+        return self._library_interpreter.get_equalization_filter_coefficients(self._repeated_capability, number_of_coefficients)
 
     def _get_error(self):
```
ni-jfitzger, line 3513: We don't seem to be calling this from anywhere. I believe this applies to the other drivers, as well.

build/templates/_library_interpreter.py/_get_error_description.py.mako
```diff
@@ -0,0 +1,21 @@
...
+            return error_string
+        except errors.Error:
+            pass
+
+        try:
+            '''
+            It is expected for _get_error to raise when the session is invalid
+            (IVI spec requires GetError to fail).
+            Use _error_message instead. It doesn't require a session.
```
ni-jfitzger, line 16: I think we should remove the leading underscore from these names, with the move to library_interpreter.

src/nitclk/templates/_library_interpreter.py/_get_error_description.py.mako
```diff
@@ -0,0 +1,15 @@
+    def get_error_description(self, error_code):
+        '''get_error_description
+
+        Returns the error description.
+        '''
+        try:
+            '''
+            It is expected for _get_error to raise when the session is invalid
+            (IVI spec requires GetError to fail).
+            Use _error_message instead. It doesn't require a session.
```
ni-jfitzger, line 10: It seems we didn't update this comment when we copied this code from the driver templates.
I think we can just delete the comment.

build/templates/_library_interpreter.py/numpy_write_method.py.mako
DavidCurtiss: Yep. But even worse, all this ivi_dance stuff is left over from a copy/paste from when I was doing all the numpy stuff without the use of fancy methods at the LibraryInterpreter layer. The numpy functions don't use ivi_dance. So I've removed all the ivi_dance code from them.

build/templates/_library_interpreter.py/numpy_read_method.py.mako
DavidCurtiss: Same here. Removed the ivi_dance code from the numpy functions.

build/templates/_library_interpreter.py/_get_error_description.py.mako
DavidCurtiss: Done

src/nitclk/templates/_library_interpreter.py/_get_error_description.py.mako
DavidCurtiss: Done

src/niscope/metadata/functions_addon.py
```diff
@@ -119,7 +120,7 @@
                 'name': 'wfmInfo',
                 'size': {
                     'mechanism': 'python-code',
-                    'value': 'self._actual_num_wfms()'
+                    'value': 'self._actual_num_wfms(session, channel_list)'
                 },
```
ni-jfitzger, line 124: Are we even using this?

src/nidigital/unit_tests/test_nidigital.py
```diff
@@ -1,20 +1,41 @@
...
-        self.patched_library_patcher = patch('nidigital._library.Library', autospec=True)
-        self.patched_library = self.patched_library_patcher.start()
-        self.patched_library_singleton_get = patch('nidigital.session._library_singleton.get', return_value=self.patched_library)
+        self.patched_library = self.PatchedLibrary(None)
+        self.patched_library_singleton_get = patch('nidigital._library_interpreter._library_singleton.get', return_value=self.patched_library)
         self.patched_library_singleton_get.start()
 
+        self.patched_tclk_library = self.PatchedTClkLibrary(None)
+        self.patched_tclk_library_singleton_get = patch('nitclk._library_interpreter._library_singleton.get', return_value=self.patched_library)
+        self.patched_tclk_library_singleton_get.start()
```
ni-jfitzger, line 37: I love that you're fixing this for us.

src/nidigital/unit_tests/test_nidigital.py
```diff
@@ -1,20 +1,41 @@
...
     def setup_method(self, method):
-        self.patched_library_patcher = patch('nidigital._library.Library', autospec=True)
-        self.patched_library = self.patched_library_patcher.start()
-        self.patched_library_singleton_get = patch('nidigital.session._library_singleton.get', return_value=self.patched_library)
+        self.patched_library = self.PatchedLibrary(None)
+        self.patched_library_singleton_get = patch('nidigital._library_interpreter._library_singleton.get', return_value=self.patched_library)
         self.patched_library_singleton_get.start()
 
+        self.patched_tclk_library = self.PatchedTClkLibrary(None)
+        self.patched_tclk_library_singleton_get = patch('nitclk._library_interpreter._library_singleton.get', return_value=self.patched_library)
```
ni-jfitzger, line 36: Shouldn't this be `return_value=self.patched_tclk_library`?

src/nifake/unit_tests/test_session.py
```diff
@@ -169,19 +173,19 @@ def test_session_context_manager_close_with_error(self):
     def test_lock_session_none(self):
         with nifake.Session('dev1') as session:
             session.lock()
-            self.patched_library.niFake_LockSession.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), None)
+            self.patched_library.niFake_LockSession.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), _matchers.ViBooleanPointerMatcher())
```
ni-jfitzger, line 176: Match any Boolean because caller_has_lock is a Boolean, right?

src/niscope/metadata/functions_addon.py
DavidCurtiss: No. I suspect it may have been used long ago, and then this was turned into a fancy function, and this size mechanism is unused. Should I remove it? (And have _no_ size mechanism??) Or fix it (removing the `session` that I added)? Or something else?

src/nifake/unit_tests/test_session.py
DavidCurtiss: No, it's any Boolean _pointer_, because the now-generic `lock` function in LibraryInterpreter returns the boolean, but it's ignored at the Session layer.

(When `lock` was implement purely as a session.py fancy function, we passed None for the boolean pointer, because that output is ignored.)

src/nidigital/unit_tests/test_nidigital.py
DavidCurtiss: Yes. Fixed.

But note that it worked anyway because all we really need to do is patch the library_singleton.get() function to return _anything_. Specifically, what we need to do is stop the get() function from trying to construct a CDLL for nitclk (which it creates to give to the Library constructor).

Side question: why does nidigital's Session object even have a self.tclk member? It never uses it (and didn't before my PR, either). Is it just a vestige of some old code, and now we just keep it around to avoid breaking backwards compatibility by removing a public member?

generated/nitclk/nitclk/session.py
DavidCurtiss: Stopped it from generating in session.py.

generated/niscope/niscope/session.py
DavidCurtiss: Stopped it from generating in session.py.

src/niscope/metadata/functions_addon.py
ni-jfitzger: I would vote for removing it, since we don't need this info for a function that only exists in Session.py. I'll defer to @marcoskirsch, though.

src/nidigital/unit_tests/test_nidigital.py
ni-jfitzger: All driver apis that support nitclk have a self.tclk member.
Here are the examples for nitclk: https://nimi-python.readthedocs.io/en/master/nitclk/examples.html

That self.tclk member is a `nitclk.SessionReference` object.
When we call an nitclk method, inside library_interpreter, we call `_converters.convert_to_nitclk_session_number_list(sessions)`. That calls `_get_tclk_session_reference` on the self.tclk obejct inside the driver API's Session object.

build/helper/codegen_helper.py
```diff
@@ -37,21 +45,29 @@
...
     },
     ParameterUsageOptions.CTYPES_ARGTYPES: {
         'skip_self': True,
         'name_to_use': 'ctypes_type_library_call',
     },
-    ParameterUsageOptions.LIBRARY_METHOD_DECLARATION: {
+    ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_DECLARATION: {
         'skip_self': False,
-        'name_to_use': 'python_name',
+        'name_to_use': 'python_name_with_default',
```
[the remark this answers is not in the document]
DavidCurtiss: At least one of the fancy functions calls into a LibraryInterpreter method and assumes the defaults.

src/niscope/metadata/functions_addon.py
DavidCurtiss: Okay. For now I've fixed it to remove the "session" argument, but we'll see what @marcoskirsch says.

src/nidigital/unit_tests/test_nidigital.py
ni-jfitzger: It's not going to be a problem that the nitclk LibraryInterpreter calls into the driver API Session object and the nitclk SessionReference object, is it?

build/templates/session.py/lock.py.mako
marcoskirsch: I'd get rid of the whole method

src/niscope/metadata/functions_addon.py
marcoskirsch: I'm all for removing dead code.

build/templates/session.py/lock.py.mako
ni-jfitzger: It's still being called by the method above it, so we can't do that without removing both methods.
This sounds related to your point about some of our lock code no longer being necessary. I'd rather a change like that be handled separately.

@marcoskirsch can you create an issue, since you understand the details?

build/helper/codegen_helper.py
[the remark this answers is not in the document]
marcoskirsch: That's surprising to me as well.
Defaults are usually for client convenience at the expense of readability / explicitness, and we probably don't need that internally.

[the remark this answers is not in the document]
DavidCurtiss: It looks like my memory may have been wrong here. I'm trying out reverting this, and so far it's working out fine. Doing a full local test run, and then I'll update this.

src/nidcpower/metadata/functions.py
ni-jfitzger: Are you planning to regenerate it and copy the output into this change?
(I'm not saying that I'm advocating that, I'm just asking).

build/helper/codegen_helper.py
ni-jfitzger: I don't have any strong opinions on the name of the Library class. I'm fine if we just leave this as is. If we do rename the Library class, that may trigger quite a few renames, but I'm prepared to review that, if necessary.

src/niscope/metadata/functions_addon.py
DavidCurtiss: Removed.

build/helper/codegen_helper.py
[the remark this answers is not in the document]
DavidCurtiss: Set it to python_name, and stopped reordering the parameters, and the build/test worked fine. So I've pushed an update with that change.

src/nidigital/unit_tests/test_nidigital.py
DavidCurtiss: I can't think of anything that would cause problems for that, no.

src/niscope/metadata/functions_addon.py
ni-jfitzger: Shouldn't we delete the whole parameter?

ni-jfitzger: Scratch that. It's probably part of the documentation.

build/helper/codegen_helper.py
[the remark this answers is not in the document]
DavidCurtiss: Ah, [there it is](https://github.com/ni/nimi-python/actions/runs/3084281989/jobs/4986250381). It's not found by the unit tests, only by the system tests.
```
    def get_error_description(self, error_code):
        '''get_error_description
    
        Returns the error description.
        '''
        try:
>           _, error_string = self.get_error()
E           TypeError: get_error() missing 1 required positional argument: 'error_description_size'
```
Should I update the get_error_description methods, or should I put the defaults back?

build/templates/session.py/lock.py.mako
marcoskirsch: I meant the caller could call LibraryInterpreter directly.
But yeah, there's some dead code there that only applied to a specific combination of NI-DCPower runtime, nimi-python, and using independent channels.

build/helper/codegen_helper.py
marcoskirsch: Improving naming and increasing consistency in how we name things is totally worth the ROI in my opinion. It makes it easier to reason about the code and onboard new developers.

build/helper/codegen_helper.py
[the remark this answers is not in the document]
ni-jfitzger: The alternative would be to look up the defaults for get_error() in get_error_description.py.mako and add it to the get_error() call. If you think that can be done without too much difficulty, it's probably the optimal solution.

build/helper/codegen_helper.py
DavidCurtiss: Would you prefer the LazyCDLL name, then?

build/templates/session.py/lock.py.mako
DavidCurtiss: Removed `_lock_session`

build/helper/codegen_helper.py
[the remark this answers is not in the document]
DavidCurtiss: Alright, kept it as 'python_name', and generated the params_snippet for the get_error() call. Also, I stopped generating the call to `error_message` when that function doesn't exist (which is also an nise-only difference, just like the get_error params).

build/helper/codegen_helper.py
marcoskirsch: I'd rather rename `ctypes_method_call_snippet`.
Why? We have a lot of internal driver code that generates equivalent classes in C++ called `Library`. We are likely to do something similar for C#. It's nice to have that consistency.

build/helper/codegen_helper.py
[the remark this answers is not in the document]
ni-jfitzger: Nice work. That error_message bit seems to be an oversight in the original code. Looks like nise was the only API that needed a default.
Shame on nise for not conforming.

src/nidcpower/metadata/functions.py
ni-jfitzger: The internal metadata for NI-Digital is currently stale.
I implemented a change earlier this year to fix it, but had to back it out because it broke niDigitalCXP behavior.

build/helper/codegen_helper.py
DavidCurtiss: Somewhere we have to make the connection that "when you call into Library, you build and pass ctypes parameters." If we don't do it here, then that connection will then instead be implied in metadata_add_all.py, where we'd fill in the library_method_call_snippet (renamed from ctypes_method_call_snippet) with ctypes-specific info.

Personally, I think that right here in codegen_helper.py's dictionary mapping is the ideal place to connect the two concepts, and let metadata_add_all deal with the lower-level concept of "you want to pass ctypes? here's how."

ni-jfitzger: I could go either way on this.
I think if we keep `'name_to_use': 'ctypes_method_call_snippet',`, though, you should add a comment `# when you call into Library, you build and pass ctypes parameters`.

As it is, it looks wrong because of how name_to_use matches up in all of the other METHOD_CALL objects.

build/templates/_library_interpreter.py/numpy_write_method.py.mako
```diff
@@ -0,0 +1,24 @@
...
+<%
+    '''Renders a LibraryInterpreter method for writing numpy.array corresponding to the passed-in function metadata.'''
+
+    import build.helper as helper
+
+    parameters = f['parameters']
+    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_DECLARATION)
+    param_names_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_METHOD_CALL)
+    param_ctypes_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.CTYPES_ARGTYPES)
+
```
ni-jfitzger, line 11: `param_ctypes_library` appears unused

build/templates/_library_interpreter.py/numpy_read_method.py.mako
```diff
@@ -0,0 +1,30 @@
...
+<%
+    '''Renders a LibraryInterpreter method for reading into a numpy.array corresponding to the passed-in function metadata.'''
+
+    import build.helper as helper
+
+    parameters = f['parameters']
+    param_names_method = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_NUMPY_INTO_METHOD_DECLARATION)
+    param_names_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_METHOD_CALL)
+    param_ctypes_library = helper.get_params_snippet(f, helper.ParameterUsageOptions.CTYPES_ARGTYPES)
+
```
ni-jfitzger, line 11: `param_ctypes_library` appears unused

build/helper/codegen_helper.py
ni-jfitzger: The main purpose of the comment would be to acknowledge that the _method_call_snippet name doesn't match and make it clear that it's not a mistake.

build/templates/_library_interpreter.py/numpy_write_method.py.mako
DavidCurtiss: Removed

build/templates/_library_interpreter.py/numpy_read_method.py.mako
DavidCurtiss: Removed

build/helper/codegen_helper.py
DavidCurtiss: Added comment

ni-jfitzger (approved):
I have reviewed all of the changes and believe this PR is in good shape.
If any new major changes are made, please re-request my review.

generated/nidcpower/nidcpower/_library_interpreter.py
```diff
@@ -0,0 +1,707 @@
...
+class LibraryInterpreter(object):
+    '''Library C<->Python interpreter.
+
+    This class handles ctypes conversions, string encoding, and raising exceptions if the library
+    function returns an error.
+    '''
+
+    def __init__(self, encoding):
+        self._encoding = encoding
+        self._library = _library_singleton.get()
```
marcoskirsch, line 60: Consider making library a `__init__` parameter.

Will clean up unit tests a bit. But more importantly it makes sense to me to have this integration policy be dictated by higher layers of the stack.

generated/nifake/nifake/unit_tests/test_session.py
```diff
@@ -9,6 +9,7 @@
 import platform
```
marcoskirsch, line 9: I think this whole file needs to be split up into `test_session` where the LibraryInterpreter is mocked and `test_library_interpreter`.

Most of the stuff here today would end up in `test_library_interpreter`... all the ctypes handling.

You can open an issue and do it separately so this can be pushed in.

src/nidmm/metadata/functions.py
```diff
@@ -1083,6 +1085,14 @@
             'description': '\nReturns the error information associated with the\n**Instrument_Handle**. This function retrieves and then clears the\nerror information for the session. If you leave the\n**Instrument_Handle** unwired, this function retrieves and then clears\nthe error information for the process.\n'
         },
         'is_error_handling': True,
+        'method_templates': [
```
marcoskirsch, line 1088: These metadata files are supposed to be copied exactly from internal NI artifacts. It appears you edited them here in source?

What's the plan?

marcoskirsch (changes requested):

marcoskirsch:
This is some major surgery and big improvement to the overall design. Really impressive! Thanks!

build/templates/session.py.mako
```diff
@@ -297,15 +235,17 @@ class Session(_SessionBase):
         ${p['python_name']} = _converters.${p['python_api_converter_name']}(${p['python_name']})
 %   endif
 % endfor
-        self._library = _library_singleton.get()
-        self._encoding = 'windows-1251'
+        self._library_interpreter = _library_interpreter.LibraryInterpreter(encoding='windows-1251')
```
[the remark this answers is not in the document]
dmondrik: I think the interpreter should own the native session handle. In the gRPC case the equivalent interpreter will own the gRPC session handle. I don't like how the session writes directly to it, though.

build/templates/errors.py.mako
```diff
@@ -103,12 +103,12 @@ class SelfTestError(Error):
 
 
 % endif
-def handle_error(session, code, ignore_warnings, is_error_handling):
+def handle_error(library_interpreter, code, ignore_warnings, is_error_handling):
```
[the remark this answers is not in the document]
dmondrik: @marcoskirsch I don't agree, I think it belongs here. The identical logic applies once we add the gRPC interpreter.

src/nidmm/metadata/functions.py
ni-jfitzger: He has an open thread that says

>FYI, I'll copy the metadata changes to the original sources after this PR is accepted.

build/templates/session.py.mako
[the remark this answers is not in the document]
ni-jfitzger: >In the gRPC case the equivalent interpreter will own the gRPC session handle.

Why? Can we change that?

build/templates/_library_interpreter.py.mako
```diff
@@ -0,0 +1,87 @@
...
+% endif
+import ${module_name}.errors as errors
+% for c in config['custom_types']:
+
+import ${module_name}.${c['file_name']} as ${c['file_name']}  # noqa: F401
+% endfor
+
+
+# Helper functions for creating ctypes needed for calling into the driver DLL
+def get_ctypes_pointer_for_buffer(value=None, library_type=None, size=None):
```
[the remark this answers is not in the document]
DavidCurtiss: Renamed

build/templates/_library_interpreter.py.mako
```diff
@@ -0,0 +1,87 @@
...
+        assert library_type is not None, 'library_type is required for list'
+        return (library_type * len(value))(*value)
+    else:
+        if library_type is not None and size is not None:
+            return (library_type * size)()
+        else:
+            return None
+
+
+def get_ctypes_and_array(value, array_type):
```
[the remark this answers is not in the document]
DavidCurtiss: Renamed

build/templates/_library_interpreter.py.mako
```diff
@@ -0,0 +1,87 @@
...
+
+    This class handles ctypes conversions, string encoding, and raising exceptions if the library
+    function returns an error.
+    '''
+
+    def __init__(self, encoding):
+        self._encoding = encoding
+        self._library = _library_singleton.get()
+        # Initialize _${config['session_handle_parameter_name']} to 0 for now.
+        # The Session will directly update it once its init function has been called.
```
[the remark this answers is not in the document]
DavidCurtiss: Done... then changed it when moving the session handle to Session.

build/templates/session.py.mako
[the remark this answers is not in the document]
DavidCurtiss: I tried it out. I also moved `encoding` back to Session, or else we can't possibly cache the LibraryInterpreter because nitclk allows the user to set the encoding. (But more on caching in another comment...) I'm not a fan of passing the session handle (and encoding) around everywhere, but let me know what you think. I can revert this change easily enough.

build/templates/_library_interpreter.py.mako
```diff
@@ -0,0 +1,87 @@
...
+        value_array = None
+
+    return value_array
+
+
+class LibraryInterpreter(object):
+    '''Library C<->Python interpreter.
+
+    This class handles ctypes conversions, string encoding, and raising exceptions if the library
+    function returns an error.
```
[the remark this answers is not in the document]
DavidCurtiss: Done

build/templates/session.py.mako
```diff
@@ -286,9 +226,7 @@ class Session(_SessionBase):
         # Initialize the superclass with default values first, populate them later
         super(Session, self).__init__(
             repeated_capability_list=[],
-            ${config['session_handle_parameter_name']}=None,
-            library=None,
-            encoding=None,
+            library_interpreter=None,
```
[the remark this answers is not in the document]
DavidCurtiss: Just following what was done before. I might change it anyway, though. I'll leave this conversation open for now.

build/templates/session.py.mako
```diff
@@ -297,15 +235,17 @@ class Session(_SessionBase):
...
+        self._library_interpreter = _library_interpreter.LibraryInterpreter(encoding='windows-1251')
 
         # Call specified init function
-        self._${config['session_handle_parameter_name']} = 0  # This must be set before calling ${init_function['python_name']}().
-        self._${config['session_handle_parameter_name']} = self.${init_function['python_name']}(${init_call_params})
+        # Note that _library_interpreter sets _${config['session_handle_parameter_name']} to 0 in its constructor, so that if
+        # ${init_function['python_name']} fails, the error handler can reference it.
+        # And then once ${init_function['python_name']} succeeds, we can update _library_interpreter._${config['session_handle_parameter_name']}
+        # with the actual session handle.
+        self._library_interpreter._${config['session_handle_parameter_name']} = self.${init_function['python_name']}(${init_call_params})
```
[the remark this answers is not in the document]
DavidCurtiss: Regarding the session handle, see my comment in the other conversation.

Regarding caching LibraryInterpreter as a singleton, I don't think that would be a good change. It would add complexity (another class) for not enough benefit. For the Library class, it's important because we don't want to load the CDLL functions more than we have to. But LibraryInterpreter doesn't do anything worth caching.

generated/nifake/nifake/unit_tests/test_session.py
DavidCurtiss: AB#2156250

build/templates/session.py.mako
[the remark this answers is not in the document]
dmondrik: I would argue that passing the session/vi and encoding to each call seems anti-object-oriented. I don't see the value in the singleton interpreter. @ni-jfitzger @marcoskirsch please provide rationale on the desire for this.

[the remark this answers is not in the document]
ni-jfitzger: > I also moved encoding back to Session, or else we can't possibly cache the LibraryInterpreter because nitclk allows the user to set the encoding.

Are we caching it? Do we want to cache it?
Why, oh why was that argument added to SessionReference when we don't use it anywhere?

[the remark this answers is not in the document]
marcoskirsch: > I would argue that passing the session/vi and encoding to each call seems anti-object-oriented. 

It depends on how you define the class' scope. Library class represenets a DLL. Session represents a driver session. LibraryInterpreter represents... a DLL but interpreted for Python!

This is not the case today, but imagine if our libraries had global functions such as "initialize library" and "finalize library" (we've discussed these for years). Where would that go?

In any case... given that we want to have the ability to sway LibraryInterpreted for one that wraps the gRPC client code, I agree with you that we are better off keeping the session handle as a member of the LibraryInterpreter.

src/nidmm/metadata/functions.py
marcoskirsch: Ah thanks! Forgot. After accepted but before merged please 😉

build/templates/session.py.mako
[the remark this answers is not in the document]
DavidCurtiss: > I agree with you that we are better off keeping the session handle as a member of the LibraryInterpreter.

Okay, I'll revert that commit that moved them to Session.

build/templates/session.py.mako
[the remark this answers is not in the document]
marcoskirsch: > Just following what was done before.

Yeah, it was always ugly.

How do you think we should change it? One half-baked idea is to add smarts to codegenerator so that LibraryInterpreter methods know to store the session handle returned from Init functions automatically. This way Session can become almost completely unaware of session handles. What do you think?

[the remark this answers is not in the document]
DavidCurtiss: > How do you think we should change it?

For this particular conversation ("Can't LibraryInterpreter be constructed as we invoke `super.__init__`?"), it's going to just be "move the object construction up a few lines, to happen while we're initializing SessionBase."

But that doesn't cover the session handle ugliness, which is a different issue. For that, I agree that doing some special Init magic at the LibraryInterpreter layer might work. A simpler solution for solving the encapsulation issue, though, is adding a set_session_handle method to the LibraryInterpreter class. I don't mind just leaving it as-is, though. LibraryInterpreter is an internal class, and if we find some reason in the future to fix the encapsulation, it'll be easy enough to change at that time.

generated/nidcpower/nidcpower/_library_interpreter.py
DavidCurtiss: I may revisit this after gRPC, if it still makes sense. But for now it seems like we'll want the library-specific layers to have that knowledge.

build/templates/session.py.mako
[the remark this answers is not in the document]
DavidCurtiss: Reverted the move-stuff-to-Session commit.

build/templates/session.py.mako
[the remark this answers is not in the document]
DavidCurtiss: Moved the LibraryInterpreter construction.

Let me know what you think about the session handle stuff. I think that's all that's left, other than the hapigen updates.

marcoskirsch (approved):

ni-jfitzger:
The latest commit adds a new entrypoint; I think that merits a changelog update.

src/nidmm/metadata/functions.py
ni-jfitzger: @marcoskirsch were you asking him to copy internal metadata here or to copy these metadata changes to our internal source?

DavidCurtiss: I'm not done yet. Just testing the updates. I'm about to post a PR for the internal source changes, and once we get exports from those changes, I'll copy it back here again.

DavidCurtiss: https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/366375

marcoskirsch: I was asking

>  to copy internal metadata here

The metadata on GitHub shouldn't be hand-edited.

DavidCurtiss:
@marcoskirsch @ni-jfitzger Apparently when I re-request a review from one of you, it unrequests the other. Whatever. I figured you'd both want to see my latest changes. (I _was_ going to wait until after the prior changes were merged and post a new PR, so it's a more limited change, but my hapigen changes are blocked by the pending version upgrades, so I can't merge this PR yet.)

build/templates/session.py/default_method.py.mako
```diff
@@ -7,6 +7,9 @@
     suffix = method_template['method_python_name_suffix']
 
     enum_input_parameters = helper.filter_parameters(f, helper.ParameterUsageOptions.INPUT_ENUM_PARAMETERS)
+    # TODO(marcoskirsch): Retrofit to call filter_parameters(function, parameter_usage_options)
```
ni-jfitzger, line 10: This is simple and easy to read, but should we be using helper functions for consistency?

src/niswitch/metadata/functions.py
```diff
@@ -1205,6 +1216,7 @@
         'documentation': {
             'description': '\nConverts an error code returned by NI-SWITCH into a user-readable\nstring. Generally this information is supplied in error out of any\nNI-SWITCH VI. Use niSwitch_error_message for a static lookup of an\nerror code description.\n'
         },
+        'grpc_name': 'ErrorMessage',
```
ni-jfitzger, line 1219: Although these came from our internal exports, we should avoid adding this unused information, if possible.

build/templates/session.py/default_method.py.mako
DavidCurtiss: Yeah, Marcos is right that we should probably use filter_parameters. That's why I copied his TODO from the location where I copied the logic. I may go ahead and do that retrofit; didn't really have time for it last night. But I'm also not the biggest fan of the giant UsageOptions dictionaries, or at least how they're filled in. I might change _that_, too. But that larger retrofit _really_ doesn't belong in this PR.

src/niswitch/metadata/functions.py
DavidCurtiss: We will use it later, just not in this PR.

build/templates/_converters.py.mako
```diff
@@ -239,7 +239,7 @@ def _convert_timedelta(value, library_type, scaling):
     if library_type in [_visatype.ViInt64, _visatype.ViInt32, _visatype.ViUInt32, _visatype.ViInt16, _visatype.ViUInt16, _visatype.ViInt8]:
```
ni-jfitzger, line 239: Should we stop mentioning ctypes if we no longer return a ctype?

DavidCurtiss: I considered that. Changing the comment to something like "convert float to int if an int is requested" would make it more generic. But on the other hand, it _is_ specifically because of ctypes that we do the conversion. So the comment as-is correctly explains the "why". If you think I should change it then I will, but so far I'm inclined to leave it alone.

build/templates/session.py/default_method.py.mako
```diff
@@ -16,4 +19,14 @@
...
         ${helper.get_enum_type_check_snippet(parameter, indent=12)}
 % endfor
-        return self._library_interpreter.${f['library_interpreter_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_CALL)})
+% for p in helper.filter_parameters(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_CALL):
+%   if 'python_api_converter_name' in p:
+        ${p['python_name']} = _converters.${p['python_api_converter_name']}(${p['python_name']})
+%   endif
+% endfor
+% if output_parameters:
+        ${output_parameters_snippet} = self._library_interpreter.${f['library_interpreter_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_CALL)})
```
ni-jfitzger, line 28: Can't we just return what's on the right-hand side of this = sign?
Doesn't get_session_method_return_snippet() just result in `return ${output_parameters_snippet}`?

I realize named output parameters makes for better documentation but is it worth all of this extra complexity?
At the very least, it looks like we're arriving at the same list of output parameters in 2 different ways?

DavidCurtiss: Often it's the same list, but when there are converters involved, the output_parameters_snippet will invoke those converters on the appropriate output parameters.

I _could_ detect when there are no converters involved, and just do the `return` directly in those cases, but since this is just generated code, I don't think it's worth the added complexity.

build/helper/codegen_helper.py
```diff
@@ -1401,34 +1398,40 @@ def get_enum_value_snippet(value):
...
 
 
-def test_get_method_return_snippet_into():
+def test_get_library_interpreter_method_return_snippet_into():
     param = [parameters_for_testing[4], parameters_for_testing[7]]
-    assert get_method_return_snippet(param, config_for_testing, use_numpy_array=True) == 'return'
+    assert get_library_interpreter_method_return_snippet(param, config_for_testing, use_numpy_array=True) == 'return'
+
+
+def test_get_session_method_return_snippet():
```
ni-jfitzger, line 1431: I think you may need more test cases for 'get_session_method_return_snippet'. You may have mostly copied the code from 'get_library_interpreter_method_return_snippet' but now that they are separate methods you can't count on testing for that method to make sure all of the behavior for this method is correct.

build/templates/session.py/default_method.py.mako
ni-jfitzger: Yeah, I'm not a big fan of some of this codegen code, either, but I agree that I don't want to add more big changes to this PR.

build/helper/codegen_helper.py
DavidCurtiss: Before this change, the get_method_return_snippet function had a matrix of [ctypes_conversions] x [converters_conversions]. Now, the library_interpreter one has just [ctypes_conversions], and the session one has just [converters_conversions]. When unit-testing the session one, I shouldn't need to test any of the ctypes_conversions, and since the _converters API is the same for all converter types, I just need one unit test for it.

The overall flow still has a matrix of [ctypes_conversions] x [converters_conversions], because the Session's conversions get layered on top of the LibraryInterpreter's conversions, but that's now the job of system tests, not unit tests.

build/helper/codegen_helper.py
```diff
@@ -147,14 +154,26 @@ def _get_output_param_return_snippet(output_parameter, parameters, config):
...
 
 # TODO(marcoskirsch): Retrofit to call filter_parameters(function, parameter_usage_options)
-def get_method_return_snippet(parameters, config, use_numpy_array=False):
-    '''Returns a string suitable to use as the return argument of a Session method, i.e. "return reading_ctype.value"'''
+def get_library_interpreter_method_return_snippet(parameters, config, use_numpy_array=False):
+    '''Returns a string suitable to use as the return argument of a LibraryInterpreter method, i.e. "return reading_ctype.value"'''
     snippets = []
     for x in parameters:
         if x['direction'] == 'out' or x['size']['mechanism'] == 'ivi-dance':
             if x['numpy'] is False or use_numpy_array is False:
```
ni-jfitzger, line 162: @marcoskirsch Is this `or use_numpy_array is False` actually providing any value?
After going through the metadata and template that call this, I don't understand the argument's intended use.

DavidCurtiss: It does. use_numpy_array is passed in by the numpy templates. The functions that have a numpy variant also have a non-numpy variant. When invoked as the non-numpy variant, the "numpy" outputs need to be included. But when invoked as the numpy variant, these outputs are inputs instead, so they're excluded here.

build/helper/codegen_helper.py
DavidCurtiss: Gotcha. I'll check on that today.

build/helper/codegen_helper.py
ni-jfitzger: Excerpt from NI-SCOPE "FetchBinary8" function metadata.

```
            {
                'direction': 'out',
                'documentation': {
                    'description': '\nReturns an array whose length is the **numSamples** times number of\nwaveforms. Call niScope_ActualNumwfms to determine the number of\nwaveforms.\n\nNI-SCOPE returns this data sequentially, so all record 0 waveforms are\nfirst. For example, with a channel list of 0,1, you would have the\nfollowing index values:\n\nindex 0 = record 0, channel 0\n\nindex *x* = record 0, channel 1\n\nindex 2\\ *x* = record 1, channel 0\n\nindex 3\\ *x* = record 1, channel 1\n\nWhere *x* = the record length\n'
                },
                'name': 'waveform',
                'numpy': True,
                'size': {
                    'mechanism': 'python-code',
                    'value': '(num_samples * self.actual_num_wfms(channel_list))'
                },
                'type': 'ViInt8[]',
                'use_array': True
            },
```
I don't think I see waveform returned anywhere, although the documentation for _fetch_into functions seem to list it anyway.

DavidCurtiss: Yes, the fact that it's listed in the docstring is a pre-existing bug. And I have a fix for it...

* Yesterday I refactored the aforementioned TODO, wherein we should be using ParameterUsageOptions and filter_parameters instead of manually filtering the parameters. You can see that refactor [here](https://github.com/DavidCurtiss/nimi-python/commits/refactor-output-params), in the Sep 28 changes. I'll post a PR for that once this one is complete. (Unless you think I should go ahead and add it to this one, since I'm still blocked on merging this one anyway.)
.
[One of the changes](https://github.com/DavidCurtiss/nimi-python/commit/467a76c830bfe85991992010868d2e16abc61b38) in that refactor adds a API_NUMPY_OUTPUT_PARAMETERS option, which I then use in `documentation_helper.get_function_docstring` for numpy functions, which fixes ni-scope's docstrings.

ni-jfitzger: No, please no more big changes in this PR. My brain might crash.

build/helper/codegen_helper.py
DavidCurtiss: Added tests for tests for use_numpy_array=True/False.

marcoskirsch (approved):

build/templates/session.py.mako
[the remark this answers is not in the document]
marcoskirsch: We can leave it alone for this PR, and revisit if we want/need to.

generated/nifake/nifake/unit_tests/test_session.py
```diff
@@ -1,116 +1,81 @@
...
-                if f.startswith("niFake_") and not f.endswith("_cfunc"):
-                    setattr(self, f, MagicMock())
+                if not f.startswith("_"):
+                    setattr(self, f, MagicMock(spec_set=getattr(self, f), side_effect=_mock_helper.MockFunctionCallError(f)))
 
     def setup_method(self, method):
-        self.patched_library = self.PatchedLibrary(None)
-        self.patched_library_singleton_get = patch('nifake._library_interpreter._library_singleton.get', return_value=self.patched_library)
+        self.patched_library_interpreter = self.PatchedLibraryInterpreter(None)
+        self.patched_library_singleton_get = patch('nifake.session._library_interpreter.LibraryInterpreter', return_value=self.patched_library_interpreter)
```
ni-jfitzger, line 26: This name `self.patched_library_singleton_get` seems wrong, now.

DavidCurtiss:
@marcoskirsch @ni-jfitzger I had to fix a couple things in the metadata, in [this commit](https://github.com/ni/nimi-python/pull/1819/commits/8338342aa296e59ba133431295c7b2fbf441762a). I'll go fix those in the source files and push another PR update.

CHANGELOG.md
```diff
@@ -42,17 +42,23 @@ All notable changes to this project will be documented in this file.
...
     * #### Removed
 * ### `nidigital` (NI-Digital Pattern Driver)
     * #### Added
     * #### Changed
     * #### Removed
 * ### `nidmm` (NI-DMM)
     * #### Added
+        * Added enum `DigitsResolution`, to clarify the available options for the `resolution_digits` attribute.
     * #### Changed
+        * The `resolution_digits` attribute's type is now `enums.DigitsResolution`.
```
DavidCurtiss, line 55: @marcoskirsch @ni-jfitzger This is a breaking change, because we require that the proper enum type is passed in; you can't just pass in a float that matches one of the enum values. Should I update the metadata to avoid it?

ni-jfitzger: I was about to bring this up. I don't think you should be making breaking changes.
i don't understand why you made this change in the first place.

generated/nidcpower/nidcpower/session.py
```diff
@@ -6271,11 +6275,11 @@ def query_output_state(self, output_state):
...
 
-                +----------------------+-------------------------------------------------------------------+
-                | OutputStates.VOLTAGE | The device maintains a constant voltage by adjusting the current. |
-                +----------------------+-------------------------------------------------------------------+
-                | OutputStates.CURRENT | The device maintains a constant current by adjusting the voltage. |
-                +----------------------+-------------------------------------------------------------------+
+                +-------------------------------+-------------------------------------------------------------------+
+                | OutputStates.CONSTANT_VOLTAGE | The device maintains a constant voltage by adjusting the current. |
+                +-------------------------------+-------------------------------------------------------------------+
+                | OutputStates.CONSTANT_CURRENT | The device maintains a constant current by adjusting the voltage. |
```
DavidCurtiss, line 6281: @marcoskirsch @ni-jfitzger The two enum values here are renamed because new enum values were added to the OutputStates enum. (Side note: I mentioned the renames and additions in the changelog.)

1. This is a breaking change, because the enum values were renamed. Does that need a special call-out?
2. The new enum values are not listed here because they're not listed in the doc table for this particular function parameter (in the metadata). Should they be?

CHANGELOG.md
DavidCurtiss: It was caused by a metadata update. In the source data, the enum has long been marked as 'codegen_method':'no', but a recent change moved it to a different layer, and I think I now need to explicitly add in the 'codegen_method':'no' at the python layer. I'll do that.

generated/nidcpower/nidcpower/session.py
ni-jfitzger: >This is a breaking change, because the enum values were renamed. Does that need a special call-out?

Hmm, the new names do seem to match (though not casing) the names used in the LabVIEW API.
https://www.ni.com/docs/en-US/bundle/ni-dcpower-labview-api-ref/page/nidcpowerviref/nidcpower_query_output_state.html

Still, I'm not sure if that meets the bar to justify breaking client code. We have documentation for these enums, if clarification is needed.

I do like the new names better, though. Can we create aliases?

CHANGELOG.md
marcoskirsch: we shouldn't introduce source breakers

generated/nidcpower/nidcpower/session.py
marcoskirsch: Yuck, the codegenerator is "smart" in that it strips the prefix of enum names if they all match. 90% of the time that's a good thing because we "namespace" the C `#define`s by using a common prefix.

In this case, I believe we should **not** have stripped the prefix. But we aren't adding more enum values so for now we should just avoid the source breaker.

marcoskirsch: I think the follow up actions are:

1. change metadata to override this so it remains as it was
2. File an issue tagged with source breaker so we can tackle this when/if we do a source breaking 2.0 release.

DavidCurtiss: > But we aren't adding more enum values so for now ...

But we _are_ adding more enum values. We're just not documenting them in this function's parameter docs.

If we intentionally don't want to support those new enum values for this function, then should we introduce a "fancy" function to explicitly reject those new enum values? Should I do something in the hapigen data to erase the added enum values at the python layer?

Alternatively, if we're okay supporting the new enum values, then I should probably just update the function parameter docs to include them.

CHANGELOG.md
DavidCurtiss: Ok, fixing in [!372493](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/372493)

generated/nidcpower/nidcpower/session.py
DavidCurtiss: In [!372484](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/372484), I took the approach of adding aliases for the enum values. Not sure if you'll like it or not.

(And I'm still hoping for an answer on whether or not the other enum values should be permitted -- or perhaps explicitly rejected -- as valid inputs for the `query_output_state` function.)

generated/nifake/nifake/unit_tests/test_session.py
DavidCurtiss: Renamed to `patched_lirary_interpreter_ctor`. (I'll push the fix when I update the dmm/dcpower metadata again.)

src/nidigital/metadata/enums.py
```diff
@@ -147,6 +165,14 @@
             }
         ]
     },
+    'HistoryRAMMaxSamplesToAcquirePerSite': {
```
ni-jfitzger, line 168: This didn't exist prior to its addition to grpc metadata. Why was it added?

generated/nifgen/nifgen/session.py
```diff
@@ -243,6 +243,19 @@ class _SessionBase(object):
 
     Controls the specified auxiliary power pin. Setting this property to TRUE energizes the auxiliary power when the session is committed. When this property is FALSE, the power pin of the connector outputs no power.
     '''
+    burst_count = _attributes.AttributeViInt32(1250350)
```
ni-jfitzger, line 246: Do we want to suddenly provide access to this attribute?
It's being added by the grpc metadata and for all I know it was omitted originally for a reason.

ni-jfitzger:
@marcoskirsch @DavidCurtiss I think it was a mistake to pull in metadata exports as part of this change.
We had carefully reviewed the (already very big and long) PR and were confident in the the changes going in.

Given the ongoing changes to internal metadata expansion, mixing metadata exports with this PR feels like an invitation for unwanted changes to our API to sneak in.  Although I do think it's the next thing that should go in, I think we should revert those changes from this PR and create a new PR to handle them.

src/nifgen/metadata/enums.py
```diff
@@ -33,6 +33,46 @@
             }
         ]
     },
+    'ArbitrarySequenceHandle': {
```
ni-jfitzger, line 36: Why are we adding all of these new Enums?

src/niswitch/metadata/enums.py
```diff
@@ -987,5 +987,21 @@
                 'value': 2
             }
         ]
+    },
+    'WireMode': {
```
ni-jfitzger, line 991: Another new enum. I found the internal discussion of it. I think this one's okay.

generated/nifgen/nifgen/session.py
marcoskirsch: It was omitted because it is only relevant to hardware that is not supported by anything today. It shouldn't be added.

marcoskirsch:
> I think it was a mistake to pull in metadata exports as part of this change.
We had carefully reviewed the (already very big and long) PR and were confident in the the changes going in.

I see your point. That would have left mainline with source breakers though which is not the end of the world but also not "correct".

src/nifgen/metadata/enums.py
DavidCurtiss: Most of them are referenced as a `grpc_enum` in some existing function or attribute. So we don't technically need them _yet_, but will soon.

There are a few, though, that were only referenced in 'codegen_method':'no' attributes, so I can mark those enums a 'codegen_method':'no', too.

ni-jfitzger: Why does grpc require us to add new enums?

DavidCurtiss: Actually, when there's not a corresponding python enum, we probably won't need that to generate the enum for `grpc_enum` functions/attributes. (We just need to know that it's defined as _some_ enum for grpc.) So I could technically mark those enums as 'codegen_method':'no', too, if you insist.

src/niswitch/metadata/enums.py
ni-jfitzger: Reopening conversation, in case we decide we can/should omit it along with other new grpc enums.

src/nifgen/metadata/enums.py
DavidCurtiss: Note that even with the enums added to the metadata, they still don't appear in the generated API (as you can see in the [diffs](https://github.com/ni/nimi-python/pull/1819/commits/dba97eb5194d0c741fb0c27b2f5de61812ae713a)), because the nimi-python codegen step throws out any unreferenced enums.

ni-jfitzger: I hadn't noticed that. I guess that's okay for now, but I think I still would like to mark them as 'codegen_method':'no' as soon as possible. It's confusing to list unused enums.

src/nidigital/metadata/enums.py
ni-jfitzger: I think we should set `'codgen_method'`: 'no' for this, but it's not being added to the generated code, so it probably doesn't need to hold up this PR.

src/niswitch/metadata/enums.py
ni-jfitzger: I think we should set 'codgen_method': 'no' for this, but it's not being added to the generated code, so it probably doesn't need to hold up this PR.

src/nifgen/metadata/enums.py
marcoskirsch: > but I would lean towards not adding

This PR shouldn't be changing customer-facing API.

src/niswitch/metadata/enums.py
marcoskirsch: If it's internal only, I'm ok to merge.

src/nidigital/metadata/enums.py
DavidCurtiss: I have a [PR](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/373719) ready to commit that will allow this. After that, I'll go apply it to all of those enums. But I'll make those changes separate from this PR.

CHANGELOG.md
DavidCurtiss: Reverted this change.

generated/nidcpower/nidcpower/session.py
DavidCurtiss: Reverted this change.

src/nifgen/metadata/enums.py
DavidCurtiss: I have a [PR](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/373719) ready to commit that will allow this. After that, I'll go apply it to all of those enums. But I'll make those changes separate from this PR.

src/niswitch/metadata/enums.py
DavidCurtiss: I have a [PR](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/373719) ready to commit that will allow this. After that, I'll go apply it to all of those enums. But I'll make those changes separate from this PR.

generated/nifgen/nifgen/session.py
DavidCurtiss: Reverted this change.

src/nidcpower/metadata/functions.py
DavidCurtiss: All metadata is now updated from internal exports.

src/nidmm/metadata/functions.py
DavidCurtiss: Done

PR #1829, opened 2022-10-15
Rename _grpc.py to _grpc_stub_interpreter.py

DavidCurtiss:

Renames `_grpc.py` to `_grpc_stub_interpreter.py`, to match the contained class name.

None

Ran the grpc-specific unit tests.


ni-jfitzger (approved):

generated/nifake/nifake/session.py
```diff
@@ -697,8 +697,8 @@ def __init__(self, resource_name, options={}, id_query=False, reset_device=False
 
         '''
         if _grpc_channel:
-            import nifake._grpc as _grpc
-            library_interpreter = _grpc.GrpcStubInterpreter(_grpc_channel)
+            import nifake._grpc_stub_interpreter as _grpc_stub_interpreter
+            library_interpreter = _grpc_stub_interpreter.GrpcStubInterpreter(_grpc_channel)
```
dmondrik, line 701: Should the session's member now just be called `interpreter`? Note the current name is now incorrect for gRPC.

dmondrik: Consider: naming the member `implementation` might be more human readable.

dmondrik (approved):

generated/nifake/nifake/session.py
dmondrik: FYI @ni-jfitzger

ni-jfitzger: I agree that the name `library_interpreter` doesn't fit the gRPC case. I don't feel great about the name `implementation`. It's not an object of some child class of an interface class and `implementation` also feels pretty vague.

`interpreter` feels like a good compromise.

Do we need an ABC Interpreter class for each of the interpreters to inherit from?

I'm interested in hearing @marcoskirsch's thoughts.

generated/nifake/nifake/session.py
```diff
@@ -697,8 +697,8 @@ def __init__(self, resource_name, options={}, id_query=False, reset_device=False
...
         else:
             library_interpreter = _library_interpreter.LibraryInterpreter(encoding='windows-1251')
```
marcoskirsch, line 703: you could move the LibraryInterpreter import in here in order to be consistent.

generated/nifake/nifake/session.py
marcoskirsch: I definitely think the varaible should be called `interpreter` and the member `_interpreter`.

marcoskirsch: I'm ok if for noisiness @DavidCurtiss prefers to do it in a separate PR.

marcoskirsch (approved):
There's one tiny change to move an import but I don't feel strongly about it and maybe you have a reason to be against it.

generated/nifake/nifake/session.py
DavidCurtiss: In general, I keep all imports at the top, but it was important to have the grpc import here because just importing it requires the grpcio package to be installed, whereas importing the library_interpreter doesn't need anything extra.

Still, if it were only this `if`/`else` block, I'd go ahead and move it for consistency, but there's also the mako `%else` block that needs the `_library_interpreter`, so I'm going to leave the import at the top.

dmondrik: To be clear, the import of gRPC must happen only inside the `if` block, by design. We do not want `nilrt` users who are just using it locally to suddenly require gRPC packages.

generated/nifake/nifake/session.py
DavidCurtiss: It _is_ a bit noisy, but it's straightforward, and won't clash with the file rename, so I'll go ahead and do it here, and reset Jay's vote.

DavidCurtiss: Done

ni-jfitzger (approved):
This looks fine to me, assuming you reran the tests locally on a Windows system (since our Windows VMs are down).

build/templates/session.py/default_method.py.mako
```diff
@@ -25,8 +25,8 @@
 %   endif
 % endfor
 % if output_parameters:
-        ${output_parameters_snippet} = self._library_interpreter.${f['library_interpreter_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_CALL)})
+        ${output_parameters_snippet} = self._interpreter.${f['library_interpreter_name']}${suffix}(${helper.get_params_snippet(f, helper.ParameterUsageOptions.LIBRARY_INTERPRETER_METHOD_CALL)})
```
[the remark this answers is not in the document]
DavidCurtiss: I'm not going to rename library_interpreter_filename here, because that will require changes to the source metadata, but I've updated the rest.

build/helper/parameter_usage_options.py
```diff
@@ -20,10 +20,10 @@ class ParameterUsageOptions(AutoNumber):
...
     SESSION_NUMPY_INTO_METHOD_DECLARATION = ()
     '''For declaring a Session method that uses numpy arrays.'''
-    LIBRARY_INTERPRETER_NUMPY_INTO_METHOD_DECLARATION = ()
-    '''For declaring a LibraryInterpreter method that uses numpy arrays.'''
-    LIBRARY_INTERPRETER_NUMPY_INTO_METHOD_CALL = ()
-    '''For calling into a LibraryInterpreter method (from Session) that uses numpy arrays.'''
+    INTERPRETER_NUMPY_INTO_METHOD_DECLARATION = ()
+    '''For declaring a **Interpreter method that uses numpy arrays.'''
+    INTERPRETER_NUMPY_INTO_METHOD_CALL = ()
+    '''For calling into a **Interpreter method (from Session) that uses numpy arrays.'''
```
ni-jfitzger, line 26: What's with the **?

ni-jfitzger: Why isn't it just one *? Is this a convention that I'm not familiar with?

DavidCurtiss: No convention, just how my brain works, I guess.

PR #1828, opened 2022-10-15
Move shared .proto files from common/metadata/ to shared_protos/

DavidCurtiss:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

~~[ ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~

~~[ ] I've added tests applicable for this pull request~~

### What does this Pull Request accomplish?

Moves session.proto and nidevice.proto to a clearer path. (Note that "metadata" has a special meaning for gRPC, so having [only] .proto files in a "metadata" folder is a bit odd.)

Also, adds a README.md describing where to update the files from.

### List issues fixed by this Pull Request below, if any.

None

### What testing has been done?

PR

PR #1830, opened 2022-10-17
Add NI-FGEN proto file to nimi-python.

tdunkle:

Showing that the existing NI-FGEN system-level tests pass is sufficient.

This pull request adds the NI-FGEN proto file to nimi-python, so that the NI-FGEN python API has a gRPC option enabled.

* [AB#2160599](https://ni.visualstudio.com/DevCentral/_workitems/edit/2160599/)

Installed python 3.7.9 and synced the nimi-python repo to a VM that has NI-FGEN 21.8 installed. Observed that I could build fgen with `tox -e py37-codegen -- nifgen`, and that I could run the system level tests with `tox -c generated\nifgen\tox-system_tests.ini`, and they all passed. Test output, for reference: [fgen_test_output.txt](https://github.com/ni/nimi-python/files/9778276/fgen_test_output.txt)


generated/nifgen/nifgen/_grpc.py
```diff
@@ -0,0 +1,514 @@
...
+
+    def is_done(self):  # noqa: N802
+        response = self._invoke(
+            self._client.IsDone,
+            grpc_types.IsDoneRequest(vi=self._vi),
+        )
+        return response.done
+
+    def lock(self):  # noqa: N802
+        response = self._invoke(
```
dmondrik, line 353: @DavidCurtiss why is the lock method for fgen not the local recursive lock?

generated/nifgen/nifgen/session.py
```diff
@@ -3100,7 +3106,9 @@ def __init__(self, resource_name, channel_name=None, reset_device=False, options
         # with the actual session handle.
         self._library_interpreter._vi = self._initialize_with_channels(resource_name, channel_name, reset_device, options)
 
-        self.tclk = nitclk.SessionReference(self._library_interpreter._vi)
+        # NI-TClk does not work over NI gRPC Device Server
```
dmondrik, line 3109: @DavidCurtiss I'm confused by this comment [given that nitclk exists there](https://github.com/ni/grpc-device/tree/main/generated/nitclk). I don't think we need to support fgen-with-tclk-over-grpc in 1.0 but I'm not sure what is expected that wouldn't work.

dmondrik: Honestly in looking at the entire file I don't see where the member is even used. Was this some planned usage that didn't get completed?

DavidCurtiss: Maybe we should say "isn't yet supported"?

DavidCurtiss: How about:
`# NI-TClk isn't yet supported in gRPC-enabled Sessions`
@marcoskirsch ?

DavidCurtiss: > I don't see where the member is even used

It's just sitting out there for users to access if they want.

DavidCurtiss:
> Should this wait for #1828?

#1828 only moves the _shared_ proto files, not the per-driver ones, so no. Or maybe you meant #1829?

ni-jfitzger:
>>Should this wait for https://github.com/ni/nimi-python/pull/1828?

>https://github.com/ni/nimi-python/pull/1828 only moves the shared proto files, not the per-driver ones, so no. Or maybe you meant https://github.com/ni/nimi-python/pull/1829?

Good to know. No, I was only concerned with 1828. For 1829, whichever PR goes in second will simply have to regenerate.

tdunkle:
Update: This is currently waiting on https://github.com/ni/grpc-device/pull/767 , and another codegen after that goes in.

dmondrik (approved):

generated/nifgen/nifgen/session.py
marcoskirsch: sounds good to me

marcoskirsch: > I don't see where the member is even used

This `tclk` member is part of the API.

marcoskirsch:
@tdunkle there's merge conflicts can you fix please?

generated/nifgen/nifgen/_grpc.py
```diff
@@ -0,0 +1,514 @@
+# -*- coding: utf-8 -*-
```
marcoskirsch, line 1: this file needs to be renamed to match the name of the class per nimi-python's convention.

generated/nifgen/nifgen/_grpc.py
marcoskirsch: You are going to get errors (if not implemented by server) or deadlocks (if yes implemented) whenever you use this, right?

That's ok for now since we are still building this. Just wanting to understand.

generated/nifgen/nifgen/nidevice_pb2.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
marcoskirsch, line 1: I see `_grpc.py` is prefixed with underscore denoting it's an internal implementation detail.
This file and others should follow suit.

generated/nifgen/nifgen/session.py
```diff
@@ -3076,12 +3076,18 @@ def __init__(self, resource_name, channel_name=None, reset_device=False, options
...
+
 
         Returns:
             session (nifgen.Session): A session object representing the device.
 
         '''
-        library_interpreter = _library_interpreter.LibraryInterpreter(encoding='windows-1251')
+        if _grpc_channel:
+            import nifgen._grpc as _grpc
+            library_interpreter = _grpc.GrpcStubInterpreter(_grpc_channel)
```
marcoskirsch, line 3088: confirm this renaming is happening in a separate PR please

generated/nifgen/tox-system_tests.ini
```diff
@@ -47,6 +47,8 @@ deps =
     nifgen-system_tests: hightime
     nifgen-system_tests: fasteners
     nifgen-system_tests: pytest-json
+    nifgen-system_tests: grpcio
```
marcoskirsch, line 50: This tells me that the wheel for `nifgen` is not going to install these things, so the client that wants to use `nifgen` with gRPC needs to explicitly install them.

Do we provide reasonable errors if this isn't the case? Maybe it's a moot point because you need these in order to create the channel?

src/nifgen/metadata/nifgen.proto
```diff
@@ -0,0 +1,2150 @@
+
```
marcoskirsch, line 1: These files got moved to a different dir.
Looks like this PR needs to merge latest main?

marcoskirsch (changes requested):

marcoskirsch:
PR does a lot more than just add the proto file.

When referencing internal NI bugs in this open source repository, make sure to call it out. We usually say something like "Internal NI bug [AB#2160599](https://ni.visualstudio.com/DevCentral/_workitems/edit/2160599/)"

generated/nifgen/nifgen/_grpc.py
DavidCurtiss: This is a generated file, of course, and Tom just needs to merge in the latest commits and re-codegen.

generated/nifgen/nifgen/nidevice_pb2.py
DavidCurtiss: This is how they're generated by protoc. Perhaps a better option would be to put them in an underscore-prefixed sub-package?

generated/nifgen/nifgen/session.py
DavidCurtiss: The rename PR has already gone in; needs to be merged in to this PR.

generated/nifgen/tox-system_tests.ini
DavidCurtiss: Correct, the client would need to explicitly install them. If you think we should auto-install it via dependencies, we can have that conversation.

You're right that the user has to create the channel, at least for now certainly.

src/nifgen/metadata/nifgen.proto
DavidCurtiss: No, only the shared .proto files got moved. (And that PR _just_ went in.) The per-driver files were left next to the .py metadata, for (hopefully) easier maintenance.

generated/nifgen/tox-system_tests.ini
marcoskirsch: No I think this is good.
We did the same for numpy.

src/nifgen/metadata/nifgen.proto
marcoskirsch: 🤔
I'm not sure which way is better. Alright this is good for this PR thanks for clarifying.

generated/nifgen/tox-system_tests.ini
dmondrik: Yeah the channel creation is up to the client, so they'll need the python gRPC package installed. They would likely get some nonsensical python error if they passed in `_grpc_channel=True` but that's fine with me.

PR #1833, opened 2022-10-19
Add proto and tests for NI-DMM

DavidCurtiss:

Adds nidmm.proto, and system tests for it.

None

Ran system tests on a machine with the required prerequisites installed. Also ran on my dev machine, where ni_grpc_device_server is not installed (but other NI software is), to see that it skips the tests.


dmondrik:
I'd like to see [my PR for gRPC session options](https://github.com/ni/nimi-python/pull/1835) to go in first.

src/nidmm/metadata/functions.py
```diff
@@ -1509,7 +1518,7 @@
         'method_templates': [
             {
                 'documentation_filename': 'lock',
-                'library_interpreter_filename': 'default_method',
```
dmondrik, line 1512: I don't think this is a new enough export. The init method starting on line 1377 should have the new initialization template stuff right?

dmondrik (changes requested):

src/nidmm/metadata/functions.py
DavidCurtiss: You're right, I need to update metadata. However, now that I do that...
`TypeError: Only simple types allowed for arrays: ViAddr`
🙁

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -1,338 +1,352 @@
...
-
-
-def test_get_cal_date_time(session):
-    last_cal = session.get_cal_date_and_time(0)
-    assert last_cal.month == 3
-    assert last_cal.day == 1
-    assert last_cal.year == 1940
-    assert last_cal.hour == 0
-    assert last_cal.minute == 0   # cal_date_and_time should be 03/01/1940:00:00 for simulated 408x devices; 407x and 4065 returns 00/00/0000:00:00
+class SystemTests:
```
marcoskirsch, line 14: it's unfortunate that these diffs are impossible to follow, we should have done the "turn into a class without functional differences" work in its own PR.

marcoskirsch:
@DavidCurtiss can you please look into getting all the PR checks to succeed?

build/helper/metadata_add_all.py
```diff
@@ -106,7 +106,10 @@ def _add_ctypes_variable_name(parameter):
 
 def _add_ctypes_type(parameter, config):
     '''Adds a ctypes_type key/value pair to the parameter metadata for calling into the library'''
-    parameter['ctypes_type'] = parameter['type']
+    if parameter['type'] == 'ViAddr' and (parameter['use_array'] or parameter['use_list']):
```
dmondrik, line 109: await: I don't agree that this hack is legitimate. The change must happen in the hapigen metadata to override the gRPC information. 

Why? Look at this 'export' structure from the .proto file:
```
message ExportAttributeConfigurationBufferResponse {
  int32 status = 1;
  repeated fixed64 configuration = 2;
}
```
It's not `repeated fixed64`, it's `bytes`.

src/nidmm/system_tests/test_system_nidmm.py
DavidCurtiss: It still doesn't help much unless you pull the branch locally and use a real diff tool. I split apart the first test change such that the system tests are first put into a class. You can see that commit in isolation [here](https://github.com/ni/nimi-python/pull/1833/commits/bb3808f2adc1a371465a5452088bb77b8ca959ce).

The next commit on top of that adds the gRPC testing, [here](https://github.com/ni/nimi-python/pull/1833/commits/35e5c4a8dc9813747a088c0ee30e2353e7ca8c4b#diff-916d5685a40780e03962ad6c406da272b854a8f97d5650a6892946f930fbcb86). Further changes to the tests can be seen [here](https://github.com/ni/nimi-python/pull/1833/commits/171928ba04deaa905b7dec77b42d0c5faf76688c), [here](https://github.com/ni/nimi-python/pull/1833/commits/d8995eba542f57891ee98573ecf4573547ae66e8), and [here](https://github.com/ni/nimi-python/pull/1833/commits/64b15b278ca4457ee6dd55689d9c3aa27dd6452d). All of these changes diff pretty well, at least.

dmondrik (changes requested):

build/helper/metadata_add_all.py
dmondrik: As you can see, 'bytes' is [what is checked into gRPC device](https://github.com/ni/grpc-device/blob/main/generated/nidmm/nidmm.proto#L952).

DavidCurtiss: Fixed the proto.

But the hack is still necessary, for the LibraryInterpreter code. I agree that the change here is not ideal, but until [AB#2185120](https://dev.azure.com/ni/DevCentral/_workitems/edit/2185120) is fixed, this is the best I could come up with. I would've preferred to do it in the hapigen data, but the code here ignores the python_type, and the PythonExpander in hapigen only looks at the cheader type. I could've modified the one of those things, of course, but those would've been larger changes that both seem like overkill for this little use case, especially since we've got that Tech Debt to change how these work.

src/nidmm/metadata/functions.py
DavidCurtiss: Done. But using non-exported metadata at the moment. I'll open a separate conversation for that.

src/nidmm/metadata/attributes.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-DMM API metadata version 23.0.0d21
+# This file is generated from NI-DMM API metadata version 23.0.0d9999
```
DavidCurtiss, line 2: Pending [AB!386120](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/386120).

DavidCurtiss: Done

dmondrik (approved):
LGTM

CHANGELOG.md
```diff
@@ -49,6 +49,10 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### `nidmm` (NI-DMM)
     * #### Added
+        * API parity with next NI-DMM release.
+            * Methods added:
+                * `convert_absolute_to_digits`
+                * `convert_digits_to_absolute`
```
marcoskirsch, line 55: these functions are ancient and were left out of `nidmm` on purpose

marcoskirsch (changes requested):

CHANGELOG.md
dmondrik: I guess we'll have to mark the grpc and python derivations as codegen `no` correct? I glanced at the history but it doesn't seem to have changed. So I'm curious what process caused them to be omitted before.

DavidCurtiss: They were added in [AB!378345](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/378345?_a=files&path=/src/nidmm/nidmmBaseDistribution/nidmmAPIMetadata/source/nidmmAPIMetadata) because it was apparently needed for another use case, but they neglected to add 'codegen_method':'no' for gRPC/Python like is done for every other function in that file.

DavidCurtiss: Removed

marcoskirsch (approved):

PR #1835, opened 2022-10-24
Add support for gRPC session options as a separate class

dmondrik:
- [X] This contribution adheres to _the spirit of_ [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).
  - Running tox failed with `py37-test` as a key.
  - There were no installers for me to use.
  - I was unable to run system tests.

~~TODO: Check the above box with an 'x' indicating you've read and followed [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).~~

TODO: Check the above box with an 'x' if considered if there were any and then documented client facing changes in [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md). Strike through if this is not a relevant client facing change.

TODO: Check the above box with an 'x' indicating you have considered and added any applicable system or unit tests. Strike through if you considered and decided additional tests were not warranted.

Multiple gRPC options are necessary to satisfy all business requirements. Putting them into a single object makes the Session constructor simpler.

[Task 2186992](https://dev.azure.com/ni/DevCentral/_workitems/edit/2186992): [nimi-python] Pass init/attach fields properly in GrpcStubInterpreter

This also includes updates to the latest exports for gRPC-enabled APIs:
`nifake_grpc_device` and `nifake_nimi_python` 23.0.0d65
`nifgen_grpc_device` and `nifgen_nimi_python` 23.0.0d54
`niswitch_grpc_device` and `niswitch_nimi_python` 23.0.0d69

### What does this Pull Request NOT accomplish?

The user can currently attach to an existing session and manually close it, but the opposite is not possible in a clean way. You could initialize the session and not put it in a context manager `with` block, but that's not necessarily recommended.

There is an outstanding question of how to allow an explicit detach.
- We could add a method (but that wouldn't be valid on native sessions). While LV is doing that, it's not inherently necessary for us to do the same here.
- We could add it to the session options (but that might not be intuitive to specify up-front what action you want taken later). An analogy is how you can specify FILE_FLAG_DELETE_ON_CLOSE when opening a file on Windows.
- It's not clear whether we need to answer the question before 1.0 ships since the automatic actions in this PR satisfy the immediate business requirements.

N/A

Test setup: Windows VM with NI-DMM installed and the latest artifact build of the gRPC Device Server. (The installer includes an older binary; Johann has since updated that.) I created a simulated 4065. I used I/O Trace to verify which calls were made from what process.

Test procedure: 5 tiny .py tests always having a `with` statement to ensure the session went out of scope properly.
- First I wrote a simple Session that read a value using the local native driver. The session was closed.
- The rest used gRPC. In the session options I specified behavior=INITIALIZE_NEW and close=False. It created a session and left it open.
- Running the same program again, it threw a python exception because the gRPC session was already initialized. No driver calls were made.
- I changed the session options to behavior=ATTACH_TO_EXISTING and close=True. It read the value using the existing session and then closed it.
- Running the same program again, it threw a python exception because there was no session to which it could attach. No driver calls were made.


build/templates/GrpcOptions.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+from enum import Enum
+
```
DavidCurtiss, line 3: Filename should be grpc_options.py

generated/nifake/nifake/GrpcOptions.py
```diff
@@ -0,0 +1,48 @@
...
+    r'''
+    This requires the NI gRPC Device Server to initialize a new session with the specified name.
+    '''
+    ATTACH_TO_EXISTING = 2
+    r'''
+    This requires the NI gRPC Device Server to attach to an existing session with the specified name.
+    '''
+
+
+class GrpcOptions(object):
```
bkeryan, line 20: Consider naming this `GrpcDeviceOptions` to disambiguate it from gRPC channel/server options such as `grpc.insecure_channel(target, options=[("grpc.max_receive_message_length", -1)])`

build/templates/session.py.mako
```diff
@@ -234,13 +234,13 @@ if grpc_supported:
...
-            'documentation': { 'description': 'MeasurementLink gRPC channel' },
+            'documentation': { 'description': 'MeasurementLink gRPC options' },
             'enum': None,
             'is_repeated_capability': False,
             'is_session_handle': False,
-            'python_name': '_grpc_channel',
+            'python_name': '_grpc_options',
             'size': {'mechanism': 'fixed', 'value': 1},
-            'type_in_documentation': 'grpc.Channel',
+            'type_in_documentation': 'GrpcOptions',
```
DavidCurtiss, line 243: Since the user will have to construct this manually, we should probably be more clear where they need to get this from: `module_name + '.grpc_options.GrpcOptions'`

build/templates/session.py.mako
```diff
@@ -249,11 +249,14 @@ if grpc_supported:
         ${helper.get_function_docstring(ctor_for_docs, False, config, indent=8)}
         '''
 % if grpc_supported:
-        if _grpc_channel:
+        if _grpc_options:
             import ${module_name}._grpc_stub_interpreter as _grpc_stub_interpreter
-            interpreter = _grpc_stub_interpreter.GrpcStubInterpreter(_grpc_channel)
+            import ${module_name}.GrpcOptions as GrpcOptions
```
DavidCurtiss, line 254: Unused import

generated/nifake/nifake/GrpcOptions.py
dmondrik: The server has the word "Device" in the name but I don't like including that in all the clients. It feels weird for DAQmx tasks or XNET sockets. Would `GrpcSessionOptions` work? Thoughts @DavidCurtiss ?

build/templates/session.py.mako
dmondrik: Is it unnecessary because the stub interpreter already imports it? The options are used here in this file (session) for the auto-close behavior for example.

generated/nifake/nifake/GrpcOptions.py
bkeryan: Sure. `GrpcSessionOptions` also describes its purpose more clearly.

build/templates/session.py.mako
DavidCurtiss: You don't ever directly use anything from inside that module. If you never use `module_name.Member` in your code, you don't need to import `module_name` at all.

You accept a `_grpc_options` parameter constructed by the user, but when you access it you'll just use duck typing to grab the fields you need. So even in the final version of this code, I don't think you'll ever need to import the module except in the examples and in tests.

src/nifake/metadata/functions.py
```diff
@@ -1326,6 +1326,14 @@
         'documentation': {
             'description': 'Creates a new IVI instrument driver session.'
         },
+        'method_templates': [
+            {
+                'documentation_filename': 'default_method',
+                'library_interpreter_filename': 'initialization_method',
```
dmondrik, line 1332: It is necessary to specify a different initialization method since it has parameters that are proto-only and not in the functions.py file. I added it manually here so you can see what the code generation looks like. Once the direction is approved, I'll go back to the hapigen metadata, add it there, then re-update this file using an official export.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
```diff
@@ -0,0 +1,24 @@
+<%page args="f, config, method_template"/>\
+<%
+    '''Renders a GrpcStubInterpreter method corresponding to the passed-in function metadata.'''
```
DavidCurtiss, line 3: nit: This docstring should probably be made specific to this file.

DavidCurtiss (approved):

generated/nifgen/nifgen/_grpc_stub_interpreter.py
```diff
@@ -8,16 +8,18 @@
...
 
 class GrpcStubInterpreter(object):
     '''Interpreter for interacting with a gRPC Stub class'''
 
-    def __init__(self, grpc_channel):
+    def __init__(self, grpc_options):
+        self._grpc_options = grpc_options
         self._lock = threading.RLock()
-        self._client = nifgen_grpc.NiFgenStub(grpc_channel)
+        self._client = nifgen_grpc.NiFgenStub(grpc_options.grpc_channel)
```
dmondrik, line 22: Note that the initialization method on line ~335 _does not use_ the session name and initialize behavior yet because I only modified the metadata for nifake. Once I have the direction approved, I will use actual exports for both.

src/nifake/metadata/functions.py
dmondrik: See [my ni-central PR](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/382959) where once submitted, that will allow me to update the metadata here with an official export.

generated/nifgen/nifgen/_grpc_stub_interpreter.py
dmondrik: See [my ni-central PR](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/382959) where once submitted, that will allow me to update the metadata here with an official export.

build/templates/_grpc_stub_interpreter.py.mako
```diff
@@ -18,6 +18,7 @@ import warnings
 from . import enums as enums
 % endif
 from . import errors as errors
+from . import grpc_session_options as grpc_session_options
```
ni-jfitzger, line 21: This import doesn't appear to be used.

build/templates/grpc_session_options.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+from enum import IntEnum
+
+
+class SessionInitializationBehavior(IntEnum):
+    UNSPECIFIED = 0
+    r'''
+    It is acceptable to either initialize a new session or attach to an existing one.
+    '''
+    INITIALIZE_NEW = 1
```
marcoskirsch, line 10: I question the value of `_NEW` and `_TO_EXISTING` in these enum values. Do they add anything?

build/templates/grpc_session_options.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+from enum import IntEnum
+
+
+class SessionInitializationBehavior(IntEnum):
+    UNSPECIFIED = 0
```
marcoskirsch, line 6: what does this really do?
Is this "attach if exists, otherwise initialize"?

build/templates/grpc_session_options.py
```diff
@@ -0,0 +1,47 @@
...
+                This is different from the resource name parameter many APIs take as a separate
+                parameter. Specifying a name makes it easy to share sessions across multiple gRPC
+                clients. You can use an empty string if you want to always initialize a new session
+                and then close it when the Python session goes out of scope.
+
+            initialization_behavior (enum): Specifies whether it is acceptable to initialize a new
+                session or attach to an existing one, or if only one of the behaviors is desired.
+                To attach to an existing session, you must specify a non-empty session name.
+
+            auto_close_grpc_session (bool): Specifies whether to have the NI gRPC Device Server close
```
marcoskirsch, line 41: I don't think we should expose this as auto-close.

A Python user does not need to think of a `nidmm.Session` object as wrapper around init/attach and close/detach.
They think of their objects as abstractions of a resource that gets cleaned up when the Python interpreter releases the object.

For that reason, I make two proposals:

1. Make it an enum with precise names TBD:
    1. AUTO (if you attached, detach; if you initialized, close)
    2. CLOSE_SERVER_SESSION
    3. DETACH_FROM_SERVER_SESSION

2. Make it so the session detaches if attached, closes if initalized. The client can always call detach() or close() as needed explicitly.

I think I like the second option best.

marcoskirsch (changes requested):

build/templates/grpc_session_options.py
dmondrik: Yes that's right. I had copied the terms from gRPC, and that term is required/meaningful there. But we should probably choose a better term for our API.

build/templates/grpc_session_options.py
dmondrik: I don't want a separate detach method which is meaningless on native sessions. I would prefer the first.

src/nifake/metadata/functions.py
dmondrik: Those went in. The latest exported metadata now specifies the new template method for initialize calls.

generated/nifgen/nifgen/_grpc_stub_interpreter.py
dmondrik: Resolved with the latest nifgen exports.

build/templates/grpc_session_options.py
dmondrik: Improved per our discussion.

build/templates/grpc_session_options.py
dmondrik: Improved per our discussion.

build/templates/grpc_session_options.py
dmondrik: Improved per our discussion.

generated/nifake/nifake/session.py
```diff
@@ -746,7 +746,8 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.close()
+        if self._interpreter._auto_close_session:
```
marcoskirsch, line 749: We really need to move away from the "auto-close" terminology, even if (or especially if) internal only.
It collides with a different concept in our drivers, which is about closing session 1 when a user initializes session 2 to the same resource.

Can you call this `_close_on_exit` or something like that?

marcoskirsch (approved):

generated/nifake/nifake/session.py
dmondrik: Good name change suggestion, I'll do that.

ni-jfitzger:
You have 2 checkboxes unmarked in your description. If they don't apply, you should strike them through and remove the TODO. Otherwise, you should take care of the TODOs.

PR #1834, opened 2022-10-24
Add NI-FGEN proto file and update generated `nifgen` files.

tdunkle:
- [X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

~- [ ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~

This change does not need to be customer-facing.

~- [ ] I've added tests applicable for this pull request~

Showing that the existing NI-FGEN system-level tests pass is sufficient.

### What does this Pull Request accomplish?

This pull request adds the NI-FGEN proto file to nimi-python, so that the NI-FGEN python API has a gRPC option enabled.
Note: This PR is a follow-on to #1830, which was accidentally closed

### List issues fixed by this Pull Request below, if any.

* [Internal NI bug AB#2160599](https://ni.visualstudio.com/DevCentral/_workitems/edit/2160599/)

### What testing has been done?

Installed python 3.7.9 and synced the nimi-python repo to a VM that has NI-FGEN 21.8 installed. Observed that I could build fgen with `tox -e py37-codegen -- nifgen`, and that I could run the system level tests with `tox -c generated\nifgen\tox-system_tests.ini`, and they all passed. Test output, for reference: [fgen_test_output.txt](https://github.com/ni/nimi-python/files/9778276/fgen_test_output.txt)

PR #1838, opened 2022-10-27
Fix system test modification pattern for `nifgen` to match other drivers and to run tests over gRPC.

tdunkle:

This pull request makes `nifgen`'s system-level tests use the same modification pattern that's been used in other drivers (e.g. NI-DMM, NI-Digital), in order to run tests both locally and over gRPC. Tests in the `SystemTests` class run both ways. Tests in the `TestLibrary` class only run locally, and tests in the `TestGrpc` class only run over gRPC.

For tests that can't run over gRPC (all of which are at the end of the `TestLibrary` class), there is a comment before each test on why it doesn't run over gRPC.

* [Internal NI bug AB#2160601](https://ni.visualstudio.com/DevCentral/_workitems/edit/2160601/)

Installed python 3.7.9 and synced the nimi-python repo to a VM that has NI-FGEN 21.8 installed. Observed that I could build fgen with `tox -e py37-codegen -- nifgen`, and that I could run the system level tests with `tox -c generated\nifgen\tox-system_tests.ini`, and they all passed. Whereas there were only 60 tests passing for #1834, there are now 112 tests passing. Test output, for reference: [fgen_test_output.txt](https://github.com/ni/nimi-python/files/9881212/fgen_test_output.txt)


tdunkle:
@DavidCurtiss FYI

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -26,538 +29,576 @@
...
+        session.write_waveform('foo', data)
+
+    # Test doesn't run over gRPC because numpy isn't supported by gRPC.
+    def test_write_named_waveform_numpy_array_int16(self, session):
+        data = numpy.ndarray(10000, dtype=numpy.int16)
+        data.fill(256)
+        session.allocate_named_waveform('foo', len(data))
+        session.write_waveform('foo', data)
+
+    # Test doesn't run over gRPC because the message SetAttributeViBooleanRequest's attribute is called attribute_value rather than attribute_value_raw.
```
DavidCurtiss, line 493: This failure should be fixed. For this one, note that [my hapigen PR yesterday](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/382146) should have the fix for it, so you should just need to grab the latest metadata.

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -26,538 +29,576 @@
...
-    data = numpy.ndarray(10000, dtype=numpy.int16)
-    data.fill(256)
-    session.allocate_named_waveform('foo', len(data))
-    session.write_waveform('foo', data)
-
-
-def test_write_waveform_wrong_type(session):
-    waveform_handle = session.allocate_waveform(100)
-    for data in invalid_waveforms:
+    # Test doesn't run over gRPC because the exception isn't caught from create_advanced_arb_sequence().
```
DavidCurtiss, line 505: Not sure what exception you're talking about, but it sounds like this is something we need to fix.

src/nifgen/system_tests/test_system_nifgen.py
tdunkle: I know you have another PR up to fix this in hapigen following our further discussion on it, but I went ahead and manually fixed it in the `.proto` and `functions.py` for now.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
```diff
@@ -0,0 +1,2 @@
+<%page args="f, config, method_template"/>\
+<%include file="/_grpc_stub_interpreter.py/default_method.py.mako" args="f=f, config=config, method_template=method_template" />\
```
DavidCurtiss, line 2: TODO: revert this file (and the next) before merging in Dan's (still pending) changes

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -26,538 +29,575 @@
...
-    assert type(session.create_waveform(data)) is int
-
-
 invalid_waveforms = ['Not waveform data',
                      numpy.zeros(100, dtype=numpy.uint16),
                      numpy.zeros(100, dtype=numpy.float32),
                      42,
                      3.14159, ]
 
+class SystemTests:
```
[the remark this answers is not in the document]
tdunkle: Sure thing! I just posted #1842 to create the SystemTests class with no functional changes first.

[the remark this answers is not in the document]
marcoskirsch: Thanks, merged!
I assume all the drivers need an equivalent change?

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
```diff
@@ -22,4 +22,4 @@
         self._close_on_exit = response.new_session_initialized
 % if return_statement:
         ${return_statement}
-% endif
+% endif
```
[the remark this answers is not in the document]
tdunkle: No. I've reverted these files so they're no longer in the PR.

src/nifgen/system_tests/test_system_nifgen.py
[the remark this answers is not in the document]
tdunkle: Yes. HW Config squad has a user story for each driver, so we'll continue to make the same change as needed.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
tdunkle: Done!

generated/nifgen/nifgen/session.py
```diff
@@ -1629,30 +1629,13 @@ def _create_waveform_f64_numpy(self, waveform_data_array):
 
         Example: :py:meth:`my_session._create_waveform_f64`
 
-        Args:
```
marcoskirsch, line 1632: this disappeared from docstring, but I don't think it should have. It's still an argument.

generated/nifgen/nifgen/session.py
```diff
@@ -1629,30 +1629,13 @@ def _create_waveform_f64_numpy(self, waveform_data_array):
...
-
-
         Returns:
             waveform_handle (int): The handle that identifies the new waveform. This handle is used later
                 when referring to this waveform.
 
         '''
         import numpy
 
-        if type(waveform_data_array) is not numpy.ndarray:
```
marcoskirsch, line 1650: This should have been caught by system tests. Are we missing coverage for these cases where the type is incorrect?

marcoskirsch (changes requested):

src/nifgen/system_tests/test_system_nifgen.py
tdunkle: I've looked into this, and it actually seems like the original test might not be working correctly even when run locally. I created a bug (internal work item 2207462) to look into this and fix it in a separate change.

marcoskirsch:
Travis CI checks failed.

>  The command "python tools/ensure_codegen_up_to_date.py" failed and exited with 1 during .
Your build has been stopped.

marcoskirsch (changes requested):
Please re-generate so Travis-CI checks all pass. We don't want to commit code that doesn't pass the checks.

generated/nifgen/nifgen/session.py
tdunkle: This was from a metadata issue involving the `numpy` field being removed. David has another PR up to fix that now.

tdunkle: Fixed!

generated/nifgen/nifgen/session.py
tdunkle: Yes, I tried commenting out this block and all of the tests still pass so we are missing coverage. I created #1847 

Marking this thread resolved, since this only affects local tests (outside the scope of the current PR) and a separate issue has been created.

generated/nifake/nifake/session.py
```diff
@@ -1357,7 +1357,7 @@ def one_input_function(self, a_number):
         self._interpreter.one_input_function(a_number)
 
     @ivi_synchronized
-    def parameters_are_multiple_types(self, a_boolean, an_int32, an_int64, an_int_enum, a_float, a_float_enum, a_string):
+    def parameters_are_multiple_types(self, a_boolean, an_int32, an_int64, an_int_enum, a_float, a_float_enum, string_size, a_string):
```
DavidCurtiss, line 1360: @marcoskirsch The `string_size` parameter got added because the `a_string` parameter used to have an `array` section that referenced `string_size` as its size but that `array` section was [removed](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/386190).

Should the `string_size` parameter be removed now, too? Should it stay, but have a calculated value? Was the removal of the `array` section incorrect??

Are there any MI driver APIs that accept a non-null-terminated string? If so, how would those be configured in the metadata?

src/nidmm/metadata/functions.py
```diff
@@ -1997,6 +2042,8 @@
                 'documentation': {
                     'description': 'Pass the value that you want to set the attribute to.'
                 },
+                'grpc_enum': 'NiDmmReal64AttributeValues',
+                'grpc_mapped_enum': 'NiDmmReal64AttributeValuesMapped',
```
DavidCurtiss, line 2046: @reckenro Why did we add both a 'grpc_enum' and a 'grpc_mapped_enum' here? Is this correct??

generated/nifake/nifake/session.py
DavidCurtiss: Alright, then I think it's inappropriate to still have the `string_size` parameter. Tom, please remove it from the metadata, and make whatever test updates are necessary (like not expecting `string_size` to be passed to the driver, if we have a test for that).

src/nidmm/metadata/functions.py
reckenro: Yes, well at least that aligns with what's in grpc-device. You can see that this param has both the `attribute_value` for the enum and `attribute_value_mapped` for the mapped_enum [here in the .proto](https://github.com/ni/grpc-device/blob/main/generated/nidmm/nidmm.proto#L1444).

For the historical context, my memory's hazy but I think the "regular" enums are where the underlying value can be represented directly in the .proto enum value (i.e., in a 32-bit int). Otherwise, it gets put in a *Mapped enum.

I don't know if that was necessarily the right way to do things but that's what that is. So in grpc-device most often you'll just see `'enum'` keys because the corresponding values can be represented naturally in the enum values but sometimes you'll see a `'mapped-enum'` key and that's what that signifies.

tdunkle:
Closing this PR, because it is no longer needed. The relevant work was already submitted in #1842, #1834, and #1861. At this time, there is only one test case that we still need to enable for gRPC, and @reckenro is currently working on a change to accomplish that.

PR #1840, opened 2022-10-29
Add proto and tests for NI-Digital

DavidCurtiss:

Adds nidigitalpattern.proto, fixes issues, and adds system tests for it.

None

Added tests


dmondrik:
I'd like to see [my PR for gRPC session options](https://github.com/ni/nimi-python/pull/1835) to go in first.

dmondrik (commented):
It appears I had pending comments that I didn't publish. Oops.

build/templates/session.py.mako
```diff
@@ -272,19 +272,19 @@ if grpc_supported:
 % endfor
 
         # Call specified init function
-        # Note that _library_interpreter sets _${config['session_handle_parameter_name']} to 0 in its constructor, so that if
+        # Note that _library_interpreter clears the session handle in its constructor, so that if
```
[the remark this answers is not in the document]
DavidCurtiss: Fixed. Now it just says `_interpreter`, not `_library_interpreter`.

generated/nidigital/nidigital/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,654 @@
...
+        response = self._invoke(
+            self._client.GetTimeSetPeriod,
+            grpc_types.GetTimeSetPeriodRequest(vi=self._vi, time_set_name=time_set_name),
+        )
+        return response.period
+
+    def init_with_options(self, resource_name, id_query, reset_device, option_string):  # noqa: N802
+        response = self._invoke(
+            self._client.InitWithOptions,
+            grpc_types.InitWithOptionsRequest(resource_name=resource_name, id_query=id_query, reset_device=reset_device, option_string=option_string),
```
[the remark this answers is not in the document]
DavidCurtiss: Done

dmondrik (approved):
LGTM

build/templates/_library_interpreter.py.mako
```diff
@@ -78,7 +78,13 @@ class LibraryInterpreter(object):
         # Initialize _${config['session_handle_parameter_name']} to 0 for now.
         # Session will directly update it once the driver runtime init function has been called and
         # we have a valid session handle.
-        self._${config['session_handle_parameter_name']} = 0
+        self._set_session_handle()
+
+    def _set_session_handle(self, value=0):
```
marcoskirsch, line 83: I'm not sure how pythonic it is to have these trivial private getter/setter methods.
We certainly don't use them throughout nimi-python. What prompted you to implement?

src/nidigital/templates/_grpc_stub_interpreter.py/fancy_fetch_capture_waveform.py.mako
```diff
@@ -0,0 +1,8 @@
+
+    def fetch_capture_waveform(self, site_list, waveform_name, samples_to_read, timeout):  # noqa: N802
+        response = self._invoke(
+            self._client.FetchCaptureWaveformU32,
+            grpc_types.FetchCaptureWaveformU32Request(vi=self._vi, site_list=site_list, waveform_name=waveform_name, samples_to_read=samples_to_read, timeout=timeout),
+        )
+        ## Modified (vs. default generated code) to return bytes() and both size outputs
+        return bytes(response.data), response.actual_num_waveforms, response.actual_samples_per_waveform
```
marcoskirsch, line 8: nimi-python has a concept of "converters" in which you can tell metadata to convert an internal type to a different type in the public API.

Did you try to use that?

marcoskirsch (changes requested):

build/templates/_library_interpreter.py.mako
DavidCurtiss: In session.py's `close()` method, it was setting `self._interpreter._vi = 0`, but for gRPC it now needs to be an object. (This was failing one of NI-Digital's system tests for the `close()` method. I looked into adding an NI-FAKE test, but there's no good way to unit test this.) So I added `_set_session_handle()` with a default value, so that `close()` can just call `self._interpreter._set_session_handle()` and not have to copy the gRPC-specific default value to session.py.

I agree that it's not pythonic to call someone else's private methods, but it's also not pythonic to modify someone else's private members like we were doing before. I could make them public methods, if you want, since the class itself is already internal. All we'd have to worry about is possibly having to rename it later if some driver added a method with a conflicting name.

src/nidigital/templates/_grpc_stub_interpreter.py/fancy_fetch_capture_waveform.py.mako
DavidCurtiss: The bigger issue here is needing to return both size outputs. I could add the converter, to make the metadata more correct, but I'd still need the custom method anyway.

build/templates/_library_interpreter.py.mako
marcoskirsch: Ok, so the only thing we get by having a method wrapping the property is the ability to have a default? I can buy that.
If the method is intended to be used outside of the (internal) class then I do think it should be "public".

src/nidigital/templates/_grpc_stub_interpreter.py/fancy_fetch_capture_waveform.py.mako
marcoskirsch: I see. I wouldn't bother then. Thanks.

generated/nidigital/nidigital/grpc_session_options.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
+from enum import IntEnum
+
+
```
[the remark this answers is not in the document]
DavidCurtiss: Done

build/templates/_library_interpreter.py.mako
DavidCurtiss: Ok, made them public.

src/nitclk/templates/session.py.mako
```diff
@@ -49,15 +49,15 @@ helper.add_attribute_rep_cap_tip(attributes[attribute], config)
 %   endif
 % endfor
 
-    def __init__(self, ${config['session_handle_parameter_name']}, encoding='windows-1251'):
+    def __init__(self, session_handle, encoding='windows-1251'):
```
marcoskirsch, line 52: Different APIs use different names for the concept of a session handle.

This change makes everything use the same. NI-ModInst uses a different name. NI-TClk uses a different concept altogether. I think this need to be reverted to the way it was (it's also a change in public API).

marcoskirsch (changes requested):

src/nitclk/templates/session.py.mako
DavidCurtiss: Updated NI-ModInst's and NI-TClk's session.py to mostly revert this rename.

generated/nidigital/nidigital/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,655 @@
...
+class GrpcStubInterpreter(object):
+    '''Interpreter for interacting with a gRPC Stub class'''
+
+    def __init__(self, grpc_options):
+        self._grpc_options = grpc_options
+        self._lock = threading.RLock()
+        self._client = nidigital_grpc.NiDigitalStub(grpc_options.grpc_channel)
+        self.set_session_handle()
+
+    def set_session_handle(self, value=session_grpc_types.Session()):
```
ni-jfitzger, line 27: Where is this class defined? I don't see it in session_pb2.

dmondrik (approved):
LGTM

generated/nidigital/nidigital/_grpc_stub_interpreter.py
DavidCurtiss: The _pb2 files are weird. Most of the stuff is instantiated from the data in the magic DESCRIPTOR string. You can see the Session definition in the .proto file, [here](https://github.com/ni/nimi-python/blob/726fad928e831f4b355919ea746d9ac89ca04e9c/src/shared_protos/session.proto#L36).

generated/nimodinst/nimodinst/session.py
```diff
@@ -146,11 +146,12 @@ def __init__(self, driver):
         self._item_count = 0
         self._current_item = 0
         self._interpreter = _library_interpreter.LibraryInterpreter('windows-1251')
-        # Note that _library_interpreter sets _handle to 0 in its constructor, so that if
+        # Note that _library_interpreter clears the session handle in its constructor, so that if
```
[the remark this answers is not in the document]
DavidCurtiss: NI-ModInst and NI-TClk have their own session.py.mako files. And since they don't yet support gRPC, using `_library_interpreter` is still fine, though it may be best to change it anyway for consistency.

build/rules.mak
```diff
@@ -61,7 +61,6 @@ $(MODULE_DIR)/%: %.mako $(BUILD_HELPER_SCRIPTS) $(METADATA_FILES)
 
 $(MODULE_DIR)/%_pb2.py: %.proto
 	$(call trace_to_console, "Generating",$@ and $(notdir $*)_pb2_grpc.py)
-	$(_hide_cmds)$(call log_command,mkdir -p $(OUTPUT_DIR)/build/proto/$(dir $*))
```
ni-jfitzger, line 64: Was this just cruft, leftover from other changes?

DavidCurtiss: Yes

build/templates/_library_interpreter.py.mako
```diff
@@ -78,7 +78,13 @@ class LibraryInterpreter(object):
         # Initialize _${config['session_handle_parameter_name']} to 0 for now.
         # Session will directly update it once the driver runtime init function has been called and
         # we have a valid session handle.
-        self._${config['session_handle_parameter_name']} = 0
+        self.set_session_handle()
+
+    def set_session_handle(self, value=0):
+        self._${config['session_handle_parameter_name']} = value
```
ni-jfitzger, line 84: Does this still default initialize to 0? I can't tell. Does it need to?

ni-jfitzger: Sorry, wrong file.

build/templates/_grpc_stub_interpreter.py.mako
```diff
@@ -33,7 +35,13 @@ class GrpcStubInterpreter(object):
         self._grpc_options = grpc_options
         self._lock = threading.RLock()
         self._client = ${module_name}_grpc.${service_class_prefix}Stub(grpc_options.grpc_channel)
-        self._${config['session_handle_parameter_name']} = 0
+        self.set_session_handle()
+
+    def set_session_handle(self, value=session_grpc_types.Session()):
+        self._${config['session_handle_parameter_name']} = value
```
ni-jfitzger, line 41: Does this still default initialize to 0? I can't tell. Does it need to?

DavidCurtiss: Yes, the id in the default-constructed session_grpc_types.Session() object is 0. We test this on line 88 of [this test](https://github.com/ni/nimi-python/blob/9201bc8259777992b502427c926c15bb901115e1/src/nifake/unit_tests/test_grpc.py#L84).

generated/nidcpower/nidcpower/session.py
```diff
@@ -6911,11 +6911,11 @@ def __init__(self, resource_name, channels=None, reset=False, options={}, indepe
...
         # Call specified init function
-        # Note that _library_interpreter sets _vi to 0 in its constructor, so that if
+        # Note that _interpreter clears the session handle in its constructor, so that if
         # _fancy_initialize fails, the error handler can reference it.
-        # And then once _fancy_initialize succeeds, we can update _library_interpreter._vi
-        # with the actual session handle.
-        self._interpreter._vi = self._fancy_initialize(resource_name, channels, reset, options, independent_channels)
+        # And then once _fancy_initialize succeeds, we can call this again with the
+        # actual session handle.
+        self._interpreter.set_session_handle(self._fancy_initialize(resource_name, channels, reset, options, independent_channels))
```
ni-jfitzger, line 6918: This change was to stop breaking encapsulation, right?
I like it overall. It may even be worth calling out in the PR description.

The comment now feels a little murky, though. `we can call this again`. Is it referring to the specified init function that the first line of the comment mentions calling or to set_session_handle? From context clues, I would say the latter, but the comment could be clearer by saying `we can call set_session_handle again`.

DavidCurtiss: Not entirely, no. It was primarily made so that Session.close() can reset the session handle without having to know that it's an int for LibraryInterpreter but a session_pb2.Session for GrpcStubInterpreter. With the udpated code, now Session.close just calls set_session_handle() without specifying any handle, so that the default value gets used to reset it.

I'll clean up the comment.

DavidCurtiss: Done

generated/nidcpower/nidcpower/session.py
```diff
@@ -6911,11 +6911,11 @@ def __init__(self, resource_name, channels=None, reset=False, options={}, indepe
         options = _converters.convert_init_with_options_dictionary(options)
 
         # Call specified init function
-        # Note that _library_interpreter sets _vi to 0 in its constructor, so that if
+        # Note that _interpreter clears the session handle in its constructor, so that if
```
ni-jfitzger, line 6914: I think this part needs some more wordsmithing. How do you clear something before it's been set? And how do you reference a value that you've cleared? Maybe it would be better to say default-initializes? @marcoskirsch may be able to help us out when he returns. I don't think we need to hold up this PR for it.

src/nidigital/system_tests/test_system_nidigital.py
```diff
@@ -12,463 +12,427 @@
...
-    if source_waveform_wrong_type == tuple:
-        source_waveform = ([i for i in range(num_samples)], [i for i in reversed(range(num_samples))])
-    elif source_waveform_wrong_type == int:
-        source_waveform = num_samples
-    elif source_waveform_wrong_type == str:
-        source_waveform = {
-            str(1): [str(i) for i in range(num_samples)],
-            str(0): [str(i) for i in reversed(range(num_samples))]}
-    else:
-        assert False, "Invalid source waveform data type: {}".format(source_waveform_type)
```
ni-jfitzger, line 372: Wow, how was this even passing, before?

DavidCurtiss: The test itself passed fine because this is only in the string of an `assert False`, so it was never actually evaluated. But it was able to pass by flake8 because there's a `def source_waveform_type` in this file. (But now that that `def` is inside the class, you'd have to update it to `self.source_waveform_type` for it to get past flake8.)

src/nidigital/system_tests/test_system_nidigital.py
```diff
@@ -477,925 +425,949 @@ def test_history_ram_cycle_information_string():
...
+    def test_close(self, grpc_channel):
+        grpc_options = nidigital.GrpcSessionOptions(grpc_channel, '')
+        session = nidigital.Session(resource_name=','.join(instruments), options='Simulate=1, DriverSetup=Model:6570', _grpc_options=grpc_options)
+        session.vil = 1
+        session.close()
+        try:
+            session.vil = 1
+            assert False
+        except nidigital.Error as e:
+            assert e.code == -1074130544
```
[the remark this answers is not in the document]
ni-jfitzger: @DavidCurtiss intends to handle this is a separate PR.

generated/nidcpower/nidcpower/session.py
ni-jfitzger: @DavidCurtiss can handle this is a separate PR. I'm going to merge.

ni-jfitzger (approved):

ni-jfitzger:
FYI @marcoskirsch merging, to get this off our plate. I feel all of your comments have been properly addressed and I have reviewed and am comfortable with the changes.

PR #1843, opened 2022-11-01
Fix NI-FAKE unit tests

DavidCurtiss:

Fixes the NI-FAKE unit tests, which were broken with the addition of GrpcSessionOptions.

None

Ran the tests

### DIFF
--- a/generated/nifgen/nifgen/session.py
+++ b/generated/nifgen/nifgen/session.py
@@ -3083,6 +3083,7 @@ class Session(_SessionBase):
             session (nifgen.Session): A session object representing the device.
 
         '''
+        self._grpc_options = _grpc_options
         if _grpc_options:
             import nifgen._grpc_stub_interpreter as _grpc_stub_interpreter
             interpreter = _grpc_stub_interpreter.GrpcStubInterpreter(_grpc_options)
@@ -3135,7 +3136,7 @@ class Session(_SessionBase):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        if self._interpreter._close_on_exit:
+        if not self._grpc_options or self._interpreter._close_on_exit:
             self.close()
 
     def initiate(self):
--- a/generated/nifgen/nifgen/session.py
+++ b/generated/nifgen/nifgen/session.py
@@ -3083,7 +3083,6 @@ class Session(_SessionBase):
             session (nifgen.Session): A session object representing the device.
 
         '''
-        self._grpc_options = _grpc_options
         if _grpc_options:
             import nifgen._grpc_stub_interpreter as _grpc_stub_interpreter
             interpreter = _grpc_stub_interpreter.GrpcStubInterpreter(_grpc_options)
@@ -3136,7 +3135,7 @@ class Session(_SessionBase):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        if not self._grpc_options or self._interpreter._close_on_exit:
+        if self._interpreter._close_on_exit:
             self.close()
 
     def initiate(self):
--- a/src/nifake/unit_tests/test_grpc.py
+++ b/src/nifake/unit_tests/test_grpc.py
@@ -58,8 +58,8 @@ class TestGrpcStubInterpreter(object):
                     error_func = _mock_helper.MockFunctionCallError(f)
                     setattr(self, f, MagicMock(spec_set=self._sample_func, side_effect=error_func))
 
-        def __call__(self, session_options):
-            self._session_options = session_options
+        def __call__(self, grpc_channel):
+            self._grpc_channel = grpc_channel
             return self
 
     def setup_method(self, method):
@@ -83,7 +83,7 @@ class TestGrpcStubInterpreter(object):
         interpreter = nifake._grpc_stub_interpreter.GrpcStubInterpreter(session_options)
         assert interpreter._client is self.patched_grpc_stub
         assert interpreter._vi == 0
-        assert self.patched_grpc_stub._session_options.grpc_channel is grpc_channel
+        assert self.patched_grpc_stub._grpc_channel is grpc_channel
         interpreter._vi = GRPC_SESSION_OBJECT_FOR_TEST
         return interpreter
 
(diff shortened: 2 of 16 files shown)


generated/nifgen/nifgen/session.py
```diff
@@ -3083,6 +3083,7 @@ def __init__(self, resource_name, channel_name=None, reset_device=False, options
             session (nifgen.Session): A session object representing the device.
 
         '''
+        self._grpc_options = _grpc_options
```
dmondrik, line 3086: This seems really odd to modify production code this way.

src/nifake/unit_tests/test_session.py
```diff
@@ -832,7 +833,7 @@ def teardown_method(self, method):
     # Session management
 
     def test_init_with_options_and_close(self):
-        session = nifake.Session('dev1', _grpc_channel=object())
+        session = nifake.Session('dev1', _grpc_options=nifake.GrpcSessionOptions(object(), ''))
```
dmondrik, line 836: You could even pass `None` instead of `object()` for the first session option.

generated/nifgen/nifgen/session.py
dmondrik: If we don't want to enforce all implementations to have it, shouldn't we just use `hasattr(_close_on_exit)`?

DavidCurtiss: This `self._grpc_options` is needed for the code in `__exit__`, so that it only checks for `self._interpreter._close_on_exit` if `self._interpreter` is a GrpcStubInterpreter -- because only GrpcStubInterpreter has a `_close_on_exit` member. Some alternatives:
1. I could instead change `__exit__` to check `isinstance(self._interpreter, GrpcStubInterpreter)`, or 
2. I could change the `_close_on_exit` check to be `getattr(self._interpreter, '_close_on_exit', False)`, or
3. I could add `self._close_on_exit` to the LibraryInterpreter constructor. (Or similar, I could add a close_on_exit function to both Interpreters, and for LibraryInterpreter always return True.)

I'm open to any of those options, I just chose what I saw to be the least disruptive.

src/nifake/unit_tests/test_session.py
DavidCurtiss: Yeah, now that we no longer check if the grpc_channel is truthy, I could change that. I think it's fine this way, too, though.

generated/nifgen/nifgen/session.py
dmondrik: My PR also added `_close_on_exit = True` to [the library interpreter](https://github.com/ni/nimi-python/blob/aa77b9334344dcb479af2a34ce155973192adca7/generated/nifgen/nifgen/_library_interpreter.py#L490). Why not propagate it elsewhere?

DavidCurtiss: Just saw your `hasattr` comment. That would be similar to option 2, and that would be fine, too.

marcoskirsch (approved):

generated/nifgen/nifgen/session.py
DavidCurtiss: Ah, I see now that it is indeed on the library interpreter, too. The problem is just that it's only set after init is called. But if init fails, session tries to query it anyway. So that's a different sort of bug. Okay, I'll fix it.

dmondrik: Ah, but here's why that shouldn't be an issue. The `__exit__` call comes only after `__enter__` which can only be done by putting the new session into a context manager after `__init__` has succeeded. If the object isn't initialized, it won't be added to the context manager to start with.

DavidCurtiss: Okay, makes sense. Done.

src/nifake/unit_tests/test_session.py
```diff
@@ -818,7 +822,12 @@ def setup_method(self, method):
         self.tclk_patched_library_singleton_get = patch('nitclk._library_interpreter._library_singleton.get', return_value=None)
         self.tclk_patched_library_singleton_get.start()
 
-        self.patched_grpc_interpreter.init_with_options.side_effect = [GRPC_SESSION_OBJECT_FOR_TEST]
+        def interpreter_init(*args, **kwargs):
+            self.patched_grpc_interpreter._close_on_exit = True
+            return GRPC_SESSION_OBJECT_FOR_TEST
+
+        self.patched_grpc_interpreter.init_with_options.side_effect = interpreter_init
```
ni-jfitzger, line 829: I'm guessing we'll never be using this for attaching to an existing session?

src/nifake/unit_tests/test_grpc.py
```diff
@@ -58,8 +58,8 @@ def __init__(self):
                     error_func = _mock_helper.MockFunctionCallError(f)
                     setattr(self, f, MagicMock(spec_set=self._sample_func, side_effect=error_func))
 
-        def __call__(self, session_options):
-            self._session_options = session_options
```
ni-jfitzger, line 62: I'm having trouble understanding this change.

src/nifake/unit_tests/test_session.py
DavidCurtiss: For the tests, yeah, we set True unconditionally. That may be a testing gap, I'm not sure. This PR is just focused on getting the tests _passing_, though.

src/nifake/unit_tests/test_grpc.py
DavidCurtiss: In Dan's PR, he [changed this](https://github.com/ni/nimi-python/commit/380fb5131b19859fe794b016f8a7da79246de6d2#diff-899f0ea835ab662dd4436c04e6f4de2ccc24050445a2871ef3bda7256bcee85b) from grpc_channel to session_options, as he started (but never finished) updating the tests. But it was an incorrect change; the Stub takes a grpc_channel. His PR was merged while tests were still failing. It was failing on line 86. So here I'm reverting that, and now the tests pass.

ni-jfitzger: So I guess you're creating self._grpc_channel (which normally doesn't exist in the class object) in order to make assertions about it later?

ni-jfitzger (approved):

src/nifake/unit_tests/test_grpc.py
DavidCurtiss: Correct

PR #1844, opened 2022-11-02
Add module constant string for use with MeasurementLink

dmondrik:

~~TODO: Check the above box with an 'x' if considered if there were any and then documented client facing changes in [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md). Strike through if this is not a relevant client facing change.~~

~~TODO: Check the above box with an 'x' indicating you have considered and added any applicable system or unit tests. Strike through if you considered and decided additional tests were not warranted.~~

This exposes a string constant that specifies the package+server instance this API uses. It is important for the MeasurementLink customer to get information about the server location (which will usually be on a dynamic port) before creating a gRPC channel.

This was requested by Jason Reding. I'm not sure whether there's a work item. If one is mentioned, I'll add it before it gets to submission.

N/A

Example test with local build:
```
>>> import nifgen
>>> print(nifgen.GRPC_SERVICE_INTERFACE)
nifgen_grpc.NiFgen
```


generated/nifake/nifake/grpc_session_options.py
```diff
@@ -1,7 +1,14 @@
 # -*- coding: utf-8 -*-
+# This file was generated
+
 from enum import IntEnum
 
 
+# This constant specifies the gRPC package and service used by this API.
+# MeasurementLink customers should pass this value to the discovery manager to resolve the server instance that provides this interface.
+GRPC_SERVICE_INTERFACE = 'nifake_grpc.NiFake'
```
dmondrik, line 9: This was the other way we were considering it:
```
class GrpcStringConstants:
  SERVICE_INSTANCE='nifake_grpc.NiFake'
  r'''Doc string'''
```
A constant of any kind in python does not support doc strings properly. Some tools like Sphinx support it though.

The [Python conventions page](https://peps.python.org/pep-0008/#constants) says the way I've got it is more desirable.

jasonmreding: The constant values themselves look correct. I'll let others who are more proficient with Python speak to the style/method used to expose the constants in Python.

jasonmreding (approved):

generated/nifake/nifake/grpc_session_options.py
marcoskirsch: Does this belong in nimi-python at all? It seems backwards for these APIs to know about MeasurementLink.

dmondrik: Yes Marcos it's important for those customers (the only ones we care about for gRPC-python in this release) to succeed. There's no other reasonable place for them to get this string since it may change from release to release. For example, it's possible the next release could require `nifake_grpc.NiFake-v2` and they'd have no other way of knowing it.

bkeryan: FYI, a future release could also switch to a different package/service naming convention like `ni.fake.v2.NiFakeService`. 😉

https://github.com/uber/prototool/blob/dev/style/README.md#package-naming
https://docs.buf.build/best-practices/style-guide#services

build/templates/grpc_session_options.py.mako
```diff
@@ -1,7 +1,17 @@
...
+# This file was generated
+<%
+    config = template_parameters['metadata'].config
+%>\
+
 from enum import IntEnum
 
 
+# This constant specifies the gRPC package and service used by this API.
+# MeasurementLink customers should pass this value to the discovery manager to resolve the server instance that provides this interface.
```
bkeryan, line 11: There is no "discovery manager". Use "discovery service".

marcoskirsch: if this gets a more official name, even "NI Discovery Service" we will need to remember to come back and update.

build/templates/grpc_session_options.py.mako
```diff
@@ -1,7 +1,17 @@
...
+GRPC_SERVICE_INTERFACE = '${config["module_name"]}_grpc.${config["grpc_service_class_prefix"]}'
```
marcoskirsch, line 12:
```suggestion
GRPC_SERVICE_INTERFACE_NAME = '${config["module_name"]}_grpc.${config["grpc_service_class_prefix"]}'
```

marcoskirsch (changes requested):

build/templates/grpc_session_options.py.mako
bkeryan: The documentation currently calls it "the MeasurementLink™ discovery service".

dmondrik: Thanks, I knew I was going to get some part of the terminology wrong.

marcoskirsch (approved):

PR #1845, opened 2022-11-03
Add proto and tests for NI-SCOPE

reckenro:

Adds niscope.proto, and system tests for it.

Related tasks:
[AB#2139819](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2139819)
[AB#2139816](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2139816)

In the process, also fixed test_reset_with_defaults() to call the correct method and use a single-instrument session (method does not support multi-instrument sessions).

None

Ran NI-SCOPE system tests on machine with required prerequsites installed.

Also ran on a machine where the ni_grpc_device_server was not installed (but other NI software is), to see that it skips the grpc versions of the NI-SCOPE system tests.


src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
 import hightime
 import math
 import niscope
 import numpy
 import os
+import pathlib
 import pytest
+import subprocess
 import tempfile
-
```
reckenro, line 10: To get a decent diff for this file, I'd recommend hiding whitespace changes:

image

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
-        assert len(waveforms[i].samples) == test_record_length
-
-
-@pytest.fixture(params=[(1000, 1000), (2000, 2000), (3000, 2000)], ids=["less_than_actual", "equal_to_actual", "greater_than_actual"])
-def measurement_wfm_length(request):
-    MeasWfmLength = collections.namedtuple('MeasurementWaveformLength', ['passed_in', 'expected'])
-    return MeasWfmLength(passed_in=request.param[0], expected=request.param[1])
-
-
-def test_fetch_array_measurement(multi_instrument_session, measurement_wfm_length):
```
reckenro, line 127: `test_fetch_array_measurement` and `test_fetch_array_measurement_defaults` are just in the SystemTests class below. There's a TODO above them to support them on grpc but right now they're broken so they're only tested against the `_library_interpreter` for now.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
-    for i in range(len(waveforms)):
-        assert len(waveforms[i].samples) == test_record_length
-
-
-@pytest.fixture(params=[numpy.int8, numpy.int16, numpy.int32, numpy.float64])
-def fetch_waveform_type(request):
-    return request.param
-
-
-def test_fetch_into(multi_instrument_session, fetch_waveform_type):
```
reckenro, line 183: `fetch_into` isn't supported on grpc-device, so this test was moved to just be in the `TestLibrary` class below.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
-    multi_instrument_session.configure_vertical(test_voltage, niscope.VerticalCoupling.AC)
-    multi_instrument_session.configure_horizontal_timing(50000000, test_record_length, 50.0, test_num_records, True)
-    with multi_instrument_session.initiate():
-        measurement_stat = multi_instrument_session.channels[test_channels].fetch_measurement_stats(niscope.enums.ScalarMeasurement.NO_MEASUREMENT)
-
-    assert isinstance(measurement_stat[0].__str__(), str)
-    assert isinstance(measurement_stat[0].__repr__(), str)
-
-
-def test_get_self_cal_last_date_time(single_instrument_session):
```
reckenro, line 318: These four tests were copied as is to `TestLibrary` class. These methods aren't included in the proto file so there are four versions of these tests in `TestGrpc` where each expects a `NotImplementedError` to be raised.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
-    deault_meas_time_histogram_high_time = multi_instrument_session.meas_time_histogram_high_time
-    assert deault_meas_time_histogram_high_time == hightime.timedelta(microseconds=500)
-    multi_instrument_session.meas_time_histogram_high_time = hightime.timedelta(microseconds=1000)
-    non_default_meas_time_histogram_high_time = multi_instrument_session.meas_time_histogram_high_time
-    assert non_default_meas_time_histogram_high_time == hightime.timedelta(microseconds=1000)
-    multi_instrument_session.reset_device()
-    assert multi_instrument_session.meas_time_histogram_high_time == hightime.timedelta(microseconds=500)
-
-
-def test_error_message():
```
reckenro, line 381: Two different versions of `test_error_message` due to the different setup. One in `TestLibrary` and one in `TestGrpc`

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
-
-def test_disable(multi_instrument_session):
-    assert multi_instrument_session.allow_more_records_than_memory is False
-    multi_instrument_session.allow_more_records_than_memory = True
-    multi_instrument_session.disable()
-    assert multi_instrument_session.allow_more_records_than_memory is False
-
-
-# Basic configuration tests
-def test_configure_ref_levels(single_instrument_session):
```
reckenro, line 454: `configure_ref_levels` isn't included in the .proto file, so two versions of `test_configure_ref_levels` exist. The original version was copied as is to `TestLibrary` while the new version in `TestGrpc` asserts that a `NotImplementedError` is raised.

src/niscope/templates/session.py/fancy_fetch.py.mako
```diff
@@ -21,7 +21,10 @@
 
         wfm, wfm_info = self._${f['python_name']}(num_samples, timeout)
 
-        mv = memoryview(wfm)
+        if isinstance(wfm, array.ArrayType):
```
reckenro, line 24: This update was to keep the `memoryview` instantiation for the `wfm` when it was coming from the `_library_interpreter`. But when coming from grpc, we should be able to just pass the returned `wfm` forward to the `_populate_samples_info` method.

build/templates/_grpc_stub_interpreter.py/default_method.py.mako
```diff
@@ -12,13 +12,18 @@
     if return_statement == 'return':
         return_statement = None
     capture_response = 'response = ' if return_statement else ''
+    included_in_proto = f.get('included_in_proto', True)
```
reckenro, line 15: This is a new field generated in the nimi-python functions.py metadata to indicate if the function is accessible through grpc-device / the proto file.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
+    def test_configure_ref_levels(self, single_instrument_session):
+        single_instrument_session._configure_ref_levels()
+        assert 90.0 == single_instrument_session.meas_chan_high_ref_level
+
+    @pytest.fixture(params=[(1000, 1000), (2000, 2000), (3000, 2000)], ids=["less_than_actual", "equal_to_actual", "greater_than_actual"])
+    def measurement_wfm_length(self, request):
+        MeasWfmLength = collections.namedtuple('MeasurementWaveformLength', ['passed_in', 'expected'])
+        return MeasWfmLength(passed_in=request.param[0], expected=request.param[1])
+
+    # TODO(RyanEckenrode): AB#2207665: Need to investigate how to get FetchArrayMeasurement working for the grpc versions of these tests
```
reckenro, line 440: Note: Added a TODO here and the task number. Basically, there's a difference in the FetchArrayMeasurement method params between nimi-python and grpc-device. In grpc-device, currently we don't have the `measurement_wfm_length` as an input in the .proto file. We just calculate it server-side by calling `ActualMeasWfmSize`. But on nimi-python, it accepts the `measurement_wfm_size` from the client.

Made the TODO task to investigate this and figure out what should be done. Part of that Task is to enable these two `fetch_array_measurement` tests for gRPC as well.

generated/niscope/niscope/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,425 @@
...
+        self._lock.release()
+
+    def close(self):  # noqa: N802
+        self._invoke(
+            self._client.Close,
+            grpc_types.CloseRequest(vi=self._vi),
+        )
+
+    def error_message(self, error_code):  # noqa: N802
+        raise NotImplementedError('error_message is not supported over gRPC')
```
dmondrik, line 412: See line 64 above. This method absolutely must be callable to get the message for a warning. The code is as I'd expect for the other gRPC-enabled APIs (nidmm, nifgen, niswitch).

generated/niscope/niscope/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,425 @@
...
+        self._invoke(
+            self._client.AutoSetup,
+            grpc_types.AutoSetupRequest(vi=self._vi),
+        )
+
+    def cal_fetch_date(self, which_one):  # noqa: N802
+        raise NotImplementedError('cal_fetch_date is not supported over gRPC')
+
+    def cal_fetch_temperature(self, which_one):  # noqa: N802
+        raise NotImplementedError('cal_fetch_temperature is not supported over gRPC')
```
dmondrik, line 113: It doesn't matter for this PR, because if it's not implemented yet, this error case makes sense. But do you know why some of these APIs aren't there for gRPC? Are these new (such that we should go back and add them to gRPC) or old (obsolete, and we shouldn't worry about it)?

dmondrik (changes requested):

generated/niscope/niscope/_grpc_stub_interpreter.py
dmondrik: The method is in gRPC-device in the proto file as of the niscope 23.0.0d58 export. Has something changed?

reckenro: Ah, good catch. And this is a bit unfortunate, in grpc-device we expose and use 'GetErrorMessage' while nimi-python uses the obsolete 'ErrorMessage' method instead which we don't use in grpc-device.

They're pretty similar except `GetErrorMessage` uses ivi-dance to determine the buffer size for the message while `ErrorMessage` uses a fixed 256 size for the message.

marcoskirsch: The reason nimi-python uses the "bad" function for messages, iirc, is that not all drivers implemented the "good" one so we had to fall back to minimum common denominator.

We could revisit this if needed. We can implement in driver runtimes that don't have them or write per-driver code when needed.

reckenro: I'm wondering here if it would be terrible for the _grpc_stub_interpreter.py for this `error_message` definition to just call grpc-device's `GetErrorMessage`. The inputs and outputs are the same, just how it calculates the error message size is different and an implementation detail. You can see nimi-python's metadata [here](https://github.com/ni/nimi-python/blob/master/src/niscope/metadata/functions.py#L3169) and grpc-device's [here for GetErrorMessage](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/niscope/functions.py#L2400)

So, I guess I could have a custom template or something that gets invoked for this NI-SCOPE to generate the call to grpc-device's `GetErrorMessage`. I guess I could also look at overriding the metadata so `grpc_name' is just 'GetErrorMessage' and 'included_in_proto' evaluates to true.

Either way, does that sound reasonable? Otherwise, I guess I'd look at exposing `GetErrorMessage` for nimi-python and using that instead of `error_message` but that seems like a more daunting change.

generated/niscope/niscope/_grpc_stub_interpreter.py
ni-jfitzger: We have not added any entrypoints to the NI-SCOPE API this year, as far as I know.

ni-jfitzger: I don't know of any reason why either of these cal_ methods shouldn't be supported by gRPC.

ni-jfitzger: Perhaps it comes from a misconception that an external cal session is needed to call them?

generated/niscope/niscope/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,425 @@
...
+        return response.waveform, [waveform_info.WaveformInfo(x) for x in response.wfm_info]
+
+    def reset_device(self):  # noqa: N802
+        self._invoke(
+            self._client.ResetDevice,
+            grpc_types.ResetDeviceRequest(vi=self._vi),
+        )
+
+    def reset_with_defaults(self):  # noqa: N802
+        raise NotImplementedError('reset_with_defaults is not supported over gRPC')
```
ni-jfitzger, line 364: Do you know why this method isn't supported?

generated/niscope/niscope/_grpc_stub_interpreter.py
reckenro: Ah, I found it in an old [deleted changes.md](https://github.com/ni/grpc-device/pull/691/files#diff-a994c5c8a10fa7b839bf24e2c7468566970a29c032be3dd4a5f16c73e640ac4a) we had originally in grpc-device to list changes from the nimi-python metadata. Looks like for these we didn't add them because they weren't in niScope.h but instead in niScopeCal.h.

I don't know if there was any other reason except that. Seems like something we could eventually support in grpc-device.

ni-jfitzger: Yeah, most of the functions in that header file probably require an external cal session, but these 2 can also be called from a regular session.

generated/niscope/niscope/_grpc_stub_interpreter.py
reckenro: So for this one I'm guessing we skipped over it because it's marked obsolete in [hapigen](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/niscope/niscopeBaseDistribution/niScopeAPIMetadata/source/niScopeAPIMetadata/base/functions/functions_obsolete_header.hapigen&version=GBmain&line=2133&lineEnd=2134&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) and the [header](https://p4.natinst.com/browser/perforce/build/exports/ni/nisc/niScopeHeaders/official/export/.latest/.latest/includes/niScopeHeaders/niScopeObsolete.h#line_361)

ni-jfitzger: That makes me wonder why we included it in the Python API. 🤷‍♂️

generated/niscope/niscope/_grpc_stub_interpreter.py
reckenro: Have an AzDo PR for the minimal fix up [here ](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/388957)

generated/niscope/niscope/_grpc_stub_interpreter.py
reckenro: Ah, I see. Yeah, sounds reasonable to try to get those added to grpc-device then. Added a separate task to look into and try to support these two methods on the grpc-device side so here we can pass through instead of raising a NotImplementedError: [AB#2210201](https://dev.azure.com/ni/DevCentral/_workitems/edit/2210201)

generated/niscope/niscope/_grpc_stub_interpreter.py
```diff
@@ -0,0 +1,425 @@
...
+            self._client.ConfigureEqualizationFilterCoefficients,
+            grpc_types.ConfigureEqualizationFilterCoefficientsRequest(vi=self._vi, channel_list=channel_list, coefficients=coefficients),
+        )
+
+    def configure_horizontal_timing(self, min_sample_rate, min_num_pts, ref_position, num_records, enforce_realtime):  # noqa: N802
+        self._invoke(
+            self._client.ConfigureHorizontalTiming,
+            grpc_types.ConfigureHorizontalTimingRequest(vi=self._vi, min_sample_rate=min_sample_rate, min_num_pts=min_num_pts, ref_position=ref_position, num_records=num_records, enforce_realtime=enforce_realtime),
+        )
+
```
ni-jfitzger, line 156: Any idea why this one isn't supported? I see we only include it in the Python API as a private method to meet the IVI Spec.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -1,527 +1,588 @@
...
-
-
-@pytest.fixture(scope='function')
-def session_5124():
-    with daqmx_sim_5124_lock:
-        with niscope.Session('5124') as simulated_session:  # 5124 is needed for video triggering
+import time
+
+
+class SystemTests:
```
ni-jfitzger, line 16: There's too many changes going on in this file. I can't follow it.
Please utilize 
```
    @pytest.fixture(scope='class')
    def session_creation_kwargs(self, grpc_channel):
        grpc_options = nidcpower.GrpcSessionOptions(grpc_channel, "")
        return {'_grpc_options': grpc_options}
```
like NI-DCPower has done to minimalize changes.

And I think it would also be best if you created the SystemTests classs in a change by itself, like #1846.

dmondrik: Good suggestion Jay. Since it looks like there'll be too much for this PR to get in today anyway (also given that Ryan's OOO today), @danestull will look into generating that separate refactoring PR.

danestull: @DavidCurtiss beat me to it haha See https://github.com/ni/nimi-python/pull/1853

generated/niscope/niscope/_grpc_stub_interpreter.py
DavidCurtiss: It was set as 'codegen_method':'no' because for nimi-python it was 'codegen_method':'private'.

I don't know why the IVI spec would cause it to be private, so that just brings up more questions for me. Should the same restriction not apply to grpc-device? Should we add it to grpc-device, but somehow make it obvious that it's intended to be "private", like by adding a leading underscore?

ni-jfitzger: My interpretation is that we didn't actually want to include it as part of our public API, but felt obligated to at least include it as a private method in order to meet the IVI spec. @marcoskirsch will likely know for sure.

ni-jfitzger: Yes, it seems to part of our obsolete header, so it makes sense that we didn't want to include it in the public API.

src/niscope/system_tests/test_system_niscope.py
danestull: Pulled Davids change from main and then added the tests again in a later commit so its easier to see

generated/niscope/niscope/_grpc_stub_interpreter.py
danestull: These fixes went in and should be pulled in by the newest update to the metadata

src/niscope/system_tests/test_system_niscope.py
danestull: These tests are not going to be enabled for grpc but there is still work to be done to have better behavior when a grpc session is used so I have left the todo in for now

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -557,32 +559,3 @@ def test_error_message(self, grpc_options):
...
-        assert exc_info.value.args[0] == 'cal_fetch_temperature is not supported over gRPC'
-        assert str(exc_info.value) == 'cal_fetch_temperature is not supported over gRPC'
-
-    def test_get_ext_cal_last_temperature(self, single_instrument_session):
-        with pytest.raises(NotImplementedError) as exc_info:
-            single_instrument_session.get_ext_cal_last_temp()
-        assert exc_info.value.args[0] == 'cal_fetch_temperature is not supported over gRPC'
-        assert str(exc_info.value) == 'cal_fetch_temperature is not supported over gRPC'
-
-    def test_configure_ref_levels(self, single_instrument_session):
```
dmondrik: The ones above are supported now, so I understand why they're no longer checking for 'not supported'. But `configure_ref_levels` is still unsupported, so shouldn't this test remain?

dmondrik (approved):
I left 1 test comment, but otherwise LGTM. Thanks for figuring out how to commit changes to his branch!

src/niscope/system_tests/test_system_niscope.py
danestull: Added this test back in, I just missed that it wasn't added

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -484,4 +339,223 @@ def test_configure_trigger_window(self, multi_instrument_session):
...
+    def test_error_message(self):
+        try:
+            # We pass in an invalid model name to force going to error_message
+            with niscope.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:invalid_model; BoardType:PXIe'):
+                assert False
+        except niscope.Error as e:
+            assert e.code == -1074118609
+            assert e.description.find('Simulation does not support the selected model and board type.') != -1
+
+    def test_get_self_cal_last_date_time(self, single_instrument_session):
```
ni-jfitzger, line 409: There's ongoing work to add grpc-devicer support for these.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -484,4 +339,228 @@ def test_configure_trigger_window(self, multi_instrument_session):
...
+
+    @pytest.fixture(scope='class')
+    def grpc_options(self, grpc_channel):
+        grpc_options = niscope.GrpcSessionOptions(grpc_channel, "")
+        yield grpc_options
+
+    @pytest.fixture(scope='function')
+    def single_instrument_session(self, grpc_options):
+        with niscope.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:5164; BoardType:PXIe', _grpc_options=grpc_options) as simulated_session:
+            yield simulated_session
```
ni-jfitzger, line 534: It would be nice to use something like
```
    @pytest.fixture(scope='class')
    def session_creation_kwargs(self, grpc_channel):
        grpc_options = nidcpower.GrpcSessionOptions(grpc_channel, "")
        return {'_grpc_options': grpc_options}
```
to reduce duplication, but it doesn't need to be done as part of this change.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -484,4 +339,228 @@ def test_configure_trigger_window(self, multi_instrument_session):
...
+        assert last_cal.hour == 0
+        assert last_cal.minute == 0
+
+    def test_get_self_cal_last_temperature(self, single_instrument_session):
+        last_cal_temp = single_instrument_session.get_self_cal_last_temp()
+        assert last_cal_temp == 25
+
+    def test_get_ext_cal_last_temperature(self, single_instrument_session):
+        last_cal_temp = single_instrument_session.get_ext_cal_last_temp()
+        assert last_cal_temp == 25
```
ni-jfitzger, line 431: @danestull please add a TODO so that we don't forget to move these tests back after the grpc-device change goes in.

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: It would be nice to add a comment above the test.
"# not supported by grpc due to numpy usage"

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: Tests for these methods are no longer present in the TestGrpc class. Since we intend to support them for gRPC very soon, I'm okay with that, so long as we have a TODO comment above them in the TestLibrary class. There is another thread where I'm requesting that comment, so I'll close this thread.

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: Acknowledged. The test creates its own session, so for the moment TestGrpc has its own copy of the method in order to pass grpc_options as session creation. This makes sense.

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: This is fine. I'm okay with not supporting this obsolete method right now. We can always change our mind and add it later. I like that we test for the NotImplementedError in TestGrpc.

src/niscope/templates/session.py/fancy_fetch.py.mako
ni-jfitzger: I've double-checked the C API documentation and thought through this and it makes sense to me. And I trust that our tests would catch if we weren't handling the data right.

build/templates/_grpc_stub_interpreter.py/default_method.py.mako
ni-jfitzger: This is actually already submitted to ni:master; it's no longer new.

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: @marcoskirsch has requested that we error when passing measurement_wfm_length while using gRPC. Since the TODO work item is parented under the overall work for adding gRPC NI-SCOPE support, though, I'll consider this resolved for the purpose of this PR.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -352,15 +220,6 @@ def test_reset_with_defaults(self, multi_instrument_session):
         multi_instrument_session.reset_device()
```
ni-jfitzger, line 220: We were supposed to have a version of this test in TestGrpc that checks for the NotImplementedError but it seems to have been deleted somewhere along the way. This test should actually be calling reset_with_defaults, not reset_device. That's why it's passing for both cases, right now. Please split the test into 2 versions and make sure it calls the right method.

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: Looks good.

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: See https://github.com/ni/grpc-device/pull/788

generated/niscope/niscope/_grpc_stub_interpreter.py
ni-jfitzger: I'm fine with not including gRPC support for this, for now. We can always add support later.

generated/niscope/niscope/_grpc_stub_interpreter.py
ni-jfitzger: @danestull we did have these NotImplementedError tests, but they seem to have gotten lost (probably during the merge of main). Given that the work to add support for these is alreeady underway, I'll allow the omission of these NotImplemented tests for grpc for this PR, so long as there's a TODO above them in the TestLibrary class. I've already requested that TODO in a separate thread, so I'll mark this thread as resolved.

src/niscope/system_tests/test_system_niscope.py
ni-jfitzger: Turns out it should also be using a single_instrument_session. Multi-instrument sessions do not support niScope_ResetWithDefaults.

generated/niscope/niscope/_grpc_stub_interpreter.py
ni-jfitzger: I'm not thrilled with the idea of having a method that's clearly intended to be mapped to a specific C API Function calling a different C API Function. I doubt users will care, though, so I'm going to allow it.

@danestull please create an internal work item, parented under this work to follow up with @marcoskirsch when he gets back, since he never responded after his initial comment. He may have a stronger opinion and I want him to have the opportunity to voice it.

src/niscope/system_tests/test_system_niscope.py
danestull: Fixed this and added tests for grpc and non grpc cases

src/niscope/system_tests/test_system_niscope.py
danestull: added this TODO

generated/niscope/niscope/_grpc_stub_interpreter.py
danestull: Created item [2213312](https://dev.azure.com/ni/DevCentral/_workitems/edit/2213312) to follow up on this at a later point

ni-jfitzger (approved):
Looks good.

ni-jfitzger:
FYI @marcoskirsch merging this to unblock important work next week. There are internal work items for anything that I felt needed follow-up.

PR #1849, opened 2022-11-04
Report gRPC-specific errors via a new errors.RpcError class

DavidCurtiss:

Provides a way for users to programmatically detect gRPC-specific errors, particularly session initialization errors.

[AB#2183743](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2183743)

None

Ran the new tests

### DIFF
--- a/build/templates/_grpc_stub_interpreter.py.mako
+++ b/build/templates/_grpc_stub_interpreter.py.mako
@@ -52,19 +52,19 @@ class GrpcStubInterpreter(object):
                         error_message += f'\nError status: {value}'
 
             grpc_error = rpc_error.code()
-            if grpc_error == grpc.StatusCode.UNAVAILABLE:
+            if grpc_error == grpc.StatusCode.NOT_FOUND:
+                raise errors.DriverTooOldError() from None
+            elif grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
+                raise ValueError(error_message) from None
+            elif grpc_error == grpc.StatusCode.UNAVAILABLE:
                 error_message = 'Failed to connect to server'
             elif grpc_error == grpc.StatusCode.UNIMPLEMENTED:
                 error_message = (
                     'This operation is not supported by the NI gRPC Device Server being used. Upgrade NI gRPC Device Server.'
                 )
-            elif grpc_error == grpc.StatusCode.NOT_FOUND:
-                raise errors.DriverTooOldError()
-            elif grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
-                raise ValueError(error_message)
 
             if error_code is None:
-                raise errors.RpcError(grpc_error, error_message)
+                raise errors.RpcError(grpc_error, error_message) from None
 
         if error_code < 0:
             raise errors.DriverError(error_code, error_message)
--- a/build/templates/_grpc_stub_interpreter.py.mako
+++ b/build/templates/_grpc_stub_interpreter.py.mako
@@ -65,7 +65,7 @@ class GrpcStubInterpreter(object):
         if grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
             raise ValueError(error_message)
         elif error_code is None:
-            raise errors.Error(error_message)
+            raise errors.RpcError(grpc_error, error_message)
         elif error_code < 0:
             raise errors.DriverError(error_code, error_message)
         elif error_code > 0:
--- a/build/templates/_grpc_stub_interpreter.py.mako
+++ b/build/templates/_grpc_stub_interpreter.py.mako
@@ -62,12 +62,11 @@ class GrpcStubInterpreter(object):
                 raise errors.DriverTooOldError()
             elif grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
                 raise ValueError(error_message)
-            elif error_code is None:
+
+            if error_code is None:
                 raise errors.RpcError(grpc_error, error_message)
 
-        if error_code is None:
-            raise errors.Error(error_message)
-        elif error_code < 0:
+        if error_code < 0:
             raise errors.DriverError(error_code, error_message)
         elif error_code > 0:
             if not error_message:
--- a/build/templates/_grpc_stub_interpreter.py.mako
+++ b/build/templates/_grpc_stub_interpreter.py.mako
@@ -36,7 +36,6 @@ class GrpcStubInterpreter(object):
         self._${config['session_handle_parameter_name']} = 0
 
     def _invoke(self, func, request):
-        grpc_error = None
         try:
             response = func(request)
             error_code = response.status
@@ -66,7 +65,9 @@ class GrpcStubInterpreter(object):
             elif error_code is None:
                 raise errors.RpcError(grpc_error, error_message)
 
-        if error_code < 0:
+        if error_code is None:
+            raise errors.Error(error_message)
+        elif error_code < 0:
             raise errors.DriverError(error_code, error_message)
         elif error_code > 0:
             if not error_message:
--- a/build/templates/_grpc_stub_interpreter.py.mako
+++ b/build/templates/_grpc_stub_interpreter.py.mako
@@ -51,22 +51,22 @@ class GrpcStubInterpreter(object):
                         error_code = int(value)
                     except ValueError:
                         error_message += f'\nError status: {value}'
+
             grpc_error = rpc_error.code()
+            if grpc_error == grpc.StatusCode.UNAVAILABLE:
+                error_message = 'Failed to connect to server'
+            elif grpc_error == grpc.StatusCode.UNIMPLEMENTED:
+                error_message = (
+                    'This operation is not supported by the NI gRPC Device Server being used. Upgrade NI gRPC Device Server.'
+                )
+            elif grpc_error == grpc.StatusCode.NOT_FOUND:
+                raise errors.DriverTooOldError()
+            elif grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
+                raise ValueError(error_message)
+            elif error_code is None:
+                raise errors.RpcError(grpc_error, error_message)
 
-        if grpc_error == grpc.StatusCode.UNAVAILABLE:
-            error_message = 'Failed to connect to server'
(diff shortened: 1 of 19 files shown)


generated/nidcpower/nidcpower/errors.py
```diff
@@ -43,6 +43,20 @@ def __init__(self, code, description):
         super(DriverWarning, self).__init__('Warning {0} occurred.\n\n{1}'.format(code, description))
 
 
+class RpcError(Error):
```
dmondrik, line 46: This should only be generated if gRPC is enabled for the API.

generated/nidmm/nidmm/_grpc_stub_interpreter.py
```diff
@@ -51,7 +51,7 @@ def _invoke(self, func, request):
         if grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
             raise ValueError(error_message)
```
dmondrik, line 52: It would seem these 2 (NOT_FOUND and INVALID_ARGUMENT) should be checked in the same if/elif block. Each raises an error based on `StatusCode`. (Existing code, but I'm evaluating the whole method since it's touched.)

dmondrik (approved):
I think it would be reasonable, when gRPC is supported, to have `__init__.py` also include this line:
`from nifgen.errors import RpcError  # noqa: F401`

generated/nidcpower/nidcpower/errors.py
DavidCurtiss: Done

generated/nidmm/nidmm/_grpc_stub_interpreter.py
DavidCurtiss: Grouped all uses of `grpc_error`.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -351,3 +351,32 @@ def test_error_message(self, grpc_channel):
...
+        init_behavior = nidmm.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION
+        grpc_options = nidmm.GrpcSessionOptions(grpc_channel, session_name, init_behavior)
+        with nidmm.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:4082; BoardType:PXIe', _grpc_options=grpc_options):
+            try:
+                with nidmm.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:4082; BoardType:PXIe', _grpc_options=grpc_options):
+                    assert False
+            except nidmm.Error as e:
+                assert e.rpc_code == expected_grpc_error
+                assert e.description == expected_error_message
+                assert str(e) == f'{expected_grpc_error}: {expected_error_message}'
```
ni-jfitzger, line 368: I wished we used fstrings in more places in this repo. (Not your problem).

src/nifake/unit_tests/test_grpc.py
```diff
@@ -163,6 +166,40 @@ def _assert_call(self, function_name, request_object):
...
+        init_behavior = nifake.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION
+        grpc_options = nifake.GrpcSessionOptions(object(), session_name, init_behavior)
+        interpreter = nifake._grpc_stub_interpreter.GrpcStubInterpreter(grpc_options)
+        try:
+            interpreter.init_with_options('dev1', False, False, '')
+            assert False
+        except nifake.Error as e:
+            assert e.rpc_code == grpc_error
+            assert e.description == error_message
+            assert str(e) == f'StatusCode.ALREADY_EXISTS: {error_message}'
```
ni-jfitzger, line 184: assert e.rpc_code == grpc_error
Minor thing: you could use `{grpc_error}` here, since this is an fstring. That would also be consistent with your change to test_system_nidmm.py.

ni-jfitzger: This applies to test_attach_to_non_existent_session, as well.

ni-jfitzger (approved):

src/nifake/unit_tests/test_grpc.py
DavidCurtiss: I did this intentionally, to 1) show you what it'll actually look like, and 2) assert that it stringifies in the way we expect, and not just as "1" or "2", etc.

build/templates/_grpc_stub_interpreter.py.mako
```diff
@@ -51,20 +50,22 @@ class GrpcStubInterpreter(object):
...
+                raise errors.DriverTooOldError()
+            elif grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
+                raise ValueError(error_message)
+            elif error_code is None:
+                raise errors.RpcError(grpc_error, error_message)
 
-        if grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
-            raise ValueError(error_message)
-        elif error_code is None:
+        if error_code is None:
```
dmondrik, line 68: Now that you check it in the 'catch' block, I don't think this check is legitimate. If the 'try' block executes to completion, the `error_code` will be set, and the status from which it was set is an integer. Right?

build/templates/errors.py.mako
```diff
@@ -55,6 +56,22 @@ class DriverWarning(Warning):
         super(DriverWarning, self).__init__('Warning {0} occurred.\n\n{1}'.format(code, description))
 
 
+% if grpc_supported:
+class RpcError(Error):
```
dmondrik, line 60: (Short non-PR-specific rant: I'm unhappy with github comments on code versus comments elsewhere, which don't appear to require being resolved and it's not trivial to comment on.)

I think it would be reasonable, when gRPC is supported, to have __init__.py also include this line:
`from nifgen.errors import RpcError # noqa: F401`

marcoskirsch: For better of for worse we don't import specific error classes today. Whether we should or not I don't know but I don't think as part of this PR.

build/templates/errors.py.mako
```diff
@@ -55,6 +56,22 @@ class DriverWarning(Warning):
...
+    '''An error specific to gRPC-based sessions'''
```
marcoskirsch, line 61: Is "gRPC-based sessions" the way we refer to these remote sessions?
I want to make sure we're 100% consistent and I see a lot of names thrown around to refer to these:

remote sessions
proxy sessions
gRPC sessions
gRPC client sessions
gRPC-based sessions

All this paraphrasing only bring ambiguity for users.

marcoskirsch (changes requested):
minor terminology comment
if it needs to be followed up separately let me know

build/templates/errors.py.mako
dmondrik: I think I have to agree with Marcos. Any customer who needs to distinguish can always import that specifically themselves. So I no longer think it's necessary to auto import it.

Interestingly, I believe the C layer will assign unique new driver error codes and map these gRPC errors in some way. It's not like they have any other option, really. But that doesn't mean we need to be 100% identical there.

build/templates/errors.py.mako
dmondrik: We have a class named `GrpcSessionOptions` and the comment on its use as a parameter says "MeasurementLink gRPC session options". @DavidCurtiss I think if you make the comment on your error class be "An error specific to gRPC sessions" we would at least be consistent. Then we can wordsmith it more later if someone really wants.

FWIW I don't see the term "gRPC client sessions" (or for that matter 'remote sessions' or 'proxy sessions') anywhere in nimi-python source, using `git grep`. Maybe those terms are prevalent in DVL efforts.

marcoskirsch: "gRPC session" works

build/templates/_grpc_stub_interpreter.py.mako
DavidCurtiss: The problem is the UNAVAILABLE/UNIMPLEMENTED use cases above. (Which we don't have a test for, except right now while NI-Switch is failing.) They set `error_message` but leave `error_code` as `None`, and fall through to right here.

But maybe those should raise an RpcError now.

dmondrik: Yeah, line 65 above should be `if` rather than `elif`.

DavidCurtiss: Done. And added unit tests.

build/templates/errors.py.mako
DavidCurtiss: Done

src/nifake/unit_tests/test_grpc.py
```diff
@@ -161,8 +164,73 @@ def _assert_call(self, function_name, request_object):
...
+            assert e.rpc_code == grpc_error
+            assert e.description == expected_error_message
+            assert str(e) == f'StatusCode.UNAVAILABLE: {expected_error_message}'
+
+    def test_function_not_implemented(self):
+        library_func = 'PoorlyNamedSimpleFunction'
+        grpc_error = grpc.StatusCode.UNIMPLEMENTED
+        expected_error_message_intro = 'This operation is not supported'
+        self._set_side_effect(library_func, side_effect=MyRpcError(None, "", grpc_error=grpc_error))
+        interpreter = self._get_initialized_library_interpreter()
```
dmondrik, line 189: This being in the gRPC section, it seems odd to use the library interpreter.

dmondrik (approved):
LGTM

src/nifake/unit_tests/test_grpc.py
DavidCurtiss: Already being changed [here](https://github.com/ni/nimi-python/pull/1840/files#diff-899f0ea835ab662dd4436c04e6f4de2ccc24050445a2871ef3bda7256bcee85b). Whichever of these gets merged first, the other will have to merge and update, and then it'll be correct.

ni-jfitzger (approved):
This looks good and enough of the tests have completed with the latest change. I'll merge.

ni-jfitzger:
FYI @marcoskirsch merged to unblock other dependent PRs, based on the resolution of your minor feedback and my own review.

PR #1837, opened 2022-10-25
Add NI-DCPower proto file and update generated `nidcpower` files

tdunkle:
- [X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

~- [ ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~

This change does not need to be customer-facing.

~- [ ] I've added tests applicable for this pull request~

Showing that the existing NI-DCPower system-level tests pass is sufficient.

### What does this Pull Request accomplish?

This pull request adds the NI-DCPower proto file to nimi-python, so that the NI-DCPower python API has a gRPC option enabled.

### List issues fixed by this Pull Request below, if any.

* [Internal NI bug AB#2149854](https://ni.visualstudio.com/DevCentral/_workitems/edit/2149854/)

### What testing has been done?

Installed python 3.7.9 and synced the nimi-python repo to a VM that has NI-DCPower 2022 Q4 installed. Observed that I could build nidcpower with `tox -e py37-codegen -- nidcpower`, and that I could run the system level tests with `tox -c generated\nidcpower\tox-system_tests.ini`, and all but one passed initially. I had to make a change to fix `test_create_and_delete_advanced_sequence_bad_type()`. Test output with all passing, for reference: [dcpower_test_output.txt](https://github.com/ni/nimi-python/files/9864065/dcpower_test_output.txt)

PR #1854, opened 2022-11-09
Enable the remaining `nidcpower` system tests and update metadata to d318 export.

tdunkle:

This change doesn't add new tests; however, it enables four test cases that only ran locally running over gRPC as well.

This pull request enables for gRPC test cases that were intentionally disabled in #1837 . It also updates the `nidcpower` metadata to the d318 export, which is why some new items show up in the attributes/config/enums/functions files.

* [Internal NI bug 2209572](https://ni.visualstudio.com/DevCentral/_workitems/edit/2209572/)
* [Internal NI bug 2208439](https://ni.visualstudio.com/DevCentral/_workitems/edit/2208439/)

Installed python 3.7.9 and synced the nimi-python repo to a VM that has NI-DCPower 2022 Q4 installed. Observed that I could build nidcpower with `tox -e py37-codegen -- nidcpower`, and that I could run the system level tests with `tox -c generated\nidcpower\tox-system_tests.ini`. All tests pass both locally and over gRPC (456 total test cases). Test output, for reference: [nidcpower_test_output.txt](https://github.com/ni/nimi-python/files/9974483/nidcpower_test_output.txt)


src/nidcpower/custom_types/lcr_load_compensation_spot.py
```diff
@@ -73,6 +73,9 @@ def __init__(self, frequency, reference_value_type, reference_value):
         self.reference_value_type = enums.LCRReferenceValueType(reference_value_type)
         self.reference_value = reference_value
 
+    def create_copy(self, target_class):
+        return target_class(frequency=self.frequency, reference_value_type=self.reference_value_type.value)
```
tdunkle, line 77: @DavidCurtiss this was the issue you and I ran into a couple of days ago with the `reference_value` property, which shows up as `reference_value_a` and `reference_value_b` in the proto. The offending test (`test_perform_lcr_load_compensation`) passes if we don't supply the reference_value at all, like I have here, but I don't think that's what we want because I can see that the test tries plugging in several reference_values. I'm looking into a better solution.

tdunkle: Hmm, well... now we're correctly pulling out the real and imaginary components, but I don't know that this is exactly how we want the test cases with non-complex numbers to work. Or maybe it is? In any case, it seems that we can't access a property called `reference_value` from this scope.

src/nifgen/metadata/nifgen.proto
```diff
@@ -1992,7 +1992,7 @@ message SetAttributeViStringRequest {
   string channel_name = 2;
   NiFgenAttribute attribute_id = 3;
   oneof attribute_value_enum {
-    string attribute_value_raw = 4;
+    string attribute_value = 4;
```
tdunkle, line 1995: I'm testing this out to see if it fixes the build (it did locally), but if the proto does need to be updated then there's a hapigen change I should make before submitting this.

tdunkle: It did fix the build, but since the proto file in the latest export of nifgen_grpc_device is identical, it seems like this is still an issue that I have to sort out in nimi-python.

generated/nidcpower/nidcpower/_grpc_stub_interpreter.py
```diff
@@ -80,7 +80,10 @@ def self_cal(self, channel_name):  # noqa: N802
         )
 
     def clear_latched_output_cutoff_state(self, channel_name, output_cutoff_reason):  # noqa: N802
-        raise NotImplementedError('clear_latched_output_cutoff_state is not supported over gRPC')
+        self._invoke(
+            self._client.ClearLatchedOutputCutoffState,
+            grpc_types.ClearLatchedOutputCutoffStateRequest(vi=self._vi, channel_name=channel_name, output_cutoff_reason=output_cutoff_reason.value),
```
dmondrik, line 85: Sorry to do this to you again, but if you could grab d288 I would appreciate it. A change I made later yesterday made it so it passes both the raw and enum as a oneof, and now our code needs to set the `output_cutoff_reason_raw` field.

src/nidcpower/metadata/functions.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-DCPower API metadata version 23.0.0d270
+# This file is generated from NI-DCPower API metadata version 23.0.0d285
```
dmondrik, line 2: In addition to these 4 metadata files, you should also include the latest exported .proto file (from the grpc export instead of the python export).

dmondrik (approved):
Once you update again, LGTM.

src/nifgen/metadata/nifgen.proto
tdunkle: Fixed this so that both the fgen and dcpower tests pass without modifying any proto files.

src/nidcpower/metadata/attributes.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-DCPower API metadata version 23.0.0d285
+# This file is generated from NI-DCPower API metadata version 23.0.0d9999
```
tdunkle, line 2: Need to use an official export for this before submitting.

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -1022,51 +1049,22 @@ class TestLibrary(SystemTests):
     def session_creation_kwargs(self):
         return {}
 
-    # Test doesn't run over gRPC due to issues with the name of the property attribute_value.
```
tdunkle, line 1025: This seems wrong, but the d305 metadata introduced this test failure for some reason.

generated/nidcpower/nidcpower/_grpc_stub_interpreter.py
tdunkle: I've since updated to d305.

src/nidcpower/system_tests/test_system_nidcpower.py
tdunkle: Talked more with @DavidCurtiss  about this. The function was very recently deprecated and its replacement isn't fully supported in the driver yet, so we're not sure what the best immediate course of action is. Need to discuss this with Jay when he's back in office next week.

tdunkle: I removed a couple of functions from grpc metadata generation since the driver doesn't fully support them yet, and then I re-enabled a couple of functions that were incorrectly marked deprecated.

src/nidcpower/metadata/attributes.py
tdunkle: Unresolving since this is an issue again with the latest commit.

tdunkle: Updated to d318 export.

src/nidcpower/metadata/functions.py
tdunkle: Updated to the d318 export's proto file.

src/nidcpower/custom_types/lcr_load_compensation_spot.py
tdunkle: David and I discussed this more, and we decided to add error handling to `create_copy()`, so that the function handles both complex and non-complex inputs in a clearer way.

docs/nidcpower/class.rst
```diff
@@ -1298,45 +1298,6 @@ get_ext_cal_recommended_interval
 
 
 
-get_lcr_compensation_data
```
marcoskirsch, line 1301: CHANGELOG.md needs corresponding change.

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
```diff
@@ -73,6 +73,12 @@ def __init__(self, frequency, reference_value_type, reference_value):
         self.reference_value_type = enums.LCRReferenceValueType(reference_value_type)
         self.reference_value = reference_value
 
+    def create_copy(self, target_class):
```
marcoskirsch, line 76: I suppose here you copy members from one type to another type that has the same `__init__` interface.
Customers shouldn't call this, right? If so add a prefix `_`.

marcoskirsch (changes requested):

generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
tdunkle: Yeah, this is just called in `metadata_add_all.py`, so I wouldn't expect customers to have a need for it. I've renamed it `_create_copy()`.

docs/nidcpower/class.rst
tdunkle: I've pushed a commit that removes `get_lcr_compensation_data` from CHANGELOG.md.

marcoskirsch:
TestGrpcStubInterpreter.test_return_multiple_types

test failing

src/nifgen/metadata/nifgen.proto
```diff
@@ -1,6 +1,10 @@
 
 //---------------------------------------------------------------------
+<<<<<<< HEAD
+// This file is generated from NI-FGEN API metadata version 23.0.0d70
+=======
 // This file is generated from NI-FGEN API metadata version 23.0.0d57
+>>>>>>> 9654eb43ff0836d48e83467184be95686c069422
```
tdunkle, line 7: bad merge

marcoskirsch (approved):

ni-jfitzger (approved):

PR #1855, opened 2022-11-09
Pass an API key for gRPC sessions

DavidCurtiss:

* Adds an api_key parameter to GrpcSessionOptions
* Makes api_key and initialization_behavior keyword-only args
* Passes the api_key to the server in the init functions

[AB#2182524](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2182524)

Created a unit test to verify the API key is passed


dmondrik (approved):
Other than the prefix possibly being `MEASUREMENTLINK`, LGTM.

build/templates/grpc_session_options.py.mako
```diff
@@ -41,24 +45,34 @@ class SessionInitializationBehavior(IntEnum):
 class GrpcSessionOptions(object):
     '''Collection of options that specifies session behaviors related to gRPC.'''
 
-    def __init__(self, grpc_channel, session_name, initialization_behavior=SessionInitializationBehavior.AUTO):
+    def __init__(
+        self,
+        grpc_channel,
+        session_name,
+        *,
+        api_key=MEASUREMENT_LINK_23Q1_NIMI_PYTHON_API_KEY,
```
dmondrik, line 53: [10:42 AM] John Bongaarts
It's `MEASUREMENTLINK_...`
Single word.

DavidCurtiss: Done

generated/nidcpower/nidcpower/_grpc_stub_interpreter.py
```diff
@@ -25,9 +25,12 @@ def __init__(self, grpc_options):
...
         self._vi = 0
 
-    def _invoke(self, func, request):
+    def _invoke(self, func, request, metadata=None):
         try:
-            response = func(request)
+            if metadata is not None:
+                response = func(request, metadata=metadata)
+            else:
+                response = func(request)
```
bkeryan, line 33: If `func` is the client stub, I don't think this if-statement is necessary. It accepts `metadata=None`.
https://grpc.github.io/grpc/python/grpc.html#multi-callable-interfaces

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
```diff
@@ -15,9 +15,13 @@
 %>\
 
     def ${full_func_name}(${method_decl_params}):  # noqa: N802
+        metadata = (
+            ('x-api-key', self._grpc_options.api_key),
+        )
         ${capture_response}self._invoke(
```
bkeryan, line 21: It looks like the API key is only used to initialize a session. After the session is initialized, another client can access the session without the API key. Is that the intent?

This might be acceptable for licensing but it would not be acceptable for authentication.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
```diff
@@ -15,9 +15,13 @@
 %>\
 
     def ${full_func_name}(${method_decl_params}):  # noqa: N802
+        metadata = (
+            ('x-api-key', self._grpc_options.api_key),
```
bkeryan, line 19: FYI, MeasurementLink is using NI-prefixed metadata headers instead of reusing standard(ish) headers: https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Source/Protos/ni/measurementlink/monitoring/v1/metadata.md

Reasoning: "In the future, if we incorporate off-the-shelf HTTP proxy or load balancing servers, these may use the standard HTTP headers differently."

This reasoning may not apply here, so this is just an FYI.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
bkeryan: You should strongly consider using an interceptor to specify metadata.

Reasons:
- An interceptor would automatically add metadata for all calls.
- You wouldn't have to modify any call sites, which would make this a lower-touch change. Instead, you would instantiate an interceptor object, call `grpc.intercept_channel(channel, interceptor)`, and use the channel that it returns.

We have been using interceptors to add metadata in C#. It's really easy with that API because it has a metadata interceptor built in and it provides an overload of `ChannelBase.Intercept()` that takes a `Func<Metadata, Metadata>`. I don't know if the gRPC Python API provides a built-in metadata interceptor, but writing a metadata interceptor for unary-unary calls should not be complicated.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
dmondrik: The invoke method is the only call site, so it doesn't seem to add much benefit.

generated/nidcpower/nidcpower/_grpc_stub_interpreter.py
DavidCurtiss: For python itself, it's not necessary, but passing along metadata=None would require us to update all the assert_called_with invocations in the tests.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
bkeryan: Here is an example of a "header adder interceptor".
https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/examples/python/interceptors/headers/header_manipulator_client_interceptor.py

DavidCurtiss: It was intentional. Zach's [User Story](https://ni.visualstudio.com/DevCentral/_workitems/edit/2182524) for this said: "It is only necessary for the Init calls, but if that is hard to scope to just those methods, all functions are fine."

The API key is indeed just for licensing -- I explicitly avoid using "auth" or "authentication" in any of my naming. Authentication would happen at the base gRPC layer, typically via SSL/TLS.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
DavidCurtiss: I don't think the _reasoning_ applies here, as this is a non-standard metadata header already. It could theoretically be (or become) a _common_ one, though. I'll see what Zach has to say about it.

DavidCurtiss: Sounds like Zach is happy with whatever. Brad, do you have a particular proposal for how to apply an NI prefix here? To be consistent with grpc-device's naming of [ni-error](https://github.com/ni/grpc-device/blob/bb15a2e7ea2549a7e4ab6fe55049fda7b550250f/source/server/converters.h#L327) trailing metdata, it should probably be ni-api-key. But grpc-device's ni-error metadata hasn't released yet, so we could change that, too, if need be.

bkeryan: `ni-api-key` sounds fine to me.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
bkeryan: > Authentication would happen at the base gRPC layer, typically via SSL/TLS.

FYI, with server-side TLS, it is common for the client to send authentication headers (such as OAuth2 tokens) with each HTTPS request.

DavidCurtiss: At the moment, we only foresee needing it during init, so I'm going to avoid adding the extra complexity of an interceptor for now, and just leave it as-is.

build/templates/_grpc_stub_interpreter.py/initialization_method.py.mako
DavidCurtiss: Renamed.

build/templates/_grpc_stub_interpreter.py.mako
```diff
@@ -43,9 +43,12 @@ class GrpcStubInterpreter(object):
...
         return self._${config['session_handle_parameter_name']}
 
-    def _invoke(self, func, request):
+    def _invoke(self, func, request, metadata=None):
         try:
-            response = func(request)
+            if metadata is not None:
+                response = func(request, metadata=metadata)
+            else:
+                response = func(request)
```
ni-jfitzger, line 51: Why do we need the if/else? Does func not already have a metadata argument with a default of None?

ni-jfitzger: I see there's already another thread about this.

DavidCurtiss: Yep. I'll copy the answer here:
For python itself, it's not necessary, but passing along metadata=None would require us to update all the assert_called_with invocations in the tests.

build/templates/grpc_session_options.py.mako
```diff
@@ -11,6 +11,10 @@ from enum import IntEnum
 # Customers can pass this value to the MeasurementLink discovery service to resolve the server instance that provides this interface.
 GRPC_SERVICE_INTERFACE_NAME = '${config["module_name"]}_grpc.${config["grpc_service_class_prefix"]}'
 
+# This constant specifies the API license key required by the NI gRPC Device Server that comes with
+# MeasurementLink 2023 Q1.
+MEASUREMENTLINK_23Q1_NIMI_PYTHON_API_KEY = 'DE10751B-3EE0-44EC-A93B-800E6A3C89E4'
```
ni-jfitzger, line 16: I read a little about API Keys and [documentation](https://cloud.google.com/endpoints/docs/grpc/when-why-api-key#security_of_api_keys) said that they're for identifying the project that's making the call. What's to stop this key from being used by other projects? Do we need to take steps to prevent that?

DavidCurtiss: No, not at this time. We might shore that up in a future release, and then deprecate the old key after a release or two has gone by.

build/templates/grpc_session_options.py.mako
```diff
@@ -41,24 +45,34 @@ class SessionInitializationBehavior(IntEnum):
...
     '''Collection of options that specifies session behaviors related to gRPC.'''
 
-    def __init__(self, grpc_channel, session_name, initialization_behavior=SessionInitializationBehavior.AUTO):
+    def __init__(
+        self,
+        grpc_channel,
+        session_name,
+        *,
+        api_key=MEASUREMENTLINK_23Q1_NIMI_PYTHON_API_KEY,
+        initialization_behavior=SessionInitializationBehavior.AUTO
```
ni-jfitzger, line 54: What's the point of specifying default values if these are going to be keyword-only arguments?

DavidCurtiss: You *must* specify default values for named keyword-only arguments. Or are you suggesting using a default of `None`? For the `None` case, you still have to turn it into the desired value, and it's just easier to specify the desired value as the default instead of having to explicitly check for `None`.

ni-jfitzger: Oh, I somehow convinced myself keyword-only meant they wouldn't be used if not passed. Nevermind.

ni-jfitzger (approved):

build/templates/grpc_session_options.py.mako
bkeryan: Specifying the desired value also informs type-checking tools like mypy and IDEs like VSCode what type to expect.

build/templates/_grpc_stub_interpreter.py.mako
marcoskirsch: > For python itself, it's not necessary, but passing along metadata=None would require us to update all the assert_called_with invocations in the tests.

Can you do this in its own follow up PR?

build/templates/grpc_session_options.py.mako
```diff
@@ -41,24 +45,34 @@ class SessionInitializationBehavior(IntEnum):
...
-            session_name (str): Specifies the **session name** to be used in the gRPC Device Server.
-                This is different from the resource name parameter many APIs take as a separate
-                parameter. Specifying a name makes it easy to share sessions across multiple gRPC
-                clients. You can use an empty string if you want to always initialize a new session
-                and then close it when the Python session goes out of scope.
+            session_name (str): Specifies the **session name** to be used in the NI gRPC Device
+                Server. This is different from the resource name parameter many APIs take as a
+                separate parameter. Specifying a name makes it easy to share sessions across
+                multiple gRPC clients. You can use an empty string if you want to always initialize
+                a new session and then close it when the Python session goes out of scope.
```
marcoskirsch, line 67:
```suggestion
a new session on the server and then close it when the Python Session instance on the client goes out of scope.
```

marcoskirsch (changes requested):

build/templates/grpc_session_options.py.mako
DavidCurtiss: Done

build/templates/_grpc_stub_interpreter.py.mako
DavidCurtiss: Did it here. Turns out it wasn't as bad as I was imagining.

marcoskirsch (approved):

PR #1852, opened 2022-11-08
Add gRPC Examples for NI-SWITCH, NI-FGEN, and NI-DMM.

danestull:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).

- [x] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.

- [] I've added tests applicable for this pull request

NA

### What does this Pull Request accomplish?

Add gRPC Examples for NI-SWITCH, NI-FGEN, and NI-DMM. These drivers have their protos and system tests added and so we are adding an example for each to show the use of the gRPC sessions. These examples are based off of existing examples so that users can see side-by-side the difference in setting up session (e.g. how easy it is). 

### What testing has been done?

Ran local server and after installing build artifacts successfully ran the gRPC examples.
Screen Shot 2022-11-08 at 8 53 08 AM
Screen Shot 2022-11-08 at 4 18 57 PM

Screen Shot 2022-11-09 at 9 19 57 AM

PR #1859, opened 2022-11-11
Add gRPC Example for NI-Digital

IPetersNI:

N/A

Add gRPC Examples for NI-Digital. These drivers have their protos and system tests added and so we are adding an example for each to show the use of the gRPC sessions. These examples are based off of existing examples so that users can see side-by-side the difference in setting up session (e.g. how easy it is).

Ran local server and after installing build artifacts successfully ran the gRPC examples.

### DIFF
--- a/src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
+++ b/src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
@@ -7,7 +7,7 @@ import os
 import sys
 
 
-def example(resource_name, options, address, port, trigger_source=None, trigger_edge=None, args.address, args.port):
+def example(resource_name, options, address, port, trigger_source=None, trigger_edge=None):
     session_name = '' # user-specified name; empty string means use a new, unnamed session
 
     # Connect to the grpc server
@@ -67,8 +67,10 @@ def _main(argsv):
 
     example(args.resource_name,
             'Simulate=1, DriverSetup=Model:6571' if args.simulate == 'True' else '',
+            args.address,
+            args.port,
             args.trigger_source if args.command == 'start-trigger' else None,
-            args.trigger_edge if args.command == 'start-trigger' else None, args.address, args.port )
+            args.trigger_edge if args.command == 'start-trigger' else None,)
 
 #TODO(danestull) Add example and main test once the gRPC server is started automatically
 
--- a/src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
+++ b/src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
@@ -7,7 +7,7 @@ import os
 import sys
 
 
-def example(resource_name, options, address, port, trigger_source=None, trigger_edge=None):
+def example(resource_name, options, address, port, trigger_source=None, trigger_edge=None, args.address, args.port):
     session_name = '' # user-specified name; empty string means use a new, unnamed session
 
     # Connect to the grpc server
@@ -70,26 +70,7 @@ def _main(argsv):
             args.trigger_source if args.command == 'start-trigger' else None,
             args.trigger_edge if args.command == 'start-trigger' else None, args.address, args.port )
 
-#TODO(IPETERSNI) Add example and main test once the gRPC server is started automatically
+#TODO(danestull) Add example and main test once the gRPC server is started automatically
 
 def main():
     _main(sys.argv[1:])
-
-
-def test_main():
-    _main([])
-    _main(['start-trigger'])
-
-
-def test_example():
-    resource_name = 'PXI1Slot2,PXI1Slot3'
-    options = {'simulate': True, 'driver_setup': {'Model': '6571'}, }
-    example(resource_name, options)
-
-    trigger_source = '/PXI1Slot2/PXI_Trig0'
-    trigger_edge = 'Rising'
-    example(resource_name, options, trigger_source, trigger_edge)
-
-
-if __name__ == '__main__':
-    main()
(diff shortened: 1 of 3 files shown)


src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/python
+
+import argparse
+import grpc
+import nidigital
+import os
+import sys
+
+
+def example(resource_name, options, address, port, trigger_source=None, trigger_edge=None):
```
danestull, line 10: You'll need to add address and port here as parameters since you're now passing them in from main()

src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
```diff
@@ -0,0 +1,95 @@
...
+    start_trigger.add_argument('-ts', '--trigger-source', default='/PXI1Slot2/PXI_Trig0', help='Source terminal for the start trigger')
+    start_trigger.add_argument('-te', '--trigger-edge', default='Rising', choices=['Rising', 'Falling'], help='Trigger on rising edge or falling edge of start trigger')
+    args = parser.parse_args(argsv)
+
+    example(args.resource_name,
+            'Simulate=1, DriverSetup=Model:6571' if args.simulate == 'True' else '',
+            args.trigger_source if args.command == 'start-trigger' else None,
+            args.trigger_edge if args.command == 'start-trigger' else None, args.address, args.port )
+
+#TODO(IPETERSNI) Add example and main test once the gRPC server is started automatically
```
danestull, line 73: You can put my name here that way all the example updates are easily searchable

src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
```diff
@@ -0,0 +1,95 @@
...
+
+def main():
+    _main(sys.argv[1:])
+
+
+def test_main():
```
danestull, line 79: This test_main and the test_example below both need to be deleted, they will be added back after the server work which is what the TODO above is about

danestull (changes requested):
I think you'll also need to pull from master to clear some of the PR build errors

src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
danestull: Clarity I would just prefer `def example(resource_name, options, trigger_source=None, trigger_edge=None, address, port):` if possible to match other examples

IPetersNI: Based on a previous conversation, it errors when in the end so I placed the parameters back in the beginning

danestull: Great thank you

src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/python
+
+import argparse
+import grpc
+import nidigital
+import os
+import sys
+
+
+def example(resource_name, options, address, port, trigger_source=None, trigger_edge=None, args.address, args.port):
```
dmondrik, line 10: There are multiple issues with the parameters:
- I don't think optional arguments can come before non-optional ones
- You might *invoke* this with `args.address` but that's not the parameter name.

src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
```diff
@@ -0,0 +1,76 @@
...
+    subparser = parser.add_subparsers(dest='command', help='Sub-command help')
+    start_trigger = subparser.add_parser('start-trigger', help='Configure start trigger')
+    start_trigger.add_argument('-ts', '--trigger-source', default='/PXI1Slot2/PXI_Trig0', help='Source terminal for the start trigger')
+    start_trigger.add_argument('-te', '--trigger-edge', default='Rising', choices=['Rising', 'Falling'], help='Trigger on rising edge or falling edge of start trigger')
+    args = parser.parse_args(argsv)
+
+    example(args.resource_name,
+            'Simulate=1, DriverSetup=Model:6571' if args.simulate == 'True' else '',
+            args.trigger_source if args.command == 'start-trigger' else None,
+            args.trigger_edge if args.command == 'start-trigger' else None, args.address, args.port )
```
dmondrik, line 71: When editing code, you must match existing style (or in rare cases be changing an entire file to match a new standard, but that's not the case here). You can see each argument to this method is on a separate line. Your new arguments should do that too.

src/nidigital/examples/burst_with_start_trigger/nidigital_burst_with_start_trigger_grpc.py
dmondrik: Also, don't forget to link the work item:
[Task 2190911](https://ni.visualstudio.com/DevCentral/_workitems/edit/2190911): Create gRPC NI-Digital example

dmondrik (approved):
For consistency with Dane's PR, a better PR title might be:
`Add gRPC Example for NI-Digital`

marcoskirsch:
Feedback from #1852 applies here as well, please look at that and apply it here.

IPetersNI:
Removed the todo and we have a work item to cover adding the tests back : https://dev.azure.com/ni/DevCentral/_workitems/edit/2222417

marcoskirsch (approved):

PR #1860, opened 2022-11-14
Add gRPC Example for NI-DCPower

danestull:

NA

Add gRPC Examples for NI-DCPower. These drivers have their protos and system tests added and so we are adding an example to show the use of the gRPC sessions. These examples are based off of existing examples so that users can see side-by-side the difference in setting up session (e.g. how easy it is).

Example runs with local grpc-device server and produces expected output.


src/nidcpower/examples/nidcpower_measure_record_grpc.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def example(resource_name, options, voltage, length, address, port):
+    session_name = ''  # user-specified name; empty string means use a new, unnamed session
+
+    # Connect to the grpc server
+    channel = grpc.insecure_channel(f'{address}:{port}')
+    session_options = nidcpower.GrpcSessionOptions(channel, session_name)
+
+    with nidcpower.Session(resource_name=resource_name, options=options, _grpc_options=session_options) as session:
```
dmondrik, line 16: The existing examples from which you copied are inconsistent in one aspect I just noticed. After the `with` statement that introduces the session, here I see a blank line (as it is in the corresponding local example). The 3 others in the PR you created last week did not have this. Thoughts @marcoskirsch @ni-jfitzger ? Should we remove the extra blank line not only here but also in the existing example?

dmondrik (approved):
LGTM. Let's see what the expert opinion is on the blank line.

CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ All notable changes to this project will be documented in this file.
         * Support for Python 3.6
 * ### `nidcpower` (NI-DCPower)
     * #### Added
+        * API parity with next NI-DCPower release.
```
ni-jfitzger, line 41: Is this a bad merge or an unrelated change?

src/nidcpower/examples/nidcpower_measure_record_grpc.py
ni-jfitzger: Yeah, that does appear to be a minor inconsistency. I wouldn't reject a cleanup of the NI-DCPower examples to be more consistent about that spacing, but it also doesn't really bother me. I don't think it's the sort of the thing that "black" would even change and it's readable either way.

danestull: I went ahead and removed the extra blank line from both.

CHANGELOG.md
danestull: Bad merge, fixed

src/nidcpower/examples/nidcpower_measure_record_grpc.py
ni-jfitzger: Please make the same change to the other examples, including the disabled example, for consistency.

danestull: Done!

ni-jfitzger (approved):

PR #1866, opened 2022-11-17
Enable NI-FGEN's test_create_advanced_arb_sequence_wrong_size for gRPC.

reckenro:

- Moves check for array sizes being equal from `_library_interpreter` up to session so that the checks are run when using gRPC as well
- Enabled `test_create_advanced_arb_sequence_wrong_size` NI-FGEN system test for gRPC
- Moved some ni-fake unit tests for these array size mismatches from the library_interpreter tests to the session tests.

Fixes [AB#2220782](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2220782)

None

FGEN system tests pass locally. Also, moved unit tests also pass.


generated/nidcpower/nidcpower/session.py
```diff
@@ -6653,6 +6653,8 @@ def set_sequence(self, values, source_delays):
                 The valid values are between 0 and 167 seconds.
 
         '''
+        if source_delays is not None and len(source_delays) != len(values):  # case S160
```
dmondrik, line 6656: It's worth noting that if you pass `values: None, source_delays: [1.0]` as the arguments then this will raise `TypeError` when calling `len(None)`. This is equally true of the existing released code, so this is not a regression. It's also not that important IMO since it should result in an exception regardless. Just something to consider.

dmondrik (approved):
LGTM. Thanks for figuring this out.

marcoskirsch (approved):

PR #1867, opened 2022-11-17
Add documentation for gRPC API elements.

dmondrik:

This adds documentation for most public-facing gRPC API elements:
- `RpcError` exception
- `SessionInitializationBehavior` enumeration
- `GrpcSessionOptions` class
- Parameter `_grpc_options` to the session constructor.

[Task 2190966](https://ni.visualstudio.com/DevCentral/_workitems/edit/2190966): Add docs for new API behaviors

Built locally. Attaching snippets of pictures of generated .html files below, using NI-DCPower as an example.

---

image

---

image

image

---

image

### DIFF
--- a/build/templates/errors.rst.mako
+++ b/build/templates/errors.rst.mako
@@ -2,10 +2,11 @@
     import build.helper as helper
 
     config = template_parameters['metadata'].config
-    module_name = config['module_name']
     driver_name = config['driver_name']
     enums = config['enums']
     extra_errors_used = config['extra_errors_used']
+    module_name = config['module_name']
+
     grpc_supported = template_parameters['include_grpc_support']
 %>\
 ${helper.get_rst_header_snippet('Exceptions and Warnings', '=')}
--- a/build/templates/errors.rst.mako
+++ b/build/templates/errors.rst.mako
@@ -6,6 +6,7 @@
     driver_name = config['driver_name']
     enums = config['enums']
     extra_errors_used = config['extra_errors_used']
+    grpc_supported = template_parameters['include_grpc_support']
 %>\
 ${helper.get_rst_header_snippet('Exceptions and Warnings', '=')}
 
@@ -78,6 +79,17 @@ ${helper.get_rst_header_snippet('SelfTestError', '-')}
     .. exception:: SelfTestError
 
         An error due to a failed self-test
+
+
+% endif
+% if grpc_supported:
+${helper.get_rst_header_snippet('RpcError', '-')}
+
+    .. py:currentmodule:: ${module_name}.errors
+
+    .. exception:: RpcError
+
+        An error specific to gRPC sessions
 
 
 % endif
--- a/build/templates/errors.rst.mako
+++ b/build/templates/errors.rst.mako
@@ -90,7 +90,7 @@ ${helper.get_rst_header_snippet('RpcError', '-')}
 
     .. exception:: RpcError
 
-        An error specific to gRPC sessions
+        An error specific to sessions to the NI gRPC Device Server
 
 
 % endif
--- a/build/templates/grpc_session_options.py.mako
+++ b/build/templates/grpc_session_options.py.mako
@@ -63,7 +63,7 @@ class GrpcSessionOptions(object):
         Args:
             grpc_channel (grpc.Channel): Specifies the channel to the NI gRPC Device Server.
 
-            session_name (str): User-specified name that identifies the driver session in the NI gRPC Device
+            session_name (str): User-specified name that identifies the driver session on the NI gRPC Device
                 Server. This is different from the resource name parameter many APIs take as a separate
                 parameter. Specifying a name makes it easy to share sessions across multiple gRPC clients.
                 You can use an empty string if you want to always initialize a new session on the server.
@@ -73,7 +73,7 @@ class GrpcSessionOptions(object):
 
             initialization_behavior (enum): Specifies whether it is acceptable to initialize a new
                 session or attach to an existing one, or if only one of the behaviors is desired.
-                The driver session exists in the NI gRPC Device Server.
+                The driver session exists on the NI gRPC Device Server.
         '''
         self.grpc_channel = grpc_channel
         self.session_name = session_name
--- a/build/templates/grpc_session_options.py.mako
+++ b/build/templates/grpc_session_options.py.mako
@@ -19,26 +19,29 @@ MEASUREMENTLINK_23Q1_NIMI_PYTHON_API_KEY = 'DE10751B-3EE0-44EC-A93B-800E6A3C89E4
 class SessionInitializationBehavior(IntEnum):
     AUTO = 0
     r'''
-    Attach to an existing session with the specified name if it exists, otherwise intialize a new session.
+    The NI gRPC Device Server will attach to an existing session with the specified name if it exists, otherwise the server
+    will initialize a new session.
 
     Note:
-    If this initializes a new session on the NI gRPC Device Server, then when the Python session goes out of scope,
-    it will automatically close the server session. If this attaches to an existing session on the NI gRPC Device Server,
-    then when the Python session goes out of scope, it will detach from the server session and leave it open.
+    When using the Session as a context manager and the context exits, the behavior depends on what happened when the constructor
+    was called. If it resulted in a new session being initialized on the NI gRPC Device Server, then it will automatically close the
+    server session. If it instead attached to an existing session, then it will detach from the server session and leave it open.
     '''
     INITIALIZE_SERVER_SESSION = 1
     r'''
     Require the NI gRPC Device Server to initialize a new session with the specified name.
 
     Note:
-    When the Python session goes out of scope, it will automatically close the server session.
+    When using the Session as a context manager and the context exits, it will automatically close the
+    server session.
     '''
     ATTACH_TO_SERVER_SESSION = 2
     r'''
     Require the NI gRPC Device Server to attach to an existing session with the specified name.
 
     Note:
-    When the Python session goes out of scope, it will detach from the server session and leave it open.
+    When using the Session as a context manager and the context exits, it will detach from the server session
+    and leave it open.
     '''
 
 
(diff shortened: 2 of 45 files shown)


ni-jfitzger:
Please change the title of your PR.

dmondrik:
> Please change the title of your PR.

Oops. Done.

build/templates/errors.rst.mako
```diff
@@ -80,6 +82,17 @@ ${helper.get_rst_header_snippet('SelfTestError', '-')}
...
 
+% endif
+% if grpc_supported:
+${helper.get_rst_header_snippet('RpcError', '-')}
+
+    .. py:currentmodule:: ${module_name}.errors
+
+    .. exception:: RpcError
+
+        An error specific to gRPC sessions
```
marcoskirsch, line 93: is this what we decided to call these? "gRPC sessions"?
Want to make sure we use consistent terminology across the MeasurementLink board.

build/templates/errors.rst.mako
```diff
@@ -80,6 +82,17 @@ ${helper.get_rst_header_snippet('SelfTestError', '-')}
         An error due to a failed self-test
 
 
+% endif
+% if grpc_supported:
+${helper.get_rst_header_snippet('RpcError', '-')}
```
marcoskirsch, line 87: Why did you choose `RpcError` without the g prefix?
I don't think it matters much, RPC being more generic. But curious. We seem to use the name of the specific technology everywhere explicitly.

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,105 @@
...
+SessionInitializationBehavior
+-----------------------------
+
+.. py:class:: SessionInitializationBehavior
+
+    .. py:attribute:: SessionInitializationBehavior.AUTO
+
+
+
+        Attach to an existing session with the specified name if it exists, otherwise intialize a new session.
```
marcoskirsch, line 29: I think this needs to specify that the existing or new session in on the server so there's no ambiguity for clients.
Do we have a term for sessions that live in the server?

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,105 @@
...
+
+        .. note:: If this initializes a new session on the NI gRPC Device Server, then when the Python session goes out of scope,
```
marcoskirsch, line 31: "the Python session" -> "the Session object".

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,105 @@
...
+            it will automatically close the server session. If this attaches to an existing session on the NI gRPC Device Server,
+            then when the Python session goes out of scope, it will detach from the server session and leave it open.
```
marcoskirsch, line 33: I've seen us talk about "Session going out of scope". I don't think this is right. `__exit()__` only gets called when the object was created as a context manager. Customers may and do sometimes use these Session objects by invoking the constructor "normally".

Consider:

```
class ContextManager():
    def __init__(self):
        print('init method called')

    def __enter__(self):
        print('enter method called')
        return self

    def __exit__(self, exc_type, exc_value, exc_traceback):
        print('exit method called')

    def __del__(self):
        print('del method called')


print("Test as context manager")

with ContextManager() as manager:
    print('with statement block')

print("\nTest as regular variable")

if True:
    c = ContextManager();

print("The end")
```

outputs:

```
Test as context manager
init method called
enter method called
with statement block
exit method called

Test as regular variable
init method called
The end
del method called
del method called
```

I think we may want to ensure that the appropriate thing happens on `__del()__` if it `__exit()__` never got called.
We also need the documentation to cover both cases.
We still (in documentation and examples) encourage customers to use as context manager.

We may need to run this through a Python guru.

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,105 @@
...
+
+        
+
+
+    :type grpc_channel: grpc.Channel
+
+    :param session_name:
+        
+
+        Specifies the **session name** to be used in the gRPC Device Server.
```
marcoskirsch, line 87:
```
      User-specified name that identifies the driver session in the gRPC Device Server.
```

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,105 @@
...
+
+                This is different from the resource name parameter many APIs take as a separate parameter. Specifying a name makes it easy to share sessions across multiple gRPC clients. You can use an empty string if you want to always initialize a new session and then close it when the Python session goes out of scope.
```
marcoskirsch, line 89: I think the above description applies to initializing the server-side session but we should also explain that the string is used at attach time to lookup an existing session.

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,105 @@
...
+
+        
+
+
+    :type session_name: str
+
+    :param initialization_behavior:
+        
+
+        Specifies whether it is acceptable to initialize a new session or attach to an existing one, or if only one of the behaviors is desired. To attach to an existing session, you must specify a non-empty session name.
```
marcoskirsch, line 99: somewhere in here you need to say that the session that is to be initialized or attached to lives in the server.

marcoskirsch: "you must specify a non-empty session name" -> "you must specify the **session name** for the session in the server that it was registered with"

(or something like that)

marcoskirsch (changes requested):
My feedback on documentation is from the angle of "what are gRPC / nimi-python developers intimately familiar with and may take for granted, but could be ambiguous or confusing to a customer trying to figure out how this works?"

I think it's pretty good overall though thanks!

build/templates/errors.rst.mako
dmondrik: Well first, I'm just documenting the code that was already submitted. But I understand that seeing it before didn't raise the trigger in your mind. We chose that name because gRPC itself [uses that exception name for Python](https://grpc.github.io/grpc/python/grpc.html#grpc-exceptions). It's also `RpcException` in C# just for reference. I think it's the appropriate choice here, although I do agree with your general guidance.

build/templates/errors.rst.mako
dmondrik: "An error specific to sessions to the NI gRPC Device Server"

build/templates/grpc_session_options.rst.mako
dmondrik: I looked at the .html documentation for Session and mimicked the verbiage about context managers. So I think it's much more technically precise now. I don't think you need to say much about what happens if the session is neither deleted nor exited. To me, that's _caveat emptor_.

build/templates/grpc_session_options.py.mako
```diff
@@ -55,23 +58,22 @@ class GrpcSessionOptions(object):
...
-                Server. This is different from the resource name parameter many APIs take as a
-                separate parameter. Specifying a name makes it easy to share sessions across
-                multiple gRPC clients. You can use an empty string if you want to always initialize
-                a new session on the server and then close it when the Python Session instance on
-                the client goes out of scope.
+            session_name (str): User-specified name that identifies the driver session in the NI gRPC Device
+                Server. This is different from the resource name parameter many APIs take as a separate
+                parameter. Specifying a name makes it easy to share sessions across multiple gRPC clients.
+                You can use an empty string if you want to always initialize a new session on the server.
+                To attach to an existing session, you must specify the session name it was initialized with.
```
ni-jfitzger, line 70: >To attach to an existing session, you must specify the session name it was initialized with.
Should this be "session name" or "resource name"?

build/templates/grpc_session_options.py.mako
```diff
@@ -55,23 +58,22 @@ class GrpcSessionOptions(object):
...
         Args:
             grpc_channel (grpc.Channel): Specifies the channel to the NI gRPC Device Server.
 
-            session_name (str): Specifies the **session name** to be used in the NI gRPC Device
-                Server. This is different from the resource name parameter many APIs take as a
-                separate parameter. Specifying a name makes it easy to share sessions across
-                multiple gRPC clients. You can use an empty string if you want to always initialize
-                a new session on the server and then close it when the Python Session instance on
-                the client goes out of scope.
+            session_name (str): User-specified name that identifies the driver session in the NI gRPC Device
```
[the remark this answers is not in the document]
ni-jfitzger: Below, you also use "in".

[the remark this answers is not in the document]
dmondrik: Changed all "in" to "on".

build/templates/grpc_session_options.py.mako
dmondrik: Discussed in chat; these are different names. No change.

ni-jfitzger: It was explained to me. The session name is different from the resource name as mentioned in the documentation above. The shared session use case is for a session already opened with GrpcSessionOptions. That is where the session name is specified for an already opened session is specified.

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,101 @@
...
+
+    .. py:attribute:: SessionInitializationBehavior.AUTO
+
+
+        The NI gRPC Device Server will attach to an existing session with the specified name if it exists,
+        otherwise the server will initialize a new session.
+
+        .. note:: When using the Session as a context manager and the context exits, the behavior depends on what happened when the constructor
+            was called. If it resulted in a new session being initialized on the NI gRPC Device Server, then it will automatically close the
+            server session. If it instead attached to an existing session, then it will detach from the server session and leave it open.
```
ni-jfitzger, line 33: Can we find a way not to replicate all of this information?

docs/nidcpower/class.rst
```diff
@@ -149,6 +149,16 @@ Session
...
 
+    :param _grpc_options:
+        
+
+        MeasurementLink gRPC session options
+
+        
+
+
+    :type _grpc_options: nidcpower.grpc_session_options.GrpcSessionOptions
```
ni-jfitzger, line 160: Can we update this so that it links to the documentation for that class? Maybe this will help: https://stackoverflow.com/a/59682507

There is `:py:class:` for referencing a class.
Feel free to close if you feel it's better not to link to it for some reason.

build/templates/grpc_session_options.rst.mako
```diff
@@ -0,0 +1,101 @@
...
+%>\
+${helper.get_rst_header_snippet('gRPC Support', '=')}
+
+Support for using ${driver_name} over gRPC
+
+.. py:currentmodule:: ${module_name}
+
+
+Enums
+=====
```
ni-jfitzger, line 17: This is showing up in the TOC for some reason.
It's kind of awkward because it doesn't show up under a toggle for gRPC Support.

image.png

ni-jfitzger: Based on https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections, maybe replace ==== with ----, ---- with ^^^^^

dmondrik: Since there's only 1 enum and 1 class, I got rid of those headers and just made the 2 entries be the subsections. It seems nice to me.

docs/nidcpower/class.rst
dmondrik: Made the links work.

build/templates/grpc_session_options.rst.mako
dmondrik: Chatted separately about this. It seems anything that doesn't come from metadata (that is, anything generated here in nimi-python itself) has some data replication. So I'm going to close this issue for now, but I agree it would be nice to improve on this in the future.

marcoskirsch (approved):

PR #1870, opened 2022-12-01
NI-DCPower: Update LCRLoadCompensationSpot byte packing alignment

olsl21:

NI-DCPower: Update LCRLoadCompensationSpot byte packing alignment

- Update the byte packing alignment of `struct_NILCRLoadCompensationSpot` from 4 bytes to 8 bytes for Linux and MacOS
  - Update CHANGELOG.md
- Create `test_lcr_load_compensation_spot_byte_packing_alignment()` unit test to test the byte packing alignment of `struct_NILCRLoadCompensationSpot`

Ran the newly created unit test on Windows 10 and Ubuntu 20.04, it passed on both OSes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ All notable changes to this project will be documented in this file.
 * ### `nidcpower` (NI-DCPower)
     * #### Added
     * #### Changed
-        * The byte packing alignment of `struct_NILCRLoadCompensationSpot` is changed from 4-byte to 8-byte in Linux.
+        * Binary compatibility change for type `LCRLoadCompensationSpot` on Linux. Client code using method `nidcpower.Session.perform_lcr_load_compensation` on Linux now requires NI-DCPower 2023 Q1 driver runtime or newer.
     * #### Removed
 * ### `nidigital` (NI-Digital Pattern Driver)
     * #### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 * ### `nidcpower` (NI-DCPower)
     * #### Added
     * #### Changed
+        * The byte packing alignment of `struct_NILCRLoadCompensationSpot` is changed from 4-byte to 8-byte in Linux.
     * #### Removed
 * ### `nidigital` (NI-Digital Pattern Driver)
     * #### Added
--- a/generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
+++ b/generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
@@ -8,10 +8,9 @@ import nidcpower.enums as enums
 # This class is an internal ctypes implementation detail that corresponds to
 # NILCRLoadCompensationSpot in the C API
 class struct_NILCRLoadCompensationSpot(ctypes.Structure):  # noqa N801
-    if platform.system() == "Windows" and platform.architecture()[0] == "64bit":
+    if (platform.system() == "Windows" and platform.architecture()[0] == "64bit") or \
+            platform.system() == "Linux" or platform.system() == "Darwin":
         _pack_ = 8
-    elif platform.system() == "Linux" or platform.system() == "Darwin":
-        _pack_ = 4
     else:
         _pack_ = 1
     _fields_ = [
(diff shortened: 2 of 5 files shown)


ni-jfitzger:
The following code generated files do not match what is commited to Git. Run codegen and include the generated files in the PR.
 M generated/nidcpower/nidcpower/lcr_load_compensation_spot.py
 M generated/nidcpower/nidcpower/unit_tests/test_nidcpower.py
 
The command "python tools/ensure_codegen_up_to_date.py" failed and exited with 1 during .

ni-jfitzger:
I don't understand why this is changing or what effect changing it will have on our users.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 * ### `nidcpower` (NI-DCPower)
     * #### Added
     * #### Changed
+        * The byte packing alignment of `struct_NILCRLoadCompensationSpot` is changed from 4-byte to 8-byte in Linux.
```
marcoskirsch, line 42: This needs to say why and what to do about it or what effects it has on customer code.

> The byte packing alignment of `struct_NILCRLoadCompensationSpot` is changed from 4-byte to 8-byte in Linux in order to match the driver runtime for NI-DCPower 2023 Q1.

marcoskirsch:
> I don't understand why this is changing or what effect changing it will have on our users.

I can explain this.

The C struct defined in `nidcpower.h` is expected to match the binary layout of the corresponding cluster in NI-DCPower's LabVIEW API. It turns out that for some reason, `nidcpower.h` screwed up and specified the wrong packing on Linux.

After much discussion, it was decided to change the packing defined in `nidcpower.h`  for Linux in order to match the LabVIEW cluster. This requires `nidcpower` Python module to be adjusted accordingly. **This is a binary breaker for NI-DCPower on Linux** but one that was deemed acceptable due to the fact that the struct is used only on one hardware product that only went on sale recently and is not believed to be used on any Linux systems.

The effect on users?

Should a customer attempt to use NI-DCPower 2023 Q1 or newer with `nidcpower` Python module 1.4.2 or older, the binary layout for these types won't match and the user will see undefined behavior.
The same is true should a customer attempt to use NI-DCPower 2022 Q4 or older together with `nidcpower` Python module 1.4.3 or newer.

@olsl21 while we don't expect this to be encountered out on the field by anyone, should it occur I think we need to provide some information. Specifically: please create an Issue in nimi-python to describe what customers will encounter. We will respond with the workaround (use matching versions) and close as `wontfix`. But we want the information out there.

CHANGELOG.md
ni-jfitzger, line 42: Do we have a special way for calling out binary breakers in the changelog?

marcoskirsch: Not really because they are soooo unusual.

I actually don't like how the change is worded because it references an internal implementation detail. That's meaningless to customers, they don't interact with `struct_NILCRLoadCompensationSpot`.

I think we need to call out the compatibility breakage here in terms customers care.

> Binary compatibility change for type `LCRLoadCompensationSpot ` on Linux. Client code using method `nidcpower.Session.perform_lcr_load_compensation` on Linux now requires NI-DCPower 2023 Q1 driver runtime or newer.

I think think this is enough. It gives customers information that is truthful, useful, and actionable to them rather than just paraphrasing a diff which a Python user won't know what to do with.

You guys like it?

marcoskirsch (changes requested):

CHANGELOG.md
olsl21: Yes, I think that sounds good. I have updated the CHANGELOG in my latest commit [6d0b3fb](https://github.com/ni/nimi-python/pull/1870/commits/6d0b3fba52ee8295e821088da8812f0da19eddec).

olsl21: I have updated the CHANGELOG statement based on [your other suggestion](https://github.com/ni/nimi-python/pull/1870#discussion_r1038439451), kindly let me know if you want both of them to be included.

olsl21:
> @olsl21 while we don't expect this to be encountered out on the field by anyone, should it occur I think we need to provide some information. Specifically: please create an Issue in nimi-python to describe what customers will encounter. We will respond with the workaround (use matching versions) and close as `wontfix`. But we want the information out there.

@marcoskirsch I've created issue #1874 and closed it with a comment regarding the workaround.

PR #1871, opened 2022-12-01
Remove gRPC Examples

danestull:

NA

NA

Removing the gRPC examples since we don't want to add server exe launch code to the examples and we are not able to add the Service to nimibot without adding circular dependencies so therefore these examples would exist without tests which is undesirable.

TODO: List of issues.

* Fix [#2222417](https://dev.azure.com/ni/DevCentral/_workitems/edit/2222417)

NA


danestull:
So these examples do not show much besides creating the session with grpc-options. If we want I'm happy to add more documentation somewhere to cover this, perhaps in the wiki?

ni-jfitzger:
>If we want I'm happy to add more documentation somewhere to cover this, perhaps in the wiki?

I believe it was decided that we didn't want to draw attention to the feature in our documentation, at the moment. I think if we add additional documentation anywhere, it should probably just be in the grpc_session_options.rst.

marcoskirsch (approved):

marcoskirsch:
> So these examples do not show much besides creating the session with grpc-options. If we want I'm happy to add more documentation somewhere to cover this, perhaps in the wiki?

If the API documentation is good, that would be enough. A page "How to use nimi-python APIs with NI gRPC Device Server" in the wiki would be really nice too.

ni-jfitzger (approved):

PR #1878, opened 2022-12-15
[Draft] Pulling in d131 FGEN exported metadata.

reckenro:
Just checking what kind of changes the recent FGEN updates have for nimi-python.

I think most (if not all) these changes are resulting from [this AzDo PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/387411)

I'm pretty sure we don't want to uptake any of these for the upcoming nimi-python release but wanted to get the diffs up here to make sure.

I think this is at least partially busted
- grpc side it's removing enums which seems wrong
- There's some of that i8 to ViAddr stuff that would need to be fixed for grpc
- It's removing a bunch of enum values and changing some enum values in nimi-python which seems problematic
- Some of this might be getting closer to the FGEN 23.0 final metadata but I'm not clear what pieces are intentional and verified vs. things we'd want to revert.


reckenro:
FYI @marcoskirsch / @dmondrik

PR #1890, opened 2023-03-02
Restore previous lock, unlock behavior

### LINKED ISSUE #1888: Deadlock on multithreaded usage due to UnlockSession always being called with callerHasLock=False
### Description of issue

I encountered a deadlock situation when using niscope from multiple threads. Starting with version 1.4.3, the Session unlock has no effect and leads to other threads hanging forever waiting for the lock.

This is due to the use of a local variable caller_has_lock_ctype which always has the default value False.

https://github.com/ni/nimi-python/blob/28f446e173f4e79b6ef97e0025fde8f2de0c490d/generated/niscope/niscope/_library_interpreter.py#L642

### System report

```
OS:
    Name:      Windows
    Version:   10.0.19041
    Bits:      64
Driver:
    Name:      NI-SCOPE
    Version:   21.0.0.49152
Module:
    Name:      niscope
    Version:   1.4.3
Python:
    Version:   3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64 bit (AMD64)]
    Bits:      64
    Is_Venv:   False
    Installed Packages:
        zipp==3.7.0
        typing-extensions==4.1.1
        six==1.16.0
        setuptools==47.1.0
        pyvisa==1.11.3
        pyvisa-py==0.5.2
        pip==22.0.3
        numpy==1.21.5
        nitclk==1.4.3
        niscope==1.4.3
        nifgen==1.4.3
        nidmm==1.4.3
        nidaqmx==0.6.0
        importlib-metadata==4.11.0
        hightime==0.2.1
```

### Steps to reproduce issue

``` 
import niscope
import threading


s = niscope.Session(resource_name="PXI2Slot7", reset_device=True)


def thread_function():

    print("Locking...")
    s.lock()
    print("Locked!")
    s.unlock()
    print("Unlocked!")



t1 = threading.Thread(target=thread_function)

t1.start()
t1.join()

t2 = threading.Thread(target=thread_function)
t2.start()
t2.join()
```

This script runs two threads sequentially. The first thread is started and does a lock/unlock.
The second thread then immediately hangs when trying to do the lock().
Output:
```
Locking...
Locked!
Unlocked!
Locking...
```

ni-jfitzger:
  - Added multi_threading system tests `test_multi_threading_ivi_synchronized_wrapper_releases_lock` that will hang if an unmanaged callerHasLock pointer is passed to _LockSession, _UnlockSession entry points.
    - Tried various timeout mechanisms, but could not find a way to test the behavior without a fail hanging the tests
  - Added multi_threading system test `test_multi_threading_lock_unlock` to validate locking behavior 

Restores the lock, unlock behavior that was present prior to the 1.4.3 release.

* Fix #1888 

* Wrote the NI-SCOPE multi_threading system test and ran without the fix.
  * Test hanged pytest
* Implemented the fix and ran the tests again
  * Test passed and did not hang.
* Wrote tests for the other APIs but did not run them locally. They will be run automatically by the PR.

### DIFF
--- a/src/nidcpower/system_tests/test_system_nidcpower.py
+++ b/src/nidcpower/system_tests/test_system_nidcpower.py
@@ -2,8 +2,6 @@ import os
 import pathlib
 import sys
 import tempfile
-import threading
-import time
 
 import grpc
 import hightime
@@ -12,7 +10,7 @@ import pytest
 import nidcpower
 
 sys.path.insert(0, str(pathlib.Path(__file__).parent.parent.parent / 'shared'))
-from system_test_utilities import GrpcServerProcess  # noqa: E402
+import system_test_utilities  # noqa: E402
 
 
 def pytest_generate_tests(metafunc):
@@ -1067,47 +1065,11 @@ class SystemTests:
 
     # Multi-Threading tests
     def test_multi_threading_lock_unlock(self, session):
-        release_lock = threading.Event()
-
-        def lock_wait_unlock():
-            session.lock()
-            release_lock.wait()
-            session.unlock()
-
-        def lock_unlock():
-            session.lock()
-            session.unlock()
-
-        # test that lock, unlock functions work properly
-        # No need to test locking for driver, but the gRPC_interpeter doesn't use the driver to lock
-        t1 = threading.Thread(target=lock_wait_unlock)
-        t2 = threading.Thread(target=lock_unlock)
-
-        t1.start()
-        t2.start()
-
-        t1.join(0.5)
-        time.sleep(0.5)
-        # t1 is blocked by the event, t2 should be blocked by t1's lock
-        assert t2.is_alive()
-        release_lock.set()
-        t2.join(0.5)
-
-        assert not t1.is_alive()
-        assert not t2.is_alive()
+        system_test_utilities.impl_test_multi_threading_lock_unlock(session)
 
     def test_multi_threading_ivi_synchronized_wrapper_releases_lock(self, session):
-        # test that the 2nd thread doesn't hang
-        t1 = threading.Thread(target=session.commit)
-        t2 = threading.Thread(target=session.commit)
-
-        t1.start()
-        t1.join(0.5)
-        assert not t1.is_alive()
-
-        t2.start()
-        t2.join(0.5)
-        assert not t2.is_alive()
+        system_test_utilities.impl_test_multi_threading_ivi_synchronized_wrapper_releases_lock(
+            session)
 
 
 class TestLibrary(SystemTests):
@@ -1119,7 +1081,7 @@ class TestLibrary(SystemTests):
 class TestGrpc(SystemTests):
     @pytest.fixture(scope='class')
     def grpc_channel(self):
-        with GrpcServerProcess() as proc:
+        with system_test_utilities.GrpcServerProcess() as proc:
             channel = grpc.insecure_channel(f"localhost:{proc.server_port}")
             yield channel
 
--- a/src/nidcpower/system_tests/test_system_nidcpower.py
+++ b/src/nidcpower/system_tests/test_system_nidcpower.py
@@ -3,6 +3,7 @@ import pathlib
 import sys
 import tempfile
 import threading
+import time
 
 import grpc
 import hightime
@@ -1065,8 +1066,38 @@ class SystemTests:
         assert last_compensation_datetime.minute == 0
 
     # Multi-Threading tests
-    def test_multi_threading(self, session):
+    def test_multi_threading_lock_unlock(self, session):
+        release_lock = threading.Event()
+
+        def lock_wait_unlock():
+            session.lock()
+            release_lock.wait()
+            session.unlock()
+
+        def lock_unlock():
+            session.lock()
+            session.unlock()
+
         # test that lock, unlock functions work properly
+        # No need to test locking for driver, but the gRPC_interpeter doesn't use the driver to lock
+        t1 = threading.Thread(target=lock_wait_unlock)
+        t2 = threading.Thread(target=lock_unlock)
+
+        t1.start()
+        t2.start()
+
+        t1.join(0.5)
+        time.sleep(0.5)
+        # t1 is blocked by the event, t2 should be blocked by t1's lock
+        assert t2.is_alive()
+        release_lock.set()
+        t2.join(0.5)
+
+        assert not t1.is_alive()
+        assert not t2.is_alive()
+
+    def test_multi_threading_ivi_synchronized_wrapper_releases_lock(self, session):
+        # test that the 2nd thread doesn't hang
         t1 = threading.Thread(target=session.commit)
         t2 = threading.Thread(target=session.commit)
 
--- a/src/nidcpower/system_tests/test_system_nidcpower.py
+++ b/src/nidcpower/system_tests/test_system_nidcpower.py
@@ -2,6 +2,7 @@ import os
 import pathlib
 import sys
 import tempfile
+import threading
 
 import grpc
 import hightime
@@ -1062,6 +1063,20 @@ class SystemTests:
         assert last_compensation_datetime.day == 1
         assert last_compensation_datetime.hour == 0
         assert last_compensation_datetime.minute == 0
+
+    # Multi-Threading tests
+    def test_multi_threading(self, session):
+        # test that lock, unlock functions work properly
+        t1 = threading.Thread(target=session.commit)
+        t2 = threading.Thread(target=session.commit)
+
+        t1.start()
+        t1.join(0.5)
+        assert not t1.is_alive()
+
+        t2.start()
+        t2.join(0.5)
+        assert not t2.is_alive()
 
 
 class TestLibrary(SystemTests):
--- a/src/nifgen/system_tests/test_system_nifgen.py
+++ b/src/nifgen/system_tests/test_system_nifgen.py
@@ -2,6 +2,7 @@ import os
 import pathlib
 import sys
 import tempfile
+import threading
 import warnings
 
 import fasteners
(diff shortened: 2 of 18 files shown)


ni-jfitzger:
```
    def test_fetch_into(self, multi_instrument_session, fetch_waveform_type):
        test_voltage = 1.0
        test_record_length = 2000
        test_num_channels = 2
        test_starting_record_number = 2
        test_num_records_to_acquire = 5
        test_num_records_to_fetch = test_num_records_to_acquire - test_starting_record_number
        waveform = numpy.ndarray(test_num_channels * test_num_records_to_fetch * test_record_length, dtype=fetch_waveform_type)
        # Initialize with NaN so we can later verify all samples were overwritten by the driver.
>       waveform.fill(float('nan'))
E       ValueError: cannot convert float NaN to integer
```

We're testing with numpy 1.21.6.
Not sure yet, what exactly changed.

ni-jfitzger:
> ```
>     def test_fetch_into(self, multi_instrument_session, fetch_waveform_type):
>         test_voltage = 1.0
>         test_record_length = 2000
>         test_num_channels = 2
>         test_starting_record_number = 2
>         test_num_records_to_acquire = 5
>         test_num_records_to_fetch = test_num_records_to_acquire - test_starting_record_number
>         waveform = numpy.ndarray(test_num_channels * test_num_records_to_fetch * test_record_length, dtype=fetch_waveform_type)
>         # Initialize with NaN so we can later verify all samples were overwritten by the driver.
> >       waveform.fill(float('nan'))
> E       ValueError: cannot convert float NaN to integer
> ```
> 
> We're testing with numpy 1.21.6. Not sure yet, what exactly changed.

It's actually 1.24, that we're testing with.
Created #1891 to fix this.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -381,6 +382,20 @@ def test_configure_trigger_window(self, multi_instrument_session):
         assert trigger_source == multi_instrument_session.trigger_source
         assert niscope.TriggerWindowMode.ENTERING == multi_instrument_session.trigger_window_mode
 
+    # Multi-Threading tests
+    def test_multi_threading(self, multi_instrument_session):
```
marcoskirsch, line 386: I'm having difficulty understanding these tests.
You call commit in two different threads on the same session. But you don't do it at the same time, do you?

I think a reasonable test for the locking would be to:

thread 1:
* lock the session
* get an attribute
* wait 0.5 seconds
* unlock the session

thread 2:
* lock the session
* get an attribute
* wait 0.5 seconds
* unlock the session

Launch both threads as close to each other as you can, confirm total time is at least 1 second and maybe that one attriubte you could get right away but the other one took about 0.5 seconds.

bkeryan: FYI, timing based tests are often flaky, especially in PR/CI pipelines. 

Some other approaches:
- Transfer control between threads by waiting "forever" on events or semaphores. To avoid literally waiting forever, use a long timeout (like 60 seconds or longer) and assert that wait() or acquire() returns True. 
- Instead of measuring time, you could do a read/modify/write operation (like incrementing a count variable) and verify that it has the correct result. Unfortunately this may still require sleeping in the middle of the read/modify/write in order to reliably fail.

ni-jfitzger: >You call commit in two different threads on the same session. But you don't do it at the same time, do you?

Not at the same time, no. That's not necessary to reproduce the issue.
threading.Thread.join() actually blocks the calling thread until the target thread completes (or a specified timeout occurs). So the first thread completes before the second thread runs.

It doesn't seem unreasonable to also test lock and unlock directly, though I think these tests still have value for their use of the ivi wrapper. I don't understand what value is added by getting an attribute in your proposed test. [The LockSession call blocks until all other threads release their locks](https://www.ni.com/docs/en-US/bundle/ni-scope-c-api-ref/page/scopefunc/cviniscope_locksession.html).

I don't like the idea of checking timing; it may introduce flakiness. Also, I don't think it's necessary to test the locking behavior so rigidly; I don't believe there's anything we could do to the _LockSession call that would allow it to complete without a lock, so long as there isn't a driver bug and our goal isn't to test the driver. As long as we trust that we're calling the right entrypoint, I think it's sufficent to have:

* a function that calls lock(), then unlock()
* 2 threads that call that function
* start thread 1
* start thread 2
* join thread 1
* join thread 2

I'm not opposed to sleeping between the lock and unlock calls.

ni-jfitzger: >Also, I don't think it's necessary to test the locking behavior so rigidly

I'm relenting on this point, because the gRPC Interpreter doesn't use the driver functions for locking.
@bkeryan's suggestion of Events was helpful. I've renamed my initial new test and added another test based on this discussion that utilizes an event to keep thread 1 from unlocking before we check the status of thread 2.

marcoskirsch: The test is specific to a bug and as written reproduced the bug, there was a deadlock. Now you confirm it's fixed and this is great.

> It doesn't seem unreasonable to also test lock and unlock directly

That is my point.

> though I think these tests still have value for their use of the ivi wrapper.

Please clarify this statement.

> I don't like the idea of checking timing; it may introduce flakiness.

Fair point. There's other, better ways to test this.

> so long as there isn't a driver bug and our goal isn't to test the driver. 

This is true. There was a time though in which lock/unlock from the Python APIs didn't actually forward it to the driver.

> As long as we trust that we're calling the right entrypoint

`nifake` unit tests can confirm this.

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -1063,6 +1065,50 @@ def test_get_lcr_compensation_last_date_and_time(self, session, compensation_typ
...
+            session.lock()
+            release_lock.wait()
+            session.unlock()
+
+        def lock_unlock():
+            session.lock()
+            session.unlock()
+
+        # test that lock, unlock functions work properly
+        # No need to test locking for driver, but the gRPC_interpeter doesn't use the driver to lock
```
marcoskirsch, line 1082: As written, I don't think this sentence adds that much value. Are you saying something like:

"We don't need to test locking on the driver because thats's out of scope for test coverage in nimi-python, but we do need to test the `gRPC_interpeter` (is that the exact name?) because it implements locking in Python" ?

You are still testing locking on the driver, aren't you?

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -1063,6 +1065,50 @@ def test_get_lcr_compensation_last_date_and_time(self, session, compensation_typ
...
+
+        t1.start()
+        t2.start()
+
+        t1.join(0.5)
+        time.sleep(0.5)
+        # t1 is blocked by the event, t2 should be blocked by t1's lock
+        assert t2.is_alive()
+        release_lock.set()
+        t2.join(0.5)
```
marcoskirsch, line 1094: don't use a timeout here, or if you insist, make it enormous like 10 s

src/nifgen/system_tests/test_system_nifgen.py
```diff
@@ -475,6 +477,50 @@ def test_create_advanced_arb_sequence_wrong_size(self, session_5421):
         assert exc_info.value.args[0] == 'Length of marker_location_array and waveform_handles_array parameters do not match.'
         assert str(exc_info.value) == 'Length of marker_location_array and waveform_handles_array parameters do not match.'
 
+    # Multi-Threading tests
```
marcoskirsch, line 480: All these tests are copied/pasted without any driver-specific customization.
Did you look into single sourcing?

marcoskirsch (changes requested):

src/niscope/system_tests/test_system_niscope.py
marcoskirsch: > I don't understand what value is added by getting an attribute in your proposed test.

Getting an attribute or making any other call is a way to prove that those functions are doing locking correctly. I proposed getting an attribute because:

1. it's identical across drivers
2. it's usually lightweight

Getting a string type attribute would be best, since it _better_ be grabbing the lock because it needs to do the IVI dance.

src/nidcpower/system_tests/test_system_nidcpower.py
ni-jfitzger: Yes, that was what I was saying. But you're right, I'm testing it anyway, so the sentence doesn't add much value.
The correct name is GrpcStubInterpreter. I should use the right name in our comments.

I'll delete this comment.

src/nidcpower/system_tests/test_system_nidcpower.py
ni-jfitzger: It feels a little uncomfortable setting no timeout for something that might block forever, but I suppose since in my testing a timeout didn't keep _LockSession from blocking, there isn't much point.

src/nifgen/system_tests/test_system_nifgen.py
ni-jfitzger: >All these tests are copied/pasted without any driver-specific customization.

Not all of them. `test_multi_threading_ivi_synchronized_wrapper_releases_lock` uses `commit` for some drivers and `abort` for those that don't have a `commit` function, though perhaps it could use `abort` for all of them.

>Did you look into single sourcing?

I didn't. Maybe I can shove them all into system_test_utilities.

src/nidcpower/system_tests/test_system_nidcpower.py
ni-jfitzger: Removed join timeouts, increased sleep time and added another threading.Event.

src/nifgen/system_tests/test_system_nifgen.py
ni-jfitzger: These tests are now single-sourced from system_test_utilities.

src/nidcpower/system_tests/test_system_nidcpower.py
ni-jfitzger: Comment deleted.

PR #1893, opened 2023-03-10
mock all library functions

ni-jfitzger:

'library-only' functions have cfunc calls in library.py, but they were not being mocked due to extra filtering of the function list with the codegen_filter used by session.py and _library_interpreter.py.
This currently affects one function: `niDigital_FetchCaptureWaveformU32`.
With this change, that function is mocked.

None

### DIFF
--- a/build/templates/_mock_helper.py.mako
+++ b/build/templates/_mock_helper.py.mako
@@ -30,7 +30,7 @@ class MockFunctionCallError(Exception):
 class SideEffectsHelper(object):
     def __init__(self):
         self._defaults = {}
-% for func_name in sorted(helper.filter_codegen_functions(functions)):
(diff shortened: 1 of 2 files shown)


build/templates/_mock_helper.py.mako
```diff
@@ -30,7 +30,7 @@ class MockFunctionCallError(Exception):
 class SideEffectsHelper(object):
     def __init__(self):
         self._defaults = {}
-% for func_name in sorted(helper.filter_codegen_functions(functions)):
+% for func_name in sorted(functions):
```
ni-jfitzger, line 33: We didn't need any filtering here or later because we already filter functions at line 15.

marcoskirsch (approved):

PR #1902, opened 2023-03-30
SetRuntimeEnv added in NI-Fake metadata and called in _library_interpreter

victorvadan:

Regenerated NI-FAKE API metadata.
Added call to <module>_SetRuntimeEnvironment in _library_singleton template, with python implementation and python version.

Added new unit tests and replaced an old one that was failing. Also updated the tests failing because of metadata changes since the last generation, unrelated to this change.
Tested on a machine with the last released nidcpower to check no exceptions are raised and block the get method. Then, reinstalled a newer version of nidcpower that has support for niDCPower_SetRuntimeEnvironment to test logging of the python interpreter and version.
The logging worked and the xml files contained the details about the interpreter and version.


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ All notable changes to this project will be documented in this file.
     * #### Removed
 * ### `nidcpower` (NI-DCPower)
     * #### Added
+        * Pass Python interpreter information to driver runtimes that support it. This is used by NI in order to better understand client usage.
```
ni-jfitzger, line 41: This is True but @marcoskirsch do you think there's a better way to explain it?

victorvadan: Rephrased it.

build/templates/_library_singleton.py.mako
```diff
@@ -55,5 +55,17 @@ def get():
...
             _instance = _library.Library(ctypes_library)
+            try:
+                runtime_env_ctype = ctypes.create_string_buffer(platform.python_implementation().encode('windows-1251'))  # case C010
+                version_ctype = ctypes.create_string_buffer(platform.python_version().encode('windows-1251'))  # case C010
+                reserved_ctype = ctypes.create_string_buffer(''.encode('windows-1251'))  # case C010
+                _instance.${config['c_function_prefix']}SetRuntimeEnvironment(
+                    runtime_env_ctype,
+                    version_ctype,
+                    reserved_ctype,
+                    reserved_ctype
```
ni-jfitzger, line 66: @marcoskirsch would it be considered best practice to have reserved1_ctype, reserved2_ctype or do you think it won't matter?

generated/nitclk/nitclk/_library_singleton.py
```diff
@@ -51,5 +51,17 @@ def get():
...
+                version_ctype = ctypes.create_string_buffer(platform.python_version().encode('windows-1251'))  # case C010
+                reserved_ctype = ctypes.create_string_buffer(''.encode('windows-1251'))  # case C010
+                _instance.niTClk_SetRuntimeEnvironment(
+                    runtime_env_ctype,
+                    version_ctype,
+                    reserved_ctype,
+                    reserved_ctype
+                )
+            except errors.DriverTooOldError:
+                pass
```
ni-jfitzger, line 65: Will NI-TClk support this in the future?

build/templates/_library_singleton.py.mako
marcoskirsch: Doesn't matter I think.

generated/nimodinst/nimodinst/_library_singleton.py
```diff
@@ -51,5 +51,17 @@ def get():
...
+                version_ctype = ctypes.create_string_buffer(platform.python_version().encode('windows-1251'))  # case C010
+                reserved_ctype = ctypes.create_string_buffer(''.encode('windows-1251'))  # case C010
+                _instance.niModInst_SetRuntimeEnvironment(
+                    runtime_env_ctype,
+                    version_ctype,
+                    reserved_ctype,
+                    reserved_ctype
+                )
+            except errors.DriverTooOldError:
+                pass
```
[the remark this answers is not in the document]
marcoskirsch: Unclear, if I had to guess I would say no. But I don't think it changes anything in this PR.

generated/nidcpower/nidcpower/_library_singleton.py
```diff
@@ -51,5 +51,17 @@ def get():
             except OSError:
                 raise errors.DriverNotInstalledError()
             _instance = _library.Library(ctypes_library)
+            try:
+                runtime_env_ctype = ctypes.create_string_buffer(platform.python_implementation().encode('windows-1251'))  # case C010
```
bkeryan, line 55: I think it's unlikely that the Python implementation and version will contain Cyrillic characters. Consider using ascii, or at least windows-1252 (Latin 1).

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,49 @@ def test_read_with_warning(self):
...
+    def test_abort_not_present_in_driver(self):
+        # Test that missing SetRuntimeEnvironment from python impl. does not affect library interpreter init
+        delattr(self.patched_library, 'niFake_Abort')
+        interpreter = self.get_initialized_library_interpreter()
+        try:
+            interpreter.abort()
+            assert False
+        except nifake.errors.DriverTooOldError as e:
+            message = e.args[0]
+            assert message == 'A function was not found in the NI-FAKE runtime. Please visit http://www.ni.com/downloads/drivers/ to download a newer version and install it.'
```
ni-jfitzger, line 453: This appears to work, but it's replacing a test for a function missing from the driver, not the Python Implementation

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,49 @@ def test_read_with_warning(self):
...
+        self.patched_library_singleton_lib.start()
+        self.get_initialized_library_interpreter()
+
+    def test_set_runtime_environment_not_present_in_driver(self):
+        # Test that missing SetRuntimeEnvironment from driver does not affect library interpreter init
+        self.patched_library_singleton_get.stop()
+        delattr(self.patched_library, 'niFake_SetRuntimeEnvironment_cfunc')
+        self.patched_library_singleton_lib = patch('nifake._library.Library', return_value=self.patched_library)
+        self.patched_library_singleton_lib.start()
+        self.get_initialized_library_interpreter()
```
ni-jfitzger, line 442: I tried having _library_singleton.get() reraise after catching a DriverTooOldError and this still didn't fail. Are you sure this properly simulates the absence of a function from the library?

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,49 @@ def test_read_with_warning(self):
             assert test_error_desc in str(w[0].message)
             assert f'Warning {test_error_code} occurred.' in str(w[0].message)
 
+    def test_library_singleton_get_returns_same_instance(self):
+        # Test that if library singleton get is called multiple times it returns the same instance
+        interpreter = nifake._library_interpreter.LibraryInterpreter('windows-1251')
```
marcoskirsch, line 414: I think this test will be more readable if you

1. create two LibraryInterpreter objects
2. assert that the LibraryInterpreter objects are not the same.
3. assert that the _library objects are the same instance

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,49 @@ def test_read_with_warning(self):
...
+        self.patched_library_singleton_get.stop()
+        self.patched_library_singleton_lib = patch('nifake._library.Library', return_value=self.patched_library)
+        self.patched_library.niFake_SetRuntimeEnvironment.side_effect = self.side_effects_helper.niFake_SetRuntimeEnvironment
+        self.patched_library_singleton_lib.start()
+        self.get_initialized_library_interpreter()
+        self.patched_library.niFake_SetRuntimeEnvironment.assert_called_once()
+        self.patched_library_singleton_lib.stop()
+        self.patched_library_singleton_get.start()
+
+    def test_set_runtime_environment_not_defined(self):
```
marcoskirsch, line 428: for consistent terminology:

`test_set_runtime_environment_not_present_in_library_interpreter`

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,49 @@ def test_read_with_warning(self):
...
+        # Test that missing SetRuntimeEnvironment from python impl. does not affect library interpreter init
+        self.patched_library_singleton_get.stop()
+        delattr(self.patched_library, 'niFake_SetRuntimeEnvironment')
+        self.patched_library_singleton_lib = patch('nifake._library.Library', return_value=self.patched_library)
+        self.patched_library_singleton_lib.start()
+        self.get_initialized_library_interpreter()
+
+    def test_set_runtime_environment_not_present_in_driver(self):
```
marcoskirsch, line 436: `test_set_runtime_environment_not_present_in_driver_runtime`

We always refer to the "driver" as driver runtime in nimi-python.

marcoskirsch (changes requested):

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,49 @@ def test_read_with_warning(self):
             assert test_error_desc in str(w[0].message)
             assert f'Warning {test_error_code} occurred.' in str(w[0].message)
 
+    def test_library_singleton_get_returns_same_instance(self):
```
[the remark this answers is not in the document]
bkeryan: Even better, use given/when/then syntax as described in NI's Python testing conventions: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions

This encourages you to test only one thing in each test case, and to clearly describe that thing in the test case name.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,49 @@ def test_read_with_warning(self):
...
+    def test_set_runtime_environment_not_present_in_driver(self):
+        # Test that missing SetRuntimeEnvironment from driver does not affect library interpreter init
+        self.patched_library_singleton_get.stop()
+        delattr(self.patched_library, 'niFake_SetRuntimeEnvironment_cfunc')
+        self.patched_library_singleton_lib = patch('nifake._library.Library', return_value=self.patched_library)
+        self.patched_library_singleton_lib.start()
+        self.get_initialized_library_interpreter()
+
+    def test_abort_not_present_in_driver(self):
+        # Test that missing SetRuntimeEnvironment from python impl. does not affect library interpreter init
```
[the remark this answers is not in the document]
ni-jfitzger: There was another test outside of the class that started failing. This is his attempt to replace that test, which was using abort.

build/templates/_library_singleton.py.mako
```diff
@@ -56,9 +56,9 @@ def get():
                 raise errors.DriverNotInstalledError()
             _instance = _library.Library(ctypes_library)
             try:
-                runtime_env_ctype = ctypes.create_string_buffer(platform.python_implementation().encode('windows-1251'))  # case C010
-                version_ctype = ctypes.create_string_buffer(platform.python_version().encode('windows-1251'))  # case C010
-                reserved_ctype = ctypes.create_string_buffer(''.encode('windows-1251'))  # case C010
+                runtime_env_ctype = ctypes.create_string_buffer(platform.python_implementation().encode('windows-1251'))  # case C020
```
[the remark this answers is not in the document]
victorvadan: To do that I would need to change the generated metadata for nidcpower (or update the hapigen files), the function is library-only.

build/templates/_library_singleton.py.mako
```diff
@@ -55,5 +55,17 @@ def get():
             except OSError:
                 raise errors.DriverNotInstalledError()
             _instance = _library.Library(ctypes_library)
+            try:
+                runtime_env_ctype = ctypes.create_string_buffer(platform.python_implementation().encode('windows-1251'))  # case C010
+                version_ctype = ctypes.create_string_buffer(platform.python_version().encode('windows-1251'))  # case C010
+                reserved_ctype = ctypes.create_string_buffer(''.encode('windows-1251'))  # case C010
```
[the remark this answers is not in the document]
victorvadan: Updated

build/templates/_library_singleton.py.mako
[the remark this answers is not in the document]
victorvadan: Also, I see we've done PR 454088 to change setRuntimeEnvironment from private to library-only in niDCPower.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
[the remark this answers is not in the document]
victorvadan: I removed this, since there is another test that does the same thing for SetRuntimeEnvironment.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
victorvadan: Reworked this test

generated/nifake/nifake/unit_tests/test_library_interpreter.py
victorvadan: I added a similar test for SetRuntimeEnvironment and deleted this one.

bkeryan (approved):

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,61 @@ def test_read_with_warning(self):
...
+        self.patched_library_singleton_lib.start()
+        nifake._library_singleton._instance = None
+        interpreter = self.get_initialized_library_interpreter()
+        try:
+            interpreter._library.niFake_SetRuntimeEnvironment('', '', '', '')
+            assert False
+        except nifake.errors.DriverTooOldError:
+            pass
+        nifake._library_singleton._instance = None
+        self.get_initialized_library_interpreter()
```
ni-jfitzger, line 462: Why call this twice?

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +409,61 @@ def test_read_with_warning(self):
...
+        delattr(self.patched_library, 'niFake_SetRuntimeEnvironment')
+        self.patched_library_singleton_lib = patch('nifake._library.Library', return_value=self.patched_library)
+        self.patched_library_singleton_lib.start()
+        nifake._library_singleton._instance = None
+        interpreter = self.get_initialized_library_interpreter()
+        try:
+            interpreter._library.niFake_SetRuntimeEnvironment('', '', '', '')
+            assert False
+        except nifake.errors.DriverTooOldError:
+            pass
```
ni-jfitzger, line 460: What's going on here?
Are you tring to confirm that DriverTooOldError is raised, as a way to validate that we're properly simulating the abasence of the function from the runtime?

If so, I recommend using `with pytest.raises(nifake.errors.DriverTooOldError):` instead of try, except.
See https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

build/templates/_library_interpreter.py.mako
```diff
@@ -75,6 +79,19 @@ class LibraryInterpreter(object):
     def __init__(self, encoding):
         self._encoding = encoding
         self._library = _library_singleton.get()
+        % if 'SetRuntimeEnvironment' in functions:
+        try:
+            runtime_env_ctype = platform.python_implementation()
+            version_ctype = platform.python_version()
```
marcoskirsch: This variable is not ctype just a Python string, no?
The name is misleading, consider just making the function call directly in the set_runtime_environment invocation.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +411,51 @@ def test_read_with_warning(self):
...
+        self.patched_library = self.PatchedLibrary(TypesLibrary)
+        self.patched_library_singleton_lib = patch('nifake._library.Library', return_value=self.patched_library)
+        self.patched_library_singleton_lib.start()
+        with pytest.raises(nifake.errors.DriverTooOldError):
+            self.get_initialized_library_interpreter()._library._get_library_function('niFake_SetRuntimeEnvironment')
+
+        self.patched_library_singleton_lib.stop()
+        self.patched_library_singleton_get.start()
+
+    def test_set_runtime_environment_not_present_in_library(self):
```
marcoskirsch, line 435: Is this still a valid unit test?
We are no longer trying to call `set_runtime_environment` when it isn't present in the interpreter. Who/what/under what circumstance would try to call it in the library class?

marcoskirsch: maybe I misunderstood the test case, use "given_when_then" naming to clarify.

marcoskirsch (changes requested):

build/templates/_library_interpreter.py.mako
```diff
@@ -28,6 +32,8 @@ import ${module_name}.errors as errors
 import ${module_name}.${c['file_name']} as ${c['file_name']}  # noqa: F401
 % endfor
 
+_was_runtime_env_set = None
```
ni-jfitzger, line 35: Let's only set this if we need it.
```suggestion
% if 'SetRuntimeEnvironment' in functions:
_was_runtime_env_set = None
% endif
```

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +411,33 @@ def test_read_with_warning(self):
...
+    def test_set_runtime_environment_is_called_if_present(self):
+        nifake._library_interpreter._was_runtime_env_set = None
+        self.get_initialized_library_interpreter()
+        self.patched_library.niFake_SetRuntimeEnvironment.assert_called_once()
+
+    def test_set_runtime_environment_not_present_in_driver_runtime(self):
+        class TypesLibrary:
+            item = ""
+
+        nifake._library_interpreter._was_runtime_env_set = None
```
ni-jfitzger, line 429: Are you allowed to do this without using the "global" keyword, first?

ni-jfitzger: Hmm, you're not initializing the interpreter first, so it's probably still None.
Maybe you should just

```suggestion
        assert nifake._library_interpreter._was_runtime_env_set is None
```

generated/nifake/nifake/unit_tests/test_library_interpreter.py
victorvadan: I left the test because I removed the one with niFake_Abort missing. I renamed it to be more specific.

src/nifake/unit_tests/test_library_interpreter.py
victorvadan: That will fail because the value is true at this point, because that part of the code was executed already. This ensures the set runtime environment function is called.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +411,35 @@ def test_read_with_warning(self):
...
+        class TypesLibrary:
+            item = ""
+
+        nifake._library_interpreter._was_runtime_env_set = None
+        self.patched_library._library = TypesLibrary()
+        interpreter = self.get_initialized_library_interpreter()
+        with pytest.raises(nifake.errors.DriverTooOldError):
+            interpreter._library._get_library_function('niFake_SetRuntimeEnvironment')
+
+    def test_abort_not_present_in_library_raises_driver_too_old_error(self):
```
marcoskirsch, line 436: Is this a valid test?
Under which circumstance would we have a function be present in the LibraryInterpreter that tries to call the Library, but the function isn't present in the library?


I wonder if we have coverage instead for the valid scenarion in which a new nimi-python module contains a function but the installed driver runtime is older and doesn't have the function.

src/nifake/unit_tests/test_library_interpreter.py
ni-jfitzger: Are you sure? I don't see us initilizing a LibraryInterpreter in `setup_method`, so how would the value have changed?

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -410,6 +411,35 @@ def test_read_with_warning(self):
...
+    def test_library_interpreter_always_uses_same_library_instance(self):
+        interpreter1 = nifake._library_interpreter.LibraryInterpreter('windows-1251')
+        interpreter2 = nifake._library_interpreter.LibraryInterpreter('windows-1251')
+        assert interpreter1 is not interpreter2
+        assert interpreter1._library is interpreter2._library
+
+    def test_set_runtime_environment_is_called_once_if_present(self):
+        nifake._library_interpreter._was_runtime_env_set = None
+        self.get_initialized_library_interpreter()
+        nifake._library_interpreter.LibraryInterpreter('windows-1251')
```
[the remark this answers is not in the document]
victorvadan: I am constructing 2, the first is inside of get_initialized_library_interpreter.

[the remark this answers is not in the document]
ni-jfitzger: I believe he has constructed multiple LibraryInterpreter objects.
`get_initialized_library_interpreter` constructs and returns one.
We could probably call that twice, though.

In fact, we could probably replace all of our direct initializations with calls to that method.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
ni-jfitzger: Yes @victorvadan is right. The above test does cover the case we care about. I think we can just delete this case.

ni-jfitzger (approved):

PR #1905, opened 2023-03-31
Add set_runtime_environment to nidcpower LibraryInterpreter class

ni-jfitzger:
- ~[] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~

This change adds set_runtime_environment to the nidcpower LibraryInterpreter class, so that we don't have to handcode the ctype manipulation.

None

### DIFF
--- a/src/nidcpower/metadata/functions.py
+++ b/src/nidcpower/metadata/functions.py
@@ -3496,7 +3496,7 @@ functions = {
         'returns': 'ViStatus'
     },
     'SetRuntimeEnvironment': {
-        'codegen_method': 'library-interpreter-and-library-only',
+        'codegen_method': 'private',
         'documentation': {
             'description': 'TBD'
         },
--- a/src/nidcpower/metadata/functions.py
+++ b/src/nidcpower/metadata/functions.py
@@ -3496,11 +3496,19 @@ functions = {
         'returns': 'ViStatus'
     },
     'SetRuntimeEnvironment': {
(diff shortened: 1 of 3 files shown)


src/nidcpower/metadata/functions.py
```diff
@@ -3496,11 +3496,19 @@
         'returns': 'ViStatus'
     },
     'SetRuntimeEnvironment': {
-        'codegen_method': 'library-only',
+        'codegen_method': 'private',
```
marcoskirsch, line 3499: I'm guessing 'private' or 'public' here makes no difference since you aren't rendering in `Session`? Maybe add a comment.

marcoskirsch (approved):

marcoskirsch (dismissed):
This is change on the copied metadata.
We should use add_on to override or upload new metadata.

ni-jfitzger:
> This is change on the copied metadata. We should use add_on to override or upload new metadata.

Done

src/nidcpower/metadata/functions.py
ni-jfitzger: Added an actual comment in the functions_addon override. The same comment will be added to our internal metadata, but it won't make it into the exported metadata.

marcoskirsch (approved):

ni-jfitzger:
Added https://github.com/ni/nimi-python/issues/1906 for tracking the TODO.

PR #1910, opened 2023-04-07
Fix niswitch multithreading test warnings

ni-jfitzger:

The [`test_multi_threading_ivi_synchronized_wrapper_releases_lock` test in niswitch system tests is currently raising `pytest.PytestUnhandledThreadExceptionWarning`](https://github.com/ni/nimi-python/actions/runs/4567686137/jobs/8209768533) in Python 3.8 (earliest version where pytest catches these exceptions) and later due to ` niswitch.errors.DriverError: -1074135023: IVI: Function or method not supported.`. This is caused by the simulated device not supporting abort. The other simulated device typically used in niswitch system tests supports abort but will also raise a DriverError if called immediately after initialization due to no scans being in progress.

This change does 2 things:
* Updates tox-system_tests.ini so that `pytest.PytestUnhandledThreadExceptionWarning` will be elevated to an error and fail a test that causes it
* Updates `impl_test_multi_threading_ivi_synchronized_wrapper_releases_lock` to accept the method to call, instead of the session
  * niswitch system tests will call `session.self_test`
  * other apis will continue to call `session.abort`

* Manually tested that the tox change would result in test failures for niswitch.
* Manually tested that the test change would fix the niswitch test failures

### DIFF
--- a/build/templates/tox-system_tests.ini.mako
+++ b/build/templates/tox-system_tests.ini.mako
@@ -105,6 +105,8 @@ passenv =
 
 [pytest]
 addopts = --verbose
+filterwarnings =
+   error::pytest.PytestUnhandledThreadExceptionWarning
 norecursedirs = .* build dist CVS _darcs {arch} *.egg venv
 junit_suite_name = nimi-python
 junit_family = xunit1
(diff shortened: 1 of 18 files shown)


marcoskirsch (approved):
I suppose all the drivers here support `self_test` (takes a long time in NI-Digital) or `reset` so you could have used that instead but this works too.

ni-jfitzger:
We are currently having issues with nimibot_win_32.
This is a simple change which has approval and the completed testing should be good enough, so I am merging this.

PR #1916, opened 2023-04-10
Put bounds on versions of grpc-related dependencies

### LINKED ISSUE #1884: Please document and/or lock version of grpcio-tools and protobuf used for codegen
### Description of issue

It is difficult to tell which version of `grpcio-tools` and `protobuf` were used for the nimi-python protobuf/gRPC codegen.

This matters because the Python `protobuf` module broke compatibility between versions 3.20 and 4.21 (hence the updated major version). This isn't a total compatibility break: 3.19-3.20 codegen works with 4.21 and 4.21 codegen works with 3.20 if you set an environment variable. See https://developers.google.com/protocol-buffers/docs/news/2022-05-06 for more info.

### System report

N/A

### Steps to reproduce issue

N/A

ni-jfitzger:

- Lock the version of grpcio-tools used in codegen to 1.49.1
  - If the latest grpcio-tools is always used, then MeasurementLink users are forced to upgrade to the corresponding grpcio version or newer.
  - [measurementlink-python currently supports grpcio >= 1.49.1](https://github.com/ni/measurementlink-python/blob/main/pyproject.toml#L34) so match grpcio-tools to that minimum version
- Lock the version of grpcio-tools used by "test" command to 1.49.1
- Add optional 'grpcio>=1.49.1,<2.0', 'protobuf>=4.21,<5.0' dependencies for packages that support grpc
  - This documents the versions of grpcio, protobuf that are supported and [gives users a way to install the correct versions of these dependencies without knowing what the dependencies or versions are](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies)
   - Use the same bounds in system tests
   - measurementlink-python currently supports protobuf >= 4.21
   - grpcio version must be >= grpio-tools version used in codegen

* Fix #1884 

The system tests confirm that the new optional dependencies work.


build/templates/tox-system_tests.ini.mako
```diff
@@ -82,8 +82,8 @@ deps =
     ${module_name}-system_tests: fasteners
     ${module_name}-system_tests: pytest-json
 % if grpc_supported:
-    ${module_name}-system_tests: grpcio
-    ${module_name}-system_tests: protobuf
+    ${module_name}-system_tests: grpcio >= 1.49.1, < 2.0
+    ${module_name}-system_tests: protobuf >= 4.21, < 5.0
 % endif
```
bkeryan, line 87: Now that the package has extras, shouldn't you use `extras = grpc` (inside an `%if grpc_supported` block) to enable them?

ni-jfitzger: Yeah, that's probably a better way to it and it would also test our optional dependencies.

generated/nidcpower/tox-system_tests.ini
```diff
@@ -41,8 +41,8 @@ deps =
     nidcpower-system_tests: hightime
     nidcpower-system_tests: fasteners
     nidcpower-system_tests: pytest-json
-    nidcpower-system_tests: grpcio
-    nidcpower-system_tests: protobuf
+    nidcpower-system_tests: grpcio >= 1.49.1, < 2.0
+    nidcpower-system_tests: protobuf >= 4.21, < 5.0
```
bkeryan, line 45: Does nimi-python run any tests without grpcio and protobuf installed?

build/templates/tox-system_tests.ini.mako
```diff
@@ -82,8 +82,8 @@ deps =
     ${module_name}-system_tests: fasteners
```
bkeryan, line 82: FYI, pytest no longer depends on py. https://github.com/advisories/GHSA-w596-4wvx-j9j6

build/templates/tox-system_tests.ini.mako
bkeryan: Likewise, if the main package depends on `hightime`, do the tests need to explicitly install it?

ni-jfitzger: >Likewise, if the main package depends on hightime, do the tests need to explicitly install it?

My philosophy on declaring dependencies is as follows: If you explicitly import something, you should declare an explicit dependency. We explicitly import hightime in our system tests.

generated/nidcpower/tox-system_tests.ini
ni-jfitzger: nimodinst, nise and nitclk all run system tests without grpcio and protobuf installed.

bkeryan: Right, but nimodinst, nise, and nitclk don't support the "grpc" extra. If a package supports extra dependencies, it ought to have some test coverage without the extra dependencies installed to ensure that it still works without them. 

When I asked @mshafer-NI whether nidaqmx-python should declare grpcio and protobuf as extra dependencies, this seemed to be his number one concern with that approach.

build/templates/tox-system_tests.ini.mako
bkeryan: > > Likewise, if the main package depends on hightime, do the tests need to explicitly install it?
> 
> My philosophy on declaring dependencies is as follows: If you explicitly import something, you should declare an explicit dependency. We explicitly import hightime in our system tests.

Ok, that makes sense.

ni-jfitzger: We now depend on .[grpc], instead, so I will mark this resolved.

generated/nidcpower/tox-system_tests.ini
ni-jfitzger: It sounds like you're concerned about a possible pre-existing test gap.
Come to think of it, we run our examples as part of our testing. Those examples don't import grpcio or protobut, but it's in the same venv as our system tests, so the extra dependencies **are** installed.

bkeryan (approved):

build/templates/tox-system_tests.ini.mako
ni-jfitzger: Good call. Created #1921 .

tox.ini
```diff
@@ -126,7 +126,7 @@ deps =
     test: mako
     test: numpy
     test: hightime
-    test: grpcio-tools
+    test: grpcio-tools == 1.49.1
```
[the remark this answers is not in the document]
marcoskirsch: try it in a separate PR, maybe

marcoskirsch (approved):

PR #1922, opened 2023-04-11
Update metadata for non-nidcpower APIs for nimi-python 1.4.4 release

ni-jfitzger:

- Metadata is updated from the latest final export or later for all APIs other than nidcpower
  - nidcpower has significant contributions, so it is handled separately
- nifake metadata is updated to remove an override in src/nifake/metadata/functions_addon.py
- proto files are updated to the latest available in https://github.com/ni/grpc-device

None

None


generated/niscope/niscope/niscope_pb2.py
```diff
@@ -15,7 +15,7 @@
...
...a .niscope_grpc.SampleModeResponse\x12O\n\nSampleRate\x12\x1f.niscope_grpc.SampleRateRequest\x1a .niscope_grpc.SampleRateResponse\x12I\n\x08SelfTest\x12\x1d.niscope_grpc.SelfTestRequest\x1a\x1e.niscope_grpc.SelfTestResponse\x12v\n\x17SendSoftwareTriggerEdge\x12,.niscope_grpc.SendSoftwareTriggerEdgeRequest\x1a-.niscope_grpc.SendSoftwareTriggerEdgeResponse\x12p\n\x15SetAttributeViBoolean\x12*.niscope_grpc.SetAttributeViBooleanRequest\x1a+.niscope_grpc.SetAttributeViBooleanResponse\x12j\n\x13SetAttributeViInt32\x12(.niscope_grpc.SetAttributeViInt32Request\x1a).niscope_grpc.SetAttributeViInt32Response\x12j\n\x13SetAttributeViInt64\x12(.niscope_grpc.SetAttributeViInt64Request\x1a).niscope_grpc.SetAttributeViInt64Response\x12m\n\x14SetAttributeViReal64\x12).niscope_grpc.SetAttributeViReal64Request\x1a*.niscope_grpc.SetAttributeViReal64Response\x12p\n\x15SetAttributeViSession\x12*.niscope_grpc.SetAttributeViSessionRequest\x1a+.niscope_grpc.SetAttributeViSessionResponse\x12m\n\x14SetAttributeViString\x12).niscope_grpc.SetAttributeViStringRequest\x1a*.niscope_grpc.SetAttributeViStringResponseB?\n\x11\x63om.ni.grpc.scopeB\x07NiScopeP\x01\xaa\x02\x1eNationalInstruments.Grpc.Scopeb\x06proto3')
```
ni-jfitzger, line 18: The changes to this file appear significant.
Should we be concerned?

src/niscope/metadata/niscope.proto
```diff
@@ -1439,6 +1439,7 @@ message FetchArrayMeasurementRequest {
     ArrayMeasurement array_meas_function = 4;
     sint32 array_meas_function_raw = 5;
   }
+  optional sint32 meas_wfm_size = 6;
```
bkeryan, line 1442: This is from https://github.com/ni/grpc-device/pull/844

https://github.com/ni/nimi-python/blob/master/generated/niscope/niscope/_grpc_stub_interpreter.py#L252 currently ignores the `measurement_waveform_size` parameter.

marcoskirsch:
Would it be too much to break this PR up?

ni-jfitzger:
>Would it be too much to break this PR up?

I can look at doing that, tomorrow. Maybe I can just pull in commits to different branches.

generated/niscope/niscope/niscope_pb2.py
bkeryan: I don't think you should worry in this case.

This byte array is a serialized description of the `niscope.proto` file. Adding a field to a message inserted some bytes in the middle, shifted the end of each line to the beginning of the next line, and increased all of the offsets listed below.

Using a different diff tool such as Beyond Compare shows this more clearly, but it's still hard to verify that most of the byte array didn't change.
image

You should worry when updating a `.proto` file that is duplicated between multiple Python packages, such as `session.proto`. If `session.proto`'s descriptor doesn't match between nimi-python, measurementlink-python, and nidaqmx-python, users will get errors when they load more than one of these packages. We have technical debt AB#2250775 about this.

ni-jfitzger:
@marcoskirsch this PR has now been split for everything but niscope. I will work on a new PR for niscope this afternoon.

ni-jfitzger:
Splitting the PR

PR #1924, opened 2023-04-12
Specify grpcio, protobuf test deps, instead of grpcio-tools

ni-jfitzger:

We were specifying grpcio-tools as a test dep for our unit tests in the tox.inis, but it's not actually used. It was only being used as an indirect way of pulling compatible grpcio and protobuf deps.

This change replaces the grpcio-tools test dep with compatible grpcio, protobuf test deps.

None

PR Checks


ni-jfitzger:
Based on the travis-ci failure, it seems we weren't just using grpcio-tools as a test dependency to get grpcio, we were also using it to get protobuf. [Both are dependencies in the setup.py.](https://github.com/grpc/grpc/blob/master/tools/distrib/python/grpcio_tools/setup.py)

ni-jfitzger:
Alternatively to this implementation, we could continue to use grpcio-tools as our test dep (maybe less work when updating dep versions) and just add a comment that it's use to pull in compatible versions of grpcio and protobuf.

marcoskirsch (approved):

PR #1948, opened 2023-04-17
Stop using deprecated setup.py commands

ni-jfitzger:

Use of setuptools for building packages is no longer recommended. See [Why you shouldn't invoke setup.py directly.](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) That page is [linked from the setuptools changelog.](https://setuptools.pypa.io/en/latest/history.html#v58-3-0) Both pages recommend use of the [`build`](https://pypa-build.readthedocs.io/en/stable/) package for building, which is what this PR sets us up to use.

In the rules.mak, we'll build the wheel directly from the sdist. Not entirely sure if that's what we were doing before, but it's the default build behavior for a reason.

None

PR Checks


generated/nidigital/tox-system_tests.ini
```diff
@@ -37,6 +37,7 @@ commands =
     nidigital-coverage: coverage report --rcfile=../../tools/coverage_system_tests.rc
 
 deps =
+    nidigital-wheel_dep: build
     nidigital-wheel_dep: packaging
```
ni-jfitzger, line 41: We might not need the packaging dep. I haven't tried removing it.

marcoskirsch: try it in separate PR

marcoskirsch (approved):

PR #1950, opened 2023-04-21
Clip the gaussian noise in nifgen_script to avoid errors

ni-jfitzger:

A PR Check recently failed with 

>        if _is_error(code):
>>           raise DriverError(code, description)
>E           nifgen.errors.DriverError: -1074135024: IVI: Invalid value for parameter or property.
E           
E           Invalid Value: 1.008848
E           Maximum Value: 1.0
E           Minimum Value: -1.0
E           Index: 825

while writing the GAUSSIAN_NOISE waveform in nifgen_script.py.

This was a random failure, because we are generating random noise, which may randomly fall outside of our bounds, despite the low standard deviation of 0.2 from the mean of 0.

To prevent this from a happening again, this change will clamp the generated values between -1.0 and 1.0.

`max(upper_bound,  min(x, lower_bound))` (the standard way of doing this) was considered but it was very hard to read because of the random.gauss() call.

None

None

### DIFF
--- a/src/nifgen/examples/nifgen_script.py
+++ b/src/nifgen/examples/nifgen_script.py
@@ -21,7 +21,7 @@ RAMP_UP = [x / (NUMBER_OF_SAMPLES) for x in range(NUMBER_OF_SAMPLES)]
 RAMP_DOWN = [-1.0 * x for x in RAMP_UP]
 SQUARE_WAVE = [1.0 if x < (NUMBER_OF_SAMPLES / 2) else -1.0 for x in range(NUMBER_OF_SAMPLES)]
 SAWTOOTH_WAVE = RAMP_UP[::2] + [(-1 + x) for x in RAMP_UP][::2]
-GAUSSIAN_NOISE = [random.gauss(0, 0.2) for x in range(NUMBER_OF_SAMPLES)]
+GAUSSIAN_NOISE = [sorted(-1.0, random.gauss(0, 0.2), 1.0)[1] for x in range(NUMBER_OF_SAMPLES)]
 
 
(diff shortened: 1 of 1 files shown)


src/nifgen/examples/nifgen_script.py
```diff
@@ -21,7 +21,7 @@
 RAMP_DOWN = [-1.0 * x for x in RAMP_UP]
 SQUARE_WAVE = [1.0 if x < (NUMBER_OF_SAMPLES / 2) else -1.0 for x in range(NUMBER_OF_SAMPLES)]
 SAWTOOTH_WAVE = RAMP_UP[::2] + [(-1 + x) for x in RAMP_UP][::2]
-GAUSSIAN_NOISE = [random.gauss(0, 0.2) for x in range(NUMBER_OF_SAMPLES)]
+GAUSSIAN_NOISE = [sorted([-1.0, random.gauss(0, 0.2), 1.0])[1] for x in range(NUMBER_OF_SAMPLES)]
```
marcoskirsch, line 24: By clipping the waveform, we are no longer generating noise with a proper gaussian distribution. That makes the example incorrect.

I think we're better off removing the gaussian noise option altogether, rather than being almost but not really correct. This also allows us to stop importing gaussian.

marcoskirsch (changes requested):

src/nifgen/examples/nifgen_script.py
ni-jfitzger: >By clipping the waveform, we are no longer generating noise with a proper gaussian distribution.

Good point.

marcoskirsch (approved):

Issue #1770, opened 2022-06-30
fetch(), read(), and friends return wrong data when called with channel ranges on multi-instrument niscope session


sbethur:
### Description of issue

On a multi-instrument session in `niscope`, any method that returns `MeasurementStats` or ` WaveformInfo` object (fetch(), read(), and friends) returns wrong data when the channel repeatability contains ranges.

### System report

`python -c "import niscope; niscope.print_diagnostic_information()"` output:

 ```
OS:
    Name:      Windows
    Version:   10.0.16299
    Bits:      64
Driver:
    Name:      NI-SCOPE
    Version:   22.5.0.43
Module:
    Name:      niscope
    Version:   1.4.1
Python:
    Version:   3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64 bit (AMD64)]
    Bits:      64
    Is_Venv:   True
    Installed Packages:
        zipp==3.8.0
        typing-extensions==4.2.0
        tomli==2.0.1
        setuptools==47.1.0
        pytest==7.1.2
        pyparsing==3.0.9
        py==1.11.0
        pluggy==1.0.0
        pip==20.1.1
        packaging==21.3
        numpy==1.21.6
        nitclk==1.4.1
        niscope==1.4.1
        nidcpower==1.4.1
        iniconfig==1.1.1
        importlib-metadata==4.12.0
        hightime==0.2.1
        fasteners==0.17.3
        colorama==0.4.5
        attrs==21.4.0
        atomicwrites==1.4.0
 ```

### Steps to reproduce issue

``` 
import niscope

with niscope.Session('FakeDevice1, FakeDevice2', False, True, 'Simulate=1, DriverSetup=Model:5164; BoardType:PXIe') as session:
    test_record_length = 2000
    test_starting_record_number = 2
    test_num_records_to_acquire = 5
    test_num_records_to_fetch = test_num_records_to_acquire - test_starting_record_number
    session.configure_vertical(1.0, niscope.VerticalCoupling.AC)
    session.configure_horizontal_timing(50000000, test_record_length, 50.0, test_num_records_to_acquire, True)
    with session.initiate():
        # Passing case
        test_channels = 'FakeDevice2/0,FakeDevice1/1'
        waveforms = session.channels[test_channels].fetch(
            num_samples=test_record_length,
            record_number=test_starting_record_number,
            num_records=test_num_records_to_fetch)
        for w in waveforms:
[...]

PR #1956, opened 2023-04-29
Test niscope fetching more thoroughly and consistently, parameterize channels tested

ni-jfitzger:

Our niscope fetch tests are inconsistent in what they test. Some, like test_read(), check very little, while others are more through. Most of them either perform the same checks or should be performing the same checks.

Additionally, we only test fetching with a single channel list.
In order to test for #1770, both issues are being addressed.

Changes:

* Move the checking of fetched data into a separate function
  * Checking MeasurementStats has a slight difference, so check the data type to determine what checks to perform
* Parameterize test_channels and test_channels_expanded for all of the basic fetch tests

None

Ran the tests locally with NI-SCOPE 23.1 driver runtime.

### DIFF
--- a/src/niscope/system_tests/test_system_niscope.py
+++ b/src/niscope/system_tests/test_system_niscope.py
@@ -19,7 +19,10 @@ import system_test_utilities  # noqa: E402
 
 instruments = ['FakeDevice1', 'FakeDevice2']
 # TODO(sbethur): Use `get_channel_names` when #1402 is fixed
-test_channels = 'FakeDevice2/0,FakeDevice1/1'
+test_channels_1 = 'FakeDevice2/0,FakeDevice1/1'
+test_channels_2 = 'FakeDevice2/0:1'
+test_channels_1_expanded = test_channels_1
+test_channels_2_expanded = 'FakeDevice2/0,FakeDevice2/1'
 
 
 # There are system tests below that need either a PXI-5124 or a PXI-5142 instead of the PXIe-5164 we use everywhere else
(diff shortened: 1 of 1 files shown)


ni-jfitzger:
flake8 failure occurred. Merging recent changes from main (assertion cleanup) will probably fix it.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -457,18 +520,29 @@ def session_creation_kwargs(self):
...
     # not supported by grpc due to numpy usage
     @pytest.mark.parametrize(
-        "fetch_waveform_type,type_min_value",
+        "fetch_waveform_type,type_min_value,test_channels,test_channels_expanded",
         [
-            (numpy.int8, numpy.iinfo(numpy.int8).min),
-            (numpy.int16, numpy.iinfo(numpy.int16).min),
-            (numpy.int32, numpy.iinfo(numpy.int32).min),
-            (numpy.float64, numpy.finfo(numpy.float64).min),
+            (numpy.int8, numpy.iinfo(numpy.int8).min, test_channels_1, test_channels_1_expanded),
```
ni-jfitzger, line 525: Given that the bug is in `fetch_into` and not the private methods that handle calls for specific types, perhaps there's no need to duplicate the numpy parameterization for each test_channels value?

marcoskirsch: Would that make the test effective only when tied to the current implementation? If so, better leave it parameterized.

marcoskirsch (approved):

PR #1981, opened 2023-06-12
Clean up niscope usage documentation

### LINKED ISSUE #1316: Documentation links on niscope PyPI page are broken
### Description of issue

On [niscope PyPI page](https://pypi.org/project/niscope/), these links under **Usage** are broken:

> The waveform returned from [fetch ](https://pypi.org/project/niscope/niscope/class.html#fetch) is a flat list of Python objects
> The waveform_infos returned from [fetch_into](https://pypi.org/project/niscope/niscope/class.html#fetch-into) is a 1D list of Python objects

@texasaggie97, do you have any suggestion on how we can fix this?

ni-jfitzger:

* fetch and fetch_into links in niscope_usage.inc were broken in the built documentation and the README. This deletes them.
  * Keeps our codegen simple
* We needlessly replicated method documentation in the usage docs. Delete the replicated documentation.
* We showed an advanced fetch use case, which was also implemented incorrectly. Create a dedicated example and refer users to that example for faster fetching.
  * Keeps the usage doc clean and simple, but calls out how to fetch faster, because it may be difficult for users to discover how to do so, otherwise.

* Fix #1316

Visually inspected the generated html documentation and README and both still looked good to me.


marcoskirsch:
Adding special case about `niscope` to the .mak and the mako template is pretty hacky. Why is NI-SCOPE different than every other driver in this regard?

ni-jfitzger:
> Adding special case about `niscope` to the .mak and the mako template is pretty hacky. Why is NI-SCOPE different than every other driver in this regard?

It has special fetch functions that it links to in the README.rst and index.rst.
Maybe I could do more generic index_links.inc and README_links.inc files for all drivers that need them?

marcoskirsch:
> It has special fetch functions that it links to in the README.rst and index.rst.

What if instead we stopped linking to the special fetch funtion from the README.rst and index.rst?

ni-jfitzger:
> > It has special fetch functions that it links to in the README.rst and index.rst.
> 
> What if instead we stopped linking to the special fetch funtion from the README.rst and index.rst? Within reason we should lean towards simplicity.

I don't mind that.
Do you also want me to remove the duplicated documentation?

marcoskirsch:
I don't think calling out `fetch_into` from the README is a good idea. I considered that to be a fairly advanced use case that we mistakenly over-emphasized. As long as we have an example, I would be all for removing that as well.

docs/_static/niscope_usage.inc
```diff
@@ -21,41 +21,8 @@ The following is a basic example of using the **niscope** module to open a sessi
...
-
-                voltage = binary data * gain factor + offset
-
-        -  **offset** (float) the offset factor of the given channel; useful for scaling binary data with the following formula:
-
-                voltage = binary data * gain factor + offset
-
-        - **samples** (array of float) floating point array of samples. Length will be of the actual samples acquired
-
-    - Such that all record 0 waveforms are first. For example, with a channel list of 0,1, you would have the following index values:
```
ni-jfitzger, line 43: The part about record and channel order isn't exactly duplicated information. We don't really document that concept well in our API documentation. Should we keep that bit or perhaps link to some our online fundamental concept documentation?

docs/_static/niscope_usage.inc
```diff
@@ -21,41 +21,8 @@ The following is a basic example of using the **niscope** module to open a sessi
...
-method takes an already allocated `numpy <http://www.numpy.org/>`_ array and puts the acquired samples in it. Data types supported:
-
-    - `numpy.float64`
-    - `numpy.int8`
-    - `numpy.in16`
-    - `numpy.int32`
+If you need more performance or need to work with `SciPy <https://www.scipy.org/>`_, you can use the **fetch_into**() method instead of **fetch**(). This
+method takes an already allocated `numpy <http://www.numpy.org/>`_ array and puts the acquired samples in it.
 
 .. code-block:: python
```
ni-jfitzger, line 27: I wasn't certain whether we should delete this whole thing. The API documentation really isn't good enough to easily determine how to actually use fetch_into and we don't use it in our examples. Without this section, I would have a hard time using it. I'd probably have to look for usage in our tests.

ni-jfitzger:
> I don't think calling out `fetch_into` from the README is a good idea. I considered that to be a fairly advanced use case that we mistakenly over-emphasized. As long as we have an example, I would be all for removing that as well.

I agree that it's too complicated to display here. I even discovered a major bug in the codeblock that we showed for it (miscalculating number of waveforms). Based on a recent escalation about fetch being slow, though, I still think we should call it out. I'll remove the codeblock and refer users to a new example if they want to fetch faster.

docs/_static/niscope_usage.inc
ni-jfitzger: We actually do use it in the fetch_forever example, but that usage is even more advanced. I created a fetch_into example and deleted the codeblock.

docs/_static/niscope_usage.inc
```diff
@@ -21,81 +21,7 @@ The following is a basic example of using the **niscope** module to open a sessi
...
-        - **samples** (numpy array of datatype used) floating point array of samples. Length will be of the actual samples acquired
-
-    - Such that all record 0 waveforms are first. For example, with a channel list of 0,1, you would have the following index values:
-
-        - index 0 = record 0, channel 0
-        - index 1 = record 0, channel 1
-        - index 2 = record 1, channel 0
-        - index 3 = record 1, channel 1
-        - etc.
+If you need more performance (faster fetching) or need to work with `SciPy <https://www.scipy.org/>`_, you can use the **fetch_into()** method instead of **fetch()**. See the fetch_into example.
```
marcoskirsch, line 24: suggestion
> If you need faster fetch performance, or to directly interface with `SciPy <https://www.scipy.org/>` ....

src/niscope/examples/niscope_fetch_info.py
```diff
@@ -0,0 +1,61 @@
...
+import niscope
+import numpy
+import pprint
+import sys
+
+pp = pprint.PrettyPrinter(indent=4, width=80)
+
+
+def example(resource_name, channels, options, length, voltage):
+    # This method of fetching is more complicated, but much faster than fetch().
```
marcoskirsch, line 13:
```suggestion
    # fetch_into() allows you to pre-allocate and re-use the destination of the fetched waveforms, which can result in better performance at the expense of the usability of fetch().
```

marcoskirsch: I see you use preallocate bellow (no hyphen). Not sure which is right just be consistent.

marcoskirsch (changes requested):

docs/_static/niscope_usage.inc
```diff
@@ -21,7 +21,7 @@ The following is a basic example of using the **niscope** module to open a sessi
         # Find all record number 3
         rec3 = [wfm for wfm in waveforms if wfm.record == 3]
 
-The waveform returned from `fetch <niscope/class.html#fetch>`_ is a flat list of Python objects
+The waveform returned from `fetch`_ is a flat list of Python objects
```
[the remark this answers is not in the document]
ni-jfitzger: I deleted this along with the duplicated documentation. Although this line technically wasn't duplicated, there's no point to it if we aren't going to detail the Waveform object, which is duplicated information.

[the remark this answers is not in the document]
ni-jfitzger: Oh, your comment was from June 16. I wonder why it wasn't resolved and was only just showing up.

build/templates/driver.rst.mako
```diff
@@ -10,6 +10,10 @@ ${"=" * len(doc_header)}
 
 .. include:: installation.inc
 
+% if module_name == "niscope":
+.. include:: ../_static/niscope_usage_docs_links.inc
+% endif
+
```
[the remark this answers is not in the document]
ni-jfitzger: I'm marking this resolved. There are no changes to this file, currently.

docs/_static/niscope_usage.inc
ni-jfitzger: The API documentation doesn't explain how it's determined what order waveforms are returned in. Although we do stick the record number and channel name in the waveform_info, it might be nice for users to know how it works.

src/niscope/examples/niscope_fetch_info.py
```diff
@@ -0,0 +1,61 @@
...
+    channels = [ch.strip() for ch in channels.split(",")]
+    num_channels = len(channels)
+    num_records = 5
+    total_num_wfms = num_channels * num_records
+    # preallocate a single array for all samples in all waveforms
+    # Supported array types are: numpy.float64, numpy.int8, numpy.int16, numpy.int32
```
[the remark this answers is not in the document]
ni-jfitzger: No. You currently cannot use array.array.
session.py will error if it's not a numpy array.

```
        if waveform.dtype == numpy.float64:
            wfm_info = self._fetch_into_numpy(num_samples=num_samples, waveform=waveform, timeout=timeout)
        elif waveform.dtype == numpy.int8:
            wfm_info = self._fetch_binary8_into_numpy(num_samples=num_samples, waveform=waveform, timeout=timeout)
        elif waveform.dtype == numpy.int16:
            wfm_info = self._fetch_binary16_into_numpy(num_samples=num_samples, waveform=waveform, timeout=timeout)
        elif waveform.dtype == numpy.int32:
            wfm_info = self._fetch_binary32_into_numpy(num_samples=num_samples, waveform=waveform, timeout=timeout)
        else:
            raise TypeError("Unsupported dtype. Is {0}, expected {1}, {2}, {3}, or {4}".format(waveform.dtype, numpy.float64, numpy.int8, numpy.int16, numpy.int32))
```
Even before it gets to raise TypeError, though, arrray.array has no dtype attribute.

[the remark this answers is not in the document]
marcoskirsch: > Even before it gets to raise TypeError, though, arrray.array has no dtype attribute.

That's not a problem, right? We could make it work with array.array if we wanted to. In any case, that's not for this PR nor do I necessarilly think there's ROI in making such change proactively.

marcoskirsch (approved):

docs/_static/niscope_usage.inc
ni-jfitzger: I guess it's fine cutting it out, for now. Information like that doesn't really belong here, anyway.

PR #2006, opened 2023-09-01
Read grpc_device_server port from any line of server output when running system_tests

ni-jfitzger:

https://github.com/ni/grpc-device/pull/989 altered the output of grpc_device_server.exe and caused our internal nimi-python test suites to fail all gRPC tests. I discussed this with ni/grpc-device owners / contributors and it was determined there is no contract for the output to follow a certain format.

A copy of https://github.com/ni/nidaqmx-python/pull/450, this change updates GrpcServerProcess to keep reading the server output until the server port is parsed.

Note: this won't fix our internal ats failures until we release.

Alternatives:
I tried reading the port from the server json config instead, but it caused the tests to fail on nimibot. Not sure if this was due to running tests in parallel for different apis on the same VM at once or due to a difference with the version of grpc_device_server.exe in use.

Internal Bug [AB#2508876](https://dev.azure.com/ni/DevCentral/_workitems/edit/2508876) (once we release)

Ran the niscope system tests without the breaking grpc-device change and with it. Tests passed in both cases.

### DIFF
--- a/src/shared/system_test_utilities.py
+++ b/src/shared/system_test_utilities.py
@@ -1,7 +1,7 @@
-import json
 import os
 import pathlib
 import pytest
+import re
 import subprocess
 import threading
 import time
@@ -11,11 +11,19 @@ class GrpcServerProcess:
     def __init__(self):
         server_exe = self._get_grpc_server_exe()
         self._proc = subprocess.Popen([str(server_exe)], stdout=subprocess.PIPE)
-        grpc_server_config = self._get_grpc_json_config()
-        self.server_port = grpc_server_config["port"]
 
-        # Discard output from server
+        # Read/parse output until we find the port number or the process exits; discard the rest.
         try:
+            self.server_port = None
+            while self.server_port is None and self._proc.poll() is None:
+                line = self._proc.stdout.readline()
+                match = re.search(rb"Server listening on port (\d+)", line)
+                if match:
+                    self.server_port = int(match.group(1))
+
+            if self._proc.poll() is not None:
+                raise RuntimeError(f"Server exited with return code {self._proc.returncode}")
+
             self._stdout_thread = threading.Thread(target=self._discard_output, args=(self._proc.stdout,), daemon=True)
             self._stdout_thread.start()
         except Exception:
@@ -43,12 +51,6 @@ class GrpcServerProcess:
         if not server_exe.exists():
             pytest.skip("NI gRPC Device Server not installed")
         return server_exe
-
-    def _get_grpc_json_config(self):
-        config_file = self._get_grpc_server_exe().parent / "server_config.json"
-        with open(config_file, "r") as f:
-            config_data = json.load(f)
-        return config_data
 
     def _discard_output(self, stdout):
         while True:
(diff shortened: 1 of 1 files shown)


bkeryan:
This doesn't work when the port number is dynamically chosen. The configuration file specifies port 0 in this case.

MeasurementLink and nidaqmx-python also read the port number from stdout/stderr. I did not expect adding an unrelated line of output to break this.

bkeryan:
MeasurementLink looks for a line of output matching `/Server listening on port (\d+)/` and should not be affected by this change.

nidaqmx-python has code that is copied from nimi-python, so it only looks at the first line of output: https://github.com/ni/nidaqmx-python/blob/master/tests/_grpc_utils.py

Cc: @zhindes

bkeryan:
BTW, I realize the tests are using the default server configuration, which uses a hardcoded port. I have considered adding a config file for nidaqmx-python to use a dynamic port so that the tests don't fail if the server is already running with the default config, but I guess I didn't do anything about this.

ni-jfitzger:
> BTW, I realize the tests are using the default server configuration, which uses a hardcoded port. I have considered adding a config file for nidaqmx-python to use a dynamic port so that the tests don't fail if the server is already running with the default config, but I guess I didn't do anything about this.

Yeah, we run the tests in a controlled setting where the server will not already be running with the default port. A person manually running the tests on their own system would have failures if the server was already running, but that's already the case. I'm not sure it's worth caring about.

ni-jfitzger:
> > BTW, I realize the tests are using the default server configuration, which uses a hardcoded port. I have considered adding a config file for nidaqmx-python to use a dynamic port so that the tests don't fail if the server is already running with the default config, but I guess I didn't do anything about this.
> 
> Yeah, we run the tests in a controlled setting where the server will not already be running with the default port. A person manually running the tests on their own system would have failures if the server was already running, but that's already the case. I'm not sure it's worth caring about.

I spoke too soon. Our tests failed. Probably because we run a bunch of them in parallel.

bkeryan:
> Yeah, we run the tests in a controlled setting where the server will not already be running with the default port. A person manually running the tests on their own system would have failures if the server was already running, but that's already the case. I'm not sure it's worth caring about.

I often ran into this with nidaqmx-python when working with examples and then running auto tests.

I've also written tests in NI internal components that bind to a hardcoded port, and these intermittently failed when run from build pipelines because *something* (possibly another instance of the same test) was using the port. Switching to a dynamic port made these tests more resilient.

Anyway, the behavior of the nimi-python tests is up to you, not me, but I wanted to share some reasons why I've considered changing that behavior for nidaqmx-python.

> I spoke too soon. Our tests failed. Probably because we run a bunch of them in parallel.

If you're using physical devices or persistent simulated devices, parallelizing tests isn't going to work even with a dynamic port, but if you're using persistent simulated devices, it could work.

bkeryan:
Here is my proposed fix for nidaqmx-python: https://github.com/ni/nidaqmx-python/pull/450

marcoskirsch (approved):

PR #2043, opened 2024-04-01
Run pyupgrade to update Python syntax used

ni-jfitzger:

Note: No hand-coded changes are included.

Using a script, I ran `pyupgrade --py38-plus` on files in nimi-python. [pyupgrade](https://pypi.org/project/pyupgrade/) is a tool used to automatically update the syntax used in Python code to syntaxes available in newer versions.

The script ran on the following folders:
- build
- docs
- source
- tools

The script ignored folder paths containing:
- .tox
- `__pycache__`

The script ignored files with the extensions:
- ..digiproj
- .mak
- .proto

The script ignored imported metadata files (the ones that say they were generated) as well as "fancy_fetch_capture_waveform.py.mako" (pyupgrade messed up the indentation) and "_get_error_description.py.mako" (pyupgrade messed up the indentation).

List of automatic changes that I noticed:
- String formatting (most of the changes were this)
  - Use fstrings where reasonable
  - delete unnecessary index arguments to {} for strings where we continue to use .format
    - pyupgrade is timid about f-string usage in order to maintain short and readable code
  - replace use of really old % format specifier with .format
- removal of coding: utf-8 comment (it's the default encoding)
- drop unnecessary Object argument from base class declarations
- Use dict comprehension rather than whatever we were doing in nifake/unit_tests/test_grpc.py
- Use set comprehension in metadata_add_all.py
- delete unnecessary arguments from open() calls

I discarded changes to test_converters.py. I believe we're not passing the right arguments for some tuple tests, so I want to handle that separately.

None

- I reviewed every line that was changed.
- PR Checks

### DIFF
--- a/build/generate_template.py
+++ b/build/generate_template.py
@@ -14,7 +14,7 @@ def generate_template(template_name, template_params, dest_file, in_zip_file=Fal
     try:
         template_params['encoding_tag'] = '# -*- coding: utf-8 -*-'
         module_name = template_params['metadata'].config['module_name']
-        lookup = TemplateLookup(directories=['src/{0}/templates'.format(module_name), 'build/templates'])
+        lookup = TemplateLookup(directories=[f'src/{module_name}/templates', 'build/templates'])
         template = Template(filename=template_name, lookup=lookup)
         rendered_template = template.render(template_parameters=template_params)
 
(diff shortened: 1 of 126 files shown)


ni-jfitzger:
I considered cleaning up some of the remaining .format() calls, but decided to leave this as only automated changes.

ni-jfitzger:
pyupgrade can be used as a pre-commit hook, but I don't think that would be a good idea, given that it broke the indentation of some of our templates.

marcoskirsch (approved):

PR #2068, opened 2025-01-25
Drop Python 3.8 Support

ni-jfitzger:

* Drop Python 3.8 support, by removing it from our testing and list of officially supported versions.
* (For the first time) enforce our version support policy at installation: disallow installation of APIs on Python versions earlier than 3.9
  * As part of this we now need to specify a list of supported Python versions for building wheel_dep in tox-system_tests.ini

None

PR Checks

### DIFF
--- a/build/templates/setup.py.mako
+++ b/build/templates/setup.py.mako
@@ -73,7 +73,6 @@ setup(
         "Operating System :: Microsoft :: Windows",
         "Operating System :: POSIX",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
--- a/build/templates/setup.py.mako
+++ b/build/templates/setup.py.mako
@@ -47,6 +47,7 @@ setup(
     license='MIT',
     include_package_data=True,
     packages=['${config['module_name']}'],
+    python_requires='>=3.9',
     install_requires=[
         'hightime>=0.2.0',
         % if config['uses_nitclk']:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ matrix:
       # We run tests on the latest supported version of Python first.
       # This is where additional tests are run so we give it more time.
       - python: "3.12"
-      - python: "3.8"
       - python: "3.9"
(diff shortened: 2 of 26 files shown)


ni-jfitzger:
Interestingly, our Changelog shows that we added 3.8 support to all APIs twice.

build/templates/setup.py.mako
ni-jfitzger, line 1: Should we add a `python_requires=">=3.9"` to prevent users from installing to versions of Python that we've dropped support for?

[python_requires](https://github.com/pypa/sampleproject/blob/db5806e0a3204034c51b1c00dde7d5eb3fa2532e/setup.py#L123C1-L123C33)

bkeryan: If you do, then installing your packages on Python 3.8 will automatically backtrack and find the latest version that still supports Python 3.8.

If you don't, then installing your packages on Python 3.8 will install the latest, and then importing it will fail (once you start using new language features).

@mshafer-NI Does/should the Python Working Group's versioning policy say something like this?

ni-jfitzger: Yeah, I don't really see any downside to it. Currently, it can proceed with an install on older Python and then a user may get a nasty surprise when an import fails. If feels kinder to enforce the support policy.

mshafer-NI: So..., relevant PWG policy is that the module must support all non-EOL versions. 3.8 is EOL, so provided that you do not have clients that need you to still support 3.8, then you may drop 3.8.

As far as the question here:
If you drop support for a Python version, update the setup.py to declare that correctly. Otherwise, you are telling pip that you support a version that you don't.

marcoskirsch (approved):
Weird about the duplication in CHANGELOG.md
If you have the energy to figure out what the right thing is (was it a different version of Python that was added? Second addition bogus) and fix it, that'd be great. You can do it in a separate PR if you prefer.

ni-jfitzger:
> Weird about the duplication in CHANGELOG.md If you have the energy to figure out what the right thing is (was it a different version of Python that was added? Second addition bogus) and fix it, that'd be great. You can do it in a separate PR if you prefer.

We didn't do things as cleanly back then, but it seems as though support was officially added in 1.1.4 and we just forgot to add the classifier for PyPI until 1.2.0.

Issue #2072, opened 2025-01-26
Cleanup enums_addon metadata that was added when fixing enum value python_name behavior


ni-jfitzger:
### Description of issue

In order to address #2066, we had to override some enum metadata to avoid altering the generated enum value names.
The `python_name` for these cases needs to be updated in internal metadata, then the overrides can be deleted when the updated metadata gets pulled in.

PR #2073, opened 2025-01-26
Fix _add_enum_value_python_name() to respect hardcoded value names

### LINKED ISSUE #2066: metadata_add_all._add_enum_value_python_name() removes common suffixes from a user-set 'python_name'.
### Description of issue

The 'python_name' key for enum values pulls double duty in `metadata_add_all._add_enum_value_python_name()`. It's intended that a user can set it to override the expanded name with the name that they want to be used in the API. In such a case, the user-set value should not be modified. But the helper method also uses the same key for the expansion process when it's not set by the user and does not track whether the user set it. As a result, common prefixes or suffixes may get removed from the user-set 'python_name', instead of using it, as is, to override the expanded name.

ni-jfitzger:

The `python_name` metadata key is supposed to allow us to hardcode the name that will be used by the Python API. For enum values, this was not being respected.

This change fixes that and adds additional unit test coverage for enum value expansion.

As part of this, we have to temporarily override some enum value metadata. Followup item: #2072

* Fix #2066

* Unit tests pass
* Codegen has run and API is unchanged
* PR Checks

### DIFF
--- a/build/helper/metadata_add_all.py
+++ b/build/helper/metadata_add_all.py
@@ -609,10 +609,11 @@ def _get_least_restrictive_codegen_method(codegen_methods):
 def _add_enum_value_python_name(enum_info, config):
     '''Add 'python_name' for all values, removing any common prefixes and suffixes'''
     for v in enum_info['values']:
+        v['user_set_python_name'] = 'python_name' in v
         if 'python_name' not in v:
             v['python_name'] = v['name'].replace('{}_VAL_'.format(config['module_name'].upper()), '')
 
-    # We are using an os.path function do find any common prefix. So that we don't
+    # We are using an os.path function to find any common prefix. So that we don't
     # get 'O' in 'ON' and 'OFF' we remove characters at the end until they are '_'
     names = [v['python_name'] for v in enum_info['values']]
     prefix = os.path.commonprefix(names)
@@ -627,13 +628,20 @@ def _add_enum_value_python_name(enum_info, config):
     # '_' only means the name starts with a number
     if len(prefix) > 0 and prefix != '_':
         for v in enum_info['values']:
+            if v['user_set_python_name']:
+                continue
             assert v['python_name'].startswith(prefix), '{} does not start with {}'.format(v['name'], prefix)
             v['prefix'] = prefix
             v['python_name'] = v['python_name'].replace(prefix, '')
 
     # Now we need to look for common suffixes
     # Using the slow method of reversing a string for readability
-    rev_names = [''.join(reversed(v['python_name'])) for v in enum_info['values']]
+    # We do not include hardcoded python names when looking for common suffixes
+    rev_names = [
+        ''.join(reversed(v['python_name']))
+        for v in enum_info['values']
+        if not v['user_set_python_name']
+    ]
     suffix = os.path.commonprefix(rev_names)
     while len(suffix) > 0 and suffix[-1] != '_':
         suffix = suffix[:-1]
@@ -649,12 +657,15 @@ def _add_enum_value_python_name(enum_info, config):
     # '_' only means the name starts with a number
     if len(suffix) > 0:
         for v in enum_info['values']:
+            if v['user_set_python_name']:
+                continue
             assert v['python_name'].endswith(suffix), '{} does not end with {}'.format(v['name'], suffix)
             v['suffix'] = suffix
             v['python_name'] = v['python_name'][:-len(suffix)]
 
     # We need to check again to see if we have any values that start with a digit
     # If we are not going to code generate this enum, we don't care about this
+    # Even hardcoded names should follow this rule
     for v in enum_info['values']:
         assert v['python_name'], enum_info
         if enum_info['codegen_method'] != 'no' and v['python_name'][0].isdigit():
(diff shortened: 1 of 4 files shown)


src/nidmm/metadata/enums_addon.py
```diff
@@ -2,5 +2,42 @@
...
 enums_override_metadata = {
+    # TODO (ni-jfitzger): delete this override once python_name is corrected for each value. See https://github.com/ni/nimi-python/issues/2072
+    'ThermistorType': {
+        'values': [
+            {
+                'documentation': {
+                    'description': 'Custom'
+                },
+                'name': 'NIDMM_VAL_TEMP_THERMISTOR_CUSTOM',
+                'python_name': 'CUSTOM',
```
ni-jfitzger, line 13: Strangely, if I set `python_name` for all the values except this one, this one came back as `TEMP`. I'm not sure why it gets cut off like that, but didn't feel it was worth further investigation.

build/helper/metadata_add_all.py
```diff
@@ -609,10 +609,11 @@ def _get_least_restrictive_codegen_method(codegen_methods):
 def _add_enum_value_python_name(enum_info, config):
     '''Add 'python_name' for all values, removing any common prefixes and suffixes'''
     for v in enum_info['values']:
+        v['user_set_python_name'] = 'python_name' in v
```
marcoskirsch, line 612: Is this `user_set` nomenclature already used?
A more descriptive term could be `explicitly_set` (vs. implicit or expanded or derived).

build/helper/metadata_add_all.py
```diff
@@ -627,13 +628,20 @@ def _add_enum_value_python_name(enum_info, config):
...
+            if v['user_set_python_name']:
+                continue
             assert v['python_name'].startswith(prefix), '{} does not start with {}'.format(v['name'], prefix)
             v['prefix'] = prefix
             v['python_name'] = v['python_name'].replace(prefix, '')
 
     # Now we need to look for common suffixes
     # Using the slow method of reversing a string for readability
-    rev_names = [''.join(reversed(v['python_name'])) for v in enum_info['values']]
+    # We do not include hardcoded python names when looking for common suffixes
```
marcoskirsch, line 639: Here you use `hardcoded` as a synonym to `user_set`. Reduce ambiguity by using one term to refer to one concept everywhere.

marcoskirsch: FWIW I like the word "explicit" for this. It means (in my mind) that it's there written in metadata.
"User set" is a bit ambiguous because it's not super clear who the user is.

src/nidmm/metadata/enums_addon.py
marcoskirsch: Does that mean that if a client wants to explicitly provide the python name for a subset of enum values, we don't know what will be the expanded value here?

I feel like it's worth getting to the bottom of, and having a test case for.

marcoskirsch (changes requested):

src/nidmm/metadata/enums_addon.py
ni-jfitzger: It was almost certainly due to common suffix removal.

The behavior was a little bit hard to follow, so I've rewritten my change and instead of using a boolean key for tracking whether to touch something, I'll have us do all of the processsing with a temporary key, '_python_name'.

I've also added another unit test related to this. There's a peculiar behavior with common prefix and common suffix removal when you only have one value to calculate the prefix or suffix from. It calculates the prefix or suffix as the entire string and we then remove almost that entire string (upto and including the first/last underscore). I setting prefix and suffix to '', instead, when there's less than 2 values to calculate the commmon prefix or suffix from, but (upsetting as it is) we have one or two enums with a single enum value that actually make use of this behavior. I can still change it if you want, though. It seems like a footgun.

build/helper/metadata_add_all.py
ni-jfitzger: I've stopped using the word hardcoded and made use of the word explicit, instead, as part of my rewrite.
I eliminated the `user_set_python_name` key. We'll do processing on the temporary key, `'_python_name'`, instead, so we won't need a special key to track whether we should expand the name.

build/helper/metadata_add_all.py
ni-jfitzger: I've eliminated the key, as noted [here](https://github.com/ni/nimi-python/pull/2073#discussion_r1929892527).

src/nidmm/metadata/enums_addon.py
ni-jfitzger: With the latest change, we no longer need to set 'python_name': 'CUSTOM', but this behavior still seems like a footgun.

marcoskirsch: If the behavior observed when only a subset of the enum values has `python_name` explicitly in metadata is hard to describle / footgun, we could consider disallowing this altogether and forcing clients to specify none or all. I think the case is rare enough that there would be very few cases of this.

Counterpoint is that bad expansion of a subset of enum values should be fairly apparent in generated code diffs.

What do you think?

ni-jfitzger: It's easy to describe at this point. Everything up through the last underscore is trimmed. This is the same behavior we would see if the values that use `python_name` didn't exist.

You're correct that is very rare and even rarer is the case where we have an enum with only one value and the one value has multiple words and that enum is used somewhere so we don't filter it out in codegen. We only seem to have a couple of those and the current value names seem okay.

I think it's fine to leave this change as is. We should be able to catch any issues in review.

PR #2080, opened 2025-02-05
Add Python 3.13 Support

### LINKED ISSUE #2075: Add Python 3.13 support and loosen protobuf version constraint
### Description of issue

The current stable release of Python is version 3.13, which was released on 10-7-2024.

There are versioning conflicts between nimi-python's `grpc` extra and Python 3.13:
- nimi-python's `grpc` extra requires `protobuf>=4.21.6,<5.0`.
- The first version of `grpcio` and `grpcio-tools` with binary wheels for Python 3.13 is [version 1.67.0](https://pypi.org/project/grpcio/1.67.0/)
- `grpcio-tools` version 1.67.0 requires `protobuf` version `<6.0dev,>=5.26.1`.

This causes issues when upgrading `ni-measurement-plugin-sdk-service`'s dev/test dependencies, since it depends on `grpcio-tools` for gRPC codegen and several nimi-python packages for testing.

```
PS C:\dev\measurement-plugin-python\packages\service> poetry lock --no-update
Resolving dependencies... (4.9s)

Because no versions of niswitch match >1.4.4,<1.4.5 || >1.4.5,<1.4.6 || >1.4.6,<1.4.7 || >1.4.7,<1.4.8 || >1.4.8
 and niswitch[grpc] (1.4.4) depends on protobuf (>=4.21,<5.0), niswitch[grpc] (>=1.4.4,<1.4.5 || >1.4.5,<1.4.6 || >1.4.6,<1.4.7 || >1.4.7,<1.4.8 || >1.4.8) requires protobuf (>=4.21,<5.0).
And because niswitch[grpc] (1.4.5) depends on protobuf (>=4.21,<5.0), niswitch[grpc] (>=1.4.4,<1.4.6 || >1.4.6,<1.4.7 || >1.4.7,<1.4.8 || >1.4.8) requires protobuf (>=4.21,<5.0).
And because niswitch[grpc] (1.4.6) depends on protobuf (>=4.21,<5.0)
 and niswitch[grpc] (1.4.7) depends on protobuf (>=4.21,<5.0), niswitch[grpc] (>=1.4.4,<1.4.8 || >1.4.8) requires protobuf (>=4.21,<5.0).
And because niswitch[grpc] (1.4.8) depends on protobuf (>=4.21.6,<5.0)
 and grpcio-tools (1.67.0) depends on protobuf (>=5.26.1,<6.0dev), niswitch[grpc] (>=1.4.4) is incompatible with grpcio-tools (1.67.0).
So, because ni-measurement-plugin-sdk-service depends on both niswitch[grpc] (>=1.4.4) and grpcio-tools (1.67.0), version solving failed.
```

The `protobuf` package follows semantic versioning, so I think it was reasonable to use `>=4.21,<5.0` based on what we knew about it, but based on what I've seen with nidaqmx-python, it appears that output from `grpcio-tools` 1.49.1 works with both `protobuf` v4 and v5. I was able to change nidaqmx-python's `protobuf` version constraint from `^4.21` (aka `>=4.21,<5.0`) to `>=4.21`: https://github.com/ni/nidaqmx-python/pull/645/

### System report

N/A, can't install

### Steps to reproduce issue

1. Download https://github.com/ni/measurement-plugin-python/tree/users/bkeryan/python-3.13
2. cd to packages/service
3. `poetry lock --no-update`

ni-jfitzger:

1. Adds official Python 3.13 support to nimi-python through testing.
   The version of Python used by readthedocs for documentation generation remains unchanged at 3.11. I prefer to be intentional about changes to the generation of documentation.

2. Update` grpcio`, `protobuf` deps for Python 3.13 compatibility

Initially attempted codegen with Python 3.13 and chose `grpcio`, `protobuf` versions for compatibility with the newer `grpcio-tools` version. We decided to continue generating pb2 files with Python 3.12 to avoid breaking compatibility with other NI Python packages. See [this](https://github.com/ni/nimi-python/pull/2080#discussion_r1943453414) discussion. Afterward, I kept the `grpcio`, `protobuf` versions in our tox .inis because I knew they were compatible with Python  3.13 and can be used if/when we begin code-generating with Python 3.13.

[grpcio-tools compatibility for version 1.67.0](https://github.com/grpc/grpc/blob/f2ffbb89a2f7456b897e5556782e8639d161a6f6/tools/distrib/python/grpcio_tools/setup.py) ([The earliest version with a 3.13 wheel](https://pypi.org/simple/grpcio-tools/))

Due to generating with the same `grpcio-tools` version as before, I had to disable the `ensure_codegen_up_to_date` check on Python 3.13, though it still runs on the other versions.`

* Fix #2075

PR Checks


build/templates/setup.py.mako
```diff
@@ -46,7 +46,7 @@ setup(
     extras_require={
         'grpc': [
             'grpcio>=1.59.0,<2.0',
-            'protobuf>=4.21.6,<5.0'
+            'protobuf>=4.21.6'
```
ni-jfitzger, line 49: This seems like it needs to be updated to 5.27.2. System Tests are randomly failing on imports when we install old protobuf versions. I'm puzzled how the system tests managed to pass in my original PR.

tox.ini
```diff
@@ -136,16 +136,16 @@ deps =
...
+    test: protobuf == 5.27.2  # Compatible with grpcio-tools 1.67.0
     build_test: pytest
     build_test: coverage
     build_test: mako
     build_test: hacking
     build_test: pep8-naming
     codegen: mako
     codegen: packaging
-    codegen: grpcio-tools == 1.59.0  # First version to support Python 3.12
+    codegen: grpcio-tools == 1.67.0  # First version to support Python 3.13
```
bkeryan, line 148: Using grpcio-tools 1.67.0 all the time breaks compatibility with grpcio <= 1.67.0 and will likely force users to update protobuf.

The technique we used for nidaqmx-python is to generate code using the oldest compatible Python and oldest compatible grpcio-tools, check the generated code into version control, then test against newer Python and grpcio versions. 
- Python 3.8 (for now): https://github.com/ni/nidaqmx-python/blob/master/.github/workflows/build.yml
- grpcio-tools 1.49.1 (as long as you are doing codegen with Python 3.8, 3.9, 3.10, or 3.11): https://github.com/ni/nidaqmx-python/blob/master/pyproject.toml#L65

bkeryan (changes requested):

tox.ini
ni-jfitzger: What a pain. I'd rather not be stuck using an old Python version for non-test steps. That's likely to accumulate tech debt.

Why should we go out of our way to avoid an update to protobuf and grpcio if the new versions are still compatible with the files generated by the old version of grpcio-tools?

bkeryan: > What a pain. I'd rather not be stuck using an old Python version for non-test steps. That's likely to accumulate tech debt.

Do the nimi-python build/codegen scripts need features or tools that are not available in Python 3.9? 

It looks like you last updated the build/codegen scripts to use Python 3.8+ syntax: https://github.com/ni/nimi-python/pull/2043

I would try it myself and see, if nimi-python used mainstream Python tooling such as Poetry, but I don't want to bother setting up WSL, GNU make, etc. Last time I tried that, I ran into conflicts between pyenv on Linux and pyenv-win on Windows.

> Why should we go out of our way to avoid an update to protobuf and grpcio if the new versions are still compatible with the files generated by the old version of grpcio-tools?

I have been avoiding this for the packages I maintain because they don't need any new protobuf/grpcio features, so it seems like updating the version constraints doesn't provide any user value.

It seems weird to say "Python 3.9 users, you have to upgrade to grpcio 1.67 or later because that's the first version that supports Python 3.13, which you are not using, but which we use in our build pipeline."

However, there are unpatched security vulnerabilities in older versions of these packages, so maybe it's fine to start requiring newer versions. Users who pin/lock their dependencies to use older versions of protobuf/grpcio can continue to use older versions of NI packages as well.

I guess the biggest concern is that nimi-python, nidaqmx-python, and measurement-plugin-python are still compatible with each other:
- Each of these packages contains a copy of `session_pb2.py`. The serialized descriptors in these files must be identical, or users will get errors when importing more than one copy. I just tried this with your branch and the nidaqmx-python master branch and I didn't get an error, so I think it's still fine.
- measurement-plugin-python currently requires `protobuf = "^4.21"`, so existing versions will be incompatible with the updated nimi-python. But I suppose that pip, Poetry, etc. should backtrack and install the old nimi-python in this scenario.

Cc: @mshafer-NI

ni-jfitzger: > Do the nimi-python build/codegen scripts need features or tools that are not available in Python 3.9?

No, we just historically run all steps on the newest Python and run a limited set of steps on the older versions.
_Until this PR, we ran all of the steps on Python 3.12, so we could continue to do that_ and make no changes to pb2 files and no changes to gRPC deps apart from editing/removing the protobuf upper bound. It just feels like tech debt.

> However, there are unpatched security vulnerabilities in older versions of these packages, so maybe it's fine to start requiring newer versions. Users who pin/lock their dependencies to use older versions of protobuf/grpcio can continue to use older versions of NI packages as well.

Sounds like a solid reason to upgrade.

> Each of these packages contains a copy of session_pb2.py. The serialized descriptors in these files must be identical, or users will get errors when importing more than one copy. I just tried this with your branch and the nidaqmx-python master branch and I didn't get an error, so I think it's still fine.

You have my attention, now. This seems like a strong reason to stick with our current version of grpcio-tools, even if our tests pass and you didn't get an error.

ni-jfitzger:
A crash in a test?

> Windows fatal exception: code 0xe0434f4e
> ...   File "C:\ar\_work\nimi-python\nimi-python\src\nidigital\system_tests\test_system_nidigital.py", line 1245 in test_specifications_levels_and_timing_multiple

build/templates/setup.py.mako
```diff
@@ -46,7 +46,7 @@ setup(
     extras_require={
         'grpc': [
             'grpcio>=1.59.0,<2.0',
-            'protobuf>=4.21.6,<5.0'
+            'protobuf>=4.21.6,<6.0'
```
bkeryan, line 49: Based on the 4.x -> 5.x upgrade, I think we should remove the `protobuf` upper bound entirely.

Thinking back to 3.x -> 4.x: it was possible to use the old 3.x codegen with 4.x, but I think you had to set an environment variable.

bkeryan (approved):
Approved with suggestions

build/templates/setup.py.mako
ni-jfitzger: Hmm, yeah, I suppose it's probably safe enough to remove.
If there's no wheel for a later version, it'll pick an older one.
If it installs but is somehow incompatible, the user can add their own installation constraints, but they can't overcome an upper bound that we include, if they need a later version.

build/templates/setup.py.mako
ni-jfitzger: >This seems like it needs to be updated to 5.27.2.

Not anymore. I reverted the changes to the pb2 files. We went back to Python 3.12 codegen.

build/templates/setup.py.mako
ni-jfitzger: Upper bound removed.

ni-jfitzger:
@marcoskirsch 
This is tested and has approvals from Brad and Tobias, so I'm submitting it without further review to avoid taking up your time.
There are no API changes, so it's less important to have your scrutiny.

PR #2065, opened 2025-01-02
Update nidcpower metadata

jkwoon-NI:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).
- [x] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.
- [ ] ~~I've added tests applicable for this pull request~~

### What does this Pull Request accomplish?

Updates metadata for nidcpower to latest 25.0 export.
Resulting changes:
- Adds `INHIBITED` to enum `OutputStates`
- Changes `OuputStates.VOLTAGE` and `OutputStates.CURRENT` enum value names to `OuputStates.CONSTANT_VOLTAGE` and `OutputStates.CONSTANT_CURRENT` respectively
- Updates documentation for `measure_buffer_size`, `measure_complete_event_delay`, `source_delay`, and `merged_channels` attributes
- Updates copyright year in the `conf.py` files in other modules so that the build passes

### List issues fixed by this Pull Request below, if any.

- Fix #2067

### What testing has been done?

PR checks

PR #2082, opened 2025-02-24
Update metadata before 1 4 9 release

ni-jfitzger:

Update metadata to latest final exports for most APIs.

Excluded:
* nise: no new final exports
* nifgen, niswitch: .proto file changes are messy, so I'm handling it in a separate PR

Resulting Changes:
* nidmm has minor documentation changes
* nidcpower: OUTPUT_STATES_NIDCPOWER_VAL_OUTPUT_INHIBITED add to nidcpower_pb2.py
  * This value was added to enum in https://github.com/ni/nimi-python/pull/2065
 
None

PR Checks


generated/nidcpower/nidcpower/nidcpower_pb2.py
```diff
@@ -15,7 +15,7 @@
 from . import session_pb2 as session__pb2
```
marcoskirsch, line 17: do you know why this changed, and are we sure it's compatbile?

src/nidigital/metadata/enums.py
```diff
@@ -792,6 +792,22 @@
             }
         ]
     },
+    'VectorMemoryType': {
```
marcoskirsch, line 795: There is no reason for these to be in the Python metadata, they aren't for public use.

marcoskirsch (changes requested):

generated/nidcpower/nidcpower/nidcpower_pb2.py
ni-jfitzger: The first large section changed because we introduced OUTPUT_STATES_NIDCPOWER_VAL_OUTPUT_INHIBITED to the .proto file.

I think that may be why the serialized start and end values changed, too.

I don't know how it impacts compatibility.
@bkeryan do you know?

ni-jfitzger: I do notice that all of the affected values increased by exactly 51.

bkeryan: The protobuf compiler parses the `.proto` file and stores it as a protobuf-serialized `FileDescriptorProto` message like this: `_descriptor_pool.Default().AddSerializedFile(b'\n\x0fnidcpower.proto\x12\x0enidcpower_grpc\x1a\x0enidevice.proto\x1a\rsession.proto\"...`

The serialized start/end values are the offsets into that serialized message for each data type.

Adding an enum shifts the serialized offset for every data type declared after it. It's perfectly normal.

src/nidigital/metadata/enums.py
ni-jfitzger: Reverted it and I've got an internal PR to update the internal metadata so that this won't be included next time.
We'll just skip updating this metadata this time around, since there's no other changes.

I considered updating the internal ~template~ expander to exclude private access functions, attributes and enums,, but then I remembered the special case of the "internal" SetRuntimeEnvironment function that we include in our metadata and didn't want to risk breaking a similar case with "private" access.

marcoskirsch (approved):

PR #2086, opened 2025-03-11
Make each product use their own gRPC Device Server ports in the System Tests

SaarthakMohanNI:

Several different gRPC device server processes are created when running system tests, but all these different processes default to the same port, 31763. This collision is causing tests to fail to run.

[AB#3048787](https://dev.azure.com/ni/DevCentral/_workitems/edit/3048787)

### DIFF
--- a/src/nidcpower/system_tests/grpc_server_config.json
+++ b/src/nidcpower/system_tests/grpc_server_config.json
@@ -1,0 +1,9 @@
+{
+    "address": "[::1]",
+    "port": 31760,
+    "security" : {
+       "server_cert": "",
+       "server_key": "",
+       "root_cert": ""
+    }
+ }
--- a/src/nidcpower/system_tests/test_system_nidcpower.py
+++ b/src/nidcpower/system_tests/test_system_nidcpower.py
@@ -1100,7 +1100,7 @@ class TestGrpc(SystemTests):
     @pytest.fixture(scope='class')
     def grpc_channel(self):
         current_directory = os.path.dirname(os.path.abspath(__file__))
-        config_file_path = os.path.join(current_directory, 'grpc_server_config,json')
+        config_file_path = os.path.join(current_directory, 'grpc_server_config.json')
         with system_test_utilities.GrpcServerProcess(config_file_path) as proc:
             channel = grpc.insecure_channel(f"localhost:{proc.server_port}")
             yield channel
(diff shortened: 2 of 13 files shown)


SaarthakMohanNI:
Note that I'm not 100% sure yet that each product doesn't run multiple gRPC tests simultaneously

src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -1099,7 +1099,9 @@ def test_lcr_compensation_data(self, session):
 class TestGrpc(SystemTests):
     @pytest.fixture(scope='class')
     def grpc_channel(self):
-        with system_test_utilities.GrpcServerProcess() as proc:
+        current_directory = os.path.dirname(os.path.abspath(__file__))
+        config_file_path = os.path.join(current_directory, 'grpc_server_config,json')
```
ni-jfitzger, line 1103: This is a comma. Should be a period.

SaarthakMohanNI: whoops. fixed

ni-jfitzger:
> Note that I'm not 100% sure yet that each product doesn't run multiple gRPC tests simultaneously

Despite claims to the contrary in the niscope system tests, I'm pretty sure we don't parallelize execution of a tests for a given API. The "parallel-mode" argument passed to `coverage` merely changes the name of the coverage report file produced. See https://coverage.readthedocs.io/en/7.6.12/cmd.html

In order to run tests in parallel for an API, we would need to be using xdist.
See https://stackoverflow.com/questions/45733763/how-to-run-pytest-tests-in-parallel

ni-jfitzger (approved):

src/nidcpower/system_tests/grpc_server_config.json
```diff
@@ -0,0 +1,9 @@
+{
+    "address": "[::1]",
+    "port": 31760,
```
bkeryan, line 3: FYI, you can configure it to use a dynamic port and read the port number from stdout. That's what InstrumentStudio's GrpcDeviceServerActivationService does.

ni-jfitzger: We considered it, but we felt this was safer, particularly with multiple instance of gRPC Device Server being kicked off at the same time.

bkeryan: That shouldn't happen. The server specifies a listening port of 0 when calling bind(), which tells the OS to allocate a port from the ephemeral port range. Then the server asks the OS which port it's using and prints that port number to stdout. This should only fail when all of the ports in the ephemeral port range are in use (and there are thousands of these).

PR #715, opened 2018-01-02
Renumber declaration case numbers

texasaggie97:
[X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).
~~[ ] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.~~
~~[ ] I've added tests applicable for this pull request~~
### What does this Pull Request accomplish?
* Requires #713 
* Renumber the different cases for ctypes declarations
  * Prefixed with 'S' or 'B' for scalar or buffer
  * Scalar begins at `S110`
  * Buffer begins at `B510`
  * Count by 10 so there is room for additions
* Add tests for all buffer cases

### What testing has been done?
* Unit
* System

PR #2098, opened 2025-05-15
Support for complex types

rahulr-NI:

It provides support for Complex128, Complex64 and Int16 data type for all the modules.

No existing issues are fixed as part of this PR.

Tested by generating a waveform with a complex datatype and returned signal is correctly generated.


build/defines.mak
```diff
@@ -59,6 +59,7 @@ DEFAULT_PY_FILES_TO_GENERATE := \
     unit_tests/_matchers.py \
     __init__.py \
     _converters.py \
+    _complextype.py \
```
ni-jfitzger, line 62: Alternatively, we could add this to `MODULE_FILES_TO_GENERATE` in nifake.mak and nirfsg.mak or filter it out of `MODULE_FILES_TO_GENERATE` for all other clients.

I'm not sure I want us to add this module to clients that don't use it.

build/helper/helper.py
```diff
@@ -7,25 +7,28 @@
...
+    'ViSession':        { 'array_type': None,     'python_type': 'int',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViAttr':           { 'array_type': None,     'python_type': 'int',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViChar':           { 'array_type': None,     'python_type': 'int',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViChar[]':         { 'array_type': None,     'python_type': 'str',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViBoolean':        { 'array_type': None,     'python_type': 'bool',  'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViRsrc':           { 'array_type': None,     'python_type': 'str',   'numpy_type': 'bool_',      },  # noqa: E201, E202, E241
+    'ComplexViReal64':  { 'array_type': None,     'python_type': None,    'numpy_type': 'complex128', },  # noqa: E201, E202, E241
+    'ComplexViReal32':  { 'array_type': None,     'python_type': None,    'numpy_type': 'complex64',  },  # noqa: E201, E202, E241
+    'ComplexViInt16':   { 'array_type': None,     'python_type': None,    'numpy_type': 'int16',      },  # noqa: E201, E202, E241
```
ni-jfitzger, line 31: Why is it int16 and not int32? We're using complex64 for ViReal32 and complex 128 for ViReal64.

build/helper/codegen_helper.py
```diff
@@ -362,7 +366,12 @@ def _get_ctype_variable_definition_snippet_for_scalar(parameter, parameters, ivi
             definition = '{}.{}({})  # case S150'.format(module_name, parameter['ctypes_type'], parameter['python_name'])
         elif corresponding_buffer_parameters and corresponding_buffer_parameters[0]['direction'] == 'in':  # We are only looking at the first one to see if it is 'in'. Assumes all are the same here, assert below if not
             # Parameter denotes the size of another (the "corresponding") parameter.
-            definitions.append(parameter['ctypes_variable_name'] + ' = {0}.{1}(0 if {2} is None else len({2}))  # case S160'.format(module_name, parameter['ctypes_type'], corresponding_buffer_parameters[0]['python_name']))
+
+            # Interleaved array length is going to be double the length of number of samples
+            if 'interleaved' in corresponding_buffer_parameters[0]['complex_type']:
+                definitions.append(parameter['ctypes_variable_name'] + ' = {0}.{1}(0 if {2} is None else len({2}) // 2)  # case S160'.format(module_name, parameter['ctypes_type'], corresponding_buffer_parameters[0]['python_name']))
```
ni-jfitzger, line 372: Where is this used?

CHANGELOG.md
```diff
@@ -1703,6 +1710,7 @@
 
 #### [niscope] Unreleased
 - Added
+  - (Common) Allow functions to have parameters that take in numpy.complex types.
```
ni-jfitzger, line 1713: The CHANGELOG is meant for end users, not developers.
This is an implementation detail with no impact on usage, because there are no functions written to accept numpy.complex types. This will just confuse users.

build/defines.mak
ni-jfitzger: You could also argue that we're importing other unused modules to session.py, which is we have `# noqa: F401` for the enums import. The difference is that enums.py could potentially be used by session.py in the future with the addition of a new method, whereas some APIs will never use these complex types.

build/helper/codegen_helper.py
vnktshr21: This is needed for one of the RFSG API which is supposed to take in complex data in the form of unscaled interleaved numpy.int16 values in a numpy array. At the C API, this is still a struct of int16 real and imaginery numbers, hence the number of samples is half the  incoming array.

build/helper/helper.py
vnktshr21: Since ComplexViInt16 is being mapped to a interleaved int16 array, it had to match the datatype of each of real or imaginery value. complex128 representing complex numbers with real and imaginery real64 values seems to be a numpy convention and its different than IVI convention which is used in the C API. I think doubling of the number in the name should happen only when a single element represents double the number of bits, which is not true in the ComplexViInt16 case.

CHANGELOG.md
vnktshr21: Makes sense. This will be fixed.

build/templates/_library_interpreter.py.mako
```diff
@@ -39,14 +40,19 @@ _was_runtime_environment_set = None
 
 % endif
 # Helper functions for creating ctypes needed for calling into the driver DLL
-def _get_ctypes_pointer_for_buffer(value=None, library_type=None, size=None):
+def _get_ctypes_pointer_for_buffer(value=None, library_type=None, size=None, complex_type='none'):
```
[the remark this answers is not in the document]
rahulr-NI: Updated

CHANGELOG.md
rahulr-NI: Updated

build/helper/codegen_helper.py
ni-jfitzger: But I didn't see it being used anywhere when I checked.
Is it for functions that haven't been added yet?

CHANGELOG.md
```diff
@@ -2140,7 +2141,6 @@
 - [0.2.0](#nise-020---2018-10-25)
 - [0.1.0](#nise-010---2018-10-17)
 
-#### [nise] Unreleased
```
ni-jfitzger: Please restore this line.

CHANGELOG.md
```diff
@@ -2296,7 +2296,6 @@
 - [0.3.0](#niswitch-030---2017-10-13)
 - [0.2.0](#niswitch-020---2017-09-20)
 
-#### [niswitch] Unreleased
```
ni-jfitzger: Please restore this line

CHANGELOG.md
```diff
@@ -2547,7 +2546,6 @@
 - [0.3.0](#nitclk-030---2019-11-19)
 - [0.1.0](#nitclk-010---2019-10-21)
 
-#### [nitclk] Unreleased
```
ni-jfitzger: Please restore this line

generated/nirfsg/nirfsg/_library_interpreter.py
```diff
@@ -18,7 +19,12 @@ def _get_ctypes_pointer_for_buffer(value=None, library_type=None, size=None):
         return ctypes.cast(addr, ctypes.POINTER(library_type))
     elif str(type(value)).find("'numpy.ndarray'") != -1:
         import numpy
-        return numpy.ctypeslib.as_ctypes(value)
+        if library_type in (_complextype.ComplexViInt16, _complextype.ComplexViReal32, _complextype.ComplexViReal64):
```
ni-jfitzger: Nitpick: It feels weird to use a tuple here, instead of a list, but I suppose it's just a matter of style. If it works, it's fine.

build/helper/codegen_helper.py
```diff
@@ -426,7 +435,10 @@ def _get_ctype_variable_definition_snippet_for_buffers(parameter, parameters, iv
     definition = None
 
     if parameter['numpy'] is True and use_numpy_array is True:
-        definition = '_get_ctypes_pointer_for_buffer(value={})  # case B510'.format(parameter['python_name'])
+        if parameter['complex_type'] == 'none':
+            definition = '_get_ctypes_pointer_for_buffer(value={})  # case B510'.format(parameter['python_name'])
+        else:
+            definition = '_get_ctypes_pointer_for_buffer(value={}, library_type={}.{})  # case B510'.format(parameter['python_name'], module_name, parameter['ctypes_type'])
```
ni-jfitzger, line 441: @marcoskirsch do you know where these BXXX sequences come from. Is this correct usage? It feels to me like it should be separate from B510.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -51,7 +51,7 @@ def niFake_read_warning(self, vi, maximum_time, reading):  # noqa: N802
         reading.contents.value = self.reading
         return self.error_code_return
 
-    def get_ctypes_pointer_for_buffer_side_effect(self, value, library_type=None):
+    def get_ctypes_pointer_for_buffer_side_effect(self, value, library_type=None, complex_type=None):
         ret_val = self.get_ctypes_pointer_for_buffer_side_effect_items[self.get_ctypes_pointer_for_buffer_side_effect_count]
```
ni-jfitzger: This `complex_type` parameter isn't being used.
Also, this function isn't used anywhere. I'll assume we wrote it for a reason, but I see no reason to add an unused parameter.

ni-jfitzger: Nevermind about it not being used. I was looking for a call, but that's not how it works.

ni-jfitzger: Still, do we need this new parameter?

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -843,6 +843,73 @@ def test_import_attribute_configuration_buffer_str(self):
...
+        import numpy as np
+
+        invalid_waveform_data = np.full(1000, 0.707 + 0.707j, dtype=np.complex128)
+        expected_error_message = "Invalid waveform data provided. Expected a non-empty array of complex64."
+        interpreter = self.get_initialized_library_interpreter()
+        self.patched_library.niFake_WriteWaveformComplexF32.side_effect = ValueError("Invalid waveform data provided. Expected a non-empty array of complex64.")
+        with pytest.raises(ValueError) as exc_info:
+            interpreter.write_waveform_complex_f32(invalid_waveform_data)
+
+        assert str(exc_info.value) == expected_error_message
```
ni-jfitzger: Rather than passing valid data and patching it to error, shouldn't you pass in invalid data, avoid using patch_library and confirm that it errors?

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -843,6 +843,73 @@ def test_import_attribute_configuration_buffer_str(self):
...
+        self.patched_library.niFake_WriteWaveformComplexF64.side_effect = self.side_effects_helper.niFake_WriteWaveformComplexF64
+        interpreter = self.get_initialized_library_interpreter()
+        interpreter.write_waveform_complex_f64(waveform_data)
+        self.patched_library.niFake_WriteWaveformComplexF64.assert_called_once_with(
+            _matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST),
+            _matchers.ViInt32Matcher(number_of_samples),
+            _matchers.ComplexViReal64PointerMatcher(waveform_data_pointer, number_of_samples)
+        )
+
+    def test_write_waveform_complex_f64_invalid_input(self):
```
ni-jfitzger: This test name is confusing because you're calling write_waveform_complex_f32, not write_waveform_complex_f64.

generated/nifake/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -843,6 +843,73 @@ def test_import_attribute_configuration_buffer_str(self):
...
+        invalid_waveform_data = np.array([], dtype=np.complex64)
+        expected_error_message = "Invalid waveform data provided. Expected a non-empty array of Int16."
+
+        interpreter = self.get_initialized_library_interpreter()
+        self.patched_library.niFake_WriteWaveformComplexI16.side_effect = ValueError("Invalid waveform data provided. Expected a non-empty array of Int16.")
+        with pytest.raises(ValueError) as exc_info:
+            interpreter.write_waveform_complex_i16(invalid_waveform_data)
+        assert str(exc_info.value) == expected_error_message
+
+    def test_write_interleaved_complexi16_valid_input(self):
```
ni-jfitzger: Again, the test name doesn't match the function being called.

generated/nidcpower/nidcpower/_complextype.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# This file was generated
+import ctypes
+import nidcpower._visatype as _visatype
+
+
+class ComplexViReal64(ctypes.Structure):
```
marcoskirsch: I really really don't want to add new classes to nimi-python for fundamental types.

I also want to point out ComplexViReal64 etc do not exist in our C codebases. Their names are misleading, I thought they were an actual VISA type or something.

Is there anything out there (Python, numtype, nitype) we can use instead of creating new classes?

If these are purely for internal use, then I can live with it. But not with their made up names. What happened to NIComplexNumberF32 and NIComplexNumber? Are these unacceptable? (they exist in C headers and they describe structures compatible with these)

build/defines.mak
marcoskirsch: Yeah, it's not great to bundle dead code with the modules. I am unsure that we want to fix that in this PR but certainly before we next ship nimi-python.

build/helper/codegen_helper.py
```diff
@@ -256,6 +256,10 @@ def get_ctype_variable_declaration_snippet(parameter, parameters, ivi_dance_step
     else:
         module_name = '_visatype'
 
+    # Use _complextype.py file for complex parameter
+    if parameter['complex_type'] != 'none':
```
marcoskirsch: Is this a new key in the metadata?

Since I can't think of a reason for this to ever be overridden in metadata, I would rather have a function that is given the type and returns true/false.

If you have a great justification for putting this in metadata, then I'd ask you to _always_ expand it to true/flase rather than look for 'none' (!?)

marcoskirsch: Could you put this in the if/else train above? why is this a separate one if it's setting the same variable?

build/helper/codegen_helper.py
marcoskirsch: Consider adding in the PR when it is needed, this PR is already massive.

build/helper/codegen_helper.py
```diff
@@ -362,7 +366,12 @@ def _get_ctype_variable_definition_snippet_for_scalar(parameter, parameters, ivi
             definition = '{}.{}({})  # case S150'.format(module_name, parameter['ctypes_type'], parameter['python_name'])
         elif corresponding_buffer_parameters and corresponding_buffer_parameters[0]['direction'] == 'in':  # We are only looking at the first one to see if it is 'in'. Assumes all are the same here, assert below if not
             # Parameter denotes the size of another (the "corresponding") parameter.
-            definitions.append(parameter['ctypes_variable_name'] + ' = {0}.{1}(0 if {2} is None else len({2}))  # case S160'.format(module_name, parameter['ctypes_type'], corresponding_buffer_parameters[0]['python_name']))
+
+            # Interleaved array length is going to be double the length of number of samples
```
marcoskirsch: Expand this comment to provide more context, that this is used for complex waveforms in which the real/imaginary parts are interleaved.

Also if possible add an assert in the if statements that checks that the type is indeed complex

build/helper/codegen_helper.py
marcoskirsch: no idea

build/helper/helper.py
```diff
@@ -7,25 +7,28 @@
...
+    'ViReal64':         { 'array_type': 'd',      'python_type': 'float', 'numpy_type': 'float64',    },  # noqa: E201, E202, E241
+    'ViStatus':         { 'array_type': None,     'python_type': 'int',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViSession':        { 'array_type': None,     'python_type': 'int',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViAttr':           { 'array_type': None,     'python_type': 'int',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViChar':           { 'array_type': None,     'python_type': 'int',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViChar[]':         { 'array_type': None,     'python_type': 'str',   'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViBoolean':        { 'array_type': None,     'python_type': 'bool',  'numpy_type': None,         },  # noqa: E201, E202, E241
+    'ViRsrc':           { 'array_type': None,     'python_type': 'str',   'numpy_type': 'bool_',      },  # noqa: E201, E202, E241
+    'ComplexViReal64':  { 'array_type': None,     'python_type': None,    'numpy_type': 'complex128', },  # noqa: E201, E202, E241
```
marcoskirsch: Use names that map to existing types in our C APIs

build/helper/helper.py
marcoskirsch: This broke my brain too.

In all the other cases you are mapping a single `<key>` to a single `<python_type>` and a single `<numpy_type>`.

But in the `ComplexViInt16` you map a single `ComplexViInt16` to two `int16`s? This isn't right. We need to figure out how to resolve this.

build/helper/metadata_add_all.py
```diff
@@ -126,6 +126,12 @@ def _add_ctypes_type(parameter, config):
         parameter['ctypes_type_library_call'] = module_name + parameter['ctypes_type']
 
 
+def _add_complex_type(parameter):
+    '''Adds a complex_type parameter to the metadata for complex numbers'''
+    if 'complex_type' not in parameter:
```
marcoskirsch, line 131: will `complex_type` ever be in `parameter` or is this an attempt at future proofing?

build/templates/_library.py.mako
```diff
@@ -18,6 +18,7 @@ import ctypes
 import ${module_name}.errors as errors
 import threading
 
+from ${module_name}._complextype import *  # noqa: F401,F403,H303
 from ${module_name}._visatype import *  # noqa: F403,H303
```
marcoskirsch, line 25: We should solve the needless import of _complextype together with needless import of _visatype (if such thing happens). In its own PR. Please open an issue.

build/templates/_matchers.py.mako
```diff
@@ -271,6 +272,65 @@ class ViReal64PointerMatcher(_PointerMatcher):
         _PointerMatcher.__init__(self, _visatype.ViReal64)
 
 
+class ComplexViReal64PointerMatcher(_PointerMatcher):
```
marcoskirsch: Can you avoid 3 distinct implementations for the matches? they appear to be almost identical.

build/unit_tests/test_metadata_add_all.py
```diff
@@ -738,6 +751,7 @@ def _compare_dicts(actual, expected):
                 'grpc_request_snippet': 'data=data',
                 'use_in_python_api': True,
                 'python_name_or_default_for_init': 'data',
+                'complex_type': 'none',
```
marcoskirsch: I only see `'none'`
Are there examples where there's another valid value here?

src/nirfsg/templates/session.py/write_arb_waveform.py.mako
```diff
@@ -0,0 +1,24 @@
+<%page args="f, config, method_template"/>\
```
marcoskirsch: Adding the fancy function to nirfsg should be a PR separate to the one adding complex type support.
This is already 71 files long and will take multiple iterations to merge. It's not realistic for reviewers to do a good job in this manner.

tox.ini
```diff
@@ -156,6 +156,7 @@ deps =
     installers: build
     flake8: hacking
     flake8: pep8-naming
+    docs: snowballstemmer == 2.2.0  # Hardcoding due to https://github.com/sphinx-doc/sphinx/issues/13533
```
marcoskirsch: what is this for?

marcoskirsch (changes requested):

generated/nidcpower/nidcpower/_complextype.py
bkeryan: My recommendation is to try using NumPy arrays of `np.complex128` when calling into C DLLs. (Disclaimer: I haven't actually done this myself.)

I have added a NumPy dtype for complex integers to nitypes: https://nitypes.readthedocs.io/en/latest/autoapi/nitypes/complex/index.html

tox.ini
ni-jfitzger: Looks like some docs dependency had broken Sphinx.
According to that link, It's resolved now, so this should be removed.

src/nifake/metadata/functions.py
```diff
@@ -2797,6 +2797,171 @@
...
+                'type': 'ComplexViReal64[]',
+                'use_in_python_api': True,
+                'use_numpy_array': True
+            }
+        ],
+        'returns': 'ViStatus'
+    },
+    'WriteWaveformComplexI16': {
+        'codegen_method': 'public',
+        'documentation': {
```
ni-jfitzger, line 2912: I know this is just nifake, but this description feels very misleading.

build/defines.mak
ni-jfitzger: If we're going to change it, I'd rather it be in this PR.
It'll greatly reduce the number of files in the PR, if we're only changing the generated output for nirfsg and nifake.

ni-jfitzger: Nevermind. I see this comment: https://github.com/ni/nimi-python/pull/2098/files#r2129228871
Let's just handle it in a followup.

build/helper/metadata_add_all.py
ni-jfitzger: They're already setting it in functions.py for nifake and nirfsg as part of this PR.

build/helper/codegen_helper.py
ni-jfitzger: >Since I can't think of a reason for this to ever be overridden in metadata, I would rather have a function that is given the type and returns true/false.

They're not setting it to True or False. They're setting it to 'numpy' and 'interleaved'.

But we shouldn't set it to 'none'. If we must, it should be set to None and we should use `is not None` here.

build/helper/codegen_helper.py
ni-jfitzger: Found it!
https://github.com/ni/nimi-python/pull/715

S is for Scalar and B is for Buffer.
We seem to have precedent for usage of a parameter to if else behavior for a case in B550, so I'm okay with not having a new case number. But there should probably be some test changes in test_codegen_helper.py.

build/helper/helper.py
vnktshr21: this is just how the rfsg writewaveformcomplexi16 api wants it. the complex_type metadata field is being used to call out this difference during codegen. When we add nitypes support to nimi-python, we could leverage the same field to handle complexint16 in one more way(probably how MI driver api would need it to be).

generated/nidcpower/nidcpower/_complextype.py
vnktshr21: These would be internal use only. We have renamed these types to match the names in C header like you pointed out.

build/helper/metadata_add_all.py
ni-jfitzger: Okay, now they're only setting it in nifake, because they yanked the nirfsg changes for a followup PR.

ni-jfitzger: Maybe we should document this new key, internally?

build/helper/metadata_filters.py
```diff
@@ -285,6 +304,23 @@
...
+        'skip_size_parameter': True,
+        'reordered_for_default_values': False,
+        'skip_repeated_capability_parameter': True,
+        'skip_non_enum_parameter': False,
+        'skip_numpy_parameters': False,
+        'skip_all_except_numpy_parameters': True,
+        'mechanism': 'any',
+        'python_api_list': True,
+        'skip_all_except_complex_type_parameters': True,
+        'skip_all_except_complex_type_parameters': True,
```
ni-jfitzger: Duplicate keys.

build/helper/metadata_filters.py
```diff
@@ -451,3 +489,13 @@ def filter_codegen_enums(enums):
     return {k: v for k, v in enums.items() if v['codegen_method'] != 'no'}
 
 
+def are_complex_parameters_used(functions):
+    '''Returns function metadata only for those functions to included the library layer (library.py and mock_helper.py)'''
```
ni-jfitzger: >'''Returns function metadata only for those functions to included the library layer (library.py and mock_helper.py)'''

Needs rephrasing.

ni-jfitzger: In fact, doesn't this return a bool?

build/templates/_library_interpreter.py.mako
```diff
@@ -20,6 +21,9 @@ import hightime  # noqa: F401
 % if 'SetRuntimeEnvironment' in functions:
 import platform
 
+% endif
+% if are_complex_parameters_used:
+import ${module_name}._complextype as _complextype  # noqa: F401
```
ni-jfitzger: I don't think we need the noqa, anymore.

src/nifake/unit_tests/test_session.py
```diff
@@ -926,6 +926,26 @@ def test_export_attribute_configuration_buffer(self):
...
+
+    def test_session_write_waveform_complex_i16_invalid_dtype(self):
+        import numpy as np
+        invalid_waveform_data = np.full(10, 1.0 + 1.0j, dtype=np.complex64)
+        expected_error_message = "waveform_data_array must be numpy.ndarray of dtype=int16, is complex64"
+        import pytest
+        with nifake.Session('dev1', grpc_options=nifake.GrpcSessionOptions(object(), '')) as session:
+            with pytest.raises(TypeError) as exc_info:
+                session.write_waveform_numpy_complex_i16(invalid_waveform_data)
+            assert str(exc_info.value) == expected_error_message
```
ni-jfitzger: Issue 1: You're only testing error behavior for 2 out of the 3 complex waveform types?
Issue 2: Why did you add these tests to TestGrpcSession instead of TestSession?
         Do you even expect to support gRPC?

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -843,6 +843,49 @@ def test_import_attribute_configuration_buffer_str(self):
             interpreter.import_attribute_configuration_buffer(configuration)
         self.patched_library.niFake_ImportAttributeConfigurationBuffer.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), _matchers.ViInt32Matcher(len(configuration)), _matchers.ViInt8BufferMatcher(expected_list))
 
+    def test_write_waveform_numpy_complex128_valid_input(self):
+        import ctypes
+        import numpy as np
```
ni-jfitzger: ctypes and numpy are already imported at the top of the module. You don't need to import them again, here.

generated/nifake/nifake/session.py
```diff
@@ -1670,6 +1670,66 @@ def write_waveform_numpy(self, waveform):
...
+        if type(waveform_data_array) is not numpy.ndarray:
+            raise TypeError('waveform_data_array must be {0}, is {1}'.format(numpy.ndarray, type(waveform_data_array)))
+        if numpy.isfortran(waveform_data_array) is True:
+            raise TypeError('waveform_data_array must be in C-order')
+        if waveform_data_array.dtype is not numpy.dtype('complex128'):
+            raise TypeError('waveform_data_array must be numpy.ndarray of dtype=complex128, is ' + str(waveform_data_array.dtype))
+        self._interpreter.write_waveform_complex_f64(waveform_data_array)
+
+    @ivi_synchronized
+    def write_waveform_numpy_complex_i16(self, waveform_data_array):
```
ni-jfitzger: Why is numpy part of the name for this method, but not the others. I don't like this inconsistency.

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -843,6 +843,49 @@ def test_import_attribute_configuration_buffer_str(self):
             interpreter.import_attribute_configuration_buffer(configuration)
         self.patched_library.niFake_ImportAttributeConfigurationBuffer.assert_called_once_with(_matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST), _matchers.ViInt32Matcher(len(configuration)), _matchers.ViInt8BufferMatcher(expected_list))
 
+    def test_write_waveform_numpy_complex128_valid_input(self):
+        import ctypes
```
ni-jfitzger: `test_write_waveform_numpy_complex128_valid_input` does not align with the name of the LibraryInterpreter method that you're testing.

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -843,6 +843,49 @@ def test_import_attribute_configuration_buffer_str(self):
...
+        interpreter = self.get_initialized_library_interpreter()
+        interpreter.write_waveform_complex_f64(waveform_data)
+        self.patched_library.niFake_WriteWaveformComplexF64.assert_called_once_with(
+            _matchers.ViSessionMatcher(SESSION_NUM_FOR_TEST),
+            _matchers.ViInt32Matcher(number_of_samples),
+            _matchers.NIComplexNumberPointerMatcher(waveform_data_pointer, number_of_samples)
+        )
+
+    def test_write_waveform_numpy_complexi16_valid_input(self):
+        import ctypes
```
ni-jfitzger: Aren't you missing a test case?
Where's `test_write_waveform_complex_f32_valid_input()`?

build/helper/metadata_filters.py
```diff
@@ -285,6 +304,22 @@
...
+        'but_keep_output_numpy_array_parameters': True,
+        'skip_size_parameter': True,
+        'reordered_for_default_values': False,
+        'skip_repeated_capability_parameter': True,
+        'skip_non_enum_parameter': False,
+        'skip_numpy_parameters': False,
+        'skip_all_except_numpy_parameters': True,
+        'mechanism': 'any',
+        'python_api_list': True,
+        'skip_all_except_complex_type_parameters': True,
```
ni-jfitzger:
```suggestion
    ParameterUsageOptions.COMPLEX_NUMBER_PARAMETERS: {
        'skip_session_handle': True,
        'skip_input_parameters': False,
        'skip_output_parameters': False,
        'but_keep_output_numpy_array_parameters': False,
        'skip_size_parameter': False,
        'reordered_for_default_values': False,
        'skip_repeated_capability_parameter': False,
        'skip_non_enum_parameter': False,
        'skip_numpy_parameters': False,
        'skip_all_except_numpy_parameters': False,
        'mechanism': 'any',
        'python_api_list': True,
        'skip_all_except_complex_type_parameters': True,
```

This seems simpler.

generated/nidcpower/nidcpower/_complextype.py
ni-jfitzger: This seems to be resolved. The new names match our existing C header type names. As mentioned, these are for internal use only, (thus the _ prefix in the file name). We can change it later if, needed.

build/helper/helper.py
ni-jfitzger: These have been renamed.
The new names follow types in our header. NIComplexNumberF32 is named such because the real and imaginary parts are each f32s.
Similarly, NIComplexI16 is named such because the real and imaginary parts are each i16s.

The numpy types choose to name their complex types differently, so a Complex Number with f64s is complex128 and a complex number with f32s is complex64.
There is no equivalent numpy complex type for i16s, so we have to create an interleaved numpy array of dtype int16, counting the length as half the number of values.

We should probably add some comments here, since it's a point of major confusion, but I don't think there are any code changes, needed.

build/helper/codegen_helper.py
ni-jfitzger: We may want to add a test case in a followup PR, but this seems okay.

ni-jfitzger (approved):

PR #2114, opened 2025-07-08
Enabling enum flags in APIs

rahulr-NI:

Support Enum flags with type ViUInt64

Ensure that there are no failures in tox build.

### DIFF
--- a/src/nifake/unit_tests/test_session.py
+++ b/src/nifake/unit_tests/test_session.py
@@ -844,18 +844,16 @@ class TestSession:
         self.patched_library_interpreter.function_with_intflag_parameter.return_value = None
         with nifake.Session('dev1') as session:
             session.function_with_intflag_parameter(flags)
-        self.patched_library_interpreter.function_with_intflag_parameter.assert_called_once_with(flags)
-        called_arg = self.patched_library_interpreter.function_with_intflag_parameter.call_args[0][0]
-        # Check if the called argument is an IntFlagEnum or its value with OR operation
-        assert called_arg == 9223372036854775809 or (hasattr(called_arg, "value") and called_arg.value == 9223372036854775809)
+        self.patched_library_interpreter.function_with_intflag_parameter.assert_called_once_with(9223372036854775809)
 
     def test_with_intflag_parameter_invalid(self):
         invalid_flag = 5
         with nifake.Session('dev1') as session:
-            with pytest.raises(TypeError) as exc_info:
+            try:
                 session.function_with_intflag_parameter(invalid_flag)
-            assert "Parameter flag must be of type" in str(exc_info.value)
-            assert "IntFlagEnum" in str(exc_info.value)
+                assert False
+            except TypeError:
+                pass
 
     def test_session_write_waveform_numpy_complex64_invalid_dtype(self):
         invalid_waveform_data = numpy.full(10, 1.0 + 1.0j, dtype=numpy.complex128)
--- a/generated/nifake/nifake/unit_tests/test_session.py
+++ b/generated/nifake/nifake/unit_tests/test_session.py
@@ -844,18 +844,16 @@ class TestSession:
         self.patched_library_interpreter.function_with_intflag_parameter.return_value = None
         with nifake.Session('dev1') as session:
             session.function_with_intflag_parameter(flags)
-        self.patched_library_interpreter.function_with_intflag_parameter.assert_called_once_with(flags)
-        called_arg = self.patched_library_interpreter.function_with_intflag_parameter.call_args[0][0]
-        # Check if the called argument is an IntFlagEnum or its value with OR operation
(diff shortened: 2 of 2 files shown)


build/templates/enums.py.mako
```diff
@@ -6,16 +6,23 @@ config = template_parameters['metadata'].config
...
-class ${enums[enum_name]['python_name']}(Enum):
+class ${enums[enum_name]['python_name']}(${enums[enum_name].get('enum_class', 'Enum')}):
 <%
     print_list = []
 %>\
-    % for enum_value in enums[enum_name]['values']:
-    % if type(enum_value['value']) is str:
+    % for i, enum_value in enumerate(enums[enum_name]['values']):
+    % if enums[enum_name].get('enum_class', 'Enum') == 'IntFlag' and enum_value['value'] == 0:
+    ${enum_value['python_name']} = 0
```
ni-jfitzger: >     % if enums[enum_name].get('enum_class', 'Enum') == 'IntFlag' and enum_value['value'] == 0:
>    ${enum_value['python_name']} = 0

I don't understand the point of these 2 lines.
If the value is 0, how is this different from what you get with

>     % elif enums[enum_name].get('enum_class', 'Enum') == 'IntFlag' and isinstance(enum_value['value'], int):
>    ${enum_value['python_name']} = ${enum_value['value']}

?

ni-jfitzger: Are you trying to allow a value of False, instead of just ints?

build/templates/enums.py.mako
```diff
@@ -6,16 +6,23 @@ config = template_parameters['metadata'].config
...
+    % elif type(enum_value['value']) is str:
     ${enum_value['python_name']} = '${enum_value['value']}'
+    % elif enums[enum_name].get('enum_class', 'Enum') == 'IntFlag' and isinstance(enum_value['value'], int):
```
ni-jfitzger: So if it's not an int, we'll omit the value?
I might prefer an error to be raised, but this is definitely better than letting an invalid value slip in to the API.

src/nifake/unit_tests/test_session.py
```diff
@@ -837,10 +838,27 @@ def test_return_timedeltas(self):
...
 
+    def test_function_with_int_flag_parameter_with_specific_flag(self):
+        flags = nifake.IntFlagEnum.C
+        self.patched_library_interpreter.function_with_int_flag_parameter.side_effect = None
+        self.patched_library_interpreter.function_with_int_flag_parameter.return_value = None
+        with nifake.Session('dev1') as session:
+            session.function_with_int_flag_parameter(flags)
+        self.patched_library_interpreter.function_with_int_flag_parameter.assert_called_once_with(flags)
+        called_arg = self.patched_library_interpreter.function_with_int_flag_parameter.call_args[0][0]
+        assert called_arg == 1 or called_arg.value == 4
```
ni-jfitzger: >         called_arg = self.patched_library_interpreter.function_with_int_flag_parameter.call_args[0][0]
>        assert called_arg == 1 or called_arg.value == 4

I don't understand what you're doing here.
You've alerady asserted that we passsed the right thing down to the library interpreter.

Now you're confirming it again?
And what's going on with the "== 1 or ... == 4"? Why would it equal 1?

ni-jfitzger:
Has there been a discussion of the desired [FlagBoundary](https://docs.python.org/3/library/enum.html#enum.FlagBoundary) behavior?

The default for IntFlag is Keep, whereas the default for Flag is Strict (which raises an error for unsupported values, before the user can ever pass it to the driver). I would think Strict would be preferred, but maybe your team has a different preference or maybe it's a per-Enum decision?

src/nifake/unit_tests/test_session.py
rahulr-NI: Yes Jay. First assertion was meant to check the API contract. Second assertion check was to ensure Enum's value.  I believe it's good to remove the second assertion as we are double checking the Enum value.   Checking with value 1 is also not required here. It should not be compared to 1.

build/templates/enums.py.mako
rahulr-NI: Hi Jay, Two conditions should be sufficient here.I have reverted the changes done based on the condition. Please let me know if any actions to be taken here.

src/nifake/unit_tests/test_session.py
```diff
@@ -837,10 +838,26 @@ def test_return_timedeltas(self):
...
+        self.patched_library_interpreter.function_with_intflag_parameter.assert_called_once_with(9223372036854775809)
+
+    def test_with_intflag_parameter_invalid(self):
+        invalid_flag = 5
+        with nifake.Session('dev1') as session:
+            try:
+                session.function_with_intflag_parameter(invalid_flag)
+                assert False
+            except TypeError:
+                pass
```
ni-jfitzger, line 856: I actually like `with pytest.raises` better than this approach, but this is fine.

PR #2115, opened 2025-07-08
Adding Write Arb API and system tests and misc fixes

vnktshr21:

- Enables write arb waveform API
  - Adds a fancy function which supports multiple complex datatypes and calls into the appropriate interpreter method based on incoming waveform datatype
  - Implementation follows nifgen write_waveform fancy function pattern (nifgen has the exact same use case)
  - This fancy function would also allow adding more complex datatypes support in future if required, with one single public write_arb_waveform function supporting all of it
  - Doc tip and function description for fancy function is tweaked to call out the supported datatypes
- Adds system tests for all the important supported API
  - Some tests can run only on real hardware and are skipped based on a single flag
- Adds arb and script mode examples
- Examples and system tests are now put through flake checks
- For set, get burst start, burst stop and marker event location APIs
  - Fixes the size parameter behavior
  - Fixes the datatype of locations
  - Marks channel name parameter as repeated capability with appropriate repeated capability type field value set
- Changes reset function case to match the header case
- Adds correct enum for parameter of configure power level type and digital trigger edge functions
- Self cal last date time function gets a module parameter matching the C API
- Removed unnecessary tabs and line breaks in some of the documentation
- Fixed the option string parameter handling for initialize
  - Metadata was missing template specification in rfsg

None

- New write arb waveform API(all datatypes) are tested to be functional with real hardware
  - Correct signal generation verified with Analyzer
- All system tests pass on both real and simulated hardware
  - Fixes for function parameters across multiple functions are all tested as part of system tests
- All examples run fine on both real and simulated hardware 
- Documentation improvements are visually inspected


src/nirfsg/examples/nirfsg_arb_waveform.py
```diff
@@ -0,0 +1,46 @@
...
+    example(args.resource_name, args.option_string, args.frequency, args.power_level, args.number_of_samples)
+
+
+def main():
+    _main(sys.argv[1:])
+
+
+def test_example():
+    options = "Simulate=1, DriverSetup=Model:5841"
+    example('5841', options, 1e9, -10.0)
```
ni-jfitzger:
```suggestion
    example('5841', options, 1e9, -10.0, 1000)
```

src/nirfsg/examples/nirfsg_script.py
```diff
@@ -0,0 +1,55 @@
...
+    example(args.resource_name, args.option_string, args.frequency, args.power_level, args.number_of_samples)
+
+
+def main():
+    _main(sys.argv[1:])
+
+
+def test_example():
+    options = "Simulate=1, DriverSetup=Model:5841"
+    example('5841', options, 1e9, -10.0)
```
ni-jfitzger:
```suggestion
    example('5841', options, 1e9, -10.0, 1000)
```

generated/nirfsg/nirfsg/session.py
```diff
@@ -6840,9 +6593,11 @@ def configure_digital_edge_script_trigger(self, trigger_id, source, edge):
 
             source (str): Specifies the source terminal for the digital edge Script Trigger. NI-RFSG sets the digital_edge_script_trigger_source property to this value.
 
-            edge (int): Specifies the active edge for the digital edge Script Trigger. NI-RFSG sets the digital_edge_script_trigger_edge property to this value.
+            edge (enums.ScriptTrigDigEdgeEdge): Specifies the active edge for the digital edge Script Trigger. NI-RFSG sets the digital_edge_script_trigger_edge property to this value.
 
         '''
+        if type(edge) is not enums.ScriptTrigDigEdgeEdge:
```
ni-jfitzger, line 6817: Repeating the word Edge is unfortunate, but I'm okay with it.

generated/nirfsg/nirfsg/session.py
```diff
@@ -6840,9 +6593,11 @@ def configure_digital_edge_script_trigger(self, trigger_id, source, edge):
...
+            raise TypeError('Parameter edge must be of type ' + str(enums.ScriptTrigDigEdgeEdge))
```
ni-jfitzger, line 6818: I think we need to talk about your enum names.

[Official guidance](https://github.com/ni/nimi-python/wiki/API-design-guidelines#enums) is to derive enum names from CVI FP and enum value names from C Vals. I can't say whether you've been doing that (even after glancing at the .fp file), but looking at the enums.py, I see a lot of unnecessary abbreviations in names like "Amp", "Anlg", "Dig", "Trig" and I think it hurts readability.

I've looked through our enums for nidcpower, nidigital, nifgen and niscope and it's pretty uncommon to abbreviate something in an enum name. My guidance would be this: If you don't abbreviate it in your function/attribute/parameter names, then don't abbreviate it in the enum name or enum value name. Don't worry about it being too verbose. Embrace readability.

generated/nirfsg/nirfsg/session.py
```diff
@@ -6840,9 +6593,11 @@ def configure_digital_edge_script_trigger(self, trigger_id, source, edge):
 
             source (str): Specifies the source terminal for the digital edge Script Trigger. NI-RFSG sets the digital_edge_script_trigger_source property to this value.
 
-            edge (int): Specifies the active edge for the digital edge Script Trigger. NI-RFSG sets the digital_edge_script_trigger_edge property to this value.
+            edge (enums.ScriptTrigDigEdgeEdge): Specifies the active edge for the digital edge Script Trigger. NI-RFSG sets the digital_edge_script_trigger_edge property to this value.
```
ni-jfitzger, line 6814: I don't want to make a big thing of this. It shouldn't hold up the PR.
But are you sure this function is valuable enough to merit inclusion in the API?
It just sets two properties.

Going by our API guidelines, it's okay to keep, because it sets 2 attributes, not one:

> * Don't include functions that are simple wrappers around setting/getting an attribute
    * Redundancy means confusion
    * Clients don't get type safety benefit that you get in C function anyway

But we could just as easily justify its exclusion.

generated/nirfsg/nirfsg/session.py
```diff
@@ -7532,16 +7299,28 @@ def get_external_calibration_last_date_and_time(self):
...
+            module (enums.Module): Specifies from which stand-alone module to retrieve the last successful self-calibration date and time.
+
 
         Returns:
             last_cal_datetime (hightime.datetime):
 
         '''
-        year, month, day, hour, minute, second = self._get_self_calibration_date_and_time()
+        year, month, day, hour, minute, second = self._get_self_calibration_date_and_time(module)
         return hightime.datetime(year, month, day, hour, minute)
```
ni-jfitzger: Minor issue: we seem to be discarding precision for no reason, here.
Most of our get_self_calibration_last_date_and_time functions across our APIs only return minute precision, but if this one is precise to the second, why not keep that precision?

We just need to update the last line of the datetime_wrappers.py.mako template to use ${output_params} for the parameter list passed to hightime.datetime(). Or use a different line based on whether "second" is part of output_params. Either would work.

Doesn't need to be done as part of this PR. In fact we could change it after the initial release.

src/nirfsg/metadata/functions.py
```diff
@@ -5025,6 +5045,7 @@
                     'description': 'Specifies the waveform name and the marker name. Example: "waveform::waveform0/marker0"'
                 },
                 'name': 'channelName',
+                'is_repeated_capability': False,
```
ni-jfitzger: Can you make these `'is_repeated_capability': False` changes in a separate PR?
I think they're going to be a hot topic.

I don't think @marcoskirsch will like this and I'm still not convinced that it makes sense.

src/nirfsg/system_tests/samples2pfile.s2p
```diff
@@ -0,0 +1,7 @@
+! 2-port S-parameter file, four frequency points
```
ni-jfitzger, line 1: Is .s2p the typical extension for a file like this?
I ask because I've never heard of it and it looks like a typical text file.

If this is typical, then just close the comment with a short response. I know we have some special file extensions for nidigital config files.

generated/nirfsg/nirfsg/session.py
```diff
@@ -8198,6 +8122,164 @@ def wait_until_settled(self, max_time_milliseconds):
...
+            else:
+                raise TypeError("Unsupported dtype. Is {}, expected {} or {} or {}".format(waveform_data_array.dtype, numpy.complex128, numpy.complex64, numpy.int16))
+        else:
+            raise TypeError("Unsupported dtype. Expected {} or {} or {}".format(numpy.complex128, numpy.complex64, numpy.int16))
+
+        return self._write_arb_waveform_complex_f64(waveform_name, waveform_data_array, more_data_pending)
+
     @ivi_synchronized
     def write_p2p_endpoint_i16(self, stream_endpoint, number_of_samples, endpoint_data):
         r'''write_p2p_endpoint_i16
```
ni-jfitzger, line 8341: `write_p2p_endpoint_i16` we still need a PR excluding this function.
Please make it a separate PR.

ni-jfitzger:
Can you include a justification for the fancy function in your PR description?
I think you're probably right to create it, but it would be nice to have the thought process written down for posterity, if it's not too much trouble.

generated/nirfsg/nirfsg/session.py
vnktshr21: This is a good point. The enum names were sourced from docsxml file while populating our internal metadata. Probably did not matter if the terms are shortened until now, but it is going to look bad in this python API case. We will not make this change as part of this PR, but we will have a follow up PR with the enum names having full terms before the first release. I have created work item to track this internally.

generated/nirfsg/nirfsg/session.py
vnktshr21: Given we have gone through the list of API with our product architect and narrowed down the ones to support for this release, I dont think this API warrants a revisit at this stage.

My thoughts, folks who are already used to the function in other language API, might start looking for the same in python too. Exclusion adds more things to the list of python specific exclusions to track with reasons. The product API guideline should definitely be considered while having this kind of function in any API in the first place. If we found this function to be worth adding in one language API, I would not want to completely remove the API on another language unless it logically doesn't make sense to add.

generated/nirfsg/nirfsg/session.py
vnktshr21: Fixed this.

src/nirfsg/metadata/functions.py
vnktshr21: For "locations" functions, I have now instead marked the applicable parameter to call out the correct repeated capability and marked is_repeated_capability as True for them. And removed authoring it as False for Save and Load configuration API. Channel name for these 2 API will be revisited before release.

The codegen and metadata lacks the capability of calling out multiple repeated capabilities for function parameters like it does for attributes. Couple of "locations" functions indeed requires nesting of repeated capability similar to attributes. Nesting and the runtime functionality works fine without issues since both single rep_cap and multiple will have same runtime behavior. Documentation snippet though will call out only one of the repeated capability. We will need to address this at some point. Not sure if other MI drivers hit upon this issue.

src/nirfsg/system_tests/samples2pfile.s2p
vnktshr21: _Is .s2p the typical extension for a file like this?_
Yes

https://en.wikipedia.org/wiki/Touchstone_file

generated/nirfsg/nirfsg/session.py
vnktshr21: Like we have already discussed on this topic, we will be having a separate PR to exclude all P2P related functions, attributes and enums. I have not included those changes as part of this PR. We have a work item to track this internally.

generated/nirfsg/nirfsg/session.py
ni-jfitzger: That's fine. Like I said, we could go either way on this one.

generated/nirfsg/nirfsg/session.py
```diff
@@ -8198,6 +8176,164 @@ def wait_until_settled(self, max_time_milliseconds):
...
+        This method accepts the complex baseband data in the form of numpy array of numpy.complex64 or numpy.complex128 or interleaved numpy array of numpy.int16. If the waveform to write is already allocated using the allocate_arb_waveform method, the **MORE_DATA_PENDING** parameter is ignored. The PXI-5670/5671 must be in the Configuration state before you call this method. When streaming is enabled, this method can be called when the PXIe-5672/5673/5673E or PXIe-5820/5830/5831/5832/5840/5841/5842/5860 is in the Generation state.
+
+        **Supported Devices** : PXIe-5644/5645/5646, PXIe-5672/5673/5673E, PXIe-5820/5830/5831/5832/5840/5841/5842/5860
+
+        **Related Topics**
+
+        `Streaming <https://www.ni.com/docs/en-US/bundle/ni-rfsg/page/streaming.html>`_
+
+        `Assigning Properties or Properties to a Waveform <https://www.ni.com/docs/en-US/bundle/ni-rfsg/page/assigning-properties-or-properties-to-a-wavef.html>`_
+
```
ni-jfitzger, line 8307: Wow, our codegen is messing up your link.

generated/nirfsg/nirfsg/session.py
```diff
@@ -8198,6 +8176,164 @@ def wait_until_settled(self, max_time_milliseconds):
...
+            if waveform_data_array.dtype == numpy.complex128:
+                return self._write_arb_waveform_complex_f64(waveform_name, waveform_data_array, more_data_pending)
+            elif waveform_data_array.dtype == numpy.complex64:
+                return self._write_arb_waveform_complex_f32(waveform_name, waveform_data_array, more_data_pending)
+            elif waveform_data_array.dtype == numpy.int16:
+                return self._write_arb_waveform_complex_i16(waveform_name, waveform_data_array)
+            else:
+                raise TypeError("Unsupported dtype. Is {}, expected {} or {} or {}".format(waveform_data_array.dtype, numpy.complex128, numpy.complex64, numpy.int16))
+        else:
+            raise TypeError("Unsupported dtype. Expected {} or {} or {}".format(numpy.complex128, numpy.complex64, numpy.int16))
```
ni-jfitzger: Is this the right error message? There may not even be a dtype if we reach this line.

src/nirfsg/metadata/functions.py
ni-jfitzger: nidigital is our only shipping API that uses nested repeated capabilities.

> Couple of "locations" functions indeed requires nesting of repeated capability similar to attributes.

I thought RF didn't need nesting for any functions or attributes.
Can you give an example?

> both single rep_cap and multiple will have same runtime behavior.

i'm not sure what you mean.

> Documentation snippet though will call out only one of the repeated capability. We will need to address this at some point. Not sure if other MI drivers hit upon this issue.

Documentation can be fixed after release. We'll need to snoop around and see what nidigital does.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+
+
+class SystemTests:
+    @pytest.fixture(scope='function')
+    def test_rfsg_device_session(self, session_creation_kwargs):
+        if use_simulated_session:
+            with nirfsg.Session("5841sim", options="Simulate=1, DriverSetup=Model:5841", **session_creation_kwargs) as sim_5841_session:
+                yield sim_5841_session
+        else:
+            with nirfsg.Session(real_hw_resource_name, options="", **session_creation_kwargs) as real_rfsg_device_session:
```
ni-jfitzger:
```suggestion
            with nirfsg.Session(real_hw_resource_name, **session_creation_kwargs) as real_rfsg_device_session:
```

Nitpick: options is an optional argument. I would omit it, since you're not passing any.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def test_set_int32_enum_attribute(self, test_rfsg_device_session):
+        test_rfsg_device_session.frequency_settling_units = nirfsg.FrequencySettlingUnits.TIME_AFTER_LOCK
+        assert test_rfsg_device_session.frequency_settling_units == nirfsg.FrequencySettlingUnits.TIME_AFTER_LOCK
+
+    def test_set_invalid_attribute_raises(self, test_rfsg_device_session):
+        with pytest.raises(AttributeError):
+            test_rfsg_device_session.non_existent_attribute = 123
+
+# Multi-threading related tests (picked from other driver tests)
```
ni-jfitzger:
```suggestion
# Multi-Threading tests
```

That extra detail is interesting for the review, but I don't think we need it, long-term. Other tweaks suggested are for consistency across all instances of these system tests.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        value = test_rfsg_device_session.external_calibration_recommended_interval
+        assert isinstance(value, int)
+
+    def test_set_int32_enum_attribute(self, test_rfsg_device_session):
+        test_rfsg_device_session.frequency_settling_units = nirfsg.FrequencySettlingUnits.TIME_AFTER_LOCK
+        assert test_rfsg_device_session.frequency_settling_units == nirfsg.FrequencySettlingUnits.TIME_AFTER_LOCK
+
+    def test_set_invalid_attribute_raises(self, test_rfsg_device_session):
+        with pytest.raises(AttributeError):
+            test_rfsg_device_session.non_existent_attribute = 123
```
ni-jfitzger, line 65:
```suggestion
```

You shouldn't need these. attribute get and set methods are covered by nifake unit tests, which we use to test commmon code.
I'm not sure we have anything in there to cover setting a non-existent attribute, but you're free to add such a test in a separate PR.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+def get_test_file_path(file_name):
+    return os.path.join(test_files_base_dir, file_name)
+
+
+sys.path.insert(0, str(pathlib.Path(__file__).parent.parent.parent / 'generated/nirfsg'))
+
+
+class SystemTests:
+    @pytest.fixture(scope='function')
+    def test_rfsg_device_session(self, session_creation_kwargs):
```
ni-jfitzger: It's weird to name a fixture "test_". That makes it sound like it's a test.
I would rename this to "session". If you do so, you'll need to replace all other instances, as well.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+
+    @pytest.mark.skipif(use_simulated_session is True, reason="Bad date returned by driver for simulated device")
+    def test_get_external_calibration_last_date_and_time(self, test_rfsg_device_session):
+        dt = test_rfsg_device_session.get_external_calibration_last_date_and_time()
+        assert isinstance(dt, hightime.datetime)
+
+    @pytest.mark.skipif(use_simulated_session is True, reason="Bad date returned by driver for simulated device")
+    def test_get_self_calibration_last_date_and_time(self, test_rfsg_device_session):
+        dt = test_rfsg_device_session.get_self_calibration_last_date_and_time(nirfsg.Module.PRIMARY_MODULE)
+        assert isinstance(dt, hightime.datetime)
```
ni-jfitzger: >Bad date returned by driver for simulated device

Bad how? Is it returning 0 for the year, month and day?

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        assert test_rfsg_device_session.waveform['mywaveform'].waveform_iq_rate == requested_waveform_iq_rate
+
+    def test_deembedding_port_rep_cap(self, test_rfsg_device_session):
+        port = test_rfsg_device_session.deembedding_port['']
+        requested_deembedding_type = nirfsg.DeembeddingTypeAttrVals.SCALAR
+        port.deembedding_type = requested_deembedding_type
+        assert port.deembedding_type == requested_deembedding_type
+
+# Configuration methods related tests
+    def test_configure_power_and_frequency(self, test_rfsg_device_session):
```
ni-jfitzger: If a test is dedicated to testing a specific function, you should name it after that function.
We're testing entrypoints, not driver behavior.

```suggestion
    def test_configure_rf(self, test_rfsg_device_session):
```

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        test_rfsg_device_session.configure_rf(2e9, -5.0)
+        assert test_rfsg_device_session.power_level == -5.0
+        assert test_rfsg_device_session.frequency == 2e9
+
+    def test_configure_output_enabled(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_output_enabled(True)
+        assert test_rfsg_device_session.output_enabled is True
+        test_rfsg_device_session.configure_output_enabled(False)
+        assert test_rfsg_device_session.output_enabled is False
```
ni-jfitzger: I think you should delete the configure_output_enabled function, as per [our API Guidelines](https://github.com/ni/nimi-python/wiki/API-design-guidelines).

> * Don't include functions that are simple wrappers around setting/getting an attribute
> * Redundancy means confusion
> * Clients don't get type safety benefit that you get in C function anyway

I feel much more strongly about this one than configure_digital_edge_script_trigger.
It appears to add no value to the API.
It's actually faster to type `session.output_enabled = True` than to call the function.

Feel free to overrule me if it does something besides just set the property.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+    def test_configure_generation_mode(self, test_rfsg_device_session):
+        # Test CW mode
+        test_rfsg_device_session.configure_generation_mode(nirfsg.GenerationMode.CW)
+        assert test_rfsg_device_session.generation_mode == nirfsg.GenerationMode.CW
+        # Test ARB_WAVEFORM mode
+        test_rfsg_device_session.configure_generation_mode(nirfsg.GenerationMode.ARB_WAVEFORM)
+        assert test_rfsg_device_session.generation_mode == nirfsg.GenerationMode.ARB_WAVEFORM
+        # Test SCRIPT mode
+        test_rfsg_device_session.configure_generation_mode(nirfsg.GenerationMode.SCRIPT)
+        assert test_rfsg_device_session.generation_mode == nirfsg.GenerationMode.SCRIPT
```
ni-jfitzger: I think you should delete the `configure_generation_mode` function, as per [our API Guidelines](https://github.com/ni/nimi-python/wiki/API-design-guidelines).

> * Don't include functions that are simple wrappers around setting/getting an attribute
> * Redundancy means confusion
> * Clients don't get type safety benefit that you get in C function anyway

I feel much more strongly about this one than `configure_digital_edge_script_trigger`.
It appears to add no value to the API.
It's actually faster to type `session.generation_mode = nirfsg.GenerationMode.CW` than to call the function.
Going by the examples, even your team seems to prefer the property.

Feel free to overrule me if it does something besides just set the property.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        test_rfsg_device_session.clear_all_arb_waveforms()
+        waveform_exists = test_rfsg_device_session.check_if_waveform_exists('mywaveform1')
+        assert waveform_exists is False
+        waveform_exists = test_rfsg_device_session.check_if_waveform_exists('mywaveform2')
+        assert waveform_exists is False
+
+    def test_allocate_arb_waveform(self, test_rfsg_device_session):
+        waveform_data = np.full(1000, 1 + 0j, dtype=np.complex128)
+        test_rfsg_device_session.allocate_arb_waveform('foo', len(waveform_data))
+        test_rfsg_device_session.write_arb_waveform('foo', waveform_data, False)
```
ni-jfitzger: I think a more correct test would be


```suggestion
        test_rfsg_device_session.allocate_arb_waveform('foo', len(waveform_data) * 2)
        test_rfsg_device_session.write_arb_waveform('foo', waveform_data, False)
        test_rfsg_device_session.write_arb_waveform('foo', waveform_data, False)
```

I'm not an expert on this API, but I did ~stay at a Holiday Inn Express last night~ read the function documentation.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+            assert False
+        except nirfsg.Error as e:
+            assert e.code == -1074101603
+            assert "A waveform matching the provided name does not exist in memory" in e.description
+
+    def test_configure_software_trigger(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_software_start_trigger()
+        assert test_rfsg_device_session.start_trigger_type == nirfsg.StartTrigType.SOFTWARE
+        test_rfsg_device_session.configure_software_script_trigger('scriptTrigger0')
+        assert test_rfsg_device_session.script_triggers[0].script_trigger_type == nirfsg.ScriptTrigType.SOFTWARE
```
ni-jfitzger: Given the existence of the `start_trigger_type` and `script_trigger_type`, `configure_software_start_trigger`, `configure_software_script_trigger`, `disable_start_trigger` and `disable_script_trigger` all seem to be of questionable value.

However, I don't feel as strongly about these as I do `configure_generation_mode` and `configure_output_enabled`, given that you may just want them to round out the API, due to the existence of `configure_digital_edge_start_trigger` and `configure_digital_edge_script_trigger`.

@marcoskirsch may feel more strongly about it than I do.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+            test_rfsg_device_session.write_script(script)
+            assert False
+        except nirfsg.Error as e:
+            assert e.code == -1074101603
+            assert "A waveform matching the provided name does not exist in memory" in e.description
+
+    def test_configure_software_trigger(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_software_start_trigger()
+        assert test_rfsg_device_session.start_trigger_type == nirfsg.StartTrigType.SOFTWARE
+        test_rfsg_device_session.configure_software_script_trigger('scriptTrigger0')
```
ni-jfitzger: > configure_software_script_trigger('scriptTrigger0')

The trigger_id shouldn't be a parameter, but a repeated capability. We have a script_triggers rep cap.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        assert test_rfsg_device_session.script_triggers[0].script_trigger_type == nirfsg.ScriptTrigType.SOFTWARE
+
+    def test_configure_digital_edge_trigger(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_digital_edge_start_trigger('PXI_Trig1', nirfsg.StartTrigDigEdgeEdge.RISING)
+        test_rfsg_device_session.configure_digital_edge_script_trigger('scriptTrigger1', 'PXI_Trig2', nirfsg.ScriptTrigDigEdgeEdge.FALLING)
```
ni-jfitzger: > configure_digital_edge_script_trigger('scriptTrigger1', 'PXI_Trig2', nirfsg.ScriptTrigDigEdgeEdge.FALLING)

The trigger_id shouldn't be a parameter, but a repeated capability. We have a script_triggers rep cap.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        assert test_rfsg_device_session.script_triggers[1].digital_edge_script_trigger_edge == nirfsg.ScriptTrigDigEdgeEdge.FALLING
+
+    def test_disable_trigger(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_software_start_trigger()
+        assert test_rfsg_device_session.start_trigger_type == nirfsg.StartTrigType.SOFTWARE
+        test_rfsg_device_session.disable_start_trigger()
+        assert test_rfsg_device_session.start_trigger_type == nirfsg.StartTrigType.NONE
+        test_rfsg_device_session.configure_software_script_trigger('scriptTrigger3')
+        assert test_rfsg_device_session.script_triggers[3].script_trigger_type == nirfsg.ScriptTrigType.SOFTWARE
+        test_rfsg_device_session.disable_script_trigger('scriptTrigger3')
```
ni-jfitzger: > disable_script_trigger('scriptTrigger3')

The trigger_id shouldn't be a parameter, but a repeated capability. We have a script_triggers rep cap.

src/nirfsg/system_tests/test_system_nirfsg.py
ni-jfitzger: In case of pushback, our design guidelines provide further guidance:

> * If there's debate on whether a function should be included or not, then don't include it
>   * Trivial to add if needed
>   * Hard to remove once added

src/nirfsg/system_tests/test_system_nirfsg.py
ni-jfitzger: In case of pushback, our design guidelines provide further guidance:

> * If there's debate on whether a function should be included or not, then don't include it
>   * Trivial to add if needed
>   * Hard to remove once added

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        assert test_rfsg_device_session.script_triggers[3].script_trigger_type == nirfsg.ScriptTrigType.NONE
+
+    def test_export_signal(self, test_rfsg_device_session):
```
ni-jfitzger: Do you really need an export_signal function?

It's redundant.

All of our other nimi-python APIs explicitly exclude this function from the Python API.
nidcpower, nidmm, nidigital, nifgen and niscope all deliberately omit it. You clearly have properties that you can set, instead.

You might feel uncomfortable removing things, because your team already reviewed your choices based on decisions made for your other APIs, assuming that things were included in those APIs for a good reason. But the truth is that our other APIs are a mess. Often, the reason was "IVI forced us to include it".

Our C APIs have too many functions and too many properties. I know because I've been digging through metadata for almost 2 years. When you have hundreds of functions and properties, it's hard to find the ones you need in the documentation. Redundancies don't make the user's life easier; they make it harder. And they give us more things to test and increase our test times.

Get rid of the function. It's unwieldy, anyway. Look at all the arguments you have to pass.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        test_rfsg_device_session.configure_rf(3e9, -15.0)
+        test_rfsg_device_session.iq_rate = 2e6
+        assert test_rfsg_device_session.frequency == 3e9
+        assert test_rfsg_device_session.power_level == -15.0
+        assert test_rfsg_device_session.iq_rate == 2e6
+        test_rfsg_device_session.load_configurations_from_file(get_test_file_path('tempConfiguration.json'))
+        assert test_rfsg_device_session.frequency == 2e9
+        assert test_rfsg_device_session.power_level == -5.0
+        assert test_rfsg_device_session.iq_rate == 1e6
+        os.remove(get_test_file_path('tempConfiguration.json'))
```
ni-jfitzger: Are these functions different from the import_attribute_configuration_file and export_attribute_configuration_file functions in our other APIs?
If not, would you consider renaming them to match? Users will probably want to use more than one of our APIs and it will be easier on them if common functions have the same names.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+
+# Basic tests for generation
+    @pytest.mark.skipif(use_simulated_session is False, reason="Test executed with status check in real hw")
```
ni-jfitzger, line 403: > Test executed with status check in real hw

What does this mean?

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+            test_rfsg_device_session.check_generation_status()
+
+    @pytest.mark.skipif(use_simulated_session is True, reason="isdone is always True on simulated device")
+    def test_cw_generation_with_status(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_rf(2e9, -5.0)
+        with test_rfsg_device_session.initiate():
+            isdone = test_rfsg_device_session.check_generation_status()
+            assert isdone is False  # isdone will never be True in CW mode
+
+    @pytest.mark.skipif(use_simulated_session is True, reason="isdone is always True on simulated device")
```
ni-jfitzger: >isdone is always True on simulated device

That shouldn't stop you from calling abort to make sure nothing bad happens.
Which is all you really need to test. Remember: this is an API test, not a behavioral test.
If simulation changes something when you call, great. Do your assert.
If not, don't sweat it. Call the function, anyway.

By the way, I consider this more of a test of the context manager.
The nifake unit test `test_acquisition_context_manager` has that covered.
You should avoid the context manager for your test and call abort explicitly, instead.

I'll admit, though, most of our APIs don't have a test_abort test, for whatever reason.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        with test_rfsg_device_session.initiate():
+            test_rfsg_device_session.send_software_edge_trigger(nirfsg.SoftwareTriggerType.START, '')
+            test_rfsg_device_session.send_software_edge_trigger(nirfsg.SoftwareTriggerType.SCRIPT, 'scriptTrigger0')
+
+    @pytest.mark.skipif(sys.platform == "linux", reason="Function not supported on Linux OS")
+    def test_deembedding_table_with_s2p_file(self, test_rfsg_device_session):
+        test_rfsg_device_session.create_deembedding_sparameter_table_s2p_file('', 'myTable1', get_test_file_path('samples2pfile.s2p'), nirfsg.SparameterOrientation.PORT2)
+        test_rfsg_device_session.create_deembedding_sparameter_table_s2p_file('', 'myTable2', get_test_file_path('samples2pfile.s2p'), nirfsg.SparameterOrientation.PORT1)
+        test_rfsg_device_session.configure_deembedding_table_interpolation_linear('', 'myTable1', nirfsg.Format.MAGNITUDE_AND_PHASE)
+        test_rfsg_device_session.deembedding_port[''].deembedding_selected_table = 'myTable1'
```
ni-jfitzger: Is deembedding_port a repeated capability or just an unnecessary parameter?
I don't see a single example of you indexing anything other than empty string.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+        with test_rfsg_device_session.initiate():
+            test_rfsg_device_session.check_generation_status()
+        test_rfsg_device_session.delete_deembedding_table('', 'myTable1')
```
ni-jfitzger: If deembeding_port really is a repeated capability, then it should be treated as one for delete_deembedding_table, too.

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+            assert 'The specified de-embedding table cannot be found' in e.description
+        test_rfsg_device_session.deembedding_port[''].deembedding_selected_table = ''
+        with test_rfsg_device_session.initiate():
+            test_rfsg_device_session.check_generation_status()
+
+    def test_read_and_download_waveform_from_file_tdms(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_generation_mode(nirfsg.GenerationMode.ARB_WAVEFORM)
+        test_rfsg_device_session.read_and_download_waveform_from_file_tdms('mywaveform', get_test_file_path('ValidWaveformTDMSFile.tdms'), 0)
+        waveformexists = test_rfsg_device_session.check_if_waveform_exists('mywaveform')
+        assert waveformexists is True
```
ni-jfitzger:
```suggestion
        waveform_exists = test_rfsg_device_session.check_if_waveform_exists('mywaveform')
        assert waveform_exists is True
```

src/nirfsg/system_tests/test_system_nirfsg.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def test_wait_until_settled(self, test_rfsg_device_session):
+        test_rfsg_device_session.configure_rf(2e9, -5.0)
+        with test_rfsg_device_session.initiate():
+            test_rfsg_device_session.wait_until_settled(15000)
```
ni-jfitzger: Will this wait 15 seconds for a simulated device, too?
If so, that's way too long for this test.
Remember, we run the tests on all supported Pythons, from the earliest (3.9) to the latest (3.13). That's 5 versions. So we have to run this test 5 times.

And we don't have a dedicated runner for every API. They're shared.

src/nirfsg/metadata/functions.py
vnktshr21: > I thought RF didn't need nesting for any functions or attributes.

That was when we had not changed the locations functions to be repeated capable during the course of this PR.

> Can you give an example?

Example channel string: "waveform::waveform1/marker0" 
There is a system test for this and the nesting is working as expected btw.

> both single rep_cap and multiple will have same runtime behavior.
> i'm not sure what you mean.

Once the channel parameter is constructed(in case of one repcap value or concatenation of multiple repcap values), then the function call happens in the same way by passing whatever channel string was constructed.

> Documentation can be fixed after release. We'll need to snoop around and see what nidigital does.

Yes, this might need a change in metadata field type itself. nidigital also seems to be specifying only one repcap in functions. I do not know if the nesting applies for functions there like it does in RFSG.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: 1809 for year and -1 for month and other fields. And hightime.datetime's constructor throws an error if any of the input is -1.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Thought process was to just ensure entrypoint call happens without any errors. Anyways changed the logic to check the driver behavior as well like you suggested.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for removal to make sure we talk about it. Will not be doing any removal as part of current PR though.

generated/nirfsg/nirfsg/session.py
vnktshr21: Fixed.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Fixed. Turns out, the option string parameter was not having a template in metadata like other MI drivers do. Without this, not passing options string was throwing error. I've fixed it now. Metadata now matches that of other MI drivers for InitWithOptions.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Fixed.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Fixed.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Fixed.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for removal to make sure we talk about it. Will not be doing any removal as part of current PR though.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for removal to make sure we talk about it. Will not be doing any removal as part of current PR though.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for making repeated capable. Will not be updating them as part of current PR though.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for making repeated capable. Will not be updating them as part of current PR though.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for making repeated capable. Will not be updating them as part of current PR though.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for removal to make sure we talk about it. Will not be doing any removal as part of current PR though.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: They are different. There are more things that we are saving and loading than just attributes in RFSG like deembedding s2p file path and waveform tdms file path.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: I do not understand the basis on which you think it could be "just an unnecessary parameter". It is being explicitly added in the config.py as a repeated capability.

The system tests for now are running for just a simulated 5841 device. There are valid non-empty values for this repeated capability for other devices. From the perspective of testing the entrypoint, I think the current test case is good enough.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Will include it in the list of API to discuss for making repeated capable. Will not be updating them as part of current PR though.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Fixed.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: 15 seconds is the timeout value while waiting for the generated signal to settle on the frequency and power level. Not the duration for which it waits.

src/nirfsg/system_tests/test_system_nirfsg.py
vnktshr21: Fixed.

generated/nirfsg/nirfsg/session.py
vnktshr21: Created an issue for this: https://github.com/ni/nimi-python/issues/2119

src/nirfsg/system_tests/test_system_nirfsg.py
ni-jfitzger: > I do not understand the basis on which you think it could be "just an unnecessary parameter".

Mostly on the basis of me not knowing your API.

src/nirfsg/system_tests/test_system_nirfsg.py
ni-jfitzger: It's not really about checking the driver behavior.
It's more about exercising the API the way you actually expect a user to.

There are 2 reasons for this:
1. It makes us feel better.
2. Due to our limited number of examples, users might sometimes look at test code as an example of how to use a function.

Sometimes there's a 3rd reason: 'Some functions may force us to set certain attributes or call certain other functions, before using them, even with a simulated device.".

Of course, this may be more of a personal preference than a nimi-python guideline.

src/nirfsg/metadata/functions.py
ni-jfitzger: > Once the channel parameter is constructed(in case of one repcap value or concatenation of multiple repcap values), then the function call happens in the same way by passing whatever channel string was constructed.

Yes, this is true. 

> nidigital also seems to be specifying only one repcap in functions. I do not know if the nesting applies for functions there like it does in RFSG.

Yeah, we don't have official support for listing multiple rep caps in our metadata, yet. We actually refer to it for nidigital as "chained" repeated capabilities. We should try to use the same terminology when discussing to avoid confusion.

> The codegen and metadata lacks the capability of calling out multiple repeated capabilities for function parameters like it does for attributes.

I couldn't tell you why that is. It works for nidigital, as far as I know, but nidigital has a few spots where it does special things and the code for repeated capabilities is so complex that I have to spend an hour making sense of it, every time I look at it.

ni-jfitzger (approved):

PR #2120, opened 2025-07-15
Including Common changes for Multidimensional array

rahulr-NI:

- Common changes required for supporting the multi-dimensional numpy arrays of numpy complex type as input and output parameter. 
  - Unit tests have been added for checking if data gets passed to library correctly with no memory copy.
- Similar no memory copy check has been added for 1d numpy array case as well.

All Unit tests pass.


build/templates/_library_interpreter.py.mako
```diff
@@ -53,8 +53,12 @@ def _get_ctypes_pointer_for_buffer(value=None, library_type=None, size=None):
         % if are_complex_parameters_used:
         if library_type in (_complextype.NIComplexI16, _complextype.NIComplexNumberF32, _complextype.NIComplexNumber):
             complex_dtype = numpy.dtype(library_type)
-            structured_array = value.view(complex_dtype)
-            return structured_array.ctypes.data_as(ctypes.POINTER(library_type))
+            if value.ndim > 1:
+                # we create a flattened view of the multi-dimensional numpy array
+                restructured_array_view = value.ravel().view(complex_dtype)
+            else:
+                restructured_array_view = value.view(complex_dtype)
```
ni-jfitzger, line 60: >            else:
>                restructured_array_view = value.view(complex_dtype)

Are we actually restructuring the array here or is this variable rename just so we can use the same return statement for either case?

src/nifake/metadata/functions.py
```diff
@@ -2990,6 +2990,49 @@
         ],
         'returns': 'ViStatus'
     },
+    'FunctionWith3dNumpyArrayOfNumpyComplex128InputParameter': {
+        'codegen_method': 'public',
```
ni-jfitzger, line 2994: Can you set python_name for this function, so that "with" and "3d" don't get joined together in the name?

build/templates/session.py/numpy_write_method.py.mako
```diff
@@ -30,6 +30,10 @@
             raise TypeError('${parameter['python_name']} must be in C-order')
         if ${parameter['python_name']}.dtype is not numpy.dtype('${parameter['numpy_type']}'):
             raise TypeError('${parameter['python_name']} must be numpy.ndarray of dtype=${parameter['numpy_type']}, is ' + str(${parameter['python_name']}.dtype))
+%   if parameter['array_dimension'] > 1:
+        if ${parameter['python_name']}.ndim != ${parameter['array_dimension']}:
+            raise TypeError('${parameter['python_name']} must be numpy.ndarray of dimension=${parameter['array_dimension']}, is ' + str(${parameter['python_name']}.ndim))
+%   endif
```
ni-jfitzger: Are you only going to do this check if you expect a multi-dimensional numpy array? What if someone tries to pass a multi-dimension numpy array when a 1D numpy array is expected. What will happen then?

build/unit_tests/test_codegen_helper.py
```diff
@@ -42,6 +42,7 @@
         'type': 'ViSession',
         'numpy': False,
         'complex_type': None,
+        'array_dimension': 1,
```
ni-jfitzger: Adding this to these dicts does absolutely nothing for our testing, because these tests don't call `add_all_function_metadata()`, even indirectly.

We're also not using a snippet for the lines you're adding to numpy_write_method.py.mako.
Is there some other usage that I'm missing?

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -899,6 +899,50 @@ def test_write_waveform_numpy_complex_interleaved_i16_valid_input(self):
...
+    def test_write_3d_numpy_array_of_numpy_complex128(self):
+        from nifake._complextype import NIComplexNumber
+
+        array_3d = numpy.full((2, 3, 4), 1.0 + 2.0j, dtype=numpy.complex128)
+        number_of_samples = array_3d.size
+        flattened_array = array_3d.flatten()
+        complex_array = (NIComplexNumber * len(flattened_array))()
+        for i, value in enumerate(flattened_array):
+            complex_array[i] = NIComplexNumber(value.real, value.imag)
+        array_3d_ptr = ctypes.cast(complex_array, ctypes.POINTER(NIComplexNumber))
```
ni-jfitzger: Wouldn't flattened_array_ptr be a more accurate name than array_3d_ptr?

src/nifake/unit_tests/test_library_interpreter.py
```diff
@@ -899,6 +899,50 @@ def test_write_waveform_numpy_complex_interleaved_i16_valid_input(self):
...
+
+    def test_no_memorycopy_with_multi_dimensional_numpy_complex128_array(self):
+        array_3d = numpy.full((2, 3, 4), 1.0 + 2.0j, dtype=numpy.complex128)
+        self.patched_library.niFake_FunctionWith3dNumpyArrayOfNumpyComplex128InputParameter.side_effect = self.side_effects_helper.niFake_FunctionWith3dNumpyArrayOfNumpyComplex128InputParameter
+        interpreter = self.get_initialized_library_interpreter()
+        interpreter.function_with3d_numpy_array_of_numpy_complex128_input_parameter(array_3d)
+        args, kwargs = self.patched_library.niFake_FunctionWith3dNumpyArrayOfNumpyComplex128InputParameter.call_args
+        actual_pointer = args[1]
+        input_address = array_3d.__array_interface__['data'][0]
+        address_passed_to_library = ctypes.addressof(actual_pointer.contents)
```
ni-jfitzger: This is good, but I want to note that it doesn't guarantee no copy will take place in real usage. `ravel` documentation says it will copy, if needed.

build/unit_tests/test_codegen_helper.py
rahulr-NI: No Jay.  I have added these lines while trying to check how this parameter works during initial trying. Later I did not have a check removing these lines. Now I have removed the default value of array_dimension in test_codegen_helper.py file

src/nifake/unit_tests/test_library_interpreter.py
rahulr-NI: Yes.. Corrected name as flattened_array_ptr

build/templates/session.py/numpy_write_method.py.mako
rahulr-NI: The condition for multidimensional numpy is array is removed. If its a NumPy array, we check the dimension. Default value is 1. If anything is set specifically, it will check for that value. This check will also be included in nifgen as we are changing the mako file.

src/nifake/metadata/functions.py
rahulr-NI: Updated

build/templates/_library_interpreter.py.mako
rahulr-NI: This is using the same variable so that a single return statement is sufficient for either case

src/nifake/unit_tests/test_library_interpreter.py
vnktshr21: You are right. Documentation hints that ravel might copy if the incoming array's memory layout isn't favorable for creating the necessary flattened view. But, in our numpy_write_method expansion template, we already check and ensure that the incoming array has C-style ordering. With this, ravel should ideally not do a copy. The ValueError we raise, should eliminate the possibility of ravel making a copy. This is what I inferred from the documentation.

Side note, seems like `numpy.isfortran` , which we use to ensure C style ordering, might not be the recommended way of checking. Recommended option seems to be numpy.flags.c_contiguous. Was this not used for any particular reason?

build/templates/_library_interpreter.py.mako
ni-jfitzger: The name feels like a lie in one of these cases.
I don't love it.

src/nifake/unit_tests/test_library_interpreter.py
ni-jfitzger: >Side note, seems like numpy.isfortran , which we use to ensure C style ordering, might not be the recommended way of checking. Recommended option seems to be numpy.flags.c_contiguous. Was this not used for any particular reason?

No clue, but @marcoskirsch added it, so he might know, if he can scour it from his memory. 8 years is a long time.

ni-jfitzger (approved):

src/nifake/unit_tests/test_library_interpreter.py
vnktshr21: Created a issue for this: https://github.com/ni/nimi-python/issues/2122

PR #2121, opened 2025-07-15
Adding and removing few methods and few more fixes in RFSG

vnktshr21:

- Enables intflag parameter for self_calibrate_range and adds reset_with_options method
  - Overriding doc_tip for stepsToOmit parameter to call out Bitwise input (this should be done in codegen itself in the future)
- Expands shortened terms in enum names to full terms
- Adds lo_channels repeated capability and changed deembedding_port into ports repcap since ports is used by API other then deembedding as well
  - Marks relevant attributes with the same
- Makes waveform repcap name plural to align with repcap naming policy
- Fixes few enum values and documentation snippets
- Removes few enums that are not used by any attribute or function
- Ensures all method description start with \n consistently
- Removes p2p related methods, enums, enum values and attributes
- Removes leftover configuration list related enum values
- Removes few missed configuration list related enums (list mode API will not be supported for 1st release)
- Makes script trigger configuration methods to have triggerID as repeated capability
- Removes methods which are configuring just one attribute
  - Removes references to these in documentation

None

System tests for newly added API and updated enum names all pass on simulated and real hardware.

### DIFF
--- a/docs/nirfsg/rep_caps.rst
+++ b/docs/nirfsg/rep_caps.rst
@@ -75,17 +75,17 @@ script_triggers
         passes a string of :python:`'scripttrigger0, scripttrigger1, scripttrigger2'` to the set attribute function.
 
 
-waveform
---------
+waveforms
+---------
 
-    .. py:attribute:: nirfsg.Session.waveform[]
+    .. py:attribute:: nirfsg.Session.waveforms[]
 
         If no prefix is added to the items in the parameter, the correct prefix will be added when
         the driver function call is made.
 
         .. code:: python
 
-            session.waveform['0-2'].channel_enabled = True
+            session.waveforms['0-2'].channel_enabled = True
 
         passes a string of :python:`'waveform::0, waveform::1, waveform::2'` to the set attribute function.
 
@@ -96,34 +96,34 @@ waveform
 
         .. code:: python
 
-            session.waveform['waveform::0-waveform::2'].channel_enabled = True
+            session.waveforms['waveform::0-waveform::2'].channel_enabled = True
 
         passes a string of :python:`'waveform::0, waveform::1, waveform::2'` to the set attribute function.
 
 
-port
-----
+ports
+-----
 
-    .. py:attribute:: nirfsg.Session.port[]
+    .. py:attribute:: nirfsg.Session.ports[]
 
         .. code:: python
 
-            session.port['0-2'].channel_enabled = True
+            session.ports['0-2'].channel_enabled = True
 
         passes a string of :python:`'0, 1, 2'` to the set attribute function.
 
 
-lo
---
+lo_channels
+-----------
 
-    .. py:attribute:: nirfsg.Session.lo[]
+    .. py:attribute:: nirfsg.Session.lo_channels[]
 
         If no prefix is added to the items in the parameter, the correct prefix will be added when
         the driver function call is made.
 
         .. code:: python
 
-            session.lo['0-2'].channel_enabled = True
+            session.lo_channels['0-2'].channel_enabled = True
 
         passes a string of :python:`'LO0, LO1, LO2'` to the set attribute function.
 
@@ -134,7 +134,7 @@ lo
 
         .. code:: python
 
-            session.lo['LO0-LO2'].channel_enabled = True
+            session.lo_channels['LO0-LO2'].channel_enabled = True
 
         passes a string of :python:`'LO0, LO1, LO2'` to the set attribute function.
 
--- a/docs/nirfsg/rep_caps.rst
+++ b/docs/nirfsg/rep_caps.rst
@@ -101,16 +101,42 @@ waveform
         passes a string of :python:`'waveform::0, waveform::1, waveform::2'` to the set attribute function.
 
 
-deembedding_port
-----------------
+port
+----
 
-    .. py:attribute:: nirfsg.Session.deembedding_port[]
+    .. py:attribute:: nirfsg.Session.port[]
 
         .. code:: python
 
-            session.deembedding_port['0-2'].channel_enabled = True
+            session.port['0-2'].channel_enabled = True
 
         passes a string of :python:`'0, 1, 2'` to the set attribute function.
+
+
+lo
+--
+
+    .. py:attribute:: nirfsg.Session.lo[]
+
+        If no prefix is added to the items in the parameter, the correct prefix will be added when
+        the driver function call is made.
+
+        .. code:: python
+
+            session.lo['0-2'].channel_enabled = True
+
+        passes a string of :python:`'LO0, LO1, LO2'` to the set attribute function.
+
+        If an invalid repeated capability is passed to the driver, the driver will return an error.
+
+        You can also explicitly use the prefix as part of the parameter, but it must be the correct prefix
+        for the specific repeated capability.
+
+        .. code:: python
+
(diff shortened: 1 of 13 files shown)


generated/nirfsg/nirfsg/enums.py
```diff
@@ -141,37 +142,6 @@ class AutomaticThermalCorrection(Enum):
...
-    '''
-    MANUAL = 0
-    CONFIGURATION_LIST_REPEAT_SINGLE = 1
-    r'''
-    NI-RFSG runs the configuration list only once.
-    '''
-    SCRIPT_TRIGGER = 1
-
-
 class DeembeddingTypeAttrVals(Enum):
```
ni-jfitzger: >class DeembeddingTypeAttrVals(Enum):

Won't necessarily hold up this PR for it, but shouldn't this be just `DeembeddingType`?

generated/nirfsg/nirfsg/enums.py
```diff
@@ -836,7 +800,7 @@ class StartTrigType(Enum):
     '''
     P2P_ENDPOINT_FULLNESS = 3
     r'''
-    The data operation does not start until the endpoint reaches the threshold specified in the p2p_endpoint_fullness_start_trigger_level property.
+    The data operation does not start until the endpoint reaches the threshold specified in the P2P_ENDPOINT_FULLNESS_START_TRIGGER_LEVEL property.
     '''
```
ni-jfitzger: You should probably exclude this value from the enum. You can disable specific enum values. Likewise, you should update the documentation table for the start_trigger_type property to not mention it.

docs/nirfsg/rep_caps.rst
```diff
@@ -101,16 +101,42 @@ waveform
         passes a string of :python:`'waveform::0, waveform::1, waveform::2'` to the set attribute function.
 
 
-deembedding_port
-----------------
+port
```
ni-jfitzger: I believe we always name our repeated capabilities plural, so 'ports', not 'port. 'waveforms', not 'waveform'.

docs/nirfsg/rep_caps.rst
```diff
@@ -101,16 +101,42 @@ waveform
...
 
         .. code:: python
 
-            session.deembedding_port['0-2'].channel_enabled = True
+            session.port['0-2'].channel_enabled = True
 
         passes a string of :python:`'0, 1, 2'` to the set attribute function.
 
 
+lo
```
ni-jfitzger: >lo

Is lo the right name for this rep cap?
We usually (always?) make our rep cap names plural and I'm not sure how I feel about abbreviating them, even if the abbreviation is very common in the API.

los?
local_oscillators?

Thoughts, @marcoskirsch?

generated/nirfsg/nirfsg/session.py
```diff
@@ -8336,43 +8200,13 @@ def write_arb_waveform(self, waveform_name, waveform_data_array, more_data_pendi
...
-        self._interpreter.write_p2p_endpoint_i16(stream_endpoint, number_of_samples, endpoint_data)
-
     @ivi_synchronized
     def write_script(self, script):
         r'''write_script
 
         Writes a script to the device to control waveform generation in Script mode.
 
-        First, configure your device for Script mode by calling the configure_generation_mode method. The NI-RFSG device must be in the Configuration state before calling the write_script method.
+        First, configure your device for Script mode by calling the ConfigureGenerationMode method. The NI-RFSG device must be in the Configuration state before calling the write_script method.
```
ni-jfitzger: ConfigureGenerationMode method no longer exists.

generated/nirfsg/nirfsg/enums.py
vnktshr21: This is a good catch! I missed this while I expanded other stuff. I've changed it to DeembeddingType.

generated/nirfsg/nirfsg/enums.py
vnktshr21: I was thinking about this. Mostly had left them since we need to ensure they don't get missed to be added, should there be a P2P python API in the future. I've now removed those anyways. We would likely not miss when adding since they carry 'P2P' in their names.

docs/nirfsg/rep_caps.rst
vnktshr21: Fixed.

docs/nirfsg/rep_caps.rst
vnktshr21: I believe its plural since most of them allow for configuring multiple values as context in one go. I renamed this one as "lo_channels". This aligns with how we are referring the repeated capability in attribute documentation and also makes it plural aligning with other repcap names.

generated/nirfsg/nirfsg/session.py
vnktshr21: Fixed. Rephrased this to instead call out the property.

ni-jfitzger:
test_waveform_rep_cap --> test_waveforms_rep_cap
test_port_rep_cap --> test_ports_rep_cap
test_lo_rep_cap --> test_lo_channels_rep_cap

ni-jfitzger:
Looking at
```
    'Lo1OutputFilter': {
        'codegen_method': 'public',
        'values': [
            {
                'documentation': {
                    'description': 'yet to be defined'
                },
                'name': 'NIRFSG_VAL_5.5GHz_HIGH_PASS',
                'value': 0
            },
            {
                'documentation': {
                    'description': 'yet to be defined'
                },
                'name': 'NIRFSG_VAL_5.5GHz_LOW_PASS',
                'value': 1
            }
        ]
    },
```

Something about our codegen that you should be aware of: by default we trim common prefixes in the enum value names for a given enum. This is how we get rid of NIRFSG_VAL in the above enum values. But this also means that "5.5GHz_" is not part of the enum value name for either. This may or may not be what's desired.

You can override this behavior by specifying the `python_name`. Or at least, that would be true, if https://github.com/ni/nimi-python/pull/2073 were in.

You may want to go through your generated enums.py to look for any enum values that aren't what you want.

src/nirfsg/metadata/enums.py
```diff
@@ -126,23 +66,23 @@
                     'description': 'Specifies a range from 100Hz to 1kHz.'
                 },
                 'name': 'NIRFSG_VAL_100HZ_TO_1KHZ',
-                'python_name': '_100hzto1khz',
+                'python_name': '_100Hzto1kHz',
```
ni-jfitzger: Why are you using a leading underscore?
Also, I don't think we should use mixed case in Enum Value Names. Upper snake case is our convention. We already have some value names in other APIs that specify meters and we capitalize the m.

ni-jfitzger:
> For enums, it turns out the enums which had numbers in names are not used by any attribute or function.

`_add_enum_codegen_method` in the build helpers will normally automatically exclude any unused enums from the generated code. This didn't happen because you set codgen_method for those enums, yourself.

I'd advise against setting codegen_method: "public" for enums, as you may be generating something that's unused.
I'm sure you review these things, but it's really better to lean on codegen than assume that your review was perfect, especially with the API being so fluid, right now.

src/nirfsg/metadata/enums.py
vnktshr21: Now adding a suitable prefix before the numbers and made the case consistent with other enum names case. Had to add few entries in enum_whitelist_suffix and enum_whitelist_prefix config fields to get this working(precedence exists for this).

PR #2124, opened 2025-08-01
NI-RFSG metadata from codegenerated files with some cleanup and dev status override logic

vnktshr21:

- Updated the get_development_status codegen helper logic to allow a override of development status via config_addon(this will be used for setting development status for RFSG release, which does not align with existing version based logic)
- Adding rfsg metadata files out of internal codegeneration tool
  - Corrected couple of enum names
  - Documentation corrections
  - functions.py and attributes.py will contain re-arranged entries, but the generated API being unchanged should validate that its just re-arranged
  - Few default value fields got removed
  - Repcap related fields for functions got added by the codegen tool
- Cleaned up get_deembedding_sparameters method to correctly have only one return value (matches .NET API)
  - Needed to add fancy function to get this working correctly
- Corrected session locking for write_arb_waveform methods 

- System tests pass with simulated and real hardware
- API codegen shows only expected updates as diff
- Verified the development_status override logic to be working fine, if we provide a override value

### DIFF
--- a/generated/nirfsg/nirfsg/session.py
+++ b/generated/nirfsg/nirfsg/session.py
@@ -7244,6 +7244,31 @@ class Session(_SessionBase):
         else:
             raise TypeError("Unsupported datatype. Expected numpy array.")
 
+    def get_deembedding_sparameters(self):
+        '''get_deembedding_sparameters
+
+        Returns the S-parameters used for de-embedding a measurement on the selected port.
+
+        This includes interpolation of the parameters based on the configured carrier frequency. This method returns an empty array if no de-embedding is done.
+
+        If you want to call this method just to get the required buffer size, you can pass 0 for **S-parameter Size** and VI_NULL for the **S-parameters** buffer.
+
+        **Supported Devices** : PXIe-5830/5831/5832/5840/5841/5842/5860
+
+        Note: The port orientation for the returned S-parameters is normalized to SparameterOrientation.PORT1_TOWARDS_DUT.
+
+        Returns:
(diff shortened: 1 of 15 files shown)


ni-jfitzger (commented):
It's not hurting anything, but the final return statement in `write_arb_waveform` is unreachable.

generated/nirfsg/nirfsg/session.py
```diff
@@ -7232,6 +7244,31 @@ def create_deembedding_sparameter_table_array(self, port, table_name, frequencie
...
+
+        Returns:
+            sparameters (numpy.array(dtype=numpy.complex128)): Returns an array of S-parameters. The S-parameters are returned in the following order: s11, s12, s21, s22.
+
+        '''
+        import numpy as np
+        number_of_ports = self._get_deembedding_table_number_of_ports()
+        sparameter_array_size = number_of_ports ** 2
+        sparameters = np.full((number_of_ports, number_of_ports), 0 + 0j, dtype=np.complex128)
+        number_of_sparameters, number_of_ports = self._get_deembedding_sparameters(sparameters, sparameter_array_size)
```
ni-jfitzger: The typical thing to do with an unused return value is to assign to '_' to indicate it's a value you don't care about and aren't going to use.

ni-jfitzger:
I didn't look too closely at the function metadata changes or the documentation changes, but I reviewed the rest and it looks okay to me.

ni-jfitzger (approved):

generated/nirfsg/nirfsg/session.py
vnktshr21: Fixed it.

PR #2130, opened 2025-09-11
RFSG examples default value fix and clean up

vnktshr21:

- Examples updated
  - Fix default value for option string in examples to be empty
  - Make resource name default as PXI1Slot2 to be consistent with other driver examples
  - Changed _options_ in test_example function to use python style objects to specify options instead of string
    - This will demonstrate the objects way of specifying Initialize options (change is internal though) 
  - Renamed the simulated device name used in test functions

None

- Examples do not default to simulation option string if not specified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1691,7 +1691,6 @@
 #### [nirfsg] Unreleased
 - Added
 - Changed
-  - Fix default value for option string in examples to be empty and make resource name as required input
   - Fixed the return type of `get_all_script_names` and `get_all_named_waveform_names` to remove the size parameter and return a list of strings instead of a comma-separated string
 - Removed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1691,6 +1691,7 @@
 #### [nirfsg] Unreleased
 - Added
 - Changed
+  - Fix default value for option string in examples to be empty and make resource name as required input
```
marcoskirsch: @ni-jfitzger examples don't ship with the modules... do we track their changes in CHANGELOG.md?

marcoskirsch:
>The option string default now aligns with other driver examples. However I do not see other drivers making resource name as required input. I made it required here mainly to catch it if its missing, at the example level itself instead of letting driver initialize error out.

I think you should stay consistent. If you think making the resource name required, then make the change consistently in another PR. But I think many drivers allow leaving resource name blank in cases of simulation so that's probably why it's not required.

CHANGELOG.md
ni-jfitzger: Generally, no, which is what I would lean toward.
We do have a history of noting the addition of examples, but you have to go all the way back to 2021 to see mentions of changes to examples and there were changes to examples after that which went unmentioned.

For example: https://github.com/ni/nimi-python/pull/1817

vnktshr21: Removed the changelog mention.

ni-jfitzger (approved):

PR #2134, opened 2025-09-24
Update the order and default values in nifake metadata 

rahulr-NI:

This PR updates the nimi-python repository metadata files with those from NI's private repository, resolves ordering issues, and addresses default values.
It contains new APIs generated during codegen but not present in nimi-python. 

Tox build is failing here as in gRPC, the abort api in generated codes is not implemented but the existing test cases are expecting it.


src/nifake/metadata/functions.py
```diff
@@ -1,12 +1,12 @@
...
-# This file is generated from NI-FAKE API metadata version 25.8.0d9999
+# This file is generated from NI-FAKE API metadata version 25.8.0d100
 functions = {
     'Abort': {
         'codegen_method': 'public',
         'documentation': {
             'description': 'Aborts a previously initiated thingie.'
         },
-        'included_in_proto': True,
+        'included_in_proto': False,
```
marcoskirsch, line 9: Abort is such a common function that it's strange not to include in the example/fake/test metadata

marcoskirsch (changes requested):

src/nifake/metadata/functions.py
rahulr-NI: Yes.. Thats why I did not modify any portion of metadata that being generated from AzDo. We will need to create some stories to issues to make the necessary changes to Match AzDo version of metadata generated from AzDo to version present in GitHub. Which also should include the necessary test cases.

Issue #2137, opened 2025-11-07
Fix size parameter handling for functions with parameters referencing multiple size parameters


batzuu:
If a function has multiple parameters and those parameters reference multiple size parameters, the codegen fails as it expects there to be only one size parameter (common for all buffers). This is an issue for `CreateDeembeddingSparameterTableArray` RFSG api which has two buffers referencing two different size parameters.

PR #2138, opened 2025-11-07
Fixing size parameter handling in len-mechanism to handle function parameters with different size parameters

batzuu:

Fixes the codegen to properly handle functions with multiple array parameters that reference different size parameters:

- **Updates `metadata_filters.py`**: Refactors the `filter_parameters` function to collect all size parameter names referenced by len-sized parameters into a `len_size_parameter_names`, replacing the previous single `len_size_parameter` approach. Removed assert to check if all size parameters are same from `filter_len_parameters`. Following this updated var names in `default_method.py.mako` and `initialization_method.py.mako` to reflect multiple possible size params.

- **Updates to nifake to enable testing**: Details covered in testing section

Previously, it was assumed all len-mechanism arrays in a function shared the same size parameter. This change enables functions like `CreateDeembeddingSparameterTableArray` in NI-RFSG where `frequencies` uses `frequenciesSize` and `sparameterTable` uses `sparameterTableSize`.

### List issues fixed by this Pull Request
Issue 2137: [Fix size parameter handling for functions with parameters referencing multiple size parameters](https://github.com/ni/nimi-python/issues/2137).
Additional PR needs to be raised to update NI-RFSG function metadata (CreateDeembeddingSparameterTableArray) to consume
these changes.

- Added `MultipleArraysDifferentSize` function to nifake metadata with two arrays (`valuesArray` using `valuesArraySize`, and `dataArray` using `dataArraySize`) to test the change in filter param logic
- Added unit test `test_multiple_arrays_different_size` in `test_library_interpreter.py` that verifies correct handling of multiple arrays with independent size parameters
- Test validates that each array's length is properly calculated and passed with its corresponding size parameter (3 elements for `valuesArray` with `valuesArraySize`, and 5 elements for `dataArray` with `dataArraySize`)
- All existing tests continue to pass.

### DIFF
--- a/build/unit_tests/test_metadata_find.py
+++ b/build/unit_tests/test_metadata_find.py
@@ -1,7 +1,7 @@
 from build.helper.metadata_find import find_len_size_parameter_names
 
 
-def _parameter(name, mechanism='fixed', size_value=None):
+def _parameter(name, mechanism='passed-in', size_value=None):
     size = {'mechanism': mechanism}
     if size_value is not None:
         size['value'] = size_value
@@ -14,11 +14,11 @@ def _parameter(name, mechanism='fixed', size_value=None):
 
 def test_find_len_size_parameter_names_multiple_sizes():
     parameters = [
-        _parameter('len_a_size', mechanism='passed-in'),
-        _parameter('len_b_size', mechanism='passed-in'),
+        _parameter('len_a_size'),
+        _parameter('len_b_size'),
         _parameter('len_a_data', mechanism='len', size_value='len_a_size'),
         _parameter('len_b_data', mechanism='len', size_value='len_b_size'),
-        _parameter('timeout', mechanism='fixed'),
+        _parameter('timeout'),
     ]
 
     size_names = find_len_size_parameter_names(parameters)
@@ -28,8 +28,8 @@ def test_find_len_size_parameter_names_multiple_sizes():
 
 def test_find_len_size_parameter_names_empty_when_no_len_parameters():
     parameters = [
-        _parameter('value', mechanism='passed-in'),
-        _parameter('timeout', mechanism='fixed'),
+        _parameter('value'),
+        _parameter('timeout'),
     ]
 
     size_names = find_len_size_parameter_names(parameters)


ni-jfitzger:
>generated/nidcpower/nidcpower/_library_singleton.py:43:5: F824 `global _instance_lock` is unused: name is never assigned in scope
    global _instance_lock

It seems we've gone from using flake8 7.1.2 to using flake8 7.3.0 and the new one has this new error code.
It doesn't like that we're using global here but never assigning a value to _instance_lock in that scope.
Though, according to the documentation, you need to declare with global before you can even use the variable.

https://docs.python.org/3/reference/simple_stmts.html#the-global-statement

This seems like something we should address in a separate PR.

bkeryan:
> Though, according to the documentation, you need to declare with global before you can even use the variable.
> 
> https://docs.python.org/3/reference/simple_stmts.html#the-global-statement
> 
> This seems like something we should address in a separate PR.

You only need the `global` keyword to assign global variables from an inner scope, not to read them.

https://stackoverflow.com/questions/291978/short-description-of-the-scoping-rules
https://docs.python.org/3/reference/executionmodel.html#resolution-of-names

build/helper/metadata_filters.py
```diff
@@ -348,23 +348,31 @@ def filter_parameters(parameters, parameter_usage_options):
...
 
     # Filter based on options
-    size_parameter = None
+    ivi_dance_size_parameter = None
+    len_size_parameter_names = set()
     size_twist_parameter = None
     # If we are being called looking for the ivi-dance, len or code param, we do not care about the size param so we do
     #  not call back into ourselves, to avoid infinite recursion
     if parameter_usage_options not in [ParameterUsageOptions.IVI_DANCE_PARAMETER, ParameterUsageOptions.LEN_PARAMETER]:
-        # Find the size parameter - we are assuming there can only be one type, either from ivi-dance or len
```
ni-jfitzger, line 356: >        # Find the size parameter - we are assuming there can only be one type, either from ivi-dance or len

You deleted this comment and the if statement guarding population of len size parameters.
But I don't see any testing for mixed usage.

build/helper/metadata_filters.py
```diff
@@ -348,23 +348,31 @@ def filter_parameters(parameters, parameter_usage_options):
...
-            size_parameter = find_size_parameter(filter_len_parameters(parameters), parameters)
+        # Determine any size parameters that should be skipped based on the presence of ivi-dance or len-sized buffers.
+        # For ivi-dance, there is a single shared size parameter; for len, there may be multiple independent size parameters.
+        ivi_dance_size_parameter = find_size_parameter(filter_ivi_dance_parameters(parameters), parameters)
+        # Collect all size parameter names referenced by len-sized parameters
+        len_params = filter_len_parameters(parameters)
+        for param in len_params:
+            len_size_param = find_size_parameter(param, parameters)
+            if len_size_param is not None:
+                len_size_parameter_names.add(len_size_param['name'])
```
ni-jfitzger: I think it would be better to push most of this logic into a `find_size_parameters` or `find_len_size_parameters` or `find_len_size_parameter_names` function in metadata_find.py.

build/helper/metadata_filters.py
```diff
@@ -443,18 +451,15 @@ def filter_ivi_dance_twist_parameters(parameters):
...
-    asserts all parameters that use len reference the same parameter
+    Note: Multiple len parameters may reference different size parameters.
     Args:
         parameters: parameters to be checked
 
     Return:
-        None if no len parameter found
-        Parameters dict if one is found
+        Empty list if no len parameter found
+        List of parameters if any are found
```
ni-jfitzger:
```suggestion
        Empty list if no len parameter found
        List of parameters dicts if any are found
```

build/templates/_library_interpreter.py/default_method.py.mako
```diff
@@ -11,8 +11,8 @@
     ivi_dance_parameters = helper.filter_ivi_dance_parameters(parameters)
     ivi_dance_size_parameter = helper.find_size_parameter(ivi_dance_parameters, parameters)
     len_parameters = helper.filter_len_parameters(parameters)
-    len_size_parameter = helper.find_size_parameter(len_parameters, parameters)
-    assert ivi_dance_size_parameter is None or len_size_parameter is None
+    len_size_parameters = helper.find_size_parameter(len_parameters, parameters)
+    assert ivi_dance_size_parameter is None or len_size_parameters is None
```
ni-jfitzger: So we're not supporting mixed usage, after all?

build/templates/_library_interpreter.py/default_method.py.mako
```diff
@@ -11,8 +11,8 @@
     ivi_dance_parameters = helper.filter_ivi_dance_parameters(parameters)
     ivi_dance_size_parameter = helper.find_size_parameter(ivi_dance_parameters, parameters)
     len_parameters = helper.filter_len_parameters(parameters)
-    len_size_parameter = helper.find_size_parameter(len_parameters, parameters)
-    assert ivi_dance_size_parameter is None or len_size_parameter is None
+    len_size_parameters = helper.find_size_parameter(len_parameters, parameters)
```
ni-jfitzger: This new variable name is a misleading. `helper.find_size_parameter` only returns a single parameter.

build/templates/_library_interpreter.py/initialization_method.py.mako
```diff
@@ -12,8 +12,8 @@
     ivi_dance_parameters = helper.filter_ivi_dance_parameters(parameters)
     ivi_dance_size_parameter = helper.find_size_parameter(ivi_dance_parameters, parameters)
     len_parameters = helper.filter_len_parameters(parameters)
-    len_size_parameter = helper.find_size_parameter(len_parameters, parameters)
-    assert ivi_dance_size_parameter is None or len_size_parameter is None
+    len_size_parameters = helper.find_size_parameter(len_parameters, parameters)
```
ni-jfitzger: This new variable name is a misleading. `helper.find_size_parameter` only returns a single parameter.

src/nifake/metadata/functions.py
```diff
@@ -3120,6 +3120,71 @@
...
+            },
+            {
+                'direction': 'in',
+                'documentation': {
+                    'description': 'Specifies the size of the data array.'
+                },
+                'name': 'dataArraySize',
+                'type': 'ViInt32',
+                'use_array': False,
+                'use_in_python_api': False
```
ni-jfitzger: `_add_use_in_python_api` should be able to set this to False for the size parameters of type 'len'. If you're having to set it yourself, I think that could be considered a bug.

rahur-NI: Removed and build tox. No issues found.  Thanks for highlighting it Jay

build/templates/_library_interpreter.py/initialization_method.py.mako
rahur-NI: I have updated the same pattern in both mako files

build/templates/_library_interpreter.py/default_method.py.mako
rahur-NI: We do support mixed usage.
That assert was removed, and mixed ivi-dance + len behavior is now covered by unit tests (including multiple len size parameters).

build/templates/_library_interpreter.py/default_method.py.mako
rahur-NI: Agreed.  I have updated the name as before

build/helper/metadata_filters.py
rahur-NI: Updated

build/helper/metadata_filters.py
rahur-NI: Yes.. I included find_len_size_parameter_names method in metadata_find.py which is used in this file.

build/helper/metadata_filters.py
rahur-NI: Thanks Jay.
I added two unit tests that verify ivi-dance and len used together, including a case with multiple len size parameters to cover the mixed usage cases.

build/unit_tests/test_metadata_filters.py
```diff
@@ -0,0 +1,52 @@
...
+        'enum': None,
+        'numpy': False,
+        'use_in_python_api': True,
+        'complex_array_representation': None,
+    }
+
+
+def test_filter_parameters_mixed_usage_ivi_dance_and_len():
+    parameters = [
+        _parameter('ivi_size', mechanism='passed-in'),
```
ni-jfitzger: From looking at nifake/functions.py, the mechanism isn't typically set for and ivi-dance size parameter or a length size parameter. In fact, not even size is set. So, what seems to end up happening is that `size` gets added for each by `metadata_add_all._add_buffer_info` with a "mechanism" of "fixed" and a "value" of 1. test_metadata_add_all shows further evidence of this.

I don't think it's important to set mechanism for 'ivi_size' or 'len_size' in this test, but if you're going to explicitly set them, you should probably set them correctly.

build/unit_tests/test_metadata_filters.py
```diff
@@ -0,0 +1,52 @@
...
+        _parameter('out_waveform', direction='out', mechanism='ivi-dance', size_value='ivi_size'),
+        _parameter('len_size', mechanism='passed-in'),
+        _parameter('len_data', mechanism='len', size_value='len_size'),
+        _parameter('timeout', mechanism='fixed'),
```
ni-jfitzger: If you're going to set `mechanism` to 'fixed' for a parameter, you set `size_value` to a number and `direction` to 'out'. If you want it to be a parameter that gets passed in, the mechanism should be 'passed-in'.

build/unit_tests/test_metadata_filters.py
ni-jfitzger: As to why it's fixed when calculate it in the library interpreter?
I think it's just an usused default value, but I haven't traced the usage, myself.

rahur-NI: Yes.  Mechanism was not needed for size parameters. I removed the mechanism field from the size parameters used by the ivi-dance and len cases. I reran tox, and both related tests are passing.

build/unit_tests/test_metadata_filters.py
rahur-NI: I have updated the default value of mechanism as 'passed-in' instead of 'fixed'

PR #2143, opened 2025-11-20
Removing RFSG methods and properties which are not applicable

vnktshr21:

- Removed the methods and properties which are present for IVI compliance reasons in RFSG driver, but are not applicable for Python.
- Removed the methods and properties which are applicable only for hardware which are not supported anymore.

None

No specific tests are run for the removed API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1694,7 +1694,55 @@
   - Fixed the return type of `get_all_script_names` and `get_all_named_waveform_names` to remove the size parameter and return a list of strings instead of a comma-separated string
 - Removed
   - Methods and properties applicable only to hardware which are not supported anymore
+    - Methods
+      - `configure_digital_modulation_user_defined_waveform`
+      - `configure_pxi_chassis_clk10`
+    - Properties
+      - `analog_modulation_fm_deviation`
+      - `analog_modulation_waveform_frequency`
+      - `arb_filter_raised_cosine_alpha`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1693,6 +1693,8 @@
 - Changed
   - Fixed the return type of `get_all_script_names` and `get_all_named_waveform_names` to remove the size parameter and return a list of strings instead of a comma-separated string
 - Removed
+  - Methods and properties applicable only to hardware which are not supported anymore
```
ni-jfitzger, line 1696: It's fine to mention why these methods and properties are being removed, as you're doing here. Please be specific, though, and list each individual method and property. Not only will it be more informative to users, it will be an excellent reference is someone opens an issue complaining that one of these methods or properties is missing from the API. A quick look at the changelog will tell us they were intentionally excluded.

ni-jfitzger (approved):
I don't see anything alarming in the removals and generally take no issue with removals from a beta API. We can always add things back later, if we need them. Just fix the changelog.

CHANGELOG.md
marcoskirsch: also helps people searching

vnktshr21: Yes it does help searching faster in the future. I chose to not list since the list was a bit bigger and bloated the changelog and thought this PR diff would anyways contain this information. But I do see that we have listed for other drivers as well over the course of the changes. I've updated the changelog.

PR #813, opened 2018-04-19
Use timedelta for cal interval

texasaggie97:
- [X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/nimi-python/blob/master/CONTRIBUTING.md).
- [X] I've updated [CHANGELOG.md](https://github.com/ni/nimi-python/blob/master/CHANGELOG.md) if applicable.
- [X] I've added/updated tests applicable for this pull request

### What does this Pull Request accomplish?
* Functions that return an interval (calibration) returns `datetime.timedelta` instead of a plain number with no units
### List issues fixed by this Pull Request below, if any.
* Fix #794 

### What testing has been done?
* Unit
* System

PR #2151, opened 2025-12-17
adding new attribute classes and converters for comma separated string and timedelta in months

vnktshr21:
  - Not updating since this is an internal change 

- Added new attribute classes for
  - Comma separated string, but exposed in python API as list of strings
  - Time interval in months, but exposed in python API as hightime.timedelta
- Added new nifake attributes for these classes and unit tests validating their behavior
- Added converters for
  - list of strings to comma separated string
  - timedelta to months
    - Using the same 365/12=30.4167 conversion that was used previously
    - This is to cater to our C API returning interval values in months and us standardizing on using hightime.timedelta type for intervals in python API
- Converters for the opposite side were already implemented
- Added tests for newly added converters

None

Build passes.

### DIFF
--- a/build/templates/_converters.py.mako
+++ b/build/templates/_converters.py.mako
@@ -266,6 +266,12 @@ def convert_month_to_timedelta(months):
     return hightime.timedelta(days=(30.4167 * months))
 
 
+# Scaling factor to apply on seconds to get months
+# would be 1/(60 seconds * 60 minutes * 24 hours * 30.4167 days)
+def convert_timedelta_to_months_int32(value):
+    return _convert_timedelta(value, _visatype.ViInt32, 1.0 / (60 * 60 * 24 * 30.4167))
+
+
 # This converter is not called from the normal codegen path for function. Instead it is
 # call from init and is a special case.
 def convert_init_with_options_dictionary(values):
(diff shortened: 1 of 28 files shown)


build/templates/_converters.py.mako
```diff
@@ -266,6 +266,12 @@ def convert_month_to_timedelta(months):
     return hightime.timedelta(days=(30.4167 * months))
 
 
+# Scaling factor to apply on seconds to get months
+# would be 1/(60 seconds * 60 minutes * 24 hours * 30.4167 days)
+def convert_timedelta_to_months_int32(value):
+    return _convert_timedelta(value, _visatype.ViInt32, 1.0 / (60 * 60 * 24 * 30.4167))
```
ni-jfitzger, line 272: I don't feel great about us coming up with some arbitrary definition for months.
There's a good reason that datetime.timedelta (and, therefore, hightime) doesn't support it. It's not a standard length of time.

Depending on how this is used, it could lead to confusion or issues filed.

Why do you want to add this?

ni-jfitzger: Nevermind. I just saw the existing `convert_month_to_timedelta`.
What's done is done, I suppose.

I'm really curious how we decided on this arbitrary value.

ni-jfitzger: I see that was added in https://github.com/ni/nimi-python/pull/813
No explanation for the number 30.4167 in that PR, but it appears to be 365 / 12.

vnktshr21: Yes, this was an existing conversion being used. Thanks for sharing the PR link. I did check the history before going ahead and using the same value, but had not found any explanation for this number either. And yes it is simply 365/12, avoiding handling specific month variations. I do see this to be a reasonable way to go about it for the use cases we have, given we have to cater to both our C API returning interval values in months and us standardizing on using hightime.timedelta type for intervals.

I will mention this in the PR description.

PR #2154, opened 2026-01-07
raise TypeError if input is not list of str for convert_list_to_comma_separated_string converter

vnktshr21:

convert_list_to_comma_separated_string converter now checks if the input is a list of str and raise a TypeError if its not, so that we do not accidentally take in any other type of input and do a join on that in case of writeable attributes using this type. 

convert_list_to_comma_separated_string was assuming a list input, which could be wrong.

Add a nifake unit test to confirm that TypeError is thrown.

### DIFF
--- a/build/templates/_converters.py.mako
+++ b/build/templates/_converters.py.mako
@@ -339,16 +339,16 @@ def convert_comma_separated_string_to_list(comma_separated_string):
 
 
 def convert_list_to_comma_separated_string(list_of_strings):
-    '''Convert a list of strings into a comma-separated string.
+    '''Convert a list or tuple of strings into a comma-separated string.
 
     Args:
-        list_of_strings (list[str]): List of strings.
+        list_of_strings (list or tuple of str): List or tuple of strings.
 
     Returns:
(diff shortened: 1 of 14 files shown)


build/templates/_converters.py.mako
```diff
@@ -347,6 +347,8 @@ def convert_list_to_comma_separated_string(list_of_strings):
     Returns:
         str: Comma-separated string.
     '''
+    if not isinstance(list_of_strings, list) or not all(isinstance(item, str) for item in list_of_strings):
```
ni-jfitzger: lists only?
Do we really want to be this strict?
What about other iterables? At the very least, shouldn't we also allow tuples?

ni-jfitzger: Looks like nirfsg is the only client using this, so no matter what we decide, I'm not too worried about breaking clients.

vnktshr21: I have changed the converter to allow tuple as well. I see that list or tuple are the common iterables. We can enable allowing other iterables if required in future since this would not be a breaking change later.

PR #2158, opened 2026-01-17
Update driver versions tested against

ni-jfitzger:

We've upgraded the driver versions that are installed to our testing VM, nimibot. This change updates our documentation to reflect the change in driver versions that we use for testing.

We've deployed the latest released driver runtime versions for everything but NI Switch Executive, which is held back by a blocking bug. We also deployed the latest gRPC Device Server, but we don't note the version of that in our documentation.

None

PR Checks


ni-jfitzger:
> Maybe we can specify the basis of the choice of driver versions in the PR description? In this case, we are upgrading to the last released versions at the time of this change, right?

For everything except NI Switch Executive, yes.

PR #2176, opened 2026-04-13
Remove duplicate session_pb2 codegen, add ni-apis submodule and dependency on ni.grpcdevice.v1.proto

### LINKED ISSUE #2146: session.proto should be replaced with a dependency on ni.grpcdevice.v1.proto
### Description of issue

The nimi-python packages contain duplicate protobuf codegen for `session.proto`. This only works when the duplicated codegen is exactly identical across all installed packages; if any package has a slightly different version of this codegen, users will get an error such as

> ERROR: Exception calling application: Couldn't build proto file into descriptor pool: duplicate symbol 'nidevice_grpc.SessionInitializationBehavior'

The path forward is to add a dependency on https://pypi.org/project/ni.grpcdevice.v1.proto/ to the `grpc` extra. This will single-source the codegen for `session.proto`. See https://github.com/ni/ni-apis-python/issues/75 for more details.

### System report

N/A

### Steps to reproduce issue

1. Rebuild one package to use the latest version of https://github.com/ni/ni-apis/blob/main/ni/grpcdevice/v1/session.proto
2. Install the rebuilt package along with one of the previously released packages.
3. Create gRPC sessions using both of the packages. You will get an error when loading the gRPC stubs.

rahur-NI:

Replaces the duplicate `session_pb2` protobuf codegen in each nimi-python driver modules with a dependency on `ni.grpcdevice.v1.proto` (similar to [ni/nidaqmx-python#859].

- Replaced relative `import from . import session_pb2` with absolute `import session_pb2` in `_grpc_stub_interpreter.py.mako`
- Install `ni.grpcdevice.v1.proto>=1.0.0` as part of installers for modules that support grpc.
- Updated mak files to remove generation of `session_pb2.py` and `session_pb2_grpc.py`
- Added third_party/ni-apis as a git submodule pointing to [https://github.com/ni/ni-apis]
- Deleted `session.proto` file

* Fix #2146 

- A successful tox.
- Codegen files have no changes other than importing from installed `ni.grpcdevice.v1.proto` package than importing from locally generated `session_pb2` file as before.


tox.ini
```diff
@@ -160,6 +161,7 @@ deps =
     flake8: pep8-naming
     docs: sphinx
     docs: sphinx-rtd-theme
+    docs: ni.grpcdevice.v1.proto >= 1.0.0
```
ni-jfitzger: Are you sure we need this dependency for generating documentation?

tools/install_local_wheel.py
```diff
@@ -51,8 +51,8 @@ def main():
         logging.error('No wheel found. Has the build run successfully?')
         sys.exit(2)
 
-    # Install/upgrade the wheel
-    call(['pip', 'install', '--upgrade', wheel])
+    # Install/upgrade the wheel with grpc extras
+    call(['pip', 'install', '--upgrade', f'{wheel}[grpc]'])
```
ni-jfitzger: Why are you changing this?
This script is for installing a locally-built wheel. That has nothing to do with an externally sourced dep like our grpc-extras. Besides, the system-tests already install grpc extras.

ni-jfitzger: Ah, I see now.
If we are testing nitclk, we (probably) need to install the grpc extras for niscope or whatever we're installing to test nitclk with.

But if we are testing niscope, is it okay to install the local nitclk wheel like this? It has no extras.

src/shared_protos/session.proto
```diff
@@ -7,88 +7,12 @@ option csharp_namespace = "NationalInstruments.Grpc.Device";
...
-
-  // Resets the server to a default state with no open sessions
-  rpc ResetServer(ResetServerRequest) returns (ResetServerResponse);
-}
-
 enum SessionInitializationBehavior {
   SESSION_INITIALIZATION_BEHAVIOR_UNSPECIFIED = 0;
   SESSION_INITIALIZATION_BEHAVIOR_INITIALIZE_NEW = 1;
   SESSION_INITIALIZATION_BEHAVIOR_ATTACH_TO_EXISTING = 2;
 }
```
ni-jfitzger: Weren't you supposed to delete the whole file?

tox.ini
rahur-NI: Yes.. I had removed this, Docs doesnt need this package.

src/shared_protos/session.proto
bkeryan: The new distribution mechanism for shared protos is:
- Delete the duplicate copy from source control
- Create a git submodule at `third_party/ni-apis` pointing to https://github.com/ni/ni-apis/tree/main
- Update your codegen scripts to use the new `--proto-path`
- Update your GitHub actions workflows to check out submodules
- Exclude third_party from lint tooling

See https://github.com/ni/nidaqmx-python/pull/859/changes for an example.

bkeryan: BTW, nidevice.proto is not in the ni-apis repo yet, so it can stay.

bkeryan: > BTW, nidevice.proto is not in the ni-apis repo yet, so it can stay.

Also, if you want to add nidevice.proto to ni-apis (and update ni-apis-python to package it), you are welcome to create a PR.

tools/install_local_wheel.py
rahur-NI: Yes.. I probably included before including in the tox.ini file and missed to remove it later.   Now i did codegen and tested , changes in install_local_wheel.py file is not required.

src/shared_protos/session.proto
rahur-NI: Thanks Brad for including the approach here. I have done the changes and tested in local.  I'm allowing travis and system tests to run through pipeline.  Will update pr description and clean up if needed post the pr build.

rahur-NI: > > BTW, nidevice.proto is not in the ni-apis repo yet, so it can stay.
> 
> Also, if you want to add nidevice.proto to ni-apis (and update ni-apis-python to package it), you are welcome to create a PR.

Ok Brad. Will look into this as well if the bandwidth permits.

rahur-NI: I have updated the description

ni-jfitzger:
~~Please revert these ni-api changes until there's actually something to pull in.~~

I've been told we actually do need these, so disregard this comment.

bkeryan:
> Please revert these ni-api changes until there's actually something to pull in.

protoc needs a copy of session.proto in order to handle `import "session.proto";`

ni-jfitzger:
> > Please revert these ni-api changes until there's actually something to pull in.
> 
> protoc needs a copy of session.proto in order to handle `import "session.proto";`

@bkeryan I don't understand. Doesn't `ni.grpcdevice.v1.proto` have the results of compiling the session.proto?

ni-jfitzger:
> > > Please revert these ni-api changes until there's actually something to pull in.
> > 
> > 
> > protoc needs a copy of session.proto in order to handle `import "session.proto";`
> 
> @bkeryan I don't understand. Doesn't `ni.grpcdevice.v1.proto` have the results of compiling the session.proto?

Oh, dang. We need it in order to compile our api-specific .proto files? That's annoying.

.github/workflows/github_actions_aws_windows_python64.yml
```diff
@@ -59,6 +59,8 @@ jobs:
       steps:
         - name: checkout repository
           uses: actions/checkout@v3
+          with:
+            submodules: true
```
ni-jfitzger: I don't think we need this.
This action is for running system tests. System tests do not involve `tox -e codegen`.

.github/workflows/check_latest_release.yml
```diff
@@ -28,6 +28,8 @@ jobs:
     steps:
       - name: checkout repository
         uses: actions/checkout@v3
+        with:
+          submodules: true
```
ni-jfitzger: Like the other workflows in this folder, this is just runs system tests and examples.
I don't think we'll need submodules for that because it shouldn't involve codegen.

build/defines.mak
```diff
@@ -9,7 +9,8 @@ DRIVER_DIR := $(ROOT_DIR)/src/$(DRIVER)
 METADATA_DIR := $(DRIVER_DIR)/metadata
 METADATA_FILES := $(wildcard $(METADATA_DIR)/*.py)
 SHARED_PROTOS_DIR := $(ROOT_DIR)/src/shared_protos
-PROTO_DIRS := $(METADATA_DIR) $(SHARED_PROTOS_DIR)
+NI_APIS_PROTOS_DIR := $(ROOT_DIR)/third_party/ni-apis/ni/grpcdevice/v1
```
ni-jfitzger, line 12: Is this the right place to get session.proto from or should it be this one?
https://github.com/ni/ni-apis-python/blob/main/proto/session.proto

The one in ni-apis is pretty bare.

rahur-NI: Hi Jay, Actually the version of session.proto file present in https://github.com/ni/ni-apis/blob/main/ni/grpcdevice/v1/session.proto is sufficient for our repo i believe. I have tried tox and grpc validation of system test for nirfsg and both were successful.

ni-jfitzger:
@rahur-NI Did you have to do anything special to fetch the submodule, locally?
Do we need to update CONTRIBUTING.md?

ni-jfitzger:
> @rahur-NI Did you have to do anything special to fetch the submodule, locally? Do we need to update CONTRIBUTING.md?

Suggested that we make this the 2nd step in our CONTRIBUTING.md

```
1. Initialize and fetch submodules (also re-run this after `git pull` to keep the submodule up to date):

       git submodule update --init --recursive

   Optionally, configure git to do this automatically after `git fetch`, `git pull` and `git checkout`:

       git config submodule.recurse true
```

ni-jfitzger:
Maybe we should also have a section that explains how to update the copy of ni-apis that's being used by our repo (`git submodule update --remote`).

I'm sure we'll need to periodically do this, but git submodule is pretty obscure stuff that I wouldn't expect even our maintainers to know or remember.

rahur-NI:
> Maybe we should also have a section that explains how to update the copy of ni-apis that's being used by our repo (`git submodule update --remote`).
> 
> I'm sure we'll need to periodically do this, but git submodule is pretty obscure stuff that I wouldn't expect even our maintainers to know or remember.

Thanks Jay.. I had pushed previous changes.. Now i updated as per your suggestion and it makes more sense to include in this way.

ni-jfitzger (approved):

PR #2181, opened 2026-06-29
Fix the python session to return correct error message when invalid session for MI drivers

zoechanzy:

- Fix get_error_description returning "Failed to retrieve error description." when the session is invalid
  - **TRY Block 1 — Try `get_error` (reads session error queue)**
If the session is valid and the error queue has not been overwritten, returned_error_code matches error_code and we return the description immediately.
    - Test: invalid resource name error
    image
   - **TRY Block 2 — Try error_message with current session**
Reached if Block 1 fell through — either `get_error` raised, or it returned a mismatched code (queue was overwritten). Unlike `get_error`, `error_message` is a direct lookup by `error_code` and does not read the error queue. If the session is still valid, this returns the correct description without touching the session handle.
    ( second try statement is exactly same with the second try statement in original file. I only updated the docstring section).
  - **TRY Block 3 — Reset session handle to 0, retry error_message**
Reached only when Block 2 also failed — meaning the error_message in TRY block 2 fail to retrieve error due to internal error or it is an invalid session. `error_message` rejects calls with an invalid session handle, but succeeds with `vi=0`. 
    -  No test for real behaviour of valid session in third try block because did not manage to trigger a real internal in valid session that failed in first and second try statement.
 - Updated the mako template and regenerated the affected drivers (nidcpower, nidigital, nidmm, nifake, nifgen, niscope, niswitch).
 - Added nifake unit test to verify the session-handle reset path calls error_message again with vi=0

* Fix [ bug 3877024](https://ni.visualstudio.com/DevCentral/_workitems/edit/3877024)
Labview error description
python error description
When session2 is opened after session1, driver will no longer have session handle of session1. It will have session handle of session2. If we use the invalidated session1, the error description between Labview and Python is different. The expected error description of python should be same with Labview, which is "(Hex 0xBFFA1190) The session handle is not valid.".

- Test of invalid session (this bug after fixing):
      image

- Unit tests and flake8 passed.

### DIFF
--- a/build/templates/_library_interpreter.py/_get_error_description.py.mako
+++ b/build/templates/_library_interpreter.py/_get_error_description.py.mako
@@ -22,10 +22,20 @@
 
         try:
             '''
+            It is possible that the session is valid but the returned_error_code unequal to error_code
+            '''
+            error_string = self.error_message(error_code)
+            return error_string
+        except errors.Error:
+            pass
+
+        try:
+            '''
             It is expected for get_error to raise when the session is invalid
             (IVI spec requires GetError to fail).
             Use error_message instead. It doesn't require a session.
             '''
+            self.set_session_handle()
             error_string = self.error_message(error_code)
             return error_string
         except errors.Error:
--- a/build/templates/_library_interpreter.py/_get_error_description.py.mako
+++ b/build/templates/_library_interpreter.py/_get_error_description.py.mako
@@ -21,9 +21,7 @@
 % if 'error_message' in config['functions']:
 
         try:
-            '''
-            It is possible that the session is valid but the returned_error_code unequal to error_code
-            '''
+            # It is possible that the session is valid but the returned_error_code unequal to error_code
             error_string = self.error_message(error_code)
             return error_string
         except errors.Error:
--- a/build/templates/_library_interpreter.py/_get_error_description.py.mako
+++ b/build/templates/_library_interpreter.py/_get_error_description.py.mako
@@ -21,7 +21,12 @@
 % if 'error_message' in config['functions']:
 
         try:
-            # It is possible that the session is valid but the returned_error_code unequal to error_code
+            # get_error reads the session's error queue, which may have been overwritten,
+            # causing it to return a mismatched error code. error_message takes the error
+            # code directly as a parameter and looks up its description without reading the
+            # queue, it will return the description for the specific error code.
+            # Use error_message in current session before the handle reset in the next block.
+
             error_string = self.error_message(error_code)
             return error_string
         except errors.Error:
@@ -29,11 +34,10 @@
 
         save_vi = self.get_session_handle()
         try:
-            '''
-            It is expected for get_error to raise when the session is invalid
-            (IVI spec requires GetError to fail).
-            Use error_message instead. It doesn't require a session.
-            '''
+            # It is expected for get_error to raise when the session is invalid
+            # (IVI spec requires GetError to fail).
+            # Use error_message instead. It doesn't require a session.
+
             self.set_session_handle()
             error_string = self.error_message(error_code)
             return error_string
--- a/build/templates/_library_interpreter.py/_get_error_description.py.mako
+++ b/build/templates/_library_interpreter.py/_get_error_description.py.mako
@@ -27,13 +27,13 @@
         except errors.Error:
             pass
 
+        save_vi = self.get_session_handle()
         try:
             '''
             It is expected for get_error to raise when the session is invalid
             (IVI spec requires GetError to fail).
             Use error_message instead. It doesn't require a session.
             '''
-            save_vi = self.get_session_handle()
             self.set_session_handle()
             error_string = self.error_message(error_code)
             return error_string
--- a/build/templates/_library_interpreter.py/_get_error_description.py.mako
+++ b/build/templates/_library_interpreter.py/_get_error_description.py.mako
@@ -33,10 +33,13 @@
             (IVI spec requires GetError to fail).
             Use error_message instead. It doesn't require a session.
             '''
+            save_vi = self.get_session_handle()
             self.set_session_handle()
             error_string = self.error_message(error_code)
             return error_string
         except errors.Error:
             pass
+        finally:
+            self.set_session_handle(save_vi)
 % endif
         return "Failed to retrieve error description."
--- a/generated/nidcpower/nidcpower/_library_interpreter.py
+++ b/generated/nidcpower/nidcpower/_library_interpreter.py
@@ -116,11 +116,14 @@ class LibraryInterpreter(object):
             (IVI spec requires GetError to fail).
             Use error_message instead. It doesn't require a session.
             '''
+            save_vi = self.get_session_handle()
             self.set_session_handle()
             error_string = self.error_message(error_code)
             return error_string
         except errors.Error:
             pass
+        finally:
+            self.set_session_handle(save_vi)
         return "Failed to retrieve error description."
 
     def abort(self, channel_name):  # noqa: N802
--- a/generated/nidcpower/nidcpower/_library_interpreter.py
+++ b/generated/nidcpower/nidcpower/_library_interpreter.py
@@ -104,7 +104,12 @@ class LibraryInterpreter(object):
             pass
 
         try:
-            # It is possible that the session is valid but the returned_error_code unequal to error_code
+            # get_error reads the session's error queue, which may have been overwritten,
+            # causing it to return a mismatched error code. error_message takes the error
+            # code directly as a parameter and looks up its description without reading the
+            # queue, it will return the description for the specific error code.
+            # Use error_message in current session before the handle reset in the next block.
+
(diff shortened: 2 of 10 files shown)


generated/nidcpower/nidcpower/_library_interpreter.py
```diff
@@ -103,12 +103,22 @@ def get_error_description(self, error_code):
...
             pass
 
+        try:
+            '''
+            It is possible that the session is valid but the returned_error_code unequal to error_code
+            '''
+            error_string = self.error_message(error_code)
+            return error_string
+        except errors.Error:
+            pass
```
ni-jfitzger, line 116: >It is possible that the session is valid but the returned_error_code unequal to error_code

If this happens, then something likely went wrong retrieving the error message. I'm not sure calling `error_message` instead of `get_error_description` is going to change anything in that case.

generated/nidcpower/nidcpower/_library_interpreter.py
```diff
@@ -103,12 +103,22 @@ def get_error_description(self, error_code):
...
+
         try:
             '''
             It is expected for get_error to raise when the session is invalid
             (IVI spec requires GetError to fail).
             Use error_message instead. It doesn't require a session.
             '''
+            self.set_session_handle()
             error_string = self.error_message(error_code)
```
ni-jfitzger, line 125: Did you add the other `error_message` call to protect against resetting the session handle in instances where the session is valid but we fail to retrieve the error message?

ni-jfitzger:
FYI @zoechanzy , our Linux VM is currently in bad shape, so Linux System tests are going to fail. We hope to have it resolved in the next week or two.

build/templates/_library_interpreter.py/_get_error_description.py.mako
```diff
@@ -20,12 +20,22 @@
...
+            '''
+            It is possible that the session is valid but the returned_error_code unequal to error_code
+            '''
+            error_string = self.error_message(error_code)
+            return error_string
+        except errors.Error:
+            pass
+
         try:
             '''
```
bkeryan: FYI, it's weird that these are multi-line string literals instead of comments. They are not in a position where they will be interpreted as docstrings.

ni-jfitzger: Yeah, that struck me as odd, too.
Apparently the Python interpreter will evaluate it as a string and then discard it?

bkeryan: Yes, it will.

BTW it's legal to write a docstring inside an init method, e.g. https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

```
self.attr5 = None
"""str: Docstring *after* attribute, with type specified."""
```

But that's not what's being done here.

generated/nidcpower/nidcpower/_library_interpreter.py
zoechanzy: If the scenario is valid session but failed to retrieve error message, it will enter third try block. Resetting the session will not affect its error_code so it will flow into the error_string = self.error_message(error_code) as well. It will still fail to retrieve the error message since the error_code is unchanged. So, it will raise execption and return "Failed to retrieve error description"

If the scenario is invalid session, it will reset the session in third try block and self.error_message(error_code) will not raise execption because session is set to 0 already. So, it will return correct "The session handle is invalid" message.

build/templates/_library_interpreter.py/_get_error_description.py.mako
zoechanzy: Will change it to comment instead of docstring

generated/nidcpower/nidcpower/_library_interpreter.py
zoechanzy: Calling get_error_description again might cause infinite loop.

ni-jfitzger: No, I wasn't suggesting that we call `get_error_description` again. I'm merely trying to think about what behavior we'll see if this new section of code executes.

Oh, i see. Sorry, I meant `get_error`, not `get_error_description`.

generated/nidcpower/nidcpower/_library_interpreter.py
ni-jfitzger: >If the scenario is valid session but failed to retrieve error message, it will enter third try block.

I'm not sure that's true.
I think the 2nd try block will return from `get_error_description` with an empty error string.
Have you done testing to determine the behavior?

If you believe we could enter the third try block and reset the session when the session is valid and we fail to retrieve the error message, then you shouldn't be resetting the session. Failure to retrieve an error message does not justify leaking a valid session handle.

zoechanzy: > I think the 2nd try block will return from get_error_description with an empty error string.
> Have you done testing to determine the behavior?

I tested with :
- Unknown error code (-9999999999)    → get_error_description returned "Explanation could not be found for the requested status code."
- VISA layer error code  (-1073807339 )→ get_error_description returned "VI_ERROR_TMO: Timeout expired before operation completed."

Since the real driver does not raise errors.Error from error_message() on a valid session either, I used a mock to cover the except errors.Error in second try block and third try block.
`with patch.object(
    type(session._interpreter),
    'error_message',
    side_effect=errors.Error(real_error_code)
):
    description = session._interpreter.get_error_description(real_error_code)
  `

It returns "Failed to retrieve error description."

>  Failure to retrieve an error message does not justify leaking a valid session handle.

I will update code to save the session handle before resetting it and restore it afterward to avoid leaking a valid session handle. We cannot selectively restore only valid sessions because we can only determine if a session handle is valid by calling a C driver function and observe whether it succeeds or fails.

adrianthongkh:
Requested to update description to be more descriptive of the error.
Provide examples if necessary.

Thanks!

jfan-1995:
Propose to rename the title to be `Fix the python session to return correct error message when invalid session for MI drivers`

In the PR description
`List issues fixed by this Pull Request below, if any.`

Can you explain with screenshot, something like:

The expected error message when invalidated NI-SWITCH session, the error message is different between Labview and Python
- Labview error message
image

- Python error message
image

build/templates/_library_interpreter.py/_get_error_description.py.mako
```diff
@@ -20,15 +20,26 @@
...
             It is expected for get_error to raise when the session is invalid
             (IVI spec requires GetError to fail).
             Use error_message instead. It doesn't require a session.
             '''
+            save_vi = self.get_session_handle()
```
jfan-1995: Hmm you are defining and calling `self.get_session_handle()` here but reference in the `finally`. My concern if `get_session_handle()` raises anything other than `errors.Error`, it may skip the rest of the `try` block, and then when run `finally` block the `save_vi` is unbound.

I think maybe we should move this `save_vi = self.get_session_handle()` before the `try` block?

```
save_vi = self.get_session_handle()     # moved outside try
try:
   self.set_session_handle()
   error_string = self.error_message(error_code)
   return error_string
except errors.Error:
   pass
finally:
   self.set_session_handle(save_vi)       # always safe or not run if save_vi failed
```

jfan-1995: Hmm I notice the `_library_interpreter.py` files on all drivers are also written like this, I not sure how the `generated` folder files being written, I suppose we only need to change this file and run some local build/script it will update for the `_library_interpreter.py` files for all drivers?

zoechanzy: > I suppose we only need to change this file and run some local build/script it will update for the`_library_interpreter.py` files for all drivers?

@jfan-1995 I updated the py.mako file only and ran tox command to generate all the files and run unit test.

generated/nidcpower/nidcpower/_library_interpreter.py
jfan-1995: I also find this comment a bit confusing....

What I understand is `get_error` reads what is currently in the session, if it is overwritten or cleared it could still successfully run but will return mismatched error code.

So this try block is to check the `error_message` in the current session first before proceed to the `set_session_handle` in the next try block.

jfan-1995: How about wording the comment to be like this?
```
# get_error reads the session's error queue, which may have been overwritten,
# causing it to return a mismatched error code. error_message takes the error
# code directly as a parameter and looks up its description without reading the
# queue, so it will just return the description for the specific error code.
# Try it here with the current session before the handle reset in the next block.
```

marcoskirsch (approved):

zoechanzy:
@vnktshr21 The PR is ready for review.

vnktshr21 (approved):

vnktshr21:
rhel builds in PR are currently undergoing some issues. Given the nature of the changes in this PR should not be impacting any OS specific behavior, I am going ahead with the merge of the changes.

PR #2207, opened 2026-08-25
Utilize nitlsconfig 1.0.0a2 + nitlsconfig's new gRPC audit session connect feature and error elaboration improvement

alexdubois-ni:

- ~~[ ] I've added tests applicable for this pull request~~

No system tests added as of this time, but we will make a subsequent PR to add system testing. My reasoning is:

1. What is being added here is fairly minor changes. Both nitlsconfig entrypoints we are utilizing inside of nimi-python itself will not change or add behavior whatsoever unless a channel is created with nitlsconfig's create_grpc_device_channel. For full system tests to succeed, we need to figure out how to do certificate provisioning, and modify the gRPC device json file to utilize nitlsconfig.
2. For now, I have opted to add unit tests though by simply creating a fake gRPC channel that is tagged with the same mechanism used in nitlsconfig to enable the auditing and error elaboration query capability and confirm the behavior. This at least gives us solid coverage of the changes proposed here.

We need to ensure we pull in the latest alpha version of nitlsconfig and integrate with latest changes which adds:

1. A new audit_session_connect function. This function when called will check whether the channel passed in via gRPC options was created from nitlsconfig's create_grpc_device_channel. If it WAS, then we will perform audit logging for whether the gRPC connection succeeded and what the target address + port for it was. This means it was an NI-sanctioned NI-TLS utilizing channel and we want to add client-side logging for it.
2. Utilize the get_tls_connection_error_elaboration function to replace the gRPC unavailable's generic message. This function returns None if the channel was not supplied by nitlsconfig's create_grpc_device_channel. The unavailable message is what gets reported when the TLS handshake fails which is VERY unhelpful otherwise. Instead, we want to tell users that they might want to verify against Hardware Manager what their TLS settings are. This matches the behavior we've encoded into our LabVIEW API.
3. nitlsconfig 1.0.0a2 also has a breaking change to improve the API naming and parameter set. We should anchor ourselves with 1.0.0a2 as our floor as this no longer intends to change (for now)
4. Added unit tests using a fake gRPC channel that is tagged with the same mechanism nitlsconfigtest uses to opt us into auditing and error elaborations and exercises our changes.

N/A. This is new development.

Ran the unit tests locally. New tests and the pre-existing tests proceed to succeed as-is

### DIFF
--- a/build/templates/_grpc_stub_interpreter.py.mako
+++ b/build/templates/_grpc_stub_interpreter.py.mako
@@ -13,6 +13,7 @@ are_complex_parameters_used = helper.are_complex_parameters_used(functions)
 
 import grpc
 import hightime  # noqa: F401
+import nitlsconfig
 import session_pb2 as session_grpc_types
 import threading
 import warnings
@@ -69,7 +70,11 @@ class GrpcStubInterpreter(object):
             elif grpc_error == grpc.StatusCode.INVALID_ARGUMENT:
                 raise ValueError(error_message) from None
             elif grpc_error == grpc.StatusCode.UNAVAILABLE:
-                error_message = 'Failed to connect to server'
+                # gRPC reports a rejected TLS handshake and an unreachable server with the
+                # same code, so ask NI-TLS whether it built this channel and can say more.
+                error_message = nitlsconfig.get_tls_connection_error_elaboration(
+                    self._grpc_options.grpc_channel
+                ) or 'Failed to connect to server'
             elif grpc_error == grpc.StatusCode.UNIMPLEMENTED:
                 error_message = (
                     'This operation is not supported by the NI gRPC Device Server being used. Upgrade NI gRPC Device Server.'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
 
 #### [nidcpower] Unreleased
 - Added
-  - (Common) Added `nitlsconfig[grpc]>=1.0.0a1` to the optional gRPC dependencies.
+  - (Common) Added `nitlsconfig[grpc]>=1.0.0a2` to the optional gRPC dependencies.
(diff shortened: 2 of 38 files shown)


build/templates/_grpc_stub_interpreter.py.mako
```diff
@@ -13,6 +13,7 @@ are_complex_parameters_used = helper.are_complex_parameters_used(functions)
 
 import grpc
 import hightime  # noqa: F401
+import nitlsconfig
```
marcoskirsch, line 16: How does it work? when client does `pip install nidmm` will `nitlsconfig` be installed for them? If not, when / how will the client learn the need to install it?

marcoskirsch (approved):

build/templates/_grpc_stub_interpreter.py.mako
alexdubois-ni: There is a grpc extra that is required to access our gRPC support:

You need to run: `pip install nidmm[grpc]` specifically, which will then automatically pull in nitlsconfig (and grpcio itself). If you fail to do this (e.g. just do pip install nidmm) and try to access our grpc support via the session constructor by passing in gRPC options, you will fail during runtime, but before nitlsconfig here even with:

ModuleNotFoundError: No module named 'grpc'

At which point, you would then subsequently fail with the same message but for nitlsconfig if you were to pip install grpcio subsequently yourself. I think this is sufficient.

This [grpc] extra requirement is a fully pre-existing behavior.

Lastly, for session.py's import, this will sequentially happen AFTER importing _grpc_stub_interpreter.py, which means that since we have the gated import nitlsconfig in the if grpc_options statement, this will have no impact here either and the grpc requirement I mentioned above is the first failure still.

vnktshr21 (approved):

PR #2208, opened 2026-08-26
Add new TLS tests to nidmm system tests

ryanwixon-emerson:

This PR upgrades the `TestGrpc` system tests to use mTLS encryption options and adds new tests for testing additional encryption options. These changes are only for DMM at the moment, but once this PR is merged, they will be expanded to cover every driver supported by `nimi-python`.

The gRPC tests in full include:
- `TestGrpcSecuredTLS` - Runs the full suite of tests with mTLS encryption, configured via `ni-tls-config`.
- `TestGrpcUnsecuredTLS` - Runs a small set of basic usability tests with `ni-tls-config`'s "Disabled" option.
- `TestGrpcNoTLS` - Runs a small set of basic usability tests with no dependency on `ni-tls-config` at all (this is effectively what is running today)
- `test_unsecured_server` - A standalone test ensuring that a TLS-enabled client will notconnect to an unsecured server.
- `test_unsercured_client` - A standalone test ensuring that a TLS-enabled server will not allow an unsecured client to connect.

In order to assist with the setup, a version of the `nitlsconfigtest` is being used. This provides us with utilities to configure the `ni-tls-config` .yml configuration files, and to provision certificates. This PR also includes two functions which wrap the Python scripts it provides and gives the tests an easy-to-use interface to interact with them.

N/A - This PR does not address any issues

N/A - These are tests themselves. However, I did verify using Wireshark that the tests really are using TLS encryption with the expected configurations.


ryanwixon-emerson:
Flaky system tests seem to be timing out. It's a different one each time and are part of test suites that this PR does not touch, but everything related to my changes is consistently passing. I'm publishing this PR now to avoid delaying feedback longer.

ni-jfitzger:
> Flaky system tests seem to be timing out. It's a different one each time and are part of test suites that this PR does not touch, but everything related to my changes is consistently passing. I'm publishing this PR now to avoid delaying feedback longer.

You don't need to commit an empty change to rerun a failed PR Check.
If you're unable to rerun it yourself, you can message me and I can rerun the failed jobs. That works better than retriggering all of the checks.

src/shared/system_test_utilities.py
```diff
@@ -1,11 +1,15 @@
...
 import os
 import pathlib
 import pytest
 import re
 import subprocess
+import sys
 import threading
 import time
 
+import nitlsconfig_32_bit_patch  # noqa: F401
```
ni-jfitzger: Why do you need to import this, if it's unused?

src/shared/system_test_utilities.py
```diff
@@ -104,3 +108,121 @@ def impl_test_multi_threading_ivi_synchronized_wrapper_releases_lock(ivi_method_
...
+
+    verbosity = max(0, min(verbosity, 4))
+    verbosity_arg = {
+        0: "-qq",
+        1: "-q",
+        3: "-v",
+        4: "-vv",
+    }.get(verbosity)
+
+    command = [sys.executable, str(pathlib.Path(script_path)), server_host_arg, server_user_arg]
```
ni-jfitzger, line 158: Why include the first 2 elements, if you aren't going to pass them?

src/shared/nitlsconfig_32_bit_patch.py
```diff
@@ -0,0 +1,32 @@
...
+
+        old = ctypes.c_void_p()
+        disabled = bool(ctypes.windll.kernel32.Wow64DisableWow64FsRedirection(ctypes.byref(old)))
+        try:
+            return original_init(self, args, *posargs, **kwargs)
+        finally:
+            if disabled:
+                ctypes.windll.kernel32.Wow64RevertWow64FsRedirection(old)
+
+    subprocess.Popen.__init__ = patched_init
```
ni-jfitzger: Wow, this is hacky.
I don't feel good about trying to call a 64-bit executable from a 32-bit process.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -327,17 +329,29 @@ def test_fetch_waveform_into(self, session):
...
+    @classmethod
+    def grpc_channel(cls):
+        system_test_utilities.configure_tls_modes(
+            service="ni-grpc-device",
+            server_host="localhost",
+            server_cert_mode="ManagedSelfSigned",
+            server_client_mode="ManagedSelfSigned",
+            client_cert_mode="Managed",
+            client_server_mode="TrustedCertificates"
+        )
```
ni-jfitzger: I'm not sure I want us to have to deal with these details in every system test module. It might be better to have some secure and insecure wrapper functions in system_test_utilities that set the right values for us.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -369,3 +383,158 @@ def test_attach_to_non_existent_session(self, grpc_channel):
...
+        with session.initiate():
+            session.fetch()
+        with session.initiate():
+            session.fetch()
+
+    def test_multi_point_acquisition(self, session):
+        session.configure_multi_point(4, 2)
+        session.configure_measurement_digits(nidmm.Function.DC_VOLTS, 1, 5.5)
+        measurements = session.read_multi_point(8)
+        assert len(measurements) == 8
```
ni-jfitzger: You're duplicating quite a bit, here.
And you're doing it twice.
We could potentially move these tests into a BasicValidationSystemTests class and have SystemTests inehrit from it to avoid the duplication.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -369,3 +383,158 @@ def test_attach_to_non_existent_session(self, grpc_channel):
...
+
+    # Attempt to connect to the server. Since the client is expecting a TLS-enabled server, this should fail.
+    with system_test_utilities.GrpcServerProcess(config_file_path) as proc:
+        unsecured_server_channel = nitlsconfig.create_grpc_device_channel('localhost', proc.server_port)
+        grpc_options = nidmm.GrpcSessionOptions(unsecured_server_channel, '')
+        try:
+            with nidmm.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:4082; BoardType:PXIe', grpc_options=grpc_options):
+                assert False
+        except nidmm.Error:
+            pass
```
ni-jfitzger: This really feels like it's creeping into the realm of testing grpc-device-server, rather than our module.

src/shared/nitlsconfig_32_bit_patch.py
ryanwixon-emerson: `nitlsconfig`'s CLI tool only builds for 64-bit today (the underlying DLL it calls builds for both). I suppose one way to avoid the hack is to update `nimibot` to copy the executable into SysWow64, but of course this doesn't address your underlying concern.

Another option could be to update `nitlsconfig` to build the CLI tool for 32-bit. I'm not sure how well this would be received or if there's a reason it's not already being done. This would force us to either install 26.8 versions of the drivers or do a separate independent install of `nitlsconfig` over the 26.5 versions as a temporary measure.

Or, if we don't care about testing the 32-bit tests, we could just disable the gRPC TLS testing for them? But with the other restructuring this PR does it would mean they'd just be doing the smoke tests.

src/nidmm/system_tests/test_system_nidmm.py
ryanwixon-emerson: Fair enough. I originally added this because I was trying to mimic the types of TLS tests that we were using for RoCo, but seeing as `nimi-python` is just a module that uses `grpc-device-server` rather than being a service itself it makes sense to remove these.

src/shared/system_test_utilities.py
ryanwixon-emerson: Importing this automatically runs the patching function that it includes. It probably will not be relevant anymore after addressing the other comments.

src/shared/system_test_utilities.py
ryanwixon-emerson: That seems to have been a strange Copilot-ism that I didn't catch when I was reviewing it. It's gone now since I refactored this approach.

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -40,6 +35,13 @@ def test_multi_point_acquisition(self, session):
         measurements = session.read_multi_point(8)
         assert len(measurements) == 8
 
+
+class SystemTests(BasicValidationTests):
+    @pytest.fixture(scope='function')
+    def session(self, session_creation_kwargs):
+        with nidmm.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:4082; BoardType:PXIe', **session_creation_kwargs) as simulated_session:
+            yield simulated_session
```
ni-jfitzger: Why not move this into BasicValidationTests, too?

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -385,7 +376,8 @@ def test_attach_to_non_existent_session(self, grpc_channel):
             assert str(e) == f'{expected_grpc_error}: {expected_error_message}'
 
 
-class TestGrpcUnsecuredTLS:
+@pytest.mark.skipif(sys.maxsize < 2**32, reason="gRPC tests are not supported in 32-bit processes")
```
ni-jfitzger: This reasoning begs questions for future reviewers. And it's also clearly not true, given that we are running TestGrpcNoTLS tests for 32-bit.
Can you be more specific about what's not supported?
Isn't it tls configuration/setup?

src/nidmm/system_tests/test_system_nidmm.py
```diff
@@ -17,13 +17,7 @@
...
 
 
-class SystemTests:
-    @pytest.fixture(scope='function')
-    def session(self, session_creation_kwargs):
-        with nidmm.Session('FakeDevice', False, True, 'Simulate=1, DriverSetup=Model:4082; BoardType:PXIe', **session_creation_kwargs) as simulated_session:
-            yield simulated_session
-
-    # Basic usability tests
+class BasicValidationTests:
```
ni-jfitzger, line 22: Maybe add a comment that these tests were split off from the SystemTests class so that we could run a small subset of tests with gRPC and TLS . With all these test classes, it may not be immediately obvious.

src/shared/system_test_utilities.py
```diff
@@ -214,15 +212,44 @@ def configure_tls_modes(
...
-        "import runpy, sys\n"
-        "import nitlsconfig_32_bit_patch\n"
-        f"sys.argv = [{script_path!r}] + {args!r}\n"
-        f"runpy.run_path({script_path!r}, run_name='__main__')\n"
+def configure_tls_modes_secure(
+    service: str,
+    server_host: str,
+    server_user: str | None = None,
+    client_host: str | None = None,
+    client_user: str | None = None,
```
ni-jfitzger: Why include these additional parameters?
We have no intention of using them, right?

The same goes for `configure_tls_modes_insecure`.

ryanwixon-emerson: We might not be using them today, but maybe someone would want to update or rework the tests to do non-localhost communication at some point in the future. My thought was that we should provide a wrapper function that allows you to use the full functionality that the `nitlsconfigtest` scripts offer.

ni-jfitzger: Given the isolated nature of our VMs and how they turn off and on, as needed, to reduce costs, I can't picture us ever testing communication across multiple systems with nimibot. I'm not sure it's even possible.

If we ever do decide to test it, we can add extra optional arguments at that time.

ni-jfitzger (approved):

PR #2214, opened 2026-09-18
Add new TLS tests to all driver API system tests

ryanwixon-emerson:

This PR implements TLS-enabled gRPC tests for every driver API which includes `grpc-device` related testing:
- `nidcpower`
- `nidigital`
- `nifgen`
- `nirfsa`
- `nirfsg`
- `niscope`
- `niswitch`

The implementations follow the pattern first established by the trial driver, NI-DMM, in a previous PR (#2208):
- Choose a small subset of tests from the main test runner to migrate to a new `BasicValidationTests` class
- Add two new test classes to test `ni-tls-config` enabled workflows (one for enabled mTLS and one for disabled mTLS)

(This PR also makes a minor change to the NI-DMM tests by adding the self test function to the `BasicValidationTests` class. This is to bring it in line with these drivers, since I decided that it would be a good idea for everything to run it as a part of all of their tests).

N/A - This PR does not address any issues.

The PR CI pipelines successfully ran and indicated that the new tests are running correctly. This is sufficient because the implementation details are identical to that of NI-DMM, which was thoroughly tested with Wireshark to validate that the tests are using the proper TLS configurations and use real encrypted communication.


src/nidcpower/system_tests/test_system_nidcpower.py
```diff
@@ -37,7 +38,9 @@ def pytest_generate_tests(metafunc):
             metafunc.parametrize('session', [True], indirect=True)
 
 
-class SystemTests:
+# Defines a subset of system tests to validate basic NI-DCPower functionality. This is run as a part of the full SystemTests class, and
```
ni-jfitzger: I would say this use of "NI-DCPower" is slightly inaccurate. The purpose of our nimi-python system tests is to validate the API, not the driver. As part of that, we do make actual driver calls, but validating the driver functionality would be much more complicated and generally requires real hardware.

The driver is NI-DCPower, but the Python API is nidcpower.
This comment applies to all of the system tests, you're touching, as well as nidmm system tests.

ni-jfitzger:
Might be worth including `test_allocate_arb_waveform` in nirfsa BasicValidationTests so that we test transferring some actual data.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -77,53 +80,19 @@ def single_instrument_session_5171(self, session_creation_kwargs):  # High chann
...
-    # Basic usability tests
-    def test_get_channel_names_with_single_instrument_session(self, single_instrument_session_5171):
-        expected_string = [f'{x}' for x in range(8)]
-        # Sanity test few different types of input. No need for test to be exhaustive
-        # since all the various types are covered by converter unit tests.
-        channel_indices = ['0-1, 2, 3:4', 5, range(6, 7), slice(7, 8)]
-        assert single_instrument_session_5171.get_channel_names(indices=channel_indices) == expected_string
+    def multi_instrument_session(self, session_creation_kwargs):
+        with niscope.Session(','.join(instruments), False, True, 'Simulate=1, DriverSetup=Model:5164; BoardType:PXIe', **session_creation_kwargs) as simulated_session:
+            yield simulated_session
```
ni-jfitzger, line 91: I'm not sure why you moved this multi_instrument_session function.

ryanwixon-emerson: Seems to have been a mistake I made when trying to migrate the fixtures to BasicValidationTests. Reverted.

src/niscope/system_tests/test_system_niscope.py
```diff
@@ -141,6 +110,20 @@ def test_read(self, multi_instrument_session, test_channels, test_channels_expan
         waveforms = multi_instrument_session.channels[test_channels].read(num_samples=test_record_length, num_records=test_num_records)
         check_fetched_data(waveforms, test_channels_expanded, test_record_length, test_num_records)
 
+    def test_get_channel_names_with_single_instrument_session(self, single_instrument_session_5171):
```
ni-jfitzger: I don't think there was any need to move `test_read` before `test_get_channel_names_with_single_instrument_session`.

ni-jfitzger:
Overall, this looks pretty good.

ni-jfitzger (approved):

marcoskirsch (approved):

